PR #184, opened 2018-08-20
better middleware checks status of service before triggering actions

julien51:


unlock-app/src/middlewares/lockMiddleware.js
```diff
@@ -78,23 +93,11 @@ export default function lockMiddleware ({ getState, dispatch }) {
...
-          // Location was changed but web3Service is not ready... so let's connect!
-          web3Service.connect({
-            network: getState().network,
-          }).then((account) => {
-            dispatch(setAccount(account))
-          }).catch(() => {
-            // we could not connect
-            // TODO: show error to user
+        // Location was changed, get the matching lock
+        const match = action.payload.pathname.match(/\/lock\/(0x[a-fA-F0-9]{40})$/)
```
benwerd: I wonder if the regex here should be moved into its own central function for later use elsewhere?

julien51: HA This is a good point. Since this PR is just moving code I'll skip it for now but adding a mental note to do this!

Issue #161, opened 2018-07-23
Show name of network on interface...


julien51:
When a user visits an unlock-protocol.com page they should instantly see which Ethereum network is being used (Main net, Rinkeby, Ropsten...)

Ideally this would take the form of a small persistent label at the bottom or top of the screen.

benwerd:
This has now been completed.

julien51:
🚀

PR #188, opened 2018-08-21
Add a simple layout engine and display current network.

benwerd:
Refs #161


benwerd:
Here's a screenshot with the network badge:

screen shot 2018-08-22 at 11 31 47 am

And I've added tests. PTAL.

PR #190, opened 2018-08-23
Upgradable Unlock Smart Contract

### LINKED ISSUE #172: Upgradable Unlock Smart Contract
The Unlock smart contract needs to be upgradable, without changing its address and its data (balances of Unlock points or deployed Locks).

The [Unlock.sol smart contract](https://github.com/unlock-protocol/unlock/blob/master/smart-contracts/contracts/Unlock.sol) has 3 main purposes:
 * Deploy locks and keep track of them
 * Keep track of the the Discount Token supply: grant them upon referrals.
 * Grant corresponding discounts on key purchases

A lot of the logic will be implemented later and will be adjusted as we make progress in understanding how the Unlock network is functioning. For that reason, it is important that we use an `Upgradeable` approach so that we can incrementally add and update the functionality of that smart contract.

For this we want to use the "Unstructured Storage Proxy" described by @leanthebean in [this article ](https://blog.zeppelinos.org/proxy-patterns/)

hardlydifficult:
Fixes #172

 - Added zos-lib.
 - Changed the Unlock constructor to initialize.  This is [required](https://docs.zeppelinos.org/docs/building.html).
 - Added a contract UnlockTestV2 to simulate an upgrade.
 - Refactored Unlock tests to use a behavior pattern, allowing tests to be reused.
 - Added an UnlockProxy test suite, demonstrating various use cases.
 - Updated lint settings to include terms used by tests.

Let me know if there are any questions or concerns.
-nick.


julien51:
Wow. This is amazing. Please give us a couple days to review all changes. Thanks a lot!

smart-contracts/contracts/UnlockTestV2.sol
```diff
@@ -0,0 +1,179 @@
...
+
+  // Events
+  event NewLock(
+    address indexed lockOwner,
+    address indexed newLockAddress
+  );
+
+  bool internal initialized;
+
+  function initialize(
```
julien51: I think this is one of the changes compared to Unlock.sol, but I fail to see the other ones?

smart-contracts/contracts/UnlockTestV2.sol
```diff
@@ -0,0 +1,179 @@
...
+    address _owner
+  )
+    public {
+      require(!initialized);
```
julien51: This too of course ;)

smart-contracts/contracts/UnlockTestV2.sol
```diff
@@ -0,0 +1,179 @@
...
+      owner = _owner;
+      grossNetworkProduct = 0;
+      totalDiscountGranted = 0;
+      initialized = true;
```
julien51: And that!

smart-contracts/test/Unlock/behaviors/createLock.js
```diff
@@ -1,44 +1,30 @@
...
-    it('should have kept track of the Lock inside Unlock with the right balances', () => {
+    it('should have kept track of the Lock inside Unlock with the right balances', async function () {
       let lock = Lock.at(transaction.logs[0].args.newLockAddress)
       // This is a bit of a dumb test because when the lock is missing, the value are 0 anyway...
-      return unlock.locks(lock.address).then(([deployed, totalSales, yieldedDiscountTokens]) => {
-        assert(deployed)
-        assert.equal(totalSales.toNumber(), 0)
-        assert.equal(yieldedDiscountTokens.toNumber(), 0)
-      })
+      let [deployed, totalSales, yieldedDiscountTokens] = await this.unlock.locks(lock.address)
```
julien51: +1 for using async/await. I should have done that earlier.... thanks!

smart-contracts/test/Unlock/UnlockProxy.js
```diff
@@ -0,0 +1,72 @@
...
+      this.unlock = await Unlock.at(this.proxy.address)
+      await this.unlock.initialize(unlockOwner)
+    })
+
+    describe('should function as a proxy', function () {
+      shared.shouldBehaveLikeV1(accounts, unlockOwner)
+
+      it('should fail if called from the proxy owner\'s account', async function () {
+        try {
+          await this.unlock.grossNetworkProduct({from: proxyOwner})
```
julien51: Do you mind explaining why this is the case?

smart-contracts/contracts/UnlockTestV2.sol
```diff
@@ -0,0 +1,179 @@
...
+  )
+    public
+    pure
+    returns (uint discount, uint tokens)
+  {
+    // an example modification
+    return (42, 42);
+  }
+
+  function testNewMethod()
```
julien51: Ha! and that too!

smart-contracts/package.json
```diff
@@ -33,7 +33,7 @@
     "solhint": "^1.2.1"
   },
   "scripts": {
-    "test": "truffle test",
+    "test": "NODE_ENV=test && truffle test",
```
julien51: Weird. Did tests no pass before that?

docker/unlock.tests.dockerfile
```diff
@@ -4,4 +4,5 @@ RUN mkdir /home/unlock
 COPY . /home/unlock
 WORKDIR /home/unlock
 
+ENV NODE_ENV=test
```
julien51: Same, did tests no pass before that?

julien51:
Again, thanks a lot for your speed and work on this.

A couple questions:

1. I see you're using zos' libraries, but my understanding is that this is not using their 'deployed' versions of these libraries, so it behaves exactly like the rest of the OpenZeppelin libs, right? (I have not used zos until now!)

2. When we create new 'versions' of the Unlock smart contract (like `UnlockTestV2`), I guess we have to include __everything__ that is in the previous versions, correct? So changes are only additive, correct? Is there anything else we need to be careful about? How does it behave if we add other `parent` contracts (`contract UnlockTestV2 is ..., Parent {`)

3. I'll submit a fix to make sure failing truffle are also triggered in CI! And I'll look into the failing tests. I guess it's a concurrency issue.

Thanks!!!

smart-contracts/contracts/UnlockTestV2.sol
hardlydifficult: This is actually on-par with Unlock now.  With the "Unstructured Storage Proxy" pattern, the only requirement for the contract itself is that instead of using a constructor, you use initialize.  

With this pattern, deploying v1 effectively creates 2 independent contracts.  You can interact with Unlock directly, as you had previously.  However if you want to upgrade in the future, you interact through the proxy.  When going through a proxy, data is stored in the proxy contract --- this means you get the logic from Unlock, including the layout of variables... but storage is completely independent.

Therefore we must change the constructor.  Anything assigned in a constructor would only impact Unlock directly... by using initialize, the same code can impact the proxy storage as well.

I'm going to submit a couple changes to clarify.  Give me a few..

smart-contracts/contracts/UnlockTestV2.sol
hardlydifficult: This has also been added to Unlock.sol as explained above.

smart-contracts/contracts/UnlockTestV2.sol
hardlydifficult: Same.  However there are changes below :)

smart-contracts/test/Unlock/behaviors/createLock.js
hardlydifficult: Good, I was a little hesitant changing styles like that without checking in with you first.  I do think this makes parts of the test much more readable.

smart-contracts/test/Unlock/UnlockProxy.js
hardlydifficult: I'm not able to explain the rational.  Not sure I understand.  However you can see this is driven by line 127 of their AdminUpgradeabilityProxy:

```
  function _willFallback() internal {
    require(msg.sender != _admin(), "Cannot call fallback function from the proxy admin");
    super._willFallback();
  }
```

hardlydifficult:
1) zos has two offerings.  Upgradability (using various patterns as the blog post you provided describes.)  The other offering a standard library.  If you are using the std lib, then the deployed version may be used.  However for Upgrades, each upgradable proxy is deployed by the proxy admin (and not reusable like the std lib is).

2) The only thing that matters is the layout of data in the contract.  I'll add another test to demonstrate this a bit.  You can add data (only by placing new data after the original list of data) and you can add, modify, or remove methods.

3) Cool.  My apologies for posting that comment on the wrong thread.

Note RE admin for this.  The ability to upgrade is effectively unlimited power.  If the keys were somehow compromised, one transaction could wipe all the data.  I would recommend making the proxy admin address another smart contract which requires multi-sig approval in order to upgrade Unlock.  ...I could help you with that, assuming you can afford my rates ;)

Okay, one more push coming in a few.

smart-contracts/contracts/UnlockTestV2.sol
julien51: Thanks!

julien51: I think we'll make it explicit when doing upgrades later by first committing a full identical copy of the smart contract and then doing a 2nd commit with only the changes. This should make things much easier to track down.

smart-contracts/package.json
julien51: hooo. ok!

docker/unlock.tests.dockerfile
julien51: I think our travis setup is [currently](https://github.com/unlock-protocol/unlock/issues/195) 💩... thanks!

julien51:
> Note RE admin for this. The ability to upgrade is effectively unlimited power. If the keys were somehow compromised, one transaction could wipe all the data. I would recommend making the proxy admin address another smart contract which requires multi-sig approval in order to upgrade Unlock. ...I could help you with that, assuming you can afford my rates ;)

Let's talk!

hardlydifficult:
Okay pushed a few changes.  Hope this helps to clarify a few things.  Let me know if there's anything more.

Email me at hardlydiff@gmail.com if you want to discuss additional projects.  

Thanks,
nick.

julien51:
Great! Thanks a lot. I'll give one more full look at this tomorrow in the morning to make sure I am not missing anything and I'll be in touch by email regarding the admin multi sig contract ;) 

Thanks a lot!

hardlydifficult:
Switched `UnlockTestV2` to inherit from the previous version.  

On second thought, this is may be a more clear pattern than we had previously.  V3 still shows the copy/paste modify approach.

By inheriting from the previous version, the variables will be in the correct order.  Then you can add more data and selectively modify methods.

smart-contracts/contracts/UnlockTestV2.sol
```diff
@@ -0,0 +1,50 @@
...
+ * @author Julien Genestoux (unlock-protocol.com)
+ * @author HardlyDifficult
+ * 
+ * This inherits from the previous version, adds data and modifies logic.
+ */
+
+import "./Unlock.sol";
+
+
+contract UnlockTestV2 is Unlock {
```
julien51: Ha! I think this is the best way to do it? Each new version `is` the previous version?

smart-contracts/contracts/UnlockTestV2.sol
```diff
@@ -0,0 +1,50 @@
...
+
+  // Example new data (which must come after the original data)
+  bool internal initializedV2;
+  uint public exampleData;
+
+  // Adding a second initialize for the new data as 'initialized' is already true when v2 is deployed.
+  function initializeV2()
+    public 
+  {
+    require(!initializedV2);
```
julien51: Is that mechanism required? (adding a check on whether the current version has been deployed)

smart-contracts/contracts/UnlockTestV3.sol
```diff
@@ -0,0 +1,87 @@
...
+ * 
+ * This is a copy paste of the current Unlock contract, with comments removed.
+ * Then a few small changes have been made, higlighted with comments.
+ */
+
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+import "./Lock.sol";
+
+
+contract UnlockTestV3 is Ownable {
```
julien51: So I guess if we were *not* using `is Unlock`, we can then re-inplement it all?

smart-contracts/contracts/UnlockTestV3.sol
```diff
@@ -0,0 +1,87 @@
...
+
+  event NewLock(
+    address indexed lockOwner,
+    address indexed newLockAddress
+  );
+
+  bool internal initialized;
+
+  // Example new data (which must come after the original data)
+  bool internal initializedV3;
```
julien51: **MUST** ?

smart-contracts/contracts/UnlockTestV2.sol
julien51: Also, which part is invoking `initializeV2` ?

smart-contracts/test/Unlock/UnlockProxy.js
julien51: Hum. Ok. I think I need to investigate this.

smart-contracts/contracts/UnlockTestV2.sol
hardlydifficult: That's a style question. The blog post you originally shared suggested this approach. It would make it easier to ensure the data is correct.

smart-contracts/contracts/UnlockTestV2.sol
hardlydifficult: We need to ensure in it only happens once.

You can do it with the same name for each contract version, but it [confuses Truffle](https://docs.zeppelinos.org/docs/advanced.html). That doc has an alternative work around we could use.

The test shows calling on upgrade.

smart-contracts/contracts/UnlockTestV3.sol
hardlydifficult: Yes. Either approach but if using this style be careful not to change the order data appears.

smart-contracts/contracts/UnlockTestV3.sol
hardlydifficult: Yes. See [Preserving the storage structure](https://docs.zeppelinos.org/docs/advanced.html)

smart-contracts/contracts/UnlockTestV3.sol
julien51: Yup. Makes sense!

smart-contracts/contracts/UnlockTestV2.sol
julien51: Cool. Thanks!

smart-contracts/contracts/UnlockTestV2.sol
julien51: Correct. I think we'll use that approach.

julien51:
Just merged it! I think this should release the funds! Thanks! 🚀

PR #191, opened 2018-08-23
Removing layout engine and replacing shell network display with an optional component.

benwerd:


benwerd:
My intention is still to add a layout engine, but this wasn't the way to do it. It looks like React, and in particular Router 4, hasn't been designed to handle multiple shell templates in this way - but it's something that makes interface development so much easier. So I'll open an issue and keep thinking about it.

PR #192, opened 2018-08-23
adding a failing test

julien51:
Please do not review. I am just testing that CI fails when a truffle test fails!


julien51:
ok, so that is the desired behavior but this explicit `exit(1)` is probably not great in the long run. Adding a note about that.

PR #193, opened 2018-08-23
Fixing minor web console errors.

benwerd:
These were throwing warnings in the console.

In the future, I think it would be good to add these sorts of things to the CI scripts / tests / linter - but I wanted to fix the warnings first.

Fixing some minor JavaScript console errors


benwerd:
@julien51 Right - I'll take a look at adding that.

Issue #195, opened 2018-08-23
A better CI setup


julien51:
We are using Travis-CI to run our tests on Pull Requests.
The current setup involves using docker-compose to build an setup in which we can do tests. This setup has 2 services: 
1. ganache-cli which runs an instance of [ganache](https://github.com/trufflesuite/ganache-cli)
2. a container with all the Unlock code with a single command to run all the tests from `tests.sh`.

When adding new test suites (we currently only test `smart-contracts` and `unlock-app`) you need to update `tests.sh`.

The travis-ci config is "strange": 
```
before_install:
- docker-compose -f docker/docker-compose.ci.yml build
- docker-compose -f docker/docker-compose.ci.yml up --abort-on-container-exit
```

We first build the cluster: so far so good, but then we also run it and ask it to terminate when a container exits. That second instruction should probably not in `before_install`. Also we should probably terminate properly.

This setup is probably what is causing #194? (the error does happen but for some reason `test.sh` continues it execution...

Finally, running CI already takes about 5 minutes while this is a very early project. There are ways to optimize that building process.

benwerd:
I think we'll gain a lot of time by caching Docker images. There's a good-looking tutorial here: http://www.rundef.com/fast-travis-ci-docker-build 

We could also do more to run tests in parallel: https://docs.travis-ci.com/user/speeding-up-the-build/#parallelizing-your-builds-across-virtual-machines

hardlydifficult:
To save on test run time, you could consider removing `smart-contracts\test\Unlock\Unlock.js`.  That file demonstrates direct access to the `Unlock` contract vs going through a proxy.  If you are moving forward with the proxy solution - there should be no benefit in maintaining both tests.  `smart-contracts\test\Unlock\UnlockProxy.js` includes all the same tests running through a proxy.

julien51:
I think out set up is actually pretty good now. Let's close!

julien51:
After meeting with @heewa this morning I start to have a better understanding of how our setup is not optimal. We probably have a decent approach but we need to improve further ;)

julien51:
So, in a nutshell, the feedback was this:

1. we need to extract the stuff from `test.sh` and pull that as much as possible into the `npm` scripts (or else that we're running for each code component.
2. we can and should run more containers... however they should probably of the same "kind" with just a different `CMD`. This way, the same image can be used to run the application, to run the test suite, or even be deployed!

julien51:
[...]

PR #196, opened 2018-08-23
Adding an explicit exit 1 to make sure that CI stops when tests are failing. Fixes #194

### LINKED ISSUE #194: CI not failing when truffle tests fail
As identified by @hardlydifficult, our CI does not fail when truffle tests are failing. This is not right.

julien51:


julien51:
Probably not the most robust approach but will do until we improve our [travis runs](https://github.com/unlock-protocol/unlock/issues/195)?

benwerd:
Can you clarify if is this supposed to be failing right now?

julien51:
No it should not be failing anymore. Basically it fixes #194 when even though tests might have been failing, the travis build would be successful.
I pushed a branch which showed that (failing test but passing build), then added this fix which made the build fail too and finally removed the failing tests.

PR #202, opened 2018-08-25
Moved ZOS env conf to the test itself

hardlydifficult:
Instead of setting in the package and the dockerfile, the config is set in the test file (line 1).


julien51:
Thanks a bunch!

PR #207, opened 2018-08-28
working on implementing new design template by @smombartz

julien51:


julien51:
This is WIP.
screen shot 2018-08-28 at 11 16 20

unlock-app/public/images/icons/about.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
+    <path fill="#A6A6A6" fill-rule="nonzero" d="M12 0C5.37 0 0 5.37 0 12s5.37 12 12 12 12-5.37 12-12S18.63 0 12 0zm1.2 18h-2.4v-7.2h2.4V18zm0-9.6h-2.4V6h2.4v2.4z"/>
```
benwerd: I really love that we're using SVG everywhere for icons. :+1: :+1:

unlock-app/public/template.html
```diff
@@ -0,0 +1,94 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+  <title>Unlock</title>
+  <link rel="stylesheet" href="template.css">
```
benwerd: Should we be calling the stylesheet something else? I'm not sure how descriptive `template` is. But it's probably fine for now.

unlock-app/public/template.html
```diff
@@ -0,0 +1,94 @@
...
+          <a class="icon"><img src="./images/icons/github.svg" /></a>
+        </header>
+
+        <div class="account">
+          <header class="account__head">
+            <h2>Account</h2>
+            <span class="network_info data">Rinkeby Test Network</span>
+          </header>
+
+          <div class="account__details">
```
benwerd: Beautifully clear HTML. There is an argument for using a table here because it's actually tabular data - but my only real concern is that it could be hard for an accessible browser to read this. I don't know if that's true. It would be nice to check that out sometime soon.

unlock-app/public/template.html
```diff
@@ -0,0 +1,94 @@
...
+
+      </div>
+      <div class="item">
+        <!-- Right -->
+      </div>
+    </div>
+  </div>
+</body>
+
+</html>
```
benwerd: Can we add a newline below the `</html>`?

PR #209, opened 2018-08-30
Consistent lock CSS grid and fonts

benwerd:


benwerd:
Building on @julien51's previous work by refining the layout of a lock based on the existing design. I included the use of the font Anonymous Pro for currencies, as Monaco was defined in the design but is a Mac-only font.

Screenshot of WIP:

screen shot 2018-08-30 at 11 28 03 am

benwerd:
Added a minor change to push the icon bar to the right:

screen shot 2018-08-30 at 11 31 23 am

benwerd:
And one further change. Now I'll leave it alone ;)

screen shot 2018-08-30 at 11 52 44 am

PR #212, opened 2018-08-31
adding button to create locks

julien51:


julien51:
Getting there!
screen shot 2018-08-31 at 11 45 41

PR #213, opened 2018-08-31
Using CSS variables for colors

benwerd:
Including editable fields for locks in editing state; names for locks in saved state; hover-over icon bar.


julien51:
And to rerun the tests, the best is to click on "details" and then "restart build" on the right.

PR #216, opened 2018-08-31
adding lock icons

julien51:


julien51:
screen shot 2018-08-31 at 15 58 28

benwerd:
This is really cool. I think a potential future upgrade might be for contrast and readability - the yellow in particular seems really faint to me. But LG!

julien51:
Great point. Maybe we can find ways to do that algorithmically.

PR #227, opened 2018-09-07
Adding stories for Lock

benwerd:


unlock-app/src/stories/creator/Lock.stories.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Lock } from '../../components/consumer/Lock'
+import { Provider } from 'react-redux'
+import createUnlockStore from '../../createUnlockStore'
+
+let store = createUnlockStore()
+
+storiesOf('Lock')
+  .addDecorator(story => <Provider store={store}>{story()}</Provider>)
```
julien51: So, I think we should use stories for components "in isolation", without the provider, but pass the `props` manually.

julien51: Ho, that's interesting. I did not think about the downstream component... I guess this is a design issue then? 

I strongly agree with you about testing include all lock states in stories.

Let's merge this for now but add a note about that because I think it is important to test the components in isolation if we can?

benwerd: Definitely agree about testing in isolation. OK - I'll merge and we can figure out the best way to get around this architecturally. I suspect we're not going to get away without some context right now - but actually, we may be able to by refactoring the components themselves to access the named exports rather than the connected versions. If we do that, we'd need to make sure we were passing explicit parameters all the way down the chain, which doesn't strike me as a bad idea.

julien51: Yeah, I think passing `props` down *was* actually the way to go...  but redux and/or the context API actually were supposed to make that easier :/ So I am lost!

PR #229, opened 2018-09-07
HeaderIcon component

julien51:


julien51:
This makes my svg sprite thing irrelevant so I'll probably get rid of it :(

julien51:
Dammit. This won't work :(

PR #233, opened 2018-09-08
adding a script to generate react components from svg files

julien51:


julien51:
After much attempts, I think this is the right approach for now until we `eject` the create-react app... which will allow us to create custom Webpack configs to do that.

Next PR will remove the previous sprite work which is unneeded.

julien51:
Agreed! I think we could do that automagically with webpack.

PR #235, opened 2018-09-10
Adding Icons and their stories

julien51:


julien51:
Please review https://github.com/unlock-protocol/unlock/pull/234 before since it includes a commit (revert) from there...

julien51:
As a FYI I think we could eventually do better on these icons and have them more "raw" with less styling so that we can re-use them in different context... but let's start with this!

PR #244, opened 2018-09-13
Initial commit for creator lock component and stories

benwerd:
This PR contains the saved and confirming states for the lock from the creator side.

There are a few known issues that will need to be added in a following PR:

* The edit and code icons are broken
* All icons are missing their grey background
* The icon bar on saved locks does not show/hide

I've also added TODOs for the exchange rate values of the lock, and have kept the styles for those values commented out for now.

### DIFF
--- a/unlock-app/src/components/creator/CreatorLock.js
+++ b/unlock-app/src/components/creator/CreatorLock.js
@@ -10,7 +10,6 @@ import Icons from '../interface/icons'
 export function CreatorLock({ lock, status = 'deployed' }) {
   // Some sanitization of strings to display
   let name = lock.name || 'New Lock'
-  let displayAddress = lock.address ? lock.address.substr(0, 16) : ''
   let outstandingKeys = lock.maxNumberOfKeys - lock.outstandingKeys || 0
 
   if (status === 'deployed') { // the transaction was mined and confirmed at least 12 times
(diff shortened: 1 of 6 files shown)


benwerd:
screen shot 2018-09-13 at 1 38 52 pm
screen shot 2018-09-13 at 1 38 00 pm

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -0,0 +1,188 @@
...
+  // Some sanitization of strings to display
+  let name = lock.name || 'New Lock'
+  let displayAddress = lock.address ? lock.address.substr(0, 16) : ''
+  let outstandingKeys = lock.maxNumberOfKeys - lock.outstandingKeys || 0
+
+  if (status === 'deployed') { // the transaction was mined and confirmed at least 12 times
+    // TODO add USD values to lock
+    // TODO add all-time balance to lock
+    return (
+      <CreatorLockRow>
```
julien51: Would it be worth splitting that in multiple files eventually? All importing the same elements from below?

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -0,0 +1,188 @@
...
+import Duration from '../helpers/Duration'
+import Balance from '../helpers/Balance'
+import Icon from '../lock/Icon'
+
+import Icons from '../interface/icons'
+
+export function CreatorLock({ lock, status = 'deployed' }) {
+  // Some sanitization of strings to display
+  let name = lock.name || 'New Lock'
+  let displayAddress = lock.address ? lock.address.substr(0, 16) : ''
```
julien51: Wouldn;t it be better to use CSS for this?

unlock-app/src/components/creator/CreatorLock.js
julien51: Also, not sure it is worth prefixing `Creator` on all of these components?

unlock-app/src/components/helpers/Balance.js
```diff
@@ -7,15 +7,19 @@ import Web3Utils from 'web3-utils'
  * @param {*} amount: the amount to convert to Eth
  * @param {string} unit: the unit of the amount to convert to Eth
  */
-export function Balance({ amount, unit = 'wei' }) {
-  const inWei = Web3Utils.toWei(amount || '0', unit)
-  const inEth = Web3Utils.fromWei(inWei, 'ether')
-  return (<span>Ξ {inEth}</span>)
+export function Balance({ amount, unit = 'wei', symbol=true }) {
```
julien51: Do you mind adding a test for this? (I think the Balance component has tests?)

unlock-app/src/components/interface/icons/Eth.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 
 const Eth = props => (
-  <svg viewBox="0 0 7 11" width="1em" height="1em" {...props}>
```
julien51: nope... these are automatically generated so we cannot change the values in there (I need to find a way to indicate that).
If you need the icons to be larger just change the `font-size` of the element in which the icons are displayed.

benwerd: Ah got it. OK!

unlock-app/src/components/helpers/Balance.js
benwerd: 👍

unlock-app/src/components/creator/CreatorLock.js
benwerd: I wanted to keep the namespace clear for user-facing components, as these are all creator-facing UI elements. But I could change some of the names?

benwerd:
Updated. New screenshots: 

screen shot 2018-09-13 at 1 54 41 pm
screen shot 2018-09-13 at 1 54 36 pm

unlock-app/src/components/creator/CreatorLock.js
benwerd: Could totally split this up into multiple files though.

benwerd: I'll do exactly that in a follow-up PR.

benwerd:
Okay. Tests and reformatting completed. I'll follow up with splitting up into smaller components. PTAL.

unlock-app/src/components/creator/CreatorLock.js
julien51: About the naming, I think one of the great thing about the styled components approach is that everything is scoped basically. Thesr components should but be reused.

PR #246, opened 2018-09-13
Splitting creator lock components into separate files

benwerd:


julien51:
I worry that we not reusing a lot of the components elements even though there mostly the same. I think we can parametrize at a lower level in the lock component?

benwerd:
That’s fair. It’s essentially the right hand side. Refactoring.

benwerd:
Revamped again to follow this advice. PTAL

unlock-app/src/components/creator/lock/CreatorLockSaved.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import styled from 'styled-components'
+import UnlockPropTypes from '../../../propTypes'
+import Icons from '../../interface/icons'
+
+export function CreatorLockSaved({ lock }) {
```
julien51: `export default` here to avoid the line below and use the import consistently (`import {CreatorLockSaved} from './lock/CreatorLockSaved' above)

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -97,28 +69,20 @@ const CreatorLockRow = styled.div`
   color: var(--slate);
   padding: 10px 0 10px 0;
   height: 64px;
-`
-
-const CreatorLockSaved = styled.div`
   display: grid;
-  grid-template-columns: 1fr 2fr repeat(4, 1fr) 2fr;
+  ${gridLayout}
```
julien51: So I think we can actually parametrize syled components which would allow you to include the logic from `gridLayout` in here.

julien51: see "Adapting based on props" on [this page](https://www.styled-components.com/docs/basics).

unlock-app/src/components/creator/lock/CreatorLockSaved.js
julien51: Actually, keep the 2 lines, I think this is better but fix the import the above for consistency please ;)

unlock-app/src/components/creator/CreatorLock.js
benwerd: :+1: adjusted to use this.

PR #247, opened 2018-09-18
Store a list of all processed transactions

benwerd:


julien51:
Yeah, React is picky on how it decides to re-render things... yet another reason to get rid of the nested stuff I built.

unlock-app/src/__tests__/reducers/accountReducer.test.js
```diff
@@ -10,7 +10,7 @@ describe('account reducer', () => {
   const balance = '1337'
 
   it('should return the initial state', () => {
-    expect(reducer(undefined, {})).toEqual({ 'locks': [], transaction: null })
+    expect(reducer(undefined, {})).toEqual({ 'locks': [], transactions: {all: {}, lastUpdated: 0, latest: null} })
```
julien51: I am asking but I have no idea if maybe it wouldn't be better to have multiple reducers, once for `transactions`, one for `lastUpdatedTransaction` and one for `latestTransaction`?

julien51: This would avoid nesting things?

unlock-app/src/components/consumer/Lock.js
```diff
@@ -74,7 +74,7 @@ const mapStateToProps = state => {
     currentKey: state.network.key, // key is a reserved props name
     account: state.network.account,
     lock: state.network.lock,
-    transaction: state.network.account && state.network.account.transaction,
+    transaction: state.network.account && state.network.account.transactions.latest,
```
julien51: I am worried that you'd need to check for `state.network.account.transactions` too?

unlock-app/src/components/creator/TransactionModal.js
```diff
@@ -35,12 +35,12 @@ export function TransactionModal({ hideTransactionModal, transaction }) {
 
 TransactionModal.propTypes = {
   hideTransactionModal: PropTypes.func,
-  transaction: UnlockPropTypes.account,
+  transaction: UnlockPropTypes.transaction,
```
julien51: 🙃

unlock-app/src/propTypes.js
```diff
@@ -18,6 +18,10 @@ export const lock = PropTypes.shape({
 
 export const transaction = PropTypes.shape({
   status: PropTypes.string,
+  confirmations: PropTypes.number,
+  createdAt: PropTypes.number,
+  hash: PropTypes.string,
+  lock: PropTypes.shape({}),
 })
```
julien51: Good!

unlock-app/src/reducers/transactionReducer.js
```diff
@@ -1,28 +1,29 @@
...
-const transactionReducer = (state = initialState, action) => {
+const transactionReducer = (transactions = initialState, action) => {
+  let newTransactions = Object.assign({}, transactions)
 
   if (action.type === SET_TRANSACTION) {
     if (action.transaction) {
-      return {
-        ...action.transaction,
-      }
+      newTransactions.latest = Object.assign({}, action.transaction)
```
julien51: Yikes. Is that because React would not refresh if you did not clone?

benwerd: Yep! I checked into it - that's expected behavior. You need to clone at all levels of nesting. Thanks, React.

unlock-app/src/__tests__/reducers/accountReducer.test.js
benwerd: That could well be - I'd like to use this as a starting point if that makes sense to you? Then we can work towards bringing more to the top level.

PR #250, opened 2018-09-18
Don't allow older transactions to hijack transactions.latest

benwerd:


benwerd:
Please check latest commit only :|

PR #251, opened 2018-09-19
Lock interface

hardlydifficult:
Here I'm proposing a design inspired by 0x as seen here: https://github.com/0xProject/0x-monorepo/blob/development/packages/contracts/src/2.0.0/protocol/Exchange/interfaces/IExchange.sol

Ultimately a Lock is also an ERC721, however interfaces cannot inherit and we do not want to copy paste.  So "ILock" is actually implemented as a contract, even though logically it is an interface.  This is working around a limitation in Solidity (the Solidity docs do say they may allow interface inheritance in the future).

All the methods for the interface are currently defined in ILockCore.  There are a couple alternatives, and I'm torn on these options ATM.

1) Remove ILockCore, using just ILock.  So we actually don't have any Lock 'interface', just the ILock contract which acts like an interface.
2) Add ILockOwner and put owner own methods in there instead of all together in ILockCore.  This approach seems to align with how 0x is designing interfaces, however it may be overkill.  This may be more clear as I better understand the different types of Locks which will be supported long term.


julien51:
I like the idea of "splitting" the various parts of a contract into sub contracts... but of course that is not always possible.

Types of lock that I can think of *today* (but of course there may be more eventually!):
* Public Lock
* Permission Lock: either thru a list of possible key owners or which delegates that "verification" to another smart contract. For example a Lock for the people who have a key to another Lock.
* Private Lock: no one can purchase key (sub case of the above lock I guess). Keys are then 'granted' by a Lock owner
* ERC20 Lock: the price of a key is not in Eth, but using an ERC20.
* Zero Knowledge Lock: a lock which does not hold an actual list of owners but has a mechanism for users to prove that they own a key, rather than be listed as an owner (not sure how purchasing the key would work tho!)

julien51:
One thing to note is that the current `ERC721.sol` is not exactly similar to the one from OpenZeppelin which we should probably use to ensure compatibility (See [this branch](https://github.com/unlock-protocol/unlock/tree/erc721Zep) of mine)

One thing that I was thinking is that we could actually use the actual implementation from OpenZeppelin (rather than inherit from it) [as explained in this document](https://docs.google.com/document/d/1mQemVkeipCgak7Mbt23Y-KQWxNIhf7i42mRgVhAnlsw/edit?usp=sharing) in the first paragraph of page 3. We would then store the extra fields (key specific) separate from the core ERC721 logic.

Does it make sense?

smart-contracts/contracts/interfaces/ERC721.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.4.24;
 * @title ERC721 interface (http://erc721.org/)
 */
 
-contract ERC721 {
+interface ERC721 {
```
julien51: It would probably make sense to check that this is the actual most recent interface used for ERC721... I believe the one I used is not 100% identical to what OpenZeppelin does.

smart-contracts/contracts/interfaces/ILock.sol
```diff
@@ -0,0 +1,20 @@
+pragma solidity 0.4.24;
+
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+import "./ERC721.sol";
+import "./ILockCore.sol";
+
+
+/**
+ * @title The Lock interface
+ * @author Julien Genestoux (unlock-protocol.com)
```
julien51: Feel free to replace these `@author` stuff with your name if you want (but keep `unlock-protocol.com`) or just `unlock-protocol.com`

hardlydifficult:
> Types of lock that I can think of _today_

This is helpful.  Looking at the types of locks here, it seems there will be no method which will be available for every Lock type.  So shall we rename ILock and ILockCore to IPublicLock and IPublicLockCore?  Then we can consider extracting a common interface which may be shared with another Lock type as we add more.  Or do you want to try another approach instead?

hardlydifficult:
> we could actually use the actual implementation from OpenZeppelin

Yes.  I believe I understand the challenge.  I could explore migrating to OpenZeppelin's ERC721 in another branch.  It looks like your branch failed with 'InternalCompilerError: Compiled contract not found.' - I'm unsure what may cause that ATM.

smart-contracts/contracts/interfaces/ERC721.sol
hardlydifficult: The `interface` vs `contract` change does align with the latest standard on http://erc721.org/ but OpenZeppelin uses a `contract` in order to compose from multiple interfaces.  I can roll this back, and then per above we then explore using OpenZeppelin directly instead of a copy paste.  If we need to diverge from their standard, we may be able to still use some of the interfaces/contracts from OpenZeppelin used to create their ERC721.

julien51:
I have seen that `InternalCompilerError: Compiled contract not found.` in the past. That happened when the contract implementing the interface was missing some functions from the interface or had a different signature (param types or return types) for some of them :/

hardlydifficult:
> I like the idea of "splitting" the various parts of a contract into sub contracts... but of course that is not always possible.

I wasn't sure how to chop up the interface.  I had owner methods separated previously, shall we go with this? https://github.com/unlock-protocol/unlock/pull/251/commits/f33afe6e681410f2aed934abc53a4e314865bf8e#diff-e95a8760be3eb6d9f79897795edeecd7 

Let me know if you have other suggestions on how this may divide up logically.  Otherwise we could just keep this in mind as things mature and maybe revisit later.

julien51:
I think we're on the right track! Lets proceed and revisit if needed later! 
Again, we could very well eventually also have PublicLock and PublicLockV1, PublickLockV2 ... etc

hardlydifficult:
I added a function for `updateKeyPrice`.  If this looks okay and the tests pass, I think we squash and merge this pull request.

julien51:
Perfect! Yes, squash and merge is the way to go ;) it makes reverts atomic just in case there is a problem with that PR.

PR #253, opened 2018-09-19
Adding new lock helper to provide accompanying transaction and confirmations

benwerd:
This PR also includes a minor change to the legacy Lock component to demonstrate how integration works, and for testing.

### DIFF
--- a/unlock-app/src/helpers/Locks.js
+++ b/unlock-app/src/helpers/Locks.js
@@ -1,0 +1,28 @@
+/**
+ * Given the address of a lock, returns its accompanying transaction
+ * @param transactionStore
+ * @param lockAddress
+ * @returns {object}
+ */
+export function getLockTransaction(transactionStore, lockAddress) {
+  if (transactionStore && transactionStore.all) {
+    for (let transaction in transactionStore.all) {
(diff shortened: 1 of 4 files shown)


benwerd:
screen shot 2018-09-19 at 11 47 51 am

This screenshot isn't meant for production (as it's on a legacy component), but more to illustrate how confirmations can be integrated.

unlock-app/src/helpers/Locks.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Given the address of a lock, returns its accompanying transaction
+ * @param transactionStore
+ * @param lockAddress
+ * @returns {object}
+ */
+export function getLockTransaction(transactionStore, lockAddress) {
+  if (transactionStore && transactionStore.all) {
+    for (let transaction in transactionStore.all) {
```
julien51: This may come back and bite us if there are 10000 transactions... but by then we'll have other problems ;)

benwerd: Yeah, I imagine that we'll adapt this over time if that becomes a problem. (I'm pretty sure we'll need a cache and a supporting database by then, for one.)

PR #254, opened 2018-09-19
adding way to parametrize Header and Layout component for content pages

julien51:


julien51:
screen shot 2018-09-19 at 15 39 36
screen shot 2018-09-19 at 15 39 31

julien51:
I am not fully happy with how we handle SVG (again!) so I'll work on that in another PR.

unlock-app/src/components/interface/Header.js
```diff
@@ -1,23 +1,29 @@
...
+      <Button><Icons.About /></Button>
+      <Button><Icons.Jobs /></Button>
+      <Button><Icons.Github /></Button>
+    </TopHeader>
+  )
 }
 
 Header.propTypes = {
   title: PropTypes.string,
+  forContent: PropTypes.bool,
```
benwerd: I'm wondering if this property should have a name that's more about what it does, vs what it's for - we might want to use this form of the header for pages other than content in the future? Your call, but wanted to call it out.

julien51: I actually tried to think about a good name that would explain what it does but since it may do multiple things I went for `forContent` because this will likely be only for content pages... Thanks!

PR #255, opened 2018-09-19
not using sized SVG components so that they're more flexible

julien51:


julien51:
So, this started as a change on how SVG are loaded so that we do not mess up with their sizes, but that obviously cascaded into a refactor of some lock stuff to make it cleaner with grids ;)

julien51:
screen shot 2018-09-19 at 18 11 56
screen shot 2018-09-19 at 18 11 49
screen shot 2018-09-19 at 18 11 42

julien51:
Sascha just provided me with cleaned SVG icons too, so I'll replace them in another PR...

julien51:
I do not understand what is going on here.

PR #274, opened 2018-09-25
Minor filename rename to let Storybook live again

benwerd:


benwerd:
This just renames `Index.js` in `pages` to `index.js`, which is what Storybook looks for.

PR #275, opened 2018-09-25
Adding create lock button

benwerd:


benwerd:
screen shot 2018-09-25 at 3 08 32 pm

unlock-app/src/components/creator/CreatorLocks.js
```diff
@@ -66,5 +70,25 @@ const LockMinorHeader = styled.div`
   color: var(--darkgrey);
 `
 
-const LockHeaderButton = styled.div`
+const LockHeaderButton = styled.div``
```
julien51: Do we need this component at all since it does not have any style info an just has the `LockHeaderButtonContainer` inside it?

unlock-app/src/components/creator/CreatorLocks.js
```diff
@@ -66,5 +70,25 @@ const LockMinorHeader = styled.div`
...
+
+const LockHeaderButtonContainer = styled.div`
+  max-height: 25px;
+  float: right;
```
julien51: Meh. I think we should be very careful not to use `float` with `grid` layout. You can achieve something similar with `align-self: end` if the parent container is a grid.

But there again I think we actually do not need that element, you could just use `ActionButton` with `align-self: end` since `LockHeaderRow` is a grid.

unlock-app/src/components/pages/Home.js
```diff
@@ -13,7 +14,7 @@ export default class Home extends PureComponent {
         </Headline>
 
         <Action>
-          <DashboardButton>Go to Your Dashboard</DashboardButton>
+          <ActionButton><HomepageButton>Go to Your Dashboard</HomepageButton></ActionButton>
```
julien51: Not sure we need 2 elements again?

benwerd:
Removed nested elements and the float. In the end I couldn't use `align-self: end` with a button, but I got around it with a `text-align` in the component cell. PTAL.

julien51:
Why couldn't you use `align-self` ?

unlock-app/src/components/creator/CreatorLocks.js
```diff
@@ -16,7 +16,9 @@ export default class CreatorLocks extends Component {
           <LockMinorHeader>Quantity</LockMinorHeader>
           <LockMinorHeader>Price</LockMinorHeader>
           <LockMinorHeader>Balance / Earnings</LockMinorHeader>
-          <LockHeaderButton>&nbsp;</LockHeaderButton>
+          <LockHeaderButtonContainer>
```
julien51: But why do you still have `LockHeaderButtonContainer` ?

julien51: ho I think this is what's causing you issues with `align-self`. Basically this would align the element in the parent... and since you have a element between the 2 that's causing you trouble ;)

unlock-app/src/components/pages/Home.js
```diff
@@ -76,28 +77,19 @@ const Action = styled.div`
...
 `
 
-const Paragraph =styled.p`
+const Paragraph = styled.p`
   font-family: 'IBM Plex Serif', serif;
   font-weight: 300;
   font-size: 20px;
 `
+
+const HomepageButton = styled(ActionButton)`
```
julien51: 👍

unlock-app/src/components/creator/CreatorLocks.js
```diff
@@ -68,23 +66,18 @@ const LockMinorHeader = styled.div`
...
   font-family: "IBM Plex Sans", sans-serif;
   border-radius: 4px;
-  grid-column: 2;
   justify-self: stretch;
   cursor: pointer;
 `
 
 const CreateButton = styled(ActionButton)`
   padding: 10px;
+  align-self: end;
```
julien51: 💯

PR #277, opened 2018-09-26
renamed Infinity into Lemniscate which is the real word for it and is not a reserved JS word

julien51:


benwerd:
I had to look up leminscate. Super cool.

julien51:
I had to look it up as well (or at least synonym of "Infinity Symbol". ∞

PR #278, opened 2018-09-26
Style tweaks for lock buttons

benwerd:
This PR shows buttons on hover and ensures they have a pointer cursor. The overall visual style is unchanged.

### DIFF
--- a/unlock-app/src/components/creator/CreatorLock.js
+++ b/unlock-app/src/components/creator/CreatorLock.js
@@ -14,7 +14,7 @@ export function CreatorLock({ lock, status = 'deployed' }) {
   let lockComponentStatusBlock
 
   if (status === 'deployed') { // the transaction was mined and confirmed at least 12 times
-    lockComponentStatusBlock = <LockIconBar lock={lock} className={'lock-icons'} />
+    lockComponentStatusBlock = <LockIconBarContainer><LockIconBar lock={lock} className={'lock-icons'} /></LockIconBarContainer>
   }
   if (status === 'confirming') { // the transaction was mined but hasn't yet been confirmed at least 12 times
     lockComponentStatusBlock = <CreatorLockConfirming lock={lock} />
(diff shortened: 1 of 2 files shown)


unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -14,7 +14,7 @@ export function CreatorLock({ lock, status = 'deployed' }) {
   let lockComponentStatusBlock
 
   if (status === 'deployed') { // the transaction was mined and confirmed at least 12 times
-    lockComponentStatusBlock = <LockIconBar lock={lock} />
+    lockComponentStatusBlock = <LockIconBar lock={lock} className={'lock-icons'} />
```
julien51: hum... I think we should not use `className` anymore with styled-components. (you'll soon hate me for being so picky).

The approach is to use   `${LockRow}:hover &` into the `IconBar` element. (Check what I did in `Button.js`)

benwerd: This approach doesn't work here, because of the way the components fit together, but I'm working on it ...

benwerd:
Removed explicit class names. PTAL

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -14,7 +14,7 @@ export function CreatorLock({ lock, status = 'deployed' }) {
   let lockComponentStatusBlock
 
   if (status === 'deployed') { // the transaction was mined and confirmed at least 12 times
-    lockComponentStatusBlock = <LockIconBar lock={lock} />
+    lockComponentStatusBlock = <LockIconBarContainer><LockIconBar lock={lock} className={'lock-icons'} /></LockIconBarContainer>
```
benwerd: Pre-emptive: I had to use a container here because you can't reference a real React component directly from a style.

julien51: Hooo... too bad! LG

PR #279, opened 2018-09-26
moving account out of the network state

julien51:


julien51:
Cool! This is a great idea...

PR #286, opened 2018-09-28
adding the demo page for each lock. We'll need to actually load the paywall next

julien51:


julien51:
Hum. something is wrong here with a commit which does not belong here. Cleaning up and opening a new one ;)

PR #288, opened 2018-09-28
updating handling of icons so that all styles are now managed with CSS

julien51:


unlock-app/src/components/interface/buttons/Button.js
```diff
@@ -3,7 +3,13 @@ import React from 'react'
 import styled from 'styled-components'
 
 export const LayoutButton = ({href, title, children}) => (
-  <Button href={href}>
+  <Button
+    href={href}
+    backgroundColor={'var(--grey)'}
```
benwerd: I love this change and use of vars - but I wonder whether this could be done via styled component?

julien51: What do you mean? This is actually done with styled components? We basically pass this as props so that we can re-use most of the `Button` ?

PR #295, opened 2018-10-01
adding FatalError Component

julien51:


julien51:
screen shot 2018-10-01 at 15 14 05
screen shot 2018-10-01 at 15 14 00

Issue #311, opened 2018-10-08
"Refreshing" data from localStorage


julien51:
Since we do not have a server side component for now, we use the blockchain as the source of truth and store (cache!) data using localStorage.
We use a redux middleware to make sure that our redux store is always 'synced' to localStorage.

The challenge is that data might have changed in the blockchain while the app is not running on the front end, and we need a mechanism to "sync" the redux store again after the web application has loaded.

julien51:
Another approach is to use our `connect` method in the `web3Service` to trigger the right actions once we have connected to right web3 provider.

This is probably more robust because we may not always load data from localstorage and yet we'd need to sync the locks again.

PR #312, opened 2018-10-09
Refreshing transactions when the app is loaded

julien51:


julien51:
This should help fix https://github.com/unlock-protocol/unlock/issues/311

PR #313, opened 2018-10-09
number of confirmations is now a setting

julien51:


julien51:
In dev, we do not need to "wait" for confirmations... because we're only mining on demand. It is actually 'working' right now because of a [bug in web3](https://github.com/ethereum/web3.js/issues/1239) which does trigger confirmations while there are none!

PR #315, opened 2018-10-09
locks are now stored as an object, indexed by their addresses

julien51:


benwerd:
This will help! I’m still wondering about self-saving objects that encapsulate their own schemas, but that might be the PHP ORM designer in me.

julien51:
I need to write about the redux state store structure!

PR #319, opened 2018-10-10
Save new lock on form submission

benwerd:
Please note: this PR doesn't yet include saving the name. That will come next.

Also, please take a look at the previous PR first - this incorporates those checkins. Here, consider the last two checkins only.


benwerd:
Form now submits and creates lock. PTAL

unlock-app/src/__tests__/utils/duration.test.js
```diff
@@ -1,4 +1,4 @@
-import {durationsAsTextFromSeconds, durations, secondsAsDays} from '../../utils/durations'
+import { durationsAsTextFromSeconds, durations, secondsAsDays } from '../../utils/durations'
```
julien51: nit! I think we use a space after brackets everywhere else? Lets maybe add a eslint rule about that.

julien51:
Yay! In the future please remember about screenshots if you can ;)

PR #320, opened 2018-10-11
updated create react app so that we have latest versions of dependencies when we eject

julien51:


julien51:
ok, that's messy. will retry later!

PR #336, opened 2018-10-12
Lock form disappears on submission and cancelation

benwerd:


benwerd:
screen shot 2018-10-12 at 8 34 52 am

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -44,6 +45,11 @@ class CreatorLockForm extends React.Component {
     if (this.props.hideAction) this.props.hideAction()
   }
 
+  handleCancel(e) {
+    e.stopPropagation()
```
benwerd: Pre-emptive: we need this line to prevent the submit method from _also_ being called.

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -44,6 +45,11 @@ class CreatorLockForm extends React.Component {
     if (this.props.hideAction) this.props.hideAction()
   }
 
+  handleCancel(e) {
+    e.stopPropagation() // This prevents submit from also being called
```
julien51: I think we actually want `preventDefault` because it's less restrictive. [See this](https://stackoverflow.com/questions/5963669/whats-the-difference-between-event-stoppropagation-and-event-preventdefault).

PR #339, opened 2018-10-12
Save and display custom names with locks

benwerd:


benwerd:
screen shot 2018-10-12 at 9 19 20 am

unlock-app/src/__tests__/helpers/Locks.test.js
```diff
@@ -144,4 +144,25 @@ describe('lockHelpers', () => {
...
+
+  it('should return the name of a lock from its transaction', () => {
+    const Lock = {
+      address: '0x1234567890',
+    }
+    const transactions = {
+      all: {
+        theTransactionWereLookingFor: {
+          hash: 'theTransactionWereLookingFor',
+          confirmations: config.requiredConfirmations-1,
```
julien51: Could you just use `config.requiredConfirmations` here so that we do not test for 2 things at once?

unlock-app/src/__tests__/helpers/Locks.test.js
```diff
@@ -1,4 +1,4 @@
-import {getLockTransaction, getLockConfirmations, getLockStatusString} from '../../helpers/Locks'
+import {getLockTransaction, getLockConfirmations, getLockStatusString, getLockName} from '../../helpers/Locks'
```
julien51: Why do you use the helper for that? Why not just get that value from the `lock` object?

unlock-app/src/helpers/Locks.js
```diff
@@ -44,3 +44,14 @@ export function getLockStatusString(transactionStore, lockAddress) {
...
+
+/**
+ * Returns a lock's user-facing name from its transaction
+ * @param transactionStore
+ * @param lockAddress
+ * @returns {string|null}
+ */
+export function getLockName(transactionStore, lockAddress) {
+  let transaction = getLockTransaction(transactionStore, lockAddress)
+  if (transaction.lockName) return transaction.lockName
```
julien51: hum... this is very strange, why would the transaction *know* about the lock's name?

unlock-app/src/services/web3Service.js
```diff
@@ -184,6 +184,7 @@ export default class Web3Service {
         } else if (event === 'NewLock') {
           return this.getLock(args.newLockAddress).then((lock) => {
             transaction.lock = lock
+            if (newLock.lockName) transaction.lockName = newLock.lockName
```
julien51: I would not attach the lock's name to the transaction but to the lock only.

benwerd:
Storing name on the lock itself. PTAL

unlock-app/src/services/web3Service.js
```diff
@@ -183,6 +183,7 @@ export default class Web3Service {
           callback(transaction)
         } else if (event === 'NewLock') {
           return this.getLock(args.newLockAddress).then((lock) => {
+            lock.lockName = newLock.lockName
```
julien51: Maybe add a comment here to indicate that since the lock's name is not stored on chain we need to pass if from the arugments provided...

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -22,7 +22,7 @@ class CreatorLockForm extends React.Component {
       keyPrice: '0.01',
       keyPriceCurrency: 'ether',
       maxNumberOfKeys: 10,
-      name: 'New Lock',
+      lockName: 'New Lock',
```
julien51: Could we keep name here? Since we're already in the `LockForm` I feel like calling it lockName is redundant. Also for consistency since no other attribute starts with `lock`...

benwerd: OK!

unlock-app/src/__tests__/helpers/Locks.test.js
```diff
@@ -1,10 +1,8 @@
-import {getLockTransaction, getLockConfirmations, getLockStatusString} from '../../helpers/Locks'
+import { getLockTransaction, getLockConfirmations, getLockStatusString } from '../../helpers/Locks'
 import configure from '../../config'
 
 const config = configure(global)
 
-describe('lockHelpers', () => {
```
julien51: Why remove that?

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -44,6 +44,7 @@ export function CreatorLock({ lock, status = 'deployed' }) {
 CreatorLock.propTypes = {
   lock: UnlockPropTypes.lock,
   status: UnlockPropTypes.status,
+  lockName: UnlockPropTypes.name,
```
julien51: eslint should complain about that ;)

julien51: Since it is not used.

unlock-app/src/__tests__/helpers/Locks.test.js
benwerd: Idiocy during a merge. Putting it back!

PR #341, opened 2018-10-12
removed eslint because it is part of react-scripts and tests will fail if there are 2 distinct versions running

julien51:


unlock-app/src/config.js
```diff
@@ -66,8 +66,8 @@ export default function configure(environment) {
     // In dev, the network can be anything above 100
     isRequiredNetwork = (networkId) => networkId > 100
 
-    // In dev, we only require 0 confirmation because we only mine when there are pending transactions
-    requiredConfirmations = 0
+    // In dev, we only require 6 confirmation because we only mine when there are pending transactions
+    requiredConfirmations = 6
```
julien51: since we're now doing real mining we can actually wait for a few blocks before confirming a lock has been deployed ;)

PR #346, opened 2018-10-12
Reverting to using e.stopPropagation

benwerd:
This seems to work and documentation I’ve found elsewhere recommends using it.


julien51:
oh :( so sorry about that... LG

benwerd:
It’s all good! JavaScript is still, frankly, completely arcane.

PR #353, opened 2018-10-15
locks are now indexed with a unique local id

julien51:


julien51:
A bunch of important changes in that PR.
First of all, each lock now has a unique id. The reason why we do that is because the address itself cannot be used: indeed there is an address only when a lock has been deployed.... which can take up to a couple minutes. So we need something else, and a unique id is a great approach.

Secondly, the locks themselves are then indexed in the redux store using that lock id.

Rather than waiting for the lock to be deployed by the service and instantiated only when that happened, we now pass the lock object to the service itself and the service will update that object based on what is happening in the service.

Finally, the lock is not passed "fully" to the transaction, only its id, which is one of the principle around designing our redux store: avoiding deep nested objects!

PR #368, opened 2018-10-16
not using the SVG icon in balance because it means styling is nearly impossible to remain consistent between the actual value and the unit

julien51:


benwerd:
STRONGLY AGREE

benwerd:
Can you add a screenshot? The eth symbol is a little weird in some typefaces.

julien51:
Oups!
screen shot 2018-10-16 at 15 10 02
screen shot 2018-10-16 at 15 09 58
screen shot 2018-10-16 at 15 09 54

Issue #325, opened 2018-10-11
Show truncated balances


julien51:
The balance on accounts is showing too many numbers. We need to only show relevant decimals.
If the user has more than 1Eth, let's show only 2 decimals.
If the user has less than 0.0001 Eth let's show ` < 0.0001` (about 2cts at current prices).
For anything in between, let's show at most 2 non zero values.

hasparus:
I could give it a shot tomorrow.

julien51:
As a follow up, this component can and will be reused in other places of the applications. Here are 2 places:
<img width="358" alt="screen shot 2018-10-16 at 14 21 17" src="https://user-images.githubusercontent.com/17735/47038166-cd7d5f00-d14e-11e8-9dc2-af8fb304e37f.png">
<img width="139" alt="screen shot 2018-10-16 at 14 21 12" src="https://user-images.githubusercontent.com/17735/47038167-cd7d5f00-d14e-11e8-8b2b-19dae45fcac1.png">

julien51:
Merged! Thanks @akeem !

PR #379, opened 2018-10-21
Truncates ETH Balances presented on front end

akeem:
As part of making the frontend more approachable, the following presentation logic has been applied where balances are currently being presented:

* Users with less than 0.0001 Eth will now see "< 0.0001"
* Users with greater than 0.0001 Eth but less than 1 ETh will now see their balance including at most two non-zero digits.
* User with > 1 Eth will see their balance truncated to two decimals.

 #325


julien51:
Thanks @akeem ! Feel free to merge ;)

akeem:
@julien51 it looks like someone with write permission will need to merge this, would you be able to merge this in?

julien51:
hoooo, so sorry about that! you're right, merging it now!

PR #387, opened 2018-10-23
adding the ability to export staticly generated site

julien51:


julien51:
Please ignore the #netlify stuff @benwerd ... but basically this PR allows us to deploy automatically from the master branch, which is quite cool!

julien51:
And this means we now need to be serious about what we put in there ;)

PR #402, opened 2018-10-24
Fixing providerReducer

julien51:


julien51:
Wait, I am not 100% sure that it was it. We shall see.

PR #404, opened 2018-10-24
when building for prod we look for the CONTEXT env variable. If set to production we consider that we are building for production

julien51:


julien51:
See this https://www.netlify.com/docs/continuous-deployment/#build-environment-variables for how netlify passes env variables!

PR #406, opened 2018-10-24
Ensuring fonts are properly loaded in server-rendered pages

### LINKED ISSUE #400: Flash of unstyled content
There is a flash of unstyled content on every reload. This is probably because the CSS are loaded *after* some of the HTML has already been rendered. We need to fix this.

benwerd:
Fixes #400

I moved the font import statements to a shareable constant and ensured they were loaded from the server side too. That's all it was - styles are being loaded properly (tested with JS off, etc).

### DIFF
--- a/unlock-app/src/pages/_document.js
+++ b/unlock-app/src/pages/_document.js
@@ -16,6 +16,11 @@ export default class MyDocument extends Document {
         <Head>
           <meta name='viewport' content='width=device-width, initial-scale=1' />
           <link rel='shortcut icon' href='/static/favicon.ico' />
+          <style>
+            @import url(https://fonts.googleapis.com/css?family=IBM+Plex+Sans:300,400,500,600,700);
+            @import url(https://fonts.googleapis.com/css?family=IBM+Plex+Mono:200i,200,500);
+            @import url(https://fonts.googleapis.com/css?family=IBM+Plex+Serif:300,400);
+          </style>
           {this.props.styleTags}
         </Head>
         <body>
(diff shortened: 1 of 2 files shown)


benwerd:
Note that I also removed apostrophes from the `@import` statements, because next.js was happily escaping them, thereby ensuring that the URLs were invalid.

julien51:
Hum.... I am not sure this is the right. The rest of the `gloablStyle` s also important to load, not just the fonts :/ For example since the css variables are in there and would not be set until JS is actually running that would still mean we have jank when some elements use *any* style defined globally?

benwerd:
(I've also checked to make sure no files are double-loaded here. The purpose of including the imports on both front end and back end is to ensure all styles are properly loaded, eg in storybook.)

benwerd:
@julien51 The rest of the global style _is_ loaded. And I've just realized I can probably simplify this. Basically only the fonts weren't being loaded on the server side.

benwerd:
Stand by while I test what could be the dumbest thing I've ever tested ...

julien51:
Buy they were loaded, right, just *after* the rest of the CSS while they should have been loaded before, right?

This means that the rest of the globalStyle rules are _also_ loaded after when they should be loaded _before_ ? I guess I am not sure why you distinguish between the fonts and the rest?

benwerd:
@julien51 indeed. See latest - I'm now adding all of the global stylesheet.

benwerd:
It is dirty, and although it's okay right now because the styles are absolute, it's not great. Theoretically the styles could just be on the server-side page, but I wanted to ensure they were okay for storybook. There's likely another way to do that just for storybook, which I'll look at in a future PR.

PR #425, opened 2018-10-27
linking transaction to locks and keys and keeping of hashes

julien51:


unlock-app/src/components/lock/Lock.js
```diff
@@ -51,20 +38,43 @@ export const Lock = ({ lock, account, keys, purchaseKey }) => {
...
   purchaseKey: (key) => dispatch(purchaseKey(key)),
 })
 
-const mapStateToProps = state => {
+export const mapStateToProps = (state, {lock}) => {
+  const account = state.network.account
+  const keys = state.keys
+
+  // If there is no account (probably not loaded yet), we do not want to create a key
+  if(!account) {
```
benwerd: World's tiniest nit: there should be a space here, right?

julien51: you are right!

PR #438, opened 2018-10-29
Refactor PublicLock.sol to be strictly Public

nfurfaro:
Eventually, there will be Private and Restricted locks. However, code
pertaining to these other types doesn't belong here.
As allowing a lock owner to change the keyReleaseMechanism of a lock is
not likely to be a common use-case, the keyReleaseMechanism has been
removed from Unlock.sol, PublicLock.sol and the lock's constructor.

## Proposed Changes

  - refactor/remove non-public modifiers
  - remove `keyReleaseMechanism`
  - refactor some tests to remove references to `keyReleaseMechanism` and to move `private` or `restricted` test cases to their own files. The new test files are named the same, "Non_Public_" prepended. There are 3 of them:
- Non_Public_approve.js
- Non_Public_transferFrom.js
- Non_Public_purchaseFor.js
The tests in these files are currently ignored with `.skip`.


smart-contracts/test/Lock/erc721/getApproved.js
```diff
@@ -24,15 +24,15 @@ contract('Lock ERC721', (accounts) => {
...
-    it('should return the address of the approved owner for a key', () => {
-      return locks['FIRST'].approve(accounts[3], accounts[3], {
-        from: lockOwner
-      }).then(() => {
-        return locks['FIRST'].getApproved(accounts[3])
-      }).then((approved) => {
-        assert.equal(accounts[3], approved)
-      })
-    })
+    // it('should return the address of the approved owner for a key', () => {
```
julien51: could you use `it.skip` rather than commenting out the whole test? This means that it will stay on our radar more easily ;)

nfurfaro: Just seeing this now. Sorry I missed this one. Will get it fixed up first thing in the morning and re-submit.

julien51: Awesome! Thanks!

PR #439, opened 2018-10-29
Display titles on pages

### LINKED ISSUE #437: Title tags not functioning on public site
To reproduce, visit any page on unlock-protocol.com. Notice that the title does not display in the browser title bar.

benwerd:
Fixes #437

  - Add title tags to every page
  - Test title tags to ensure they don't disappear again


unlock-app/src/pages/about.js
```diff
@@ -1,10 +1,14 @@
 import styled from 'styled-components'
 import React from 'react'
+import Head from 'next/head'
 import Layout from '../components/interface/Layout'
 import { Section, Title, Headline, SubTitle, CallToAction, ThreeColumns, Column, TwoColumns } from '../components/Components'
 
 export const About = () => (
   <Layout forContent={true}>
+    <Head>
+      <title>About | Unlock: The Web&#39;s new business model</title>
```
julien51: Thinking we should probably templatize this in constants.js? something like 
```
export pageTitle = (title) => `${title} | Unlock: The Web&#39;s new business model` 
```
and then use that in all pages like

```
<title>{pageTitle('About')}</title>
```

This would make things DRY-er and easier to change when we start tinkering with SEO or i18n?

benwerd: 👍

Issue #39, opened 2018-04-25
Stricter Lint rules


julien51:
The current lint rules are fairly limited. We need a set of much stricter rules.

kphed:
Have you considered enforcing JS coding practices based on an existing style guide? For example, AirBnB's is widely used for good reason and it's regularly maintained and kept up-to-date: https://github.com/airbnb/javascript

julien51:
Hey! Thanks for following up on this issue. Right now we use eslint with a [few defaults](https://github.com/unlock-protocol/unlock/blob/master/unlock-app/.eslintrc.js) but I think we need to tighten things up a little bit. I considered using Airbnb but I am not a semicolon person :/ 
Do you think we could just use AirBnb and tweak a couple rules? Is that good practice?

kphed:
>Hey! Thanks for following up on this issue. Right now we use eslint with a few defaults but I think we need to tighten things up a little bit. I considered using Airbnb but I am not a semicolon person :/
Do you think we could just use AirBnb and tweak a couple rules? Is that good practice?

You could definitely tweak the rules to suit your coding preferences, Julien. It's a good base to start from though.

benwerd:
Worth checking to see if there's a rule about `import` order. It's a bit all over the place right now - and I think it'd be a good thing to enforce if we can. I'll take a look after my current batch of PRs too.

julien51:
^ that PR does have a rule for import orders ;)

PR #440, opened 2018-10-29
Making our eslint setup stricter... for now a bunch of rules are disabled to avoid doing a monster PR but we'll slow renable them!

julien51:


julien51:
This will fix https://github.com/unlock-protocol/unlock/issues/39

julien51:
lol. 🤞

PR #453, opened 2018-10-31
disabling no-empty-blocks on ILockPublic

julien51:
Fixes master which is currently broken because of that solhint rule


julien51:
cc @nfurfaro 
Could you please review this :)?
I cannot formally add you as a reviewer until you accepted the invite I sent you.

PR #454, opened 2018-10-31
Issue245

### LINKED ISSUE #245: add USD values to lock
https://github.com/unlock-protocol/unlock/blob/4eb93e28cdde85ad3c9e88523572bd3008e0b769/unlock-app/src/components/creator/CreatorLock.js#L16-L21

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 4eb93e28cdde85ad3c9e88523572bd3008e0b769 when #244 was merged. cc @unlock-protocol.

cellog:
Fixes #245 

This implements an automatic currency conversion from ether to US Dollars based on the current market rate.

## Proposed Changes

  - store the ether->USD conversion rate in the redux store (new action, reducer)
  - use a new simple middleware to retrieve the current conversion rate every 10 seconds (this can be done outside middleware, but it makes it easier to find where side effects occur if we group them together)
  - encapsulate all display logic inside the `Balance` component. This allows us to pass in the input to display in the CreatorLockForm instead of just a static value
  - As a result, CreatorLockForm updates the USD value as you type the ether value
  - All places Balance is used, the conversion happens, on the Dashboard as well as for each individual lock
  - full tests and storybook stories for all changes are implemented.
  - Balance display is fancier for dollars, and supports abbreviating as the number increases, from normal "0.03" "123.03" to thousands "3,456" 100 thousands "234k" millions "1.2m" and billion+ "123.4b" This can (and should) be modified as desired, as it introduces issues of locale and translation. The current solution is American English-centric
 
### Caveats:

  - Although the design of the reducer plans for future conversion to a user's preferred currency, this would require rewriting. I chose simplicity over flexibility. Rewriting will be relatively straightforward
  - The lock form could choose to implement a kind of toggle that allows users to enter the USD value and auto-calculate the ether value instead. This would involve a button that swaps the input so that we enter the dollar amount and calculate the ether amount in gDSFP and would be relatively straightforward to implement.
  - The CSS is rather inelegant, perhaps there is a better way to do this?
  - a new dependency on axios is introduced (and as a consequence, jest-axios-mock). If Unlock will limit dashboard users to browsers that support fetch, the dependency can be removed
  - I really don't like `BalancePresenter` now, it does two things and looks odd. I recommend splitting off into a separate file and removing the clever use of switch/case and instead using if/else
  - testing via enzyme is a bad idea, enzyme has had a lot of trouble keeping pace with changes in React. A better choice is react-testing-library. In my experience converting the react-redux tests from enzyme to rtl, it was only about an hour's work, and they have many more tests than Unlock does currently. With this switch, the tests for Balance could become more fine-grained quite easily, so we can test the eth and dollar amounts separately instead of in that mish-mash you see now.
  - naming: I am not sure the names I have chosen are ideal with both `currency` and `conversion` used. Perhaps you have an opinion on whether they should be more unified into some kind of naming convention?

P.S. happy halloween!

### DIFF
--- a/unlock-app/package.json
+++ b/unlock-app/package.json
@@ -7,6 +7,7 @@
     "@storybook/addon-jest": "^4.0.0",
     "@storybook/react": "4.0.0",
     "@svgr/cli": "^3.1.0",
+    "axios": "^0.18.0",
     "babel-eslint": "^10.0.1",
     "babel-jest": "^23.6.0",
     "babel-loader": "^8.0.4",
@@ -91,5 +92,8 @@
     "not dead",
     "not ie <= 11",
     "not op_mini all"
-  ]
+  ],
+  "devDependencies": {
+    "jest-mock-axios": "^2.1.11"
+  }
 }
--- a/unlock-app/src/__tests__/components/creator/CreatorAccount.test.js
+++ b/unlock-app/src/__tests__/components/creator/CreatorAccount.test.js
@@ -12,8 +12,12 @@ describe('CreatorAccount', () => {
     const network = {
       name: 4,
     }
+    const conversion = {
+      USD: 195.99
+    }
 
-    const wrapper = shallow(<CreatorAccount account={account} network={network} />)
+    const wrapper = shallow(<CreatorAccount account={account} network={network} conversion={conversion} />)
     expect(wrapper.find('Balance').first().props().amount).toEqual('17.73')
+    expect(wrapper.find('Balance').first().props().conversion).toEqual({ USD: 195.99 })
   })
 })
--- a/unlock-app/src/__tests__/components/helpers/Balance.test.js
+++ b/unlock-app/src/__tests__/components/helpers/Balance.test.js
@@ -9,13 +9,22 @@ describe('Balance Component', () => {
   describe('when the balance is 0 Eth', () => {
     const amount = '0'
 
-    const wrapper = render(<Balance
-      amount={amount}
-      unit={unit}
-    />)
+    it('no conversion data available', () => {
+      const wrapper = render(<Balance
+        amount={amount}
+        unit={unit}
+        conversion={{ USD: undefined }}
+      />)
 
-    it('shows the value of 三 0', () => {
-      expect(wrapper.text()).toEqual('三 0')
+      expect(wrapper.text()).toEqual('0---')
+    })
+    it('USD conversion data available', () => {
+      const wrapper = render(<Balance
+        amount={amount}
+        unit={unit}
+        conversion={{ USD: 195.99 }}
+      />)
+      expect(wrapper.text()).toEqual('00')
     })
   })
 
@@ -25,10 +34,11 @@ describe('Balance Component', () => {
     const wrapper = render(<Balance
       amount={amount}
       unit={unit}
+      conversion={{ USD: 195.99 }}
     />)
 
     it('shows the default minimum value of 三 < 0.0001', () => {
-      expect(wrapper.text()).toEqual('三 < 0.0001')
+      expect(wrapper.text()).toEqual('< 0.00010.014')
     })
   })
 
@@ -38,10 +48,11 @@ describe('Balance Component', () => {
     const wrapper = render(<Balance
       amount={amount}
       unit={unit}
+      conversion={{ USD: 195.99 }}
     />)
 
     it('shows the balance in Eth to two decimal places', () => {
-      expect(wrapper.text()).toEqual('三 0.076')
+      expect(wrapper.text()).toEqual('0.07614.86')
     })
   })
 
@@ -51,10 +62,11 @@ describe('Balance Component', () => {
     const wrapper = render(<Balance
       amount={amount}
       unit={unit}
+      conversion={{ USD: 195.99 }}
     />)
 
     it('shows the balance in Eth to two decimal places', () => {
-      expect(wrapper.text()).toEqual('三 2.00')
+      expect(wrapper.text()).toEqual('2.00391.98')
     })
   })
 })
--- a/unlock-app/src/__tests__/components/helpers/Balance.test.js
+++ b/unlock-app/src/__tests__/components/helpers/Balance.test.js
@@ -5,6 +5,7 @@ import { Balance } from '../../../components/helpers/Balance'
 
 describe('Balance Component', () => {
   const unit = 'szabo'
+  const conversion = { USD: 195.99 }
 
   describe('when the balance is 0 Eth', () => {
     const amount = '0'
@@ -22,7 +23,7 @@ describe('Balance Component', () => {
       const wrapper = render(<Balance
         amount={amount}
         unit={unit}
-        conversion={{ USD: 195.99 }}
+        conversion={conversion}
       />)
       expect(wrapper.text()).toEqual('00')
     })
@@ -34,7 +35,7 @@ describe('Balance Component', () => {
     const wrapper = render(<Balance
       amount={amount}
       unit={unit}
-      conversion={{ USD: 195.99 }}
+      conversion={conversion}
     />)
 
     it('shows the default minimum value of 三 < 0.0001', () => {
@@ -48,7 +49,7 @@ describe('Balance Component', () => {
     const wrapper = render(<Balance
       amount={amount}
       unit={unit}
-      conversion={{ USD: 195.99 }}
+      conversion={conversion}
     />)
 
     it('shows the balance in Eth to two decimal places', () => {
@@ -62,11 +63,81 @@ describe('Balance Component', () => {
     const wrapper = render(<Balance
       amount={amount}
       unit={unit}
-      conversion={{ USD: 195.99 }}
+      conversion={conversion}
     />)
 
     it('shows the balance in Eth to two decimal places', () => {
       expect(wrapper.text()).toEqual('2.00391.98')
+    })
+  })
+
+  describe('when the balance converts to > $1000 ', () => {
+    const amount = '20000000'
+
+    const wrapper = render(<Balance
+      amount={amount}
+      unit={unit}
+      conversion={conversion}
+    />)
+
+    it('shows the balance in dollars in locale format without decimal', () => {
+      expect(wrapper.text()).toEqual('20.003,920')
+    })
+  })
+
+  describe('when the balance converts to > $100k ', () => {
+    const amount = '2000000000'
+
+    const wrapper = render(<Balance
+      amount={amount}
+      unit={unit}
+      conversion={conversion}
+    />)
+
+    it('shows the balance in thousands of dollars postfixed with k', () => {
+      expect(wrapper.text()).toEqual('2000.00392k')
+    })
+  })
+
+  describe('when the balance converts to > $1m ', () => {
+    const amount = '20000000000'
+
+    const wrapper = render(<Balance
+      amount={amount}
+      unit={unit}
+      conversion={conversion}
+    />)
+
+    it('shows the balance in millions of dollars postfixed with m', () => {
+      expect(wrapper.text()).toEqual('20000.003.9m')
+    })
+  })
+
+  describe('when the balance converts to > $1b ', () => {
+    const amount = '20000000000000'
+
+    const wrapper = render(<Balance
+      amount={amount}
+      unit={unit}
+      conversion={conversion}
+    />)
+
+    it('shows the balance in billions of dollars postfixed with b', () => {
+      expect(wrapper.text()).toEqual('20000000.003.9b')
+    })
+  })
+
+  describe('when the balance converts to > $1b, unit is eth (used in lock form)', () => {
+    const amount = '9999999'
+
+    const wrapper = render(<Balance
+      amount={amount}
+      unit='eth'
+      conversion={conversion}
+    />)
+
+    it('shows the balance in billions of dollars postfixed with b', () => {
+      expect(wrapper.text()).toEqual('9999999.002b')
     })
   })
 })
--- a/unlock-app/src/components/creator/CreatorLock.js
+++ b/unlock-app/src/components/creator/CreatorLock.js
@@ -15,6 +15,7 @@ export class CreatorLock extends React.Component {
     super(props)
     this.state = {
       showEmbedCode: false,
+      usdConvertedPrice: 1
     }
     this.toggleEmbedCode = this.toggleEmbedCode.bind(this)
   }
@@ -29,7 +30,7 @@ export class CreatorLock extends React.Component {
     // TODO add USD values to lock
     // TODO add all-time balance to lock
 
-    const { lock, transaction, config } = this.props
+    const { lock, transaction, config, conversion } = this.props
 
     // Some sanitization of strings to display
     let name = lock.name || 'New Lock'
@@ -67,8 +68,8 @@ export class CreatorLock extends React.Component {
 /
           {lock.maxNumberOfKeys}
         </LockKeys>
-        <Balance amount={lock.keyPrice} />
-        <Balance amount={lock.balance} />
+        <Balance amount={lock.keyPrice} conversion={conversion}/>
+        <Balance amount={lock.balance} conversion={conversion} />
         {lockComponentStatusBlock}
         {status === 'deployed' && this.state.showEmbedCode &&
           <LockCode>
@@ -85,6 +86,7 @@ CreatorLock.propTypes = {
   lock: UnlockPropTypes.lock,
   transaction: UnlockPropTypes.transaction,
   config: UnlockPropTypes.configuration,
+  conversion: UnlockPropTypes.conversion
 }
 
 const mapStateToProps = (state, { lock }) => {
@@ -92,6 +94,7 @@ const mapStateToProps = (state, { lock }) => {
   return {
     transaction,
     lock,
+    conversion: state.currency
   }
 }
 
@@ -161,13 +164,5 @@ const LockValueSub = styled.div`
   font-size: 0.6em;
   color: var(--grey);
   margin-top: 5px;
-`
-*/
-
-/* Saving for use with sub-values that need to be added in a future PR
-const LockValueUsd = styled.div`
-  &:before {
-    content: "$ ";
-  }
 `
 */
--- a/unlock-app/src/components/helpers/Balance.js
+++ b/unlock-app/src/components/helpers/Balance.js
@@ -8,9 +8,14 @@ import Web3Utils from 'web3-utils'
  * @param {*} amount: the amount to convert to Eth
  * @param {string} unit: the unit of the amount to convert to Eth
  */
-export function Balance({ amount, unit = 'wei', conversion = { USD: undefined } }) {
-  const inWei = Web3Utils.toWei(amount || '0', unit)
-  const inEth = Web3Utils.fromWei(inWei, 'ether')
+export function Balance({ amount, unit = 'wei', conversion = { USD: undefined }, EthComponent = ({ value }) => value }) {
+  let inEth
+  if (unit === 'wei') {
+    const inWei = Web3Utils.toWei(amount || '0', unit)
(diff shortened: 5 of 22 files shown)


julien51:
Wow! Thanks a lot @cellog ! 
I will review all this very soon, but as you might have seen, I merged yesterday a PR which had issues and caused our master to break. Please bear with us!

cc @asood123

cellog:
@julien51 no worries, tend to your collective heartburn first :)

A few additional notes I didn't put in my original comment:

## Balance component

The primary cleverness I wrote was making the Balance component usable in both a `CreatorLock` and in the `CreatorLockForm`. This is accomplished by replacing the simple display of eth value:

```js
        <BalanceWithUnit>
          {ethWithPresentation}
        </BalanceWithUnit>
```

with a component that is passed in as a prop:

```js
        <BalanceWithUnit>
          <EthComponent value={ethWithPresentation} />
        </BalanceWithUnit>
```

The default value for EthComponent is simply a component that returns the text:

```js
export function Balance({ amount, unit = 'wei', conversion = { USD: undefined }, EthComponent = ({ value }) => value }) {
```

Doing it this way allows us to pass in the form input field instead:

```js
    this.ethPrice = ({ value }) => (
      <FormBalanceWithUnit>
        <input type="text" id="keyPrice" onChange={this.handleChange} defaultValue={value} />
      </FormBalanceWithUnit>
    )

...

        <Balance unit='eth' amount={this.state.keyPrice} conversion={conversion} EthComponent={this.ethPrice}/>
```

When coupled with a computed state value for the USD conversion rate:

```js
  static getDerivedStateFromProps(props, state) {
    if (!state.keyPrice.length || props.conversion.USD === 'undefined' || isNaN(+state.keyPrice)) {
      return {
        keyPriceUSD: '---',
      }
    }
    return {
      keyPriceUSD: +state.keyPrice * props.conversion.USD
    }
  }
```

Voila, we get instant calculation of the USD value of eth when creating a lock, updated every 10 seconds to keep up with the current exchange rate.

Also, a lock that is not being edited will update its USD value every 10 seconds (if the rate changes) so that this is always accurate to within 10 seconds.

Because I did it this way, we get the USD conversion anywhere Balance was used, in the Dashboard summary, lock creation form, and lock display, with minimal change to source. Most of the changes involve adding the currency conversion table to mapStateToProps and adding it as a prop to the dumb components. This could be refactored to change Balance to a smart component that retrieves the conversion table and reduce some code size. Up to you if you want me to do that prior to merge, I would recommend it, but I'm less familiar with the codebase.

## extracted logic naming and directory location

Currently, you have a directory `utils` which contains a hodgepodge of things, some are components for handling shared state, and some are transformers of input for display. I think this will quickly be hard to maintain (and hard to find things). I suggest that `duration.js` be moved to the new `selectors` directory, which is where I put the formatting helpers for currency display. I would rename `utils` to `stateManagers` or something along those lines.

That's all. Good luck with the integration testing!

cellog:
P.S. I just saw the linting errors. I will fix those.

Do you have any aversion to using prettier? This would enforce a coding style and make linting less necessary

julien51:
I have no aversion to using prettier ;) Just make the case for it and show us the path forward ;)
I'd say the first step is probably opening an issue so we can keep track of it and more people can chime in?

cellog:
P.P.S. After looking at linting, I realized that by abstracting `Balance` out, the gDSFP code I mentioned is no longer necessary, so it is getting stripped. The logic is simpler than I realized :). Commits forthcoming

julien51:
Greg: thanks a lot for your very quick work on this. 
Now onto the review ;)

First: I renamed the PR to something more obvious. It's good to link to the issue you're solving from inside the PR comments but the title should be more explicit.
A second thing, which I'll ask you to do on the "formal" aspect is to add a couple screenshot to your PR so that reviewers can easily see the change. I should have mentioned that earlier (but [just opened a PR](https://github.com/unlock-protocol/unlock/pull/463)). Sorry for this.

Please see comments inline:

>  - store the ether->USD conversion rate in the redux store (new action, reducer)
✅

>  - use a new simple middleware to retrieve the current conversion rate every 10 seconds (this can be done outside middleware, but it makes it easier to find where side effects occur if we group them together)
✅

>  - encapsulate all display logic inside the `Balance` component. This allows us to pass in the input to display in the CreatorLockForm instead of just a static value
✅

>  - As a result, CreatorLockForm updates the USD value as you type the ether value
✅

>  - All places Balance is used, the conversion happens, on the Dashboard as well as for each individual lock
✅

>  - full tests and storybook stories for all changes are implemented.
✅(minus screenshots ;p)

>  - Balance display is fancier for dollars, and supports abbreviating as the number increases, from normal "0.03" "123.03" to thousands "3,456" 100 thousands "234k" millions "1.2m" and billion+ "123.4b" This can (and should) be modified as desired, as it introduces issues of locale and translation. The current solution is American English-centric
 ✅ 💯for considering locales! I certainly hope we will soon have more languages supported. I have a pretty good idea about the first one 🇫🇷

>  - Although the design of the reducer plans for future conversion to a user's preferred currency, this would require rewriting. I chose simplicity over flexibility. Rewriting will be relatively straightforward
✅I think this is a good default for now.

>  - The lock form could choose to implement a kind of toggle that allows users to enter the USD value and auto-calculate the ether value instead. This would involve a button that swaps the input so that we enter the dollar amount and calculate the ether amount in gDSFP and would be relatively straightforward to implement.
✅Great idea for a future addition.

>  - The CSS is rather inelegant, perhaps there is a better way to do this?
✅ Ok I'll look into this and leave inline comments.

>  - a new dependency on axios is introduced (and as a consequence, jest-axios-mock). If Unlock will limit dashboard users to browsers that support fetch, the dependency can be removed
✅ Good point, I think at this point we use other "edgy" technologies like CSS grids which may actually have an even lower coverage than fetch (caniuse shows 87% for it)

>  - testing via enzyme is a bad idea, enzyme has had a lot of trouble keeping pace with changes in React. A better choice is react-testing-library. In my experience converting the react-redux tests from enzyme to rtl, it was only about an hour's work, and they have many more tests than Unlock does currently. With this switch, the tests for Balance could become more fine-grained quite easily, so we can test the eth and dollar amounts separately instead of in that mish-mash you see now.
✅ This is a very useful point/comment. Would you be ok to open an issue and expand on that? I picked enzyme as it seemed like a popular option which is probably not a good enough factor! Since change seems to be fairly easy right now, it's probably better to do it early.

>  - naming: I am not sure the names I have chosen are ideal with both `currency` and `conversion` used. Perhaps you have an opinion on whether they should be more unified into some kind of naming convention?
✅ It is a good question. I am looking at the code next and will share my opinion in there. Generally if things are really modular, names tend to be highly local which means they can be fairly generic if what they refer too in the module's context is obvious.

unlock-app/src/__tests__/components/helpers/Balance.test.js
```diff
@@ -5,17 +5,27 @@ import { Balance } from '../../../components/helpers/Balance'
...
-      expect(wrapper.text()).toEqual('三 0')
+      expect(wrapper.text()).toEqual('0---')
+    })
+    it('USD conversion data available', () => {
+      const wrapper = render(<Balance
+        amount={amount}
+        unit={unit}
+        conversion={conversion}
+      />)
+      expect(wrapper.text()).toEqual('00')
```
julien51: I wonder if there is no better way than testing the `text()` value here. This worked fine when using just Eth but here, this is probably not good enough.

unlock-app/src/middlewares/currencyConversionMiddleware.js
```diff
@@ -0,0 +1,19 @@
+import axios from 'axios'
+
```
julien51: I think here would be a good place to indicate that we could use `Fetch` if there was a larger support so that we can reconsider this in the future when support for Fetch is large enough.

Another option would also to be using a Polyfill since we already use [Babel](https://babeljs.io/docs/en/babel-polyfill) which could replace `Fetch` when it does not exist?

unlock-app/src/components/helpers/Balance.js
```diff
@@ -3,50 +3,84 @@ import React from 'react'
...
-    return parseFloat(numericalEth.toPrecision(SIGNIFICANT_DIGITS))
-  default:
-    return numericalEth.toFixed(DECIMAL_PLACES)
+export function Balance({ amount, unit = 'wei', conversion = { USD: undefined }, EthComponent = ({ value }) => value }) {
+  let inEth
+  if (unit !== 'dollars' && unit !== 'eth') {
+    const inWei = Web3Utils.toWei(amount || '0', unit)
+    inEth = Web3Utils.fromWei(inWei, 'ether')
+  } else {
+    inEth = +amount
```
julien51: Isn't that happening either when unit is `dollars` or `eth`?

unlock-app/src/components/helpers/Balance.js
```diff
@@ -3,50 +3,84 @@ import React from 'react'
...
 
+export const BalanceWithConversion = styled.div`
+  display: flex;
+  flex-direction: column;
+`
+
+export const Currency = styled.span`
+  display: flex;
+  flex-direction: row;
+
```
julien51: Nit: remove this whitespace please :)

unlock-app/src/__tests__/components/creator/CreatorAccount.test.js
```diff
@@ -12,8 +12,12 @@ describe('CreatorAccount', () => {
     const network = {
       name: 4,
     }
+    const conversion = {
+      USD: 195.99,
+    }
 
-    const wrapper = shallow(<CreatorAccount account={account} network={network} />)
+    const wrapper = shallow(<CreatorAccount account={account} network={network} conversion={conversion} />)
```
julien51: I am not sure it makes a lot of sense to pass that `conversion` props here. It feels like it should probably only be loaded inside the `Balance` component?

unlock-app/src/__tests__/components/helpers/Balance.test.js
```diff
@@ -25,10 +35,11 @@ describe('Balance Component', () => {
     const wrapper = render(<Balance
       amount={amount}
       unit={unit}
+      conversion={conversion}
     />)
 
     it('shows the default minimum value of 三 < 0.0001', () => {
-      expect(wrapper.text()).toEqual('三 < 0.0001')
+      expect(wrapper.text()).toEqual('< 0.00010.014')
```
julien51: Sam here as above. Maybe this time instead of checking for the text value we check for the actual markup? (and I start to see where your comments about Enzyme are coming from!)

unlock-app/src/__tests__/components/helpers/Balance.test.js
```diff
@@ -51,10 +63,81 @@ describe('Balance Component', () => {
...
+    const amount = '9999999'
+
+    const wrapper = render(<Balance
+      amount={amount}
+      unit='eth'
+      conversion={conversion}
+    />)
+
+    it('shows the balance in billions of dollars postfixed with b', () => {
+      expect(wrapper.text()).toEqual('9999999.002b')
```
julien51: 💰

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -163,11 +164,3 @@ const LockValueSub = styled.div`
   margin-top: 5px;
 `
 */
-
```
julien51: Thanks for the cleanup ;)

unlock-app/src/stories/creator/CreatorLock.stories.js
```diff
@@ -46,7 +49,7 @@ storiesOf('CreatorLock', CreatorLock)
       confirmations: 2,
     }
     return (
-      <CreatorLock lock={lock} transaction={transaction} />
+      <CreatorLock lock={lock} transaction={transaction}  />
```
julien51: NIT: white space here

unlock-app/package.json
```diff
@@ -91,5 +92,8 @@
     "not dead",
     "not ie <= 11",
     "not op_mini all"
-  ]
+  ],
+  "devDependencies": {
```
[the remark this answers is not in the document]
cellog: I don't think this is wholly accurate. This article explains the reasoning: https://silvenon.com/blog/dependencies-or-devdependencies

Basically, jest, jest-mock-axios, and all the testing things are not needed in order to actually build unlock-app or to execute its code, so they should go into devDependencies. If you feel strongly after reading this, I'll happily move it. I suspect that the only disadvantage will be build time, since webpack prunes dead code now, but that could be significant as the app gets larger.

unlock-app/src/__tests__/components/creator/CreatorAccount.test.js
cellog: We think alike: I had previously said:

> Because I did it this way, we get the USD conversion anywhere Balance was used, in the Dashboard summary, lock creation form, and lock display, with minimal change to source. Most of the changes involve adding the currency conversion table to mapStateToProps and adding it as a prop to the dumb components. This could be refactored to change Balance to a smart component that retrieves the conversion table and reduce some code size. Up to you if you want me to do that prior to merge, I would recommend it, but I'm less familiar with the codebase.

:)

I'll make it so.

unlock-app/src/__tests__/components/helpers/Balance.test.js
cellog: this is part of why I recommend using rtl for testing, it is easier to wiggle through the DOM, since all the search facilities are literally querySelector-based, but I can see if enzyme has a better way.

This brings up another issue with testing. When using styled-components, how do you search for a component? It may be possible to search for the component itself using `find()` I'll try that out. In other projects, I have preferred using css modules, because all the CSS is built statically, and so it's much easier to search for a component. However, this is a conversation for a different issue. :) For now, I'll find a solution that works with enzyme

cellog:
screenshots added. will finish up code changes later tonight.

cellog:
just realized `Lock` also uses Balance, and looks horrible with the changes I made. I'll fix soon.

unlock-app/src/__tests__/components/helpers/Balance.test.js
cellog: The most recent commit resolves this issue. Unfortunately, enzyme was incapable of resolving it (believe me, I tried, spent about 20 minutes with the manual, testing different things, and nada). After switching the test to react-testing-library, I had it resolved in less than 10 seconds. I was forced to move both the Balance and CreatorAccount tests over to rtl in order to test the new behavior.

unlock-app/src/middlewares/currencyConversionMiddleware.js
cellog: added a TODO comment here

unlock-app/src/components/helpers/Balance.js
cellog: amount is usually a string, formatEth expects a number

unlock-app/src/components/helpers/Balance.js
cellog: done

unlock-app/src/__tests__/components/creator/CreatorAccount.test.js
cellog: done

unlock-app/src/__tests__/components/helpers/Balance.test.js
cellog: fixed

cellog:
just rebased against master so automatic merge is possible

unlock-app/src/__tests__/components/helpers/Balance.test.js
julien51: hooo sorry about that :( 
I just opened #467 for this.

unlock-app/src/middlewares/currencyConversionMiddleware.js
julien51: Thanks!

unlock-app/package.json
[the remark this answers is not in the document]
julien51: This is great feedback. I opened an issue so we start thinking about migrating all other `dependencies` which are not required to build to `devDependencies `

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -35,10 +35,9 @@ export class CreatorLock extends React.Component {
   }
 
   render() {
-    // TODO add USD values to lock
     // TODO add all-time balance to lock
 
-    const { lock, transaction, config } = this.props
+    const { lock, transaction, config, conversion } = this.props
```
julien51: Can you get rid of `conversion` here since it is not used at all?

unlock-app/src/components/helpers/Balance.js
julien51: oh yes of course, but I meant that `inEth` is probably a poor choice for a value which could be actually dollars?

unlock-app/src/pages/dashboard.js
```diff
@@ -9,7 +9,7 @@ import CreatorLocks from '../components/creator/CreatorLocks'
 import withConfig from '../utils/withConfig'
 import { pageTitle } from '../constants'
 
-export const Dashboard = ({account, network, locks}) => {
+export const Dashboard = ({ account, network, locks, conversion }) => {
```
julien51: Please cleanup `conversion` from this file too ;)

unlock-app/src/components/helpers/Balance.js
```diff
@@ -2,54 +2,91 @@ import styled from 'styled-components'
...
+  }
+`
+
 export const BalanceWithUnit = styled.span`
   white-space: nowrap;
   text-transform: uppercase;
 `
 
-export default Balance
+export default connect(state => ({ conversion: state.currency }))(Balance)
```
julien51: Nit: maybe use a function (`mapStateToProps`) for this for clarity?

unlock-app/src/components/creator/CreatorLock.js
cellog: done

unlock-app/src/pages/dashboard.js
cellog: done

unlock-app/src/components/helpers/Balance.js
cellog: done

unlock-app/src/components/creator/CreatorLock.js
cellog: 👍

unlock-app/src/components/helpers/Balance.js
cellog: I see. I renamed it to currency, hopefully that resolves this :)

cellog:
updated screenshots, these are clearer

unlock-app/src/stories/creator/Dashboard.stories.js
```diff
@@ -49,6 +52,6 @@ storiesOf('Dashboard', Dashboard)
       },
     }
     return (
-      <Dashboard network={network} account={account} transactions={transactions} locks={locks} />
+      <Dashboard network={network} account={account} transactions={transactions} locks={locks} conversion={conversion} />
```
julien51:
```suggestion
      <Dashboard network={network} account={account} transactions={transactions} locks={locks} />
```

unlock-app/src/stories/creator/Dashboard.stories.js
```diff
@@ -15,6 +15,9 @@ storiesOf('Dashboard', Dashboard)
     const network = {
       name: 4,
     }
+    const conversion = {
```
julien51: We need to remove this :(

cellog:
strange, I think my rebase must have reverted some of the changes I made. No matter, I'm going to make those changes and then carefully look over the changed files and see if anything else escaped!

cellog:
ok, ready to go from my end

PR #456, opened 2018-10-31
Fixing console errors on lock creation

benwerd:
This seems to remove most of the discrepancies, although I think there's more to come.

### DIFF
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -16,12 +16,12 @@ class CreatorLockForm extends React.Component {
   constructor (props, context) {
     super(props)
     this.state = {
-      keyReleaseMechanism: 0, // Public
-      expirationDuration: 30,
-      expirationDurationUnit: 86400, // Days
+      keyReleaseMechanism: '0', // Public
+      expirationDuration: '30',
+      expirationDurationUnit: '86400', // Days
       keyPrice: '0.01',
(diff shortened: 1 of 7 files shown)


unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -16,12 +16,12 @@ class CreatorLockForm extends React.Component {
   constructor (props, context) {
     super(props)
     this.state = {
-      keyReleaseMechanism: 0, // Public
-      expirationDuration: 30,
-      expirationDurationUnit: 86400, // Days
+      keyReleaseMechanism: '0', // Public
```
julien51: Nah, I don't think this is a problem and I actually think we need to keep the initial types here :(

unlock-app/src/components/creator/lock/EmbedCodeSnippet.js
```diff
@@ -10,9 +10,9 @@ export function EmbedCodeSnippet({ lock }) {
 <script src="https://unlock-protocol.com/unlock.js"></script>
 
 <!-- Lock elements by wrapping them in this div -->
-<div unlock-lock="${lock.address}">
+<div unlock-lock="${lock.transaction}">
```
julien51: This is also probably not the right thing. The lock.transaction value is actually the hash of the transaction, I do not think we can replace this.

benwerd:
I'm going to close this and hope the upcoming reversion helps. I have a little less hair today ...

PR #457, opened 2018-10-31
Revert "Refactor PublicLock.sol to be strictly Public (#446)"

julien51:
Fixes master by reverting "Refactor PublicLock.sol to be strictly Public (#446)"


julien51:
@nfurfaro I think we(I) merged this PR too fast. This changed a bunch of things (obviously!) on the web app itself and until we fix these I am reverting this PR. Sorry about that!

nfurfaro:
No worries... Sorry I broke unlock!
Have you ever used the gitflow workflow? It's pretty sweet, and merges features to develop first. Might help catch these sort of things moving forward?

julien51:
Please tell more? I am actually ok with breaking master ;) It's not prod' yet!

nfurfaro:
So, Gitflow is both a strict branching strategy and a set of tools to make the strategy easy to use. It can be used strictly on the command line, but I've also used it in Gitkrakken (GUI).
The biggest difference is the addition of a "develop" branch which becomes the default for all new work. Master only ever contains production-ready code. 
I really like it, and find it quite simple and intuitive. The "develop" branch can be a good place to bring features together and make sure they play nicely together before releasing them.
That said, using gitflow wouldn't eliminate the need for integration-testing, and probably has as many critics as it does supporters.

julien51:
Thanks for this! 
Right now, I am actually ok having master being broken because we do not actually run prod on master directly. We have a staging env that is being deployed from master and then we can manually trigger the publication into the prod env.

PR #460, opened 2018-11-01
fixing how state is loaded

julien51:


julien51:
I manually tested this since we do not have better tests for this yet but I think this works as expected!

unlock-app/src/utils/withReduxStore.js
```diff
@@ -40,11 +40,13 @@ export default (App) => {
       }
     }
 
+    /**
+     * initialReduxState is actually coming from the server side rendered content
+     * We want to "merge" it with the data that might be stored in localStorage
+     */
```
benwerd: 👍 👍 👍

unlock-app/src/createUnlockStore.js
```diff
@@ -15,7 +14,7 @@ import lockMiddleware from './middlewares/lockMiddleware'
 
 const config = configure(global)
 
-export default function createUnlockStore(defaultState = loadState()) {
```
julien51: This did not work because `defaultState` was indeed defined and passed as the "brand new" default state as loaded in the HTML via server side rendering.

unlock-app/src/services/localStorageService.js
```diff
@@ -11,11 +13,10 @@ export const loadState = () => {
 }
 
 export const saveState = (state) => {
-
   try {
     const serializedState = JSON.stringify(state)
     localStorage.setItem('state', serializedState)
   } catch (err) {
-    // TODO log/warn user
+    console.error(err)
```
julien51: This would not have helped us here but would help in other cases.

unlock-app/src/utils/withReduxStore.js
```diff
@@ -40,11 +40,13 @@ export default (App) => {
...
     constructor (props) {
       super(props)
-
-      // For some reason this argument is being ignored... we probably need to use it!
-      this.reduxStore = getOrCreateStore(props.initialReduxState)
+      this.reduxStore = getOrCreateStore(Object.assign(props.initialReduxState, loadState()))
```
julien51: I like the symmetry of having loadState() and getState() be in the same code block!

PR #462, opened 2018-11-01
List keys on lock

benwerd:
- Adds new KeyList component
  - Displays component as a pane on a CreatorLock when the lock is deployed and keys exist for the lock
  - Adds a new expirationAsDate utility function
  - Tests and stories for the above

### DIFF
--- a/unlock-app/src/components/creator/CreatorLock.js
+++ b/unlock-app/src/components/creator/CreatorLock.js
@@ -30,7 +30,7 @@ export class CreatorLock extends React.Component {
     // TODO add USD values to lock
     // TODO add all-time balance to lock
 
-    const { lock, transaction, keys, config } = this.props
+    const { lock, transaction, config } = this.props
 
     // Some sanitization of strings to display
     let name = lock.name || 'New Lock'
@@ -39,23 +39,19 @@ export class CreatorLock extends React.Component {
       <LockIconBar lock={lock} toggleCode={this.toggleEmbedCode} />
     </LockIconBarContainer>)
 
-    let status
     if (!transaction) {
       // We assume that the lock has been succeesfuly deployed?
       // TODO if the transaction is missing we should try to look it up from the lock address
     } else if (transaction.status === 'submitted') {
-      status = 'submitted'
       lockComponentStatusBlock = <CreatorLockStatus lock={lock} status="Submitted" />
     } else if (transaction.status === 'mined' &&
         transaction.confirmations < config.requiredConfirmations) {
-      status = 'confirming'
       lockComponentStatusBlock = <CreatorLockStatus
         lock={lock}
         status="Confirming"
         confirmations={transaction.confirmations}
       />
-    } else if (transaction.status === 'mined' &&
-      transaction.confirmations >= config.requiredConfirmations) status = 'deployed'
+    }
 
     return (
       <LockRow>
@@ -75,16 +71,16 @@ export class CreatorLock extends React.Component {
         <Balance amount={lock.keyPrice} />
         <Balance amount={lock.balance} />
         {lockComponentStatusBlock}
-        {status === 'deployed' && this.state.showEmbedCode &&
+        {status === this.state.showEmbedCode &&
           <LockPanel>
             <LockDivider />
             <EmbedCodeSnippet lock={lock} />
           </LockPanel>
         }
-        {!this.state.showEmbedCode && status === 'deployed' &&
+        {!this.state.showEmbedCode &&
           <LockPanel>
             <LockDivider />
-            <KeyList keys={keys} lock={lock} />
+            <KeyList lock={lock} />
           </LockPanel>
         }
       </LockRow>
@@ -104,7 +100,6 @@ const mapStateToProps = (state, { lock }) => {
   const keys = state.keys
   return {
     transaction,
-    keys,
     lock,
   }
 }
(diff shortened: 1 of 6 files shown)


benwerd:
screen shot 2018-11-01 at 11 47 06 am

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -38,19 +39,23 @@ export class CreatorLock extends React.Component {
       <LockIconBar lock={lock} toggleCode={this.toggleEmbedCode} />
     </LockIconBarContainer>)
 
+    let status = 'Not found'
```
julien51: hum. What is this? It looks like if there is no transaction you would mark the lock as not found when this should not be the case. We can have cases where the app would now know about the transaction fromn which this lock is from and still the lock would be "found" ;)

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -71,10 +76,16 @@ export class CreatorLock extends React.Component {
         <Balance amount={lock.balance} />
         {lockComponentStatusBlock}
         {status === 'deployed' && this.state.showEmbedCode &&
-          <LockCode>
+          <LockPanel>
```
julien51: I actually think it is ok to show an empty list when the lock has not been deployed?

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -84,13 +95,17 @@ export class CreatorLock extends React.Component {
...
   lock: UnlockPropTypes.lock,
   transaction: UnlockPropTypes.transaction,
+  keys: UnlockPropTypes.keys,
   config: UnlockPropTypes.configuration,
 }
 
 const mapStateToProps = (state, { lock }) => {
   const transaction = state.transactions[lock.transaction]
+  const keys = {}
+  if (state.keys) Object.values(state.keys).filter((key) => key.lockAddress === lock.address).forEach((item) => keys[item.id] = item)
```
julien51: Why not move that logic down to the `KeyList` component?

unlock-app/src/components/creator/lock/KeyList.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react'
+import styled from 'styled-components'
+import UnlockPropTypes from '../../../propTypes'
+import { expirationAsDate } from '../../../utils/durations'
+
+// TODO add pagination
+export function KeyList ({ keys }) {
+  return (
+    <KeyTable>
```
julien51: You and the prefixes ;) I think there is no need to prefix everything with `Key` here because all of this is only local to that file ;p

benwerd:
Changed based on comments (subcomponent name changes, logic fixes, removing the not found label). PTAL

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -84,13 +95,16 @@ export class CreatorLock extends React.Component {
 CreatorLock.propTypes = {
   lock: UnlockPropTypes.lock,
   transaction: UnlockPropTypes.transaction,
+  keys: UnlockPropTypes.keys,
   config: UnlockPropTypes.configuration,
 }
 
 const mapStateToProps = (state, { lock }) => {
   const transaction = state.transactions[lock.transaction]
+  const keys = state.keys
```
julien51: hum... why not extract the keys from state inside the component which actually needs them?

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -29,7 +30,7 @@ export class CreatorLock extends React.Component {
     // TODO add USD values to lock
     // TODO add all-time balance to lock
 
-    const { lock, transaction, config } = this.props
+    const { lock, transaction, keys, config } = this.props
```
julien51: I see you still pass keys here but they're not needed here. I was thinking you could extract them in the `KeyList` component altogether!

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -38,19 +39,23 @@ export class CreatorLock extends React.Component {
       <LockIconBar lock={lock} toggleCode={this.toggleEmbedCode} />
     </LockIconBarContainer>)
 
+    let status
```
julien51: hum. why do you need `status`? Maybe get rid of it since it is not being used anymore?

unlock-app/src/utils/durations.js
```diff
@@ -63,3 +65,18 @@ export function durationsAsTextFromSeconds(seconds) {
 export function secondsAsDays(seconds) {
   return Math.ceil(seconds / 86400).toString()
 }
+
+/**
+ * Given an epoch timestamp, returns a string of the form 'Month Day, Year' (eg, 'Dec 31, 1980')
+ * @param timestamp
+ * @returns {string}
+ */
+export function expirationAsDate(timestamp) {
```
julien51: Could you add a TODO to make this a little bit smarter to include hours/minutes when the lock is about to expire?

benwerd: I went ahead and just implemented this.

benwerd:
* Keys are now extracted from state in KeyList
* That status variable is removed (although I still think there's room to have a consistent way to determine state, which could maybe be a utility function)
* Improved the expiration date function

PTAL.

PR #478, opened 2018-11-05
using an explicit mnemonic to make sure that the accounts stay consistent

julien51:


julien51:
This is useful for integration tests ;)

PR #479, opened 2018-11-05
using name instead of id in forms

julien51:


julien51:
This will make it easier to write functional tests and is more inline with good web practices ;)

PR #480, opened 2018-11-05
Add Deployment.md and edit .gitignore

nfurfaro:
Fixes #

## Proposed Changes

  - Added `build/` directory to `.gitignore`.
  - Added `zos.dev-*.json` and `.zos.session` to `.gitignore`.
  - Added Deployment.md file to document necessary steps after deployment.

While using zOS makes it simpler to track contract deployments by network, `build/` is still needed because it contains the ABI & bytecode (zos doesn't manage this on its own).

It should be noted that trying to use zos right now, for example `zos push`, results in the following message:
```zos version identifier not found in zos.json. This means the project was built with an older version of zos (1.x), and needs to be upgraded. Please refer to the documentation at https://docs.zeppelinos.org for more info.```
I'm not aware of any reason to not upgrade to 2.0, especially if zos was not used to deploy the existing contracts on the testnet. An upgrade of zOS is not currently part of this PR, but could be added.
Changes introduced can be seen here:
https://docs.zeppelinos.org/docs/new_2.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,10 @@
 .vscode
 tmp_site/
 node_modules/
+
+# ZeppelinOS
+zos.dev-*.json
+.zos.session
+
+#Truffle
+build/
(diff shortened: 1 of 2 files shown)


julien51:
Let me know if you do not have the rights to merge but otherwise: go ahead ;) 🍾

PR #501, opened 2018-11-08
Deploys humans.txt and robots.txt

### LINKED ISSUE #474: people.txt
We need to implement ways to invite people to contribute.
One of the ways us to use `people.txt` (on top of robots.txt). another one is to add a message in the JS console like @gitcoin does:

screen shot 2018-11-05 at 09 28 25

benwerd:
Fixes #474

There's no hard and fast way to do this in next.js. This should hopefully copy humans.txt and robots.txt to the root in production and staging environments.

  - Copy robots.txt and humans.txt as part of exportPathMap

### DIFF
--- a/unlock-app/next.config.js
+++ b/unlock-app/next.config.js
@@ -1,5 +1,18 @@
+const fs = require('fs')
+const { join } = require('path')
+const { promisify } = require('util')
+
+const copyFile = promisify(fs.copyFile)
+
 module.exports = {
-  exportPathMap: function () {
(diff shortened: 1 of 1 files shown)


julien51:
Hum... how does https://staging.unlock-protocol.com/static/favicon.ico work?

benwerd:
Next.js has a concept of a static folder - but robots.txt and humans.txt need to be accessed from the root. That was definitely my starting point!

Issue #506, opened 2018-11-09
Broken logo


julien51:
It looks like the logo is broken. See :
<img width="1166" alt="screen shot 2018-11-08 at 19 43 48" src="https://user-images.githubusercontent.com/17735/48236093-b36e1f80-e38e-11e8-8c34-5dd79fd3924e.png">

This is certainly due to this https://github.com/unlock-protocol/unlock/pull/503

The solution is to not treat all SVG in the same way.

PR #508, opened 2018-11-09
Quick fix for logo dimensions

benwerd:
Refs #506

  - Adjusts wordmark logo viewBox so full logo is on display

There are still some alignment issues to fix but this solves the most obvious optical problem for now.

### DIFF
--- a/unlock-app/src/components/interface/svg/UnlockWordMark.js
+++ b/unlock-app/src/components/interface/svg/UnlockWordMark.js
@@ -1,7 +1,7 @@
 import React from 'react'
 
 const SvgUnlockWordMark = props => (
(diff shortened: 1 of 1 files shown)


benwerd:
Never mind. There was an alignment issue. I've just removed the viewBox entirely for now.

benwerd:
screen shot 2018-11-08 at 5 36 41 pm

julien51:
That won't work because these files are generated from the SVG file so next time we trying the script to generate them, this will be dropped. I'll submit another solution tomorrow...

julien51:
I think you can close this one @benwerd ;)

benwerd:
Indeed.

PR #511, opened 2018-11-09
now refreshing the locks when the application restarts

julien51:


julien51:
I think eventually we'll have to rethink how redux actions are triggered.

PR #515, opened 2018-11-09
Showing correct number of keys on lock

benwerd:


julien51:
I wish we had tests for this :/

benwerd:
Added a test.

PR #516, opened 2018-11-09
Buttons now stop propagation by default

benwerd:
- Buttons now have an "action" property that takes an optional function as value
- Regardless of whether action is set, buttons will stop propagation
- href links, eg to the test page, still work


benwerd:
😅

PR #535, opened 2018-11-12
ensuring that we use lowercase when looking at the env variable values

julien51:


julien51:
Note: env vars are lowercase in node: https://github.com/nodejs/node/issues/9157

julien51:
👋 @benwerd

PR #539, opened 2018-11-13
Fixing layout button fill color

benwerd:
- Sets default fillColor to white instead of grey


benwerd:
screen shot 2018-11-12 at 4 10 58 pm

benwerd:
Whooops - left a giant code comment in. Will re-submit.

PR #541, opened 2018-11-13
not overwritting state with null values

julien51:


unlock-app/src/createUnlockStore.js
```diff
@@ -28,6 +28,10 @@ export default function createUnlockStore(defaultState) {
     error: errorReducer,
   }
 
+  // Cleanup the defaultState to remove all null values so that we do not overwrite existing
+  // values with null
+  Object.keys(defaultState).forEach((k) => (defaultState[k] == null) && delete defaultState[k])
```
benwerd: I wonder if there's a case for including an `if(defaultState)` here? Although your default value above should protect against this, there's still a chance of a `null` being passed?

julien51: Well this function actually filters out null values, so we never overwrite the default values with `null`.

PR #542, opened 2018-11-13
Fixing layout button fill color

benwerd:
Default fillColor is now white, not grey.


benwerd:
screen shot 2018-11-12 at 4 17 43 pm

PR #546, opened 2018-11-13
Replace Lock.json references with PublicLock.json

nfurfaro:
## Proposed Changes

There are some old references to `Lock.json` in some of the files in `unlock-app/src`. They need to be changed to `PublicLock.json` to reflect the new contract name. The old way still works because `Lock.json` still exists in the artifacts (`build/contracts`). 
This has been fixed in the following files:
- `./src/services/web3Service.js`
- `src/__tests__/services/web3Service.test.js`


julien51:
There must be something wrong here because none of the tests work :/

julien51:
hum, I think I see th problem: we need these files `PublicLock.json` and `Unlock.json` to exist... but since they're not in git, we need to "compile" the smart contracts before running the tests.

julien51:
Let's see what this yields.

nfurfaro:
I compiled all contracts, and ran git add -f  on the `PublicLock.json` file because the build dir is ignored by git. Committed  and pushed again, with a new build error:
```
$ git checkout -qf 979e08c479da253b0e54bd16a83203c3d548f241
fatal: reference is not a tree: 979e08c479da253b0e54bd16a83203c3d548f241
The command "git checkout -qf 979e08c479da253b0e54bd16a83203c3d548f241" failed and exited with 128 during .
```

julien51:
Ok, the good news is that tests are passing. The bad news is that this cannot be deployed because we are not compiling the the smart contracts when we deploy to netlify.

This I think confirms definitely that we *need* to have the artifacts in version control because we cannot expect to have truffle/solc available on the platforms which are just hosting the web app. (until we have a different build process)

So, for now, we *have to* go back to putting the artifacts into git.

nfurfaro:
Yes, this makes sense to me.

PR #556, opened 2018-11-14
refactored provider page to make it slightly better (this is a tool only useful for debugging)

julien51:


julien51:
screen shot 2018-11-13 at 20 19 14

Issue #529, opened 2018-11-12
Dashboard Withdrawing Funds from Lock


smombartz:
## Description
The different Lock states and messaging when withdrawing funds from a Lock.

Specs: https://zpl.io/brNeBPr

## Screenshot
<img width="1157" alt="dashboard withdrawing funds from lock" src="https://user-images.githubusercontent.com/1377389/48366376-b2353f00-e67b-11e8-917e-d95223fb50fa.png">

julien51:
@benwerd This is not working for me on master. When I click on the button, nothing happens.
We should also probably show an error message when the balance is 0 and/or disable the button?

benwerd:
Yeah, I’m still working on this. At the point where the buttons are rendered, most of lock is currently not populated (including the balance) and I’m having trouble telling the component to rerender it. Should have a fix in this morning.

benwerd:
The two stages of this that are not currently implemented are:

1. The "pending browser wallet" status. I can go looking, but I'm not sure how to reliably detect this (beyond maybe triggering it when the button is clicked, but this wouldn't detect cancels).
2. Withdrawal complete. I'm worried about the status sticking around long after the transaction is confirmed - but maybe we display it when the confirmations are above the minimum but below a defined threshold. I'll add this today.

julien51:
1. I looked at that for other transactions (local creatiomn, key purchase) and I found that it was a mess... so I skipped for other things. The fact that the browser wallet window will open is probably good enough?

2. Yes I think that would work. We should also soon have the list of transactions implemented which will be helpeful.

julien51:
@benwerd Do you mind providing an update on that front?

julien51:
@benwerd Can we close this?

benwerd:
I think we can close this. Some other related issues have been reported elsewhere and will be reported there.

PR #558, opened 2018-11-14
Withdraw balance from lock

benwerd:
Refs #529

## Proposed Changes

- Wires up "withdrawal" button to create a transaction that withdraws funds
- Adds withdrawal transaction to redux store so it can be traced
- Displays notice on lock while withdrawal is in progress, both for submission and confirmations
- Does not attempt to withdraw funds if balance on lock is 0

### DIFF
--- a/unlock-app/src/components/creator/CreatorLock.js
+++ b/unlock-app/src/components/creator/CreatorLock.js
@@ -121,7 +121,7 @@ CreatorLock.defaultProps = {
 
 const mapStateToProps = (state, { lock }) => {
   const transaction = state.transactions[lock.transaction]
-  const account = state.network.account
+  const account = state.account
   return {
     transaction,
     account,
--- a/unlock-app/src/components/creator/CreatorLock.js
+++ b/unlock-app/src/components/creator/CreatorLock.js
@@ -27,8 +27,9 @@ export class CreatorLock extends React.Component {
   }
 
   startWithdrawal() {
-    if (this.props.lock.balance > 0) {
-      this.props.withdraw(this.props.lock, this.props.account)
+    const { lock, account, withdraw } = this.props
+    if (lock.balance > 0) {
+      withdraw(lock, account)
     }
   }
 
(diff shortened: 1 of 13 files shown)


benwerd:
screen shot 2018-11-14 at 3 36 01 pm

screen shot 2018-11-14 at 3 36 18 pm

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -37,16 +48,43 @@ export class CreatorLock extends React.Component {
...
+    }
+
     // Some sanitization of strings to display
     let name = lock.name || 'New Lock'
     let outstandingKeys = lock.outstandingKeys || 0
     let lockComponentStatusBlock = (
-      <LockIconBarContainer>
-        <LockIconBar lock={lock} toggleCode={this.toggleEmbedCode} />
-      </LockIconBarContainer>)
+      <StatusBlock>
```
julien51: What do you think of moving all of the withdraw logic inside `LockIconBar`?

unlock-app/src/components/interface/buttons/lock/Withdraw.js
```diff
@@ -2,10 +2,12 @@ import React from 'react'
 import Svg from '../../svg'
 import { LockButton } from '../Button'
 
-const Withdraw = (props) => (
-  <LockButton title='Withdraw balance' {...props}>
-    <Svg.Withdraw name="Withdraw" />
-  </LockButton>
-)
+export function Withdraw(props) {
```
julien51: huh, why did you move back to a named function here? ES6 syntax is much cleaner?

unlock-app/src/pages/dashboard.js
```diff
@@ -30,13 +30,14 @@ export const Dashboard = ({account, network, locks}) => {
 Dashboard.displayName = 'Dashboard'
 
 Dashboard.propTypes = {
-  account: UnlockPropTypes.account.isRequired,
+  account: UnlockPropTypes.account, // We can't mark this as required because it's undefined when the dashboard is loading
```
julien51: Yeah, I think we need to fix that in another PR. I think we actually need to keep it required but not render if it is missing ;)

unlock-app/src/propTypes.js
```diff
@@ -38,7 +38,7 @@ export const mechanism = PropTypes.oneOf(['0', '1', '2', undefined])
 
 export const layout = PropTypes.instanceOf(Function) //PropTypes.instanceOf(React.Component)
 
-export const locks = PropTypes.arrayOf(PropTypes.object)
```
julien51: Actually,
Would this `export const locks = PropTypes.arrayOf(lock)` work? ( I think we define the type `lock`)

unlock-app/src/stories/creator/CreatorLock.stories.js
```diff
@@ -32,6 +42,7 @@ storiesOf('CreatorLock', CreatorLock)
   .addDecorator(getStory => <Provider store={store}>{getStory()}</Provider>)
   .add('Deployed', () => {
     const lock = {
+      id: '0xab',
```
julien51: This is a a giant nit but the id is not the address and starting with `0x` implies an eth address? Would you be ok to use another random id?

unlock-app/src/propTypes.js
benwerd: It wouldn't right now, because it's not an array - it's an object where the properties are lock IDs. We could shift that in another PR.

julien51: ohhh I see! no I think this is right it is better if it is an object whose keys are lock ids. sorry about that!
We can probably do something like that though:

```
PropTypes.objectOf(lock)
```

benwerd:
Moved the withdrawal logic, updated based on feedback. PTAL

julien51:
Wait, did you push your changes? I still see a lot of stuff in `CreatorLock`?

benwerd:
Moved significantly more logic out of `CreatorLock`.

julien51:
Ok you are going to hate me for the next 12 years, but I see a lot of things going on which I believe being in there own PR. Would you mind splitting things? Or would make reviewing much much easier :/

Also, I see you moved a lot of logic down into the lockIconBar (including the status stuff) which is great but then the name is probably not appropriate anymore? I was thinking you could even move the withdraw logic down to the withdraw button (buy that means you'll have to pass the transaction status one level up... Which is I'd say is ok but I'll let you decide).

Also it looks like it's missing tests?

Again, I understand I'm being picky here but I think it's for the best and should make things easier to review and improve in the future!

Thanks

benwerd:
Closing this PR as all component elements have now been submitted separately.

PR #563, opened 2018-11-17
Refactored lockMiddleware and web3Service to use events

julien51:


julien51:
This is a significant refactor which I believes brings us to a much cleaner place.
It was tedious to write all the tests but I believe we also now have a very good coverage for these.

benwerd:
Nice work! Reviewing this morning.

unlock-app/src/middlewares/lockMiddleware.js
```diff
@@ -3,123 +3,124 @@
...
 import { setError } from '../actions/error'
 import { SET_PROVIDER } from '../actions/provider'
-import { REFRESH_TRANSACTION, addTransaction, refreshTransaction, updateTransaction, deleteTransaction } from '../actions/transaction'
+import { addTransaction, updateTransaction } from '../actions/transaction'
 
 import Web3Service from '../services/web3Service'
 
 // This middleware listen to redux events and invokes the services APIs.
+// It also listen to events from web3Service and dispatches corresponding actions
+// TODO: consider if on events we should only trigger more actions instead of calling web3Service directly
```
benwerd: I'd be strongly in favor of this. That way we can hook into these actions and change the logic over time. Doesn't need to be a now thing though.

unlock-app/src/middlewares/lockMiddleware.js
```diff
@@ -3,123 +3,124 @@
...
+  /**
+   * When a key was saved, we refresh the balance of the lock
+   * and the balance of the account!
+   */
+  web3Service.on('key.saved', (key) => {
+    web3Service.getLock(getState().locks[key.lock])
+    web3Service.refreshOrGetAccount(getState().account)
+  })
+
+  web3Service.on('key.updated', (key) => {
```
benwerd: I think it'd be nice to have comments here as above for consistency. Spell it out.

unlock-app/src/services/web3Service.js
```diff
@@ -141,61 +154,50 @@ export default class Web3Service {
...
+      gas: 2000000,
+      contractAbi: UnlockContract.abi,
+    }, (error, { event, args } = {}) => {
+      if (error) {
+        this.emit('error', error)
+      }
+      if (event === 'transactionHash') {
+        transaction.hash = args.hash
+        transaction.status = 'submitted'
+        lock.transaction = transaction.hash
```
benwerd: I kind of wish there was a more syntactically elegant way to do this - maybe a helper function that applies the properties of a new object given as a parameter to an existing object?

julien51: Yeah, agreed. I think we need to think about  our "models" more!

PR #566, opened 2018-11-18
Refactored how <Balance /> works to use a provider and render props

julien51:


julien51:
This is a follow up to @cellog 's work on the Balance stuff.

julien51:
Yes! I think the new hotness in react-land is react hooks and they provide an even more elegant approach to that... but they have not landed yet in a stable react version.

PR #573, opened 2018-11-19
Improving alignment of lock status block

benwerd:
Minor improved display change.


benwerd:
screen shot 2018-11-19 at 12 12 04 pm

PR #576, opened 2018-11-19
Allow an infinite number of keys

nfurfaro:
Fixes #

## Proposed Changes

- allow a value of -1 to be passed to the constructor for `PublicLock.sol`, setting the value of `maxNumberOfKeys` to be 2^256-1, or infinite for all intents and purposes.
-
-


nfurfaro:
One more small fix coming

julien51:
Feel free to merge ;)

PR #581, opened 2018-11-20
Implementing multiple lock statuses for overlay

julien51:


julien51:
screen shot 2018-11-20 at 13 54 35
screen shot 2018-11-20 at 13 54 31
screen shot 2018-11-20 at 13 54 26
screen shot 2018-11-20 at 13 54 21

PR #587, opened 2018-11-21
avoid infinite loop to connect if we are trying to reconnect to the same provider

julien51:
Fixes infinite connect loop

### DIFF
--- a/unlock-app/src/__tests__/services/web3Service.test.js
+++ b/unlock-app/src/__tests__/services/web3Service.test.js
@@ -133,6 +133,26 @@ describe('Web3Service', () => {
       web3Service.connect(Object.assign({}, defaultState))
     })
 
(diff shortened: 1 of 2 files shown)


unlock-app/src/__tests__/services/web3Service.test.js
```diff
@@ -133,6 +133,26 @@ describe('Web3Service', () => {
       web3Service.connect(Object.assign({}, defaultState))
     })
 
+    it('should silently ignore requests to connect again to the same provider', done => {
```
benwerd: Why does this happen?

julien51: So, there is an infinite loop when the `connect` method` triggers an error since the `middleware` will intercept all events and while we're not connected, will issue a new `connect` call.

PR #588, opened 2018-11-21
Withdrawal logic is now attached to withdraw button

benwerd:
The button initiates the action, and this PR also adds a test for it. The final pieces of the interface will be added to this in the following PR.

This also includes an automatic prettify change.

### DIFF
--- a/unlock-app/src/__tests__/components/creator/CreatorLock.test.js
+++ b/unlock-app/src/__tests__/components/creator/CreatorLock.test.js
@@ -1,7 +1,6 @@
 import React from 'react'
 import * as rtl from 'react-testing-library'
 import { Provider } from 'react-redux'
-import EventEmitter from 'events'
 
 import { CreatorLock } from '../../../components/creator/CreatorLock'
 import configure from '../../../config'
@@ -9,25 +8,6 @@ import createUnlockStore from '../../../createUnlockStore'
 
 jest.mock('next/link', () => {
   return ({ children }) => children
-})
-
-/**
- * Mocking web3Service
- * Default objects yielded by promises
- */
-class MockWebService extends EventEmitter {
-  constructor() {
-    super()
-    this.ready = true
-  }
-}
-
-let mockWeb3Service = new MockWebService()
-
-jest.mock('../../../services/web3Service', () => {
-  return function() {
-    return mockWeb3Service
-  }
 })
 
 const lock = {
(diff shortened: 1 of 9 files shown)


unlock-app/src/__tests__/components/creator/CreatorLock.test.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import * as rtl from 'react-testing-library'
 import { Provider } from 'react-redux'
+import EventEmitter from 'events'
```
julien51: Do you need this since it looks like we actually do not need a full class?

unlock-app/src/__tests__/components/creator/CreatorLock.test.js
```diff
@@ -10,6 +11,25 @@ jest.mock('next/link', () => {
...
+ * Default objects yielded by promises
+ */
+class MockWebService extends EventEmitter {
+  constructor() {
+    super()
+    this.ready = true
+  }
+}
+
+let mockWeb3Service = new MockWebService()
```
julien51: You can use just a regular object? Also I think we should actually not "integrate" this test so deeply. (see comments below)

```suggestion
let mockWeb3Service = {} 
```

unlock-app/src/__tests__/components/creator/CreatorLock.test.js
```diff
@@ -84,4 +103,39 @@ describe('CreatorLock', () => {
...
+      transactions: {
+        transactionid: transaction,
+      },
+      locks: {
+        lockid: keylock,
+      },
+      account: account,
+    })
+
+    mockWeb3Service.withdrawFromLock = jest.fn()
```
julien51: Since you do that you can *definetely* remove the class stuff above ;)

julien51:
So... I think you should not write tests for `CreatorLock.test.js` but you should write tests for `lock/Withdraw.js` because this is where the withdrawal is happening.

Also this means your tests will likely be simpler...

benwerd:
Swapped test out for something much simpler in `Withdraw.test.js`. PTAL

unlock-app/src/components/interface/buttons/lock/Withdraw.js
```diff
@@ -1,11 +1,53 @@
 import React from 'react'
+import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
+
 import Svg from '../../svg'
 import { LockButton } from '../Button'
+import UnlockPropTypes from '../../../../propTypes'
+import { withdrawFromLock } from '../../../../actions/lock'
+
+export function Withdraw({ lock, withdraw, account, ...props }) {
```
julien51: Nit: I think you could use a stateless function here and just put the if statement in the onclick?

No strong opinion.

unlock-app/src/components/interface/buttons/lock/Withdraw.js
```diff
@@ -1,11 +1,53 @@
...
+  account: UnlockPropTypes.account,
+}
+
+Withdraw.defaultProps = {
+  transaction: null,
+  account: null,
+  withdraw: () => {},
+}
+
+const mapStateToProps = (state, { lock }) => {
```
julien51: Nit: destructure state?

PR #593, opened 2018-11-22
objects should be immutable in react/redux

julien51:


julien51:
This is the first of a long chain of PR which aim at making our data store cleaner and more robust.

Issue #431, opened 2018-10-29
Hamburger Menu on mobile


julien51:
Our mobile layout needs a hamburger menu!
Right now, when the screen is small we show the following header for content pages:
<img width="431" alt="screen shot 2018-10-28 at 20 09 58" src="https://user-images.githubusercontent.com/17735/47623976-8055a180-daed-11e8-9b71-93fcc07799ec.png">

It is likely that we'll add more icons to that bar, which means that we may eventually lack space.
Our designer @smombartz came up with that design:
<img width="391" alt="screen shot 2018-10-28 at 20 11 22" src="https://user-images.githubusercontent.com/17735/47624004-b2670380-daed-11e8-9716-9c5b17694860.png">

Which will allow us to expand into the following when the user clicks on the hamburger icon. (when the user clicks on the chevron the content of the menu should collapse back into the hamburger icon)
<img width="370" alt="screen shot 2018-10-28 at 20 11 31" src="https://user-images.githubusercontent.com/17735/47624008-babf3e80-daed-11e8-89a8-1beaf164541c.png">

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


**1) [gyrevik](https://gitcoin.co/profile/gyrevik) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/unlock-protocol/unlock/431/1733?mutate_worker_action=approve&worker=gyrevik) | [reject worker](https://gitcoin.co/issue/unlock-protocol/unlock/431/1733?mutate_worker_action=reject&worker=gyrevik))_.**

I have not started running any of the unlock code yet so I'm not sure about how mobile is handled in unlock. I will try to run the code and may have a question at some point about how the app handles mobile.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/unlock-protocol/unlock/431/1733).

gitcoinbot:
@gyrevik Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
* [x] reminder (3 days)
* [ ] escalation to mods (6 days)

[...]

PR #595, opened 2018-11-23
Mobile Menu

ChrisCates:
Fixes https://github.com/unlock-protocol/unlock/issues/431

## Proposed Changes

Mobile Responsive Menu with nice animation

## Needs Comments

1. SVG Assets for Menu.

2. Correct URLs for each item.

### DIFF
--- a/unlock-app/src/components/interface/Header.js
+++ b/unlock-app/src/components/interface/Header.js
@@ -4,23 +4,51 @@ import styled from 'styled-components'
 import Link from 'next/link'
 import Svg from './svg'
 import Buttons from './buttons/layout'
+import { Button } from './buttons/Button'
 
-export default function Header({ forContent, title }) {
-  return (
-    <TopHeader>
-      {!forContent && <Title>{title}</Title>}
-      {!!forContent && (
-        <Link href="/">
-          <a>
-            <Svg.UnlockWordMark height="28px" width="100%" name="Unlock" />
-          </a>
-        </Link>
-      )}
-      <Buttons.About />
-      <Buttons.Jobs />
-      <Buttons.Github />
-    </TopHeader>
-  )
+export default class Header extends React.PureComponent {
+  constructor(props) {
+    super(props)
+    this.forContent = props.forContent
+    this.title = props.title
+    this.toggleMenu = this.toggleMenu.bind(this)
+    this.state = { menu: false }
+  }
+
+  toggleMenu() {
+    this.setState(prevState => ({ menu: !prevState.menu }))
+  }
+
+  render() {
+    const { menu } = this.state
+
+    return (
+      <TopHeader>
+        {!this.forContent && <Title>{this.title}</Title>}
+        {!!this.forContent && (
+          <Link href="/">
+            <a>
+              <Svg.UnlockWordMark height="28px" width="100%" name="Unlock" />
+            </a>
+          </Link>
+        )}
+        <DesktopButtons>
+          <Buttons.About />
+          <Buttons.Jobs />
+          <Buttons.Github />
+        </DesktopButtons>
+        <MobileToggle>
+          <Buttons.Bars onClick={this.toggleMenu} size="48px" />
+          <Buttons.ChevronUp onClick={this.toggleMenu} size="48px" />
+        </MobileToggle>
+        <MobilePopover visibilityToggle={menu ? true : false}>
+          <Buttons.About size="48px" />
+          <Buttons.Jobs size="48px" />
+          <Buttons.Github size="48px" />
+        </MobilePopover>
+      </TopHeader>
+    )
+  }
 }
 
 Header.propTypes = {
@@ -40,8 +68,102 @@ const TopHeader = styled.header`
   grid-auto-flow: column;
   align-items: center;
   height: 70px;
+
+  @media (max-width: 600px) {
+    display: flex;
+  }
 `
 
 const Title = styled.h1`
   color: var(--grey);
+`
+
+const DesktopButtons = styled.div`
+  display: grid;
+  grid-gap: 16px;
+  grid-template-columns: 1fr repeat(3, 24px);
+  grid-auto-flow: column;
+  align-items: center;
+  height: 100%;
+
+  @media (max-width: 600px) {
+    display: none;
+  }
+`
+
+const MobileToggle = styled.div`
+  display: none;
+  flex-direction: row;
+  align-items: center;
+  height: 100%;
+
+  ${Button} {
+    position: absolute;
+    right: 5px;
+    top: 0px;
+
+    background-color: var(--white);
+    padding: 15px;
+
+    transition: all 300ms cubic-bezier(0.165, 0.84, 0.44, 1);
+    opacity: 1;
+    pointer-events: visible;
+
+    ${props => (props.visibilityToggle ? '&:nth-child(1)' : '&:nth-child(2)')} {
+      opacity: 0;
+      pointer-events: none;
+    }
+
+    > svg {
+      fill: var(--grey);
+    }
+
+    &:hover {
+      > svg {
+        fill: var(--link);
+      }
+    }
+  }
+
+  @media (max-width: 600px) {
+    display: block;
+  }
+`
+
+const MobilePopover = styled.div`
+  background-color: var(--white);
+  position: absolute;
+  left: 0;
+  top: 70px;
+  width: 100%;
+  height: auto;
+  z-index: var(--foreground);
+  padding-bottom: 30px;
+
+  display: none;
+  grid-template-columns: 92px 92px 92px;
+  grid-gap: 0px;
+  align-items: center;
+  justify-content: center;
+
+  transition: all 300ms cubic-bezier(0.165, 0.84, 0.44, 1);
+  ${props =>
+    props.visibilityToggle
+      ? 'opacity: 1; pointer-events: visible;'
+      : 'opacity: 0; pointer-events: none;'} ${Button} {
+    margin: 24px;
+
+    small {
+      display: block;
+      color: var(--grey);
+      text-align: center;
+      top: 5px;
+      width: 100%;
+      text-align: center;
+    }
+  }
+
+  @media (max-width: 600px) {
+    display: grid;
+  }
 `
--- a/unlock-app/src/components/interface/buttons/Button.js
+++ b/unlock-app/src/components/interface/buttons/Button.js
@@ -10,15 +10,7 @@ const clickAction = (e, action) => {
 
 export const LayoutButton = ({ href, title, children, action, ...props }) => {
   const button = (
-    <Button
-      href={href}
-      backgroundColor="var(--grey)"
-      fillColor="white"
-      backgroundHoverColor="var(--link)"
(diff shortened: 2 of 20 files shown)


julien51:
@ChrisCates Thanks a lot! Do you mind removing the `*.json` files from your PR?

Also, please submit screenshots...

julien51:
Thanks a lot @ChrisCates. First of all, please, let's fix the tests here. It needs to be all green to merge.

Also I see you still have the `*.json` files in there (with different permissions). Please remove them completely.
Similarly `smart-contracts/npm-shrinkwrap.json` and `unlock-app/package-lock.json` don't belong in this PR. 

Now, let's not care about the newsletter, telegram and twitter buttons. You can remove them fro your work. It will be easy enough to add them when we need them.

Finally, I don't think it is a great idea to write custom `XXXLarge` components because it will be a lot more code to maintain when most of it is identical to the non large version. Couldn't we just use CSS to indicate that when a button is inside `MobilePopover` it needs to be displayed at a larger size?

unlock-app/src/components/interface/Header.js
```diff
@@ -40,8 +71,64 @@ const TopHeader = styled.header`
   grid-auto-flow: column;
   align-items: center;
   height: 70px;
+
+  @media (max-width: 600px) {
+    display: flex;
```
julien51: Would there be a way to use `grid` here too to be consistent with the rest of the UI?

unlock-app/src/components/interface/Header.js
```diff
@@ -40,8 +71,64 @@ const TopHeader = styled.header`
...
+`
+
+const MobileToggle = styled.div`
+  display: none;
+  flex-direction: row;
+  align-items: center;
+  height: 100%;
+
+  @media (max-width: 600px) {
+    display: flex;
```
julien51: Same here? `grid` would be more consistent...

unlock-app/src/components/interface/Header.js
```diff
@@ -40,8 +71,64 @@ const TopHeader = styled.header`
...
+  }
+`
+
+const MobilePopover = styled.div`
+  background-color: var(--white);
+  position: absolute;
+  left: 0;
+  top: 70px;
+  z-index: 9001;
```
julien51: Can use a CSS var for this? like `--foreground`.

unlock-app/src/components/interface/Header.js
```diff
@@ -40,8 +71,64 @@ const TopHeader = styled.header`
...
+
+  display: none;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: center;
+  width: 100%;
+  height: auto;
+
+  transition: all 300ms cubic-bezier(0.165, 0.84, 0.44, 1);
```
julien51: 🎊

unlock-app/src/components/interface/buttons/Button.js
```diff
@@ -29,6 +29,58 @@ export const LayoutButton = ({ href, title, children, action, ...props }) => {
   return button
 }
 
+export const MobileToggleButton = ({
```
julien51: Probably worth defining in its own file using the same approach as the other buttons.

unlock-app/src/components/interface/buttons/Button.js
```diff
@@ -116,6 +168,70 @@ export const Button = styled.a`
...
+  }
+
+  &:hover {
+    > svg {
+      fill: ${props => props.fillHoverColor || 'var(--link)'};
+    }
+  }
+`
+
+export const MobileButton = styled.a`
```
julien51: Not a huge fan to have yet another type of button when most of the code is the same as for `LayoutButton`.

ChrisCates:
Hi, thanks for the detailed review:

1. I updated the .gitignore, but, then decided to just restore it to the original state. I'll make sure to add a .gitignore to ignore those files (**in all folders where applicable**).

2. I agree in regards to the extra boilerplate with `.js` files. It will probably be better to just specify the `props.size` to those icons instead.

3. Perhaps I can use some sort of ternary operator for the `MobileButton` code to make it utilize the same `styled-component` class. This way we can remove that code altogether and just keep it as `LayoutButton`.

Will update tonight. Thanks!~

ChrisCates:
@julien51 here's some more updates for review.
One of the assets sent did not reflect the screens given.

But the update (below) shows the chevron used instead:

ChrisCates:
@julien51, the @netlify tests will not pass unless you have those build files available. I have restored the original permissions...

Perhaps it would be best to run a `truffle compile` in the test in order to do those tests.
Never used netlify before, so helping with the process to change that would be helpful!~

julien51:
Hey, you need to compile the files, but do not include them in your commit ;)

julien51:
Also the error you're seeing is actually that you need to update the jest snapshots. Check your logs ;)

julien51:
You can run tests locally with `npm test`.

ChrisCates:
@julien51, thanks for clarifying. Did some reading into Jest.

julien51:
Generally we're big fans of 1 PR <=> 1 PuRpose and I see a lot of files/changes in this PR which are not related to what you're trying to achieve here. Make sure you only commit changes which are relevant.

.gitignore
```diff
@@ -6,3 +6,5 @@ node_modules/
 # ZeppelinOS
 zos.dev-*.json
 .zos.session
+
+package-lock.json
```
julien51: Please remove this from this PR.

smart-contracts/.gitignore
```diff
@@ -0,0 +1,4 @@
+package-lock.json
```
julien51: Please remove this from this PR.

julien51:
I see you have removed the artifacts files altogether in this PR which is not right.  Just remove them from the commit, but keep them in git ;)

unlock-app/package.json
```diff
@@ -74,8 +74,9 @@
     "deploy-staging": "npm run build && TARGET=staging npm run build-unlockjs && aws s3 sync build/ s3://staging.unlock-protocol.com --profile unlock --no-progress | awk '{print $2;}' | cut -d'/' -f2- | awk '{print \"/\"$0}' | xargs aws cloudfront create-invalidation --distribution-id E1ZWU0E90B9EBN --paths --profile unlock",
     "build-unlockjs": "./node_modules/webpack/bin/webpack.js --config webpack-unlockjs.config.js",
     "serve-unlockjs": "TARGET=dev node_modules/.bin/webpack-dev-server --config webpack-unlockjs.config.js &",
-    "test": "NODE_ENV=test jest --env=jsdom",
+    "test": "cd ../smart-contracts && truffle compile && cd ../unlock-app && NODE_ENV=test jest --env=jsdom",
```
julien51: Please do not change this test script.
If you need, use `pretest` in another PR.

ChrisCates:
@julien51, let me squash all the commits. I'll untrack the artifacts too.

unlock-app/src/components/interface/Header.js
```diff
@@ -5,22 +5,68 @@ import Link from 'next/link'
...
+            className={
+              'mobile-toggle ' +
+              (menu ? 'mobile-toggle-visible' : 'mobile-toggle-invisible')
+            }
+          />
+        </MobileToggle>
+        <MobilePopover
+          className={menu ? 'popover-visible' : 'popover-invisible'}
+        >
+          <Buttons.About size="48px" className="mobile-menu" />
```
julien51: Let's not use `className`. I think you can probably use a CSS rule on `MobilePopover` for is `Button` Children.

unlock-app/src/components/interface/Header.js
```diff
@@ -40,8 +86,66 @@ const TopHeader = styled.header`
...
+
+  transition: all 300ms cubic-bezier(0.165, 0.84, 0.44, 1);
+  opacity: 1;
+  pointer-events: visible;
+
+  &.popover-invisible {
+    opacity: 0;
+    pointer-events: none;
+  }
+
```
julien51: You can apply properties to the buttons here using something like this:

```
${Button} {
... // CSS rules to apply to the `Button` component when they are inside a `MobilePopover` element. 
}
```

unlock-app/src/components/interface/Header.js
```diff
@@ -5,22 +5,68 @@ import Link from 'next/link'
...
+            }
+          />
+        </MobileToggle>
+        <MobilePopover
+          className={menu ? 'popover-visible' : 'popover-invisible'}
+        >
+          <Buttons.About size="48px" className="mobile-menu" />
+          <Buttons.Jobs size="48px" className="mobile-menu" />
+          <Buttons.Github size="48px" className="mobile-menu" />
+          <Buttons.Newsletter size="48px" className="mobile-menu" />
```
julien51: You can remove the Newsletter, Telegram and Twitter buttons because we actually do not use them yet... (I know the design brief is wrong on that... sorry!)

unlock-app/src/components/interface/Header.js
```diff
@@ -5,22 +5,68 @@ import Link from 'next/link'
...
+        <DesktopButtons>
+          <Buttons.About />
+          <Buttons.Jobs />
+          <Buttons.Github />
+        </DesktopButtons>
+        <MobileToggle>
+          <Buttons.Bars
+            onClick={this.toggleMenu}
+            size="48px"
+            className={
```
julien51: Same here, let's not use `className`.

julien51: You can pass props to styled-components.

ChrisCates:
@julien51, there we go. It's all in one commit. Let's just make sure tests pass.

Something seemed to be wrong with the docker image in the Travis-CI test though.

unlock-app/package.json
```diff
@@ -76,6 +76,7 @@
     "serve-unlockjs": "TARGET=dev node_modules/.bin/webpack-dev-server --config webpack-unlockjs.config.js &",
     "test": "NODE_ENV=test jest --env=jsdom",
     "lint": "./node_modules/eslint/bin/eslint.js .",
+    "lint-fix": "./node_modules/eslint/bin/eslint.js . --fix",
```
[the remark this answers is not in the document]
ChrisCates: Had no idea about the additional `--` prefix. Handy, thanks for letting me know!

unlock-app/src/components/interface/Header.js
ChrisCates: No worries! They can easily be popped in later. It's nice to have them available anyway.

unlock-app/src/components/interface/Header.js
ChrisCates: Makes sense!~

julien51:
Looks like tests are passing on travis now! Please see a couple more comments/simplifications!

ChrisCates:
@julien51, okay, just restored the original package locks... If tests pass. Everything should be good to go.

ChrisCates:
Looks good to go @julien51, if you need anything from me. Let me know!
Elsewise, I'll mark as done!~

julien51:
Great work! Thanks a lot @ChrisCates !

PR #598, opened 2018-11-24
Fixing storybook stories.

julien51:


unlock-app/src/stories/creator/LockIconBar.stories.js
```diff
@@ -1,14 +1,21 @@
...
-  return <LockIconBar lock={lock} />
-})
+storiesOf('LockIconBar', LockIconBar)
+  .addDecorator(getStory => <Provider store={store}>{getStory()}</Provider>)
+  .add('LockIconBar', () => {
+    const lock = {
+      keyPrice: '10000000000000000000',
+      expirationDuration: '172800',
+      maxNumberOfKeys: '240',
+      outstandingKeys: '3',
```
benwerd: I'll go back and fix this - my oversight - but we should be using integers now.

PR #607, opened 2018-11-27
updated web3Service to process events on transactions when refreshed

julien51:


julien51:
Another fairly big PR (sorry about that!) but I think I start to like how we architect things between the `web3Service`, the `lockMiddleware` and the redux store.

One of the benefits of this new approach is that even if we lose the data for the lock it can be "retrieved" again from the transaction data (except for the title... which is not stored anywhere yet.).

PR #615, opened 2018-11-28
drying up code for buttons

julien51:


unlock-app/src/components/creator/FatalError.js
```diff
@@ -84,7 +84,7 @@ export const WrongNetwork = ({ currentNetwork, requiredNetwork }) => (
     illustration="/static/images/illustrations/network.svg"
   >
     <p>
-      You’re currently on the
+      You’re currently on the
```
benwerd: Worth seeing if your editor can automatically trim whitespace from the end of lines. It's not a big deal but it'd be nice to keep the source clean.

unlock-app/src/components/interface/buttons/Button.js
```diff
@@ -8,63 +8,59 @@ const clickAction = (e, action) => {
...
+  children,
+  backgroundColor,
+  backgroundHoverColor,
+  fillColor,
+  fillHoverColor,
+  ...props
+}) => {
+  return (
+    <BaseButton
+      backgroundColor={backgroundColor || 'var(--grey)'}
```
benwerd: Do we need this if `defaultProps` is being set below?

unlock-app/src/components/interface/buttons/layout/About.js
```diff
@@ -3,7 +3,7 @@ import Svg from '../../svg'
 import { LayoutButton } from '../Button'
 
 const About = props => (
-  <LayoutButton href="/about" title="About" {...props}>
+  <LayoutButton href="/about" label="About" {...props}>
```
benwerd: Makes sense to use label here on the layout buttons - but let's continue to use `title` for UI elements?

unlock-app/src/components/creator/FatalError.js
julien51: Meh. this is weird. I thought prettify would take care of that?

unlock-app/src/components/interface/buttons/Button.js
julien51: I think you're right. Let me remove that.

unlock-app/src/components/interface/buttons/layout/About.js
julien51: ok!

PR #641, opened 2018-11-30
Showing all SVG on single story

julien51:


julien51:
screen shot 2018-11-30 at 12 28 31


Per @smombartz's request!

When you hover on any, it shows its name!

PR #644, opened 2018-12-01
Fixing new lock address reference issue

benwerd:
Typo fix. I think there must have been some search and replacing done on lockAddress => lock. The effect was that saved locks had an undefined address.


benwerd:
Absolutely not dumb! I've done this a million times. Easy to fix. I do have a question about how best to add a test for this - will bump to an issue.

PR #646, opened 2018-12-01
Better handle long lock names in key purchase UI

benwerd:
I was seeing this kind of thing in the demo UI:

screen shot 2018-11-30 at 5 51 58 pm

This update ensures long names wrap more cleanly:

screen shot 2018-11-30 at 6 38 29 pmscreen shot 2018-11-30 at 6 38 22 pmscreen shot 2018-11-30 at 6 45 46 pmscreen shot 2018-11-30 at 6 45 51 pm


benwerd:
That’s correct - basically only one or two word lock names fit properly. I’m going to merge this but assume that the design will change again! @smombartz I’m eager for your thoughts.

PR #651, opened 2018-12-03
Disabled withdrawal button displayed when balance is 0

benwerd:
Disabled buttons are lighter (cc @smombartz for design feedback) and don't trigger any attached action.

screen shot 2018-12-03 at 12 35 15 pm

### DIFF
--- a/unlock-app/src/__tests__/components/interface/Button.test.js
+++ b/unlock-app/src/__tests__/components/interface/Button.test.js
@@ -2,7 +2,9 @@ import React from 'react'
 import * as rtl from 'react-testing-library'
 import 'jest-dom/extend-expect'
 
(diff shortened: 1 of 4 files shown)


unlock-app/src/__tests__/components/interface/Button.test.js
```diff
@@ -52,4 +54,21 @@ describe('Button', () => {
...
     expect(wrapperClicked).toEqual(false)
   })
+  it('should not run an action from a disabled button', () => {
+    let buttonClicked = false
+
+    const buttonAction = () => {
+      buttonClicked = true
+    }
+
+    // role and onKeyDown action here are to comply with accessibility rules
```
julien51: 🙏

PR #655, opened 2018-12-04
removed localStorage

julien51:


julien51:
Yeah, I am *half-expecting* that we'll have to revert that at some point to improve the experience...

PR #659, opened 2018-12-04
better default state when loading

julien51:


julien51:
screen shot 2018-12-04 at 14 01 32

I think this is better than showing a bunch of 0s in weird places.

PR #663, opened 2018-12-05
Update icons with more shapes, rotation, and mirroring.

### LINKED ISSUE #219: Customize position of circles
In order to get more unique lock icons we should also position (x, y and z) the inner circles

---

https://github.com/unlock-protocol/unlock/blob/b31941ac3a2a778f2c755951c838e0048ab181a2/unlock-app/src/components/lock/Icon.js#L9-L14

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `@todo` comment in b31941ac3a2a778f2c755951c838e0048ab181a2 when #216 was merged. cc @unlock-protocol.

cnasc:
This pull request modifies the Icon component to include more variation in shapes (6 total, including the original icon).

The new shapes are as follows:
screen shot 2018-12-04 at 6 59 25 pm
_stripedIcon_

screen shot 2018-12-04 at 7 07 36 pm
_chompIcon_

screen shot 2018-12-04 at 7 06 43 pm
_biteIcon_

screen shot 2018-12-04 at 7 09 31 pm
_tailIcon_

screen shot 2018-12-04 at 7 09 14 pm
and _triadIcon_

(I hope these names make it easier to refer to a given icon easily)

Please let me know if there are any edits I should make. I suspect we may want to update the storybook to show an example of at least each different shape.

Fixes #219 

## Proposed Changes

- Add new shapes
- Add rotation of inner part of icon, in 10-degree increments based on lock address
- Add horizontal mirroring of inner part of icon, based on lock address
- Fix existing examples in Storybook for icons (looks like they were constructed based on an older revision of the icon component and needed to have a lock object passed in rather than a bare address)

### DIFF
--- a/unlock-app/src/components/lock/Icon.js
+++ b/unlock-app/src/components/lock/Icon.js
@@ -3,12 +3,66 @@ import ColorScheme from 'color-scheme'
 
 import UnlockPropTypes from '../../propTypes'
 
+const originalIcon = [
+  { x: 195.75, y: 114.75 },
+  { x: 33.75, y: 162 },
+  { x: 121.5, y: 0 },
+]
+
+const stripedIcon = [{ x: 108, y: 108 }, { x: 146, y: 147 }, { x: 216, y: 216 }]
+
+const chompIcon = [{ x: 108, y: 108 }, { x: 108, y: -64 }, { x: 108, y: 280 }]
+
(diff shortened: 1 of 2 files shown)


julien51:
I think the errors you're seeing are due to the address of pending locks (being deployed) is null... Could you look into that? I think the best approach is to pick a random shape and make it grey?

PR #667, opened 2018-12-05
Forcing withdrawal button to update

benwerd:
So it turns out that:

1. `CreatorLock` is initially drawn before `lock.balance` is known
2. When `CreatorLock` updates to show a balance, React decides that there's no need to update the withdrawal button, because it thinks no relevant props have changed

In order to fix that, I'm now passing balance as a separate prop. And I hate myself for it a little bit. But here it is.

This PR restores the "withdraw" button when lock has a balance.

### DIFF
--- a/unlock-app/src/components/creator/CreatorLocks.js
+++ b/unlock-app/src/components/creator/CreatorLocks.js
@@ -39,7 +39,7 @@ export class CreatorLocks extends React.Component {
         <Error />
         {showDashboardForm && <CreatorLockForm hideAction={this.toggleForm} />}
         {Object.values(locks).map(lock => {
-          return <CreatorLock key={lock.address} lock={lock} />
+          return <CreatorLock key={JSON.stringify(lock)} lock={lock} />
         })}
       </Locks>
     )


julien51:
Hum... this is an antipattern. There must be a better way. Let me look this up.

julien51:
So, I think there must be *something* else that is wrong, because in practice, why would `LockIconBar` be re-rendered and not `Withdraw`? They're both sub-components of something which knows of the state, right?

benwerd:
That's what I thought, but I've been fighting this for a while. It could be that my React knowledge isn't quite strong enough. I have definitely (through lots and lots of `console.log`) established that the icon bar is only rendered once, at a time when balance isn't populated.

julien51:
My understanding is that react will re-render any component only if the props have changed (be different objects). So if it is not re-rendering it means that the object passed as props is the same (even with different keys/values).

The best way to see this is that if you remove all of your changes but just keep the following diff:
change: 
```
return <CreatorLock key={lock.address} lock={lock} />
```
into:
```
return <CreatorLock key={lock.address} lock={{ ...lock }} />
```

The withdraw button down the line will be re-render as expected. so now, we just have to find out why we are not passing a *new* version of the `lock` object down the line!

benwerd:
Wait. Could this be as simple as a pass by reference error somewhere down the line?! So lock gets updated and we're returning the modified lock object rather than a copy of that object with new values ...

julien51:
After diving a bit deeper (and reading react docs ;p) it looks like our use of `key` is the culprit. Basically, this `key` is a way for React to know if/when an element is different. since we use the `id` which never changes, React assume the `lock` object being passed is the same... when it was actually changed ;)

benwerd:
Got it. So key should maybe be a hash of the object's important values (or of the serialization of the whole object)?

benwerd:
Trying this in a separate branch ...

julien51:
Exactly, I am trying with `          return <CreatorLock key={JSON.stringify(lock)} lock={lock} />
` and that works as expected. I can push to your branch if that's ok!

PR #675, opened 2018-12-06
using a local ganache module to avoid discrepancy avoid versions

julien51:


julien51:
I also re-compiled the smart contracts to make sure we are consistently using the right version.

Now I need to check that if there is a globally installed version of ganache, it won't override this.

julien51:
This version *forces* the use of an older version of ganache-cli. Once we know this is being consistently used by everyone, we can upgrade to the latest version and find what is going on with it.
[this should at least with https://github.com/unlock-protocol/unlock/issues/465]

PR #677, opened 2018-12-06
Reduce ganache blocktime to 0 for local testing

nfurfaro:
## Proposed Changes

- Removes `-b 3` from  the npm script "start-ganache" 
- Adds `-- -b 3` to the npm script "setup-dev": "cd .. && (npm run start-ganache -- -b 3 &) && npm run deploy-lock",
-

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   },
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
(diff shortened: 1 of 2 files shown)


nfurfaro:
Yes, 3-second block time on dev is confirmed, and tests are fast, block time  = 0 seconds

PR #692, opened 2018-12-10
fixing order of people and layout

julien51:


julien51:
screen shot 2018-12-10 at 17 37 03

PR #699, opened 2018-12-12
Embeddable paywall code

benwerd:
This PR creates an iframe using the latest code that displays a paywall lock over existing content.

We may (I think we should) want to refactor the way the code detects locks in the third-party page HTML in a future PR. Rather than hiding elements on the page, we may want to include a `<meta name="lock" value="{lock id}" />` tag in the head, to reflect the fact that we're going with a full-page overlay rather than locks that can cover divs arbitrarily.

Here's what it looks like on a test site:

screen shot 2018-12-11 at 3 59 24 pm


julien51:
Great comments! Please open issues/add to board!

unlock-app/src/_server.js
```diff
@@ -18,7 +18,10 @@ function _server(port, dev) {
           // assigning `query` into the params means that we still
           // get the query string passed to our application
           const path = pathname.split('/')[1]
-          if (path === 'lock') {
+          if (path === 'paywall') {
+            const params = route('/paywall/:lockAddress')(pathname)
+            app.render(req, res, '/paywall', Object.assign(params, query))
+          } else if (path === 'lock') {
```
julien51: Are we using `/lock` anymore?

unlock-app/src/components/lock/ShowUnlessUserHasKeyToAnyLock.js
```diff
@@ -16,6 +17,7 @@ export const ShowUnlessUserHasKeyToAnyLock = ({
 
   // There is a valid key, but we shown the modal previously
   if (keys.length > 0 && !!modalShown) {
+    unlockPage()
```
julien51: I actually have a PR coming that refactors all this because I think we need to use `lifecycle` functions here. Please bear with me ;)

unlock-app/src/pages/paywall.js
```diff
@@ -0,0 +1,44 @@
...
+    return {
+      lockAddress: lockAddress,
+    }
+  }
+
+  render() {
+    const { lock } = this.props
+
+    return (
+      <>
```
julien51: Do you need this if you're using `<NoSSR>`?

PR #701, opened 2018-12-12
Form validates input on change, doesn't submit invalid values

### LINKED ISSUE #688: Validate inputs on form when creating lock
When creating a lock but entering values which are not of the right type (strings instead of numbers for example), nothing happens. We should probably show an error indicating that and/or update the form field to show it's not valid? (I like when form are circles in red to indicate trouble: @smombartz does it make sense).

benwerd:
I went with a red background for now, but obviously this can be changed.

Fixes #688.

screen shot 2018-12-11 at 5 36 56 pm
screen shot 2018-12-11 at 5 37 19 pm
screen shot 2018-12-11 at 5 37 27 pm
screen shot 2018-12-11 at 5 37 44 pm
screen shot 2018-12-11 at 5 37 53 pm

### DIFF
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -35,21 +35,9 @@ class CreatorLockForm extends React.Component {
   }
 
   validate(name, value) {
-    switch (name) {
-      case 'name':
-        // TODO do we want to impose a maximum lock name length?
-        if (typeof value === 'string' && value.length > 0) return true
-        break
-      case 'expirationDuration':
-      case 'maxNumberOfKeys':
-        if (!isNaN(value) && value > 0) return true
-        break
-      case 'keyPrice':
-        if (!isNaN(value) && value >= 0) return true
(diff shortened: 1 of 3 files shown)


benwerd:
cc @smombartz for design feedback

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -34,11 +34,20 @@ class CreatorLockForm extends React.Component {
...
+      (name === 'keyPrice' && !isNaN(value) && value >= 0)
+  }
+
   handleChange(event) {
+    event.target.dataset.valid = this.validate(event.target.name, event.target.value)
     this.setState({ [event.target.name]: event.target.value })
   }
 
   handleSubmit() {
+    if (document.querySelector('[data-valid="false"]')) return false
```
julien51: I think we could maybe even show an error message in the error panel here later (I worry that just showing an error state may not *tell* what the error actually is)? But @smombartz can weigh on this!

PR #703, opened 2018-12-12
Creator Log Page

njmyers:
Fixes #

## Proposed Changes

- Create Log Page
- Fix durations function to pass in seconds instead of milliseconds
- New durations function to create strings with date and time
- Add log button to interface/buttons
- Add TRANSACTION_TYPES constant
- Add transaction types to web3Service file

nick-log-page


unlock-app/src/utils/durations.js
```diff
@@ -82,10 +82,33 @@ export function expirationAsDate(timestamp) {
     return durationsAsTextFromSeconds(timestamp)
 
   let expirationDate = new Date(0)
-  expirationDate.setUTCSeconds(timestamp)
+  expirationDate.setUTCSeconds(timestamp / 1000)
```
benwerd: See the error in `duration.test.js` - it looks like something is dividing by 1000 one too many times (and resulting in January 20, 1970, which is roughly July 7, 2022 divided by 1000).

njmyers:
Hello @benwerd. Happy to make these changes. 

Regarding the date function, it seems like the paradigm you are working from is to make the date functions accept timestamps in terms of seconds (like epoch timestamp). Since the data that we get from web3Service is in milliseconds I wanted to match the pattern of the application so it is clear for future consumers of both the preexisting function and the one that I created. Please see the below pseudocode for illustrative purposes.

```
// define a formatting function
function dateString(seconds){
  // format
}

// using a formatting function with web3 data.
dateString(web3Milliseconds / 1000) = 'correctly formatted string'
```
In that case I would define my function to accepts seconds and then do any prerequisite division at the call site.

Thanks for your feedback and comments!

unlock-app/src/components/creator/CreatorLog.js
```diff
@@ -0,0 +1,135 @@
...
+/** Define default string values for the log so we don't crash the app */
+const logDefaults = {
+  DATE: 'Date Unknown',
+  NAME: '',
+  TYPE: 'TRANSACTION',
+  OWNER: '',
+  DATA: '',
+}
+
+function snakeToTitleCase(string) {
```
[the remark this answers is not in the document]
julien51: I also think we could move this function into an `strings.js` in `utils` ? This feels like the kind of function that we could re-use in many places.

unlock-app/src/components/creator/CreatorLog.js
```diff
@@ -0,0 +1,135 @@
...
+  return string
+    .split('_')
+    .map(word => word.slice(0, 1).toUpperCase() + word.slice(1).toLowerCase())
+    .join(' ')
+}
+
+export class CreatorLog extends React.Component {
```
julien51: Could we actually rename this `TransactionLog` (and the corresponding file). I think that would be a better description..

unlock-app/src/pages/log.js
```diff
@@ -0,0 +1,54 @@
...
+import withConfig from '../utils/withConfig'
+import { pageTitle } from '../constants'
+
+export const Log = ({ account, network, transactions, locks, state }) => {
+  if (!account) {
+    return null //loading
+  }
+
+  return (
+    <Layout title="Creator Log">
```
julien51: `Transaction log` ?

julien51:
Hello @njmyers... please let us know if you have time to make progress on this.
Thanks

julien51:
@njmyers I have added the transaction type extraction in its own PR so you can probably remove this from your branch and use ours! Let me know if you need help.

njmyers:
@julien51 Hi Julien. Sorry for the delay. Had a busy week with some other things but I'm happy to finish up with the stories, tests and refactoring next week if that is alright with you?

Thanks I'll check out the types PR!

Issue #331, opened 2018-10-11
Group pending and saved locks, rather than displaying them by time


benwerd:
The order is jarring - really pending locks should show up at the top, newest first.

julien51:
I think this is now done ;) Thanks @benwerd !

PR #707, opened 2018-12-13
Display newest locks first

### LINKED ISSUE #330: Visually replace form with pending lock on submission
Currently new locks show up at the bottom of the screen.

benwerd:
Displays locks in `CreatorLocks` in reverse chronological order. Fixes #330 _and_ #331.

### DIFF
--- a/unlock-app/src/components/creator/CreatorLocks.js
+++ b/unlock-app/src/components/creator/CreatorLocks.js
@@ -24,6 +24,7 @@ export class CreatorLocks extends React.Component {
   render() {
     const { locks } = this.props
     const { showDashboardForm } = this.state
+    let lockFeed = Object.values(locks).reverse() // We want to display newer locks first
 
     return (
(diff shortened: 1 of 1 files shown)


unlock-app/src/components/creator/CreatorLocks.js
```diff
@@ -24,6 +24,7 @@ export class CreatorLocks extends React.Component {
   render() {
     const { locks } = this.props
     const { showDashboardForm } = this.state
+    let lockFeed = Object.values(locks).reverse() // We want to display newer locks first
```
julien51: Is the order right now based on the creation date?

benwerd: Unfortunately we don't seem to have access to creation dates for locks or transactions - but the effect is that they're retrieved in created order. This may not always be the case, so some kind of stored created date would be useful.

PR #709, opened 2018-12-13
ran npm run reformat

julien51:


benwerd:
Interestingly, when I ran reformat locally, I also got these changes. So although reformatting _is_ happening on commit for me, I wonder if it's not running on the same rules.

benwerd:
Never mind, it stopped running reformat on commit for some reason and I didn't notice. I'll try and figure out why.

PR #711, opened 2018-12-13
Refactored web3Service for keys calls

julien51:


julien51:
There is a second coming after which fetches the keys using life cycle methods on `ShowUnlessUserHasKeyToAnyLock`.

PR #715, opened 2018-12-14
refreshing a lock's data after it has been deployed

julien51:
Fixes bug where previously deployed locks did not show any value for expiration, price of keys ... etc

### DIFF
--- a/unlock-app/src/__tests__/middlewares/lockMiddleware.test.js
+++ b/unlock-app/src/__tests__/middlewares/lockMiddleware.test.js
@@ -141,17 +141,18 @@ describe('Lock middleware', () => {
   })
 
   it('it should handle lock.saved events triggered by the web3Service', () => {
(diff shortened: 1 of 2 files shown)


unlock-app/src/__tests__/middlewares/lockMiddleware.test.js
```diff
@@ -141,17 +141,18 @@ describe('Lock middleware', () => {
   })
 
   it('it should handle lock.saved events triggered by the web3Service', () => {
-    expect.assertions(2)
+    expect.assertions(3)
     const { store } = create()
     const lock = {}
     const address = '0x123'
     mockWeb3Service.refreshAccountBalance = jest.fn()
-    mockWeb3Service.getKeyByLockForOwner = jest.fn()
```
julien51: Not sure why that was there :/

PR #716, opened 2018-12-14
running reformat and failing if some files have been changed

julien51:
On each PR, this script will now run `npm run reformat` and make sure that nothing was changed. If the styles are changed, the PR will fail. This should at least enforce that no code is ever merged that has a different format.

I believe an important next step here would be to find a way to run `npm install` any time the versions in `package-lock.json` has been changed.


julien51:
Ho,ooo, this failed and could very well be the same problem that you're seeing @benwerd ?

benwerd:
This is a very different problem to what I'm seeing. Here reformat seems to be being run on text files and markdown?!

julien51:
This is failing because of an out of date package-lock.json file. See https://github.com/unlock-protocol/unlock/pull/726

julien51:
Ok, @benwerd I think this should be mostly good now? PTAL?

julien51:
FYI I added 2 files whose styles were not right :/

PR #718, opened 2018-12-14
Introduces eslint-check to check if eslint rules conflict with prettier

benwerd:


benwerd:
@akeem Great call!

julien51:
huhhhh, sorry I did not hit submit on this one but I had a tension :(
We already use `"lint": "./node_modules/eslint/bin/eslint.js .",` Maybe worth modifying it rather than introduing a new one?

Issue #363, opened 2018-10-16
Add /smart-contracts/build to .gitignore


hakimelek:
I am not sure if you think that this is worth doing or not but I think `/smart-contracts/build` is being generate on every `npm prestart` thus have different versions for different users. Maybe we can move this to .gitignore?

julien51:
Yeah, this is an issue, but we cannot put this into .gitignore either because these artifacts actually also keep track of the smart contracts being deployed on the actual networks (which we absolutely want to keep track of!).
We need to research this further. Ideally, we'd have multiple artifacts file for each smart contract, on each network.

julien51:
I think @nfurfaro is looking into this!

cellog:
Are you assuming that users will install from git? The solution to this issue could be to instead release unlock as an npm package. This would also address #468.

Either way, transitive files like this should not be stored in the application, but in a user-configured location or perhaps in a database (my opinion)

Adding a script that installs and configures unlock (a la create-react-app) would be an elegant solution to both.

nfurfaro:
I'm looking into this. 
As pointed out above by @hakimelek, `/build/contracts` is indeed regenerated/overwritten each time `truffle migrate` is called. `truffle deploy` is an alias for `truffle migrate`. Given the same source to compile (our contracts) and the same compiler version, the output in `/build/contracts` will always be the same.

julien51:
Well, I think the artifacts also kept track of the address where to the smart contract is deployed which means that every time it is being deployed we get a new "version" of the file.
Since we all use different ganache instances we get a different version of the file every time.

We need to find a solution to that.

julien51:
@cellog the react app needs to know the address of the deploy smart contracts but also the "abi" (which is really a description of the interfaces offered by the smart contract). Both of these things are included in the artifacts files.

So in a way the artifacts are more "config" files for the react app than transient files which we could discard easily. The challenge comes from the fact that for each smart contract, there is only a single artifact file which contains the addresses of the deployed smart contract on each network (the test/local ones... which are not important to keep, but also the "production" versions which are extremely important to keep).

I hope this clarifies things.

nfurfaro:
> Yeah, this is an issue, but we cannot put this into .gitignore either because these artifacts actually also keep track of the smart contracts being deployed on the actual networks (which we absolutely want to keep track of!).
> We need to research this further. Ideally, we'd have multiple artifacts file for each smart contract, on each network.

[...]

PR #721, opened 2018-12-14
RFC: Introduces an initial foray into a backend service for unlock

akeem:
Disclaimer: This is very rough the following items are known as needing to be addressed:

* Updated test for the locksReducer
* Stronger tests for the REST interface around data store
* The URL and port are fixed to support initial development
* Staging & Production deployments will need real world databases and accompanying deployment scripts

With that inconsideration, here is general idea around this rough draft. We have an immediate use case of supporting the storage and presentation of Lock names for our initial release. The data layer supports this use case with an approach that should ideally make it trivial/easy to expand.

In this initial pass security has been punted on, but should be implemented before making it widely available in order to avoid any unintended consequences.

The service is provided as node application with a REST interface servicing lock reads, creations, and updates (data being names and addresses in the case of an initial block deployment).
At the moment, the platform functions with the blockchain being it's only datasource. We should keep this in account and ensure that app remains as functional as possible without the data service

### DIFF
--- a/locksmith/__tests__/app.test.js
+++ b/locksmith/__tests__/app.test.js
@@ -1,0 +1,84 @@
+const request = require("supertest")
+const app = require("../app")
+const Lock = require("../sequelize")
+
+beforeAll(() => {
+    Lock.create({ name: "Test Lock", address: "0xbac" })
+})
+
+afterAll((done) => {
+    Lock.truncate()
+    done()
+})
+
+describe('Requesting lock details', () => {
+    describe("when the lock details are available", () => {
+        test('it should return the name of the lock', () => {
+            return request(app).get('/lock/0xbac').then(response => {
+                expect(response.body).toMatchObject({ "name": "Test Lock" })
+            })
+        })
+
+        test('should return an OK status code', () => {
+            return request(app).get('/lock/0xbac').then(response => {
+                expect(response.statusCode).toBe(200);
+            })
+        })
+    })
+
+    describe("when the lock details are unavailable", () => {
+        test('it should returns an appropriate error code', () => {
+            return request(app).get('/lock/0xdeadbeef').then((response) => {
+                expect(response.statusCode).toBe(404);
+            })
+        })
+    })
+
+    describe('Setting lock details', () => {
+        test("it saves the passed the information", () => {
+            return request(app).post('/lock')
+                .set('Accept', /json/)
+                .send({ address: '0xnewaddress', name: 'creation' }).then((response) => {
+                    Lock.findOne({ where: { address: '0xnewaddress' } }).then(record => {
+                        expect(record.name).toBe('creation')
+                    })
+                })
+        })
+
+        test("it return a OK status code", () => {
+            return request(app).post('/lock').
+                send({ address: '0xnewaddress2', name: 'creation2' }).then((response) => {
+                    expect(response.statusCode).toBe(200)
+                })
+            
+        })
+    })
+
+    describe('Updating lock details', () => {
+        describe('when the lock exists', () => {
+            test('it updates the lock details', (done) => {
+                let update = { address: "0xNewAddress", currentAddress: "0xbac" };
+                request(app).put('/lock')
+                    .send(update)
+                    .then(response => {
+                        Lock.findOne({ where: { address: "0xNewAddress" } }).then(record => {
+                            expect(response.statusCode).toBe(202)
+                            console.log(record)
+                        })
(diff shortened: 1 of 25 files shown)


unlock-app/src/services/storageService.js
```diff
@@ -0,0 +1,21 @@
+import axios from 'axios'
+
+export default class StorageService {
+  HOST = 'http://localhost:4000'
```
akeem: Will definitely be removed and replaced with proper configuration aware lookup

julien51:
Hey, please remove the `*.json` files from your commit. We have an [outstanding issue](https://github.com/unlock-protocol/unlock/issues/363) with them. :(

locksmith/migrations/20181212150352-create-lock.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
akeem: This migration isn't particularly tightened for production, but should be adjustable within a few minutes when have a agreement that this general approach is a good path forward

julien51:
Ha! I should have been more clear :/ I meant you need to remove them from the commit but not delete them. They are actually required :(

akeem:
No sorry, that as on me. Getting this sorted so that the branch in green

unlock-app/src/middlewares/storageMiddleware.js
```diff
@@ -0,0 +1,39 @@
...
+      switch (action.type) {
+      case CREATE_LOCK:
+        storageService.storeLockDetails(action.lock)
+        break
+      case UPDATE_LOCK:
+        storageService.lockLookUp(action.address).then(results => {
+          dispatch(setLockName(action.address, results.data.name))
+        })
+        break
+      case ADD_LOCK:
```
julien51: I wonder if we should maybe move the `UPDATE_LOCK` processing down here. Since we just need the address and the `ADD_LOCK` action has it, maybe we can invoke `lockLookUp` here?

unlock-app/src/services/storageService.js
julien51: + security ;)

locksmith/sequelize.js
```diff
@@ -0,0 +1,4 @@
+const Sequelize = require('sequelize')
```
julien51: I am not sure this is being used below?

locksmith/migrations/20181212150352-create-lock.js
```diff
@@ -0,0 +1,30 @@
...
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('Locks', {
+      id: {
```
julien51: Would we want to use the `lock`'s `address` here has primary key?

locksmith/app.js
```diff
@@ -0,0 +1,56 @@
+const express = require('express')
+var cors = require('cors');
+var bodyParser = require('body-parser');
+const models = require('./models')
+const app = express()
+const Lock = require('./sequelize')
+
+var router = express.Router();
+
+router.route('/lock').put(function (req, res, next) {
```
julien51: I wonder if it is actually even worth "replacing" the name for a lock? Couldn't we just add a new record with the deployed address (and keep the old one even if never used ever again).

akeem: This is a really good point!

akeem:
@julien51 Thank you for the feedback, it is super helpful! This is not merge ready yet, but wanted to ensure that we liked the general direction before piling on more code.

julien51:
Maybe @benwerd can chime on this too?

locksmith/__tests__/app.test.js
```diff
@@ -0,0 +1,84 @@
...
+            return test('it returns a 412 status code', () => {
+                request(app).put('/lock')
+                    .set('Accept', /json/)
+                    .send('{}').then((response) => {
+                        expect(response.statusCode).toBe(412)
+                    })
+            })
+        })
+    })
+})
```
benwerd: Tiny nit: let's leave an extra line break at the bottom of files. Thanks!

locksmith/app.js
benwerd: Careful about primary key here though - I agree with using `address`, but we'd want to make sure the new record with the stored name was the one that was always retrieved.

julien51:
@akeem What is the status of that PR?

akeem:
@julien51 closing now as this PR has been been succeeded by #883 & #888

Issue #730, opened 2018-12-16
On Windows "npm run dev" fails, '&' command not supported


swkidd:
Running 'npm run dev' from Windows console fails with error below. This is because starting a command in background is not supported in windows using the '&' command. 

To Reproduce: 
    1. Be running a windows computer with default command prompt
    2. clone commit with hash 9d840a56
    3. from unlock folder run 'npm install'
    4. change directory to unlock\unlock-app and run 'npm run dev' 

> unlock-app@0.1.0 dev C:\Users\skidd9\bounty_playground\unlock\unlock-app
> npm run setup-dev && nodemon src/server.js

> unlock-app@0.1.0 setup-dev C:\Users\skidd9\bounty_playground\unlock\unlock-app
> cd .. && (npm run start-ganache -- -b 3 &) && npm run deploy-lock

) was unexpected at this time.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! unlock-app@0.1.0 setup-dev: `cd .. && (npm run start-ganache -- -b 3 &) && npm run deploy-lock`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the unlock-app@0.1.0 setup-dev script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\skidd9\AppData\Roaming\npm-cache\_logs\2018-12-16T02_17_38_320Z-debug.lognpm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! unlock-app@0.1.0 dev: `npm run setup-dev && nodemon src/server.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the unlock-app@0.1.0 dev script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\skidd9\AppData\Roaming\npm-cache\_logs\2018-12-16T02_17_38_498Z-debug.log

cellog:
this is a unix-specific thing, yes. Perhaps we should also develop using parallels or vmware to ensure windows is supported?

julien51:
@cellog yes! I believe it would be good to deploy/run on CI setup on a windows env... maybe in the context of our work on https://github.com/unlock-protocol/unlock/issues/195 ?

julien51:
[...]

PR #733, opened 2018-12-16
fixed windows 'npm run dev' error using run-script-os

swkidd:
Per issue #730 

## Proposed Changes
Added dependency to 'run-script-os'
Split run-dev command into two versions based OS version
     - windows uses 'START /b <command>' as equivalent for unix based '&' command


julien51:
Thanks a lot! I have no way to test on Windows right now... but I just tested on my local OSX machine and this works as expected. LG!

PR #735, opened 2018-12-17
create symlink during 'npm run dev', os compat with 'run-script-os'

swkidd:
Fixes #
#594 
## Proposed Changes
create symlink during 'npm run dev' command
    - if symlink file exists, delete it and create a new one
    - symlink folder 'smart-contracts/build' with target 'unlock-app/src/artifacts'
different command versions for unix/win using run-script-os

## Tested for
- ubuntu 18.04.1 LTS
- Windows 10 (home)

### DIFF
--- a/unlock-app/package.json
+++ b/unlock-app/package.json
@@ -67,7 +67,10 @@
   },
   "engineStrict": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


julien51:
@cellog LG

PR #743, opened 2018-12-17
Only rm artifacts/ if it exists

jedahan:
I needed to do this to stop errors with `npm run-dev` on macOS.


julien51:
Thanks @jedahan ! I should have been more careful with https://github.com/unlock-protocol/unlock/pull/735

PR #749, opened 2018-12-17
check for and call metamask enable() if user is in privacy mode

### LINKED ISSUE #447: New Metamask API
We are using a deprecated Metamask API: they show this message in the console:
> ATTENTION: In an effort to improve user privacy, MetaMask will stop exposing user accounts to dapps by default beginning November 2nd, 2018. Dapps should call provider.enable() in order to view and use accounts. Please see https://bit.ly/2QQHXvF for complete information and up-to-date example code.

November 2nd is... in 3 days!

cellog:
Fixes #447

## Proposed Changes

- if current provider has an "enable()" method, we call it
- minor fixes to remove errors
- minor formatting

I can't get the "/provider" page to load any more, not sure why, so please verifty that before merging

### DIFF
--- a/unlock-app/src/__tests__/services/web3Service.test.js
+++ b/unlock-app/src/__tests__/services/web3Service.test.js
@@ -223,7 +223,7 @@ describe('Web3Service', () => {
     })
   })
 
-  describe('fail while enabling access to acount', () => {
+  describe('fail while enabling access to account', () => {
     const netVersion = Math.floor(Math.random() * 100000)
     let enable, error
 
--- a/unlock-app/src/__tests__/services/web3Service.test.js
+++ b/unlock-app/src/__tests__/services/web3Service.test.js
@@ -187,25 +187,83 @@ describe('Web3Service', () => {
       })
       web3Service.connect({ provider: 'CLOUD' })
     })
+  })
 
-    it('should call enable on a provider that supplies it', async done => {
+  describe('request enabling access to account', () => {
+    const netVersion = Math.floor(Math.random() * 100000)
+    let enable
+
+    beforeEach(done => {
       const { providers } = configure(global)
-      expect.assertions(1)
-      try {
-        providers.HTTP.enable = jest.fn(() => Promise.resolve())
+      nock.cleanAll()
 
-        web3Service = new Web3Service(providers)
+      enable = providers.HTTP.enable = jest.fn(() => Promise.resolve())
+      web3Service = new Web3Service(providers)
+      netVersionAndYield(netVersion)
 
-        await web3Service.enable({ provider: 'test' })
-        expect(providers.test.enable).toHaveBeenCalled()
-        done()
-      } finally {
-        delete providers.HTTP.enable
-        done()
+      UnlockContract.networks = {
+        [netVersion]: {
+          events: {},
+          links: {},
+          address: '0x3f313221a2af33fd8a430891291370632cb471bf',
+          transactionHash:
+            '0x8545541749873b42c96e1699c2e62f0f4062ca57f3398270423c1089232ef7dd',
+        },
       }
+
+      web3Service.on('network.changed', () => {
+        done()
+      })
+
+      return web3Service.connect(Object.assign({}, defaultState))
+    })
+
+    it('should call enable on a provider that supplies it', () => {
+      expect(enable).toHaveBeenCalled()
     })
   })
 
+  describe('fail while enabling access to acount', () => {
+    const netVersion = Math.floor(Math.random() * 100000)
+    let enable, errors
+
+    beforeEach(done => {
+      const { providers } = configure(global)
+      nock.cleanAll()
+
+      enable = providers.HTTP.enable = jest.fn(() => Promise.reject())
+      errors = []
+
+      web3Service = new Web3Service(providers)
+      netVersionAndYield(netVersion)
+
+      UnlockContract.networks = {
+        [netVersion]: {
+          events: {},
+          links: {},
+          address: '0x3f313221a2af33fd8a430891291370632cb471bf',
+          transactionHash:
+            '0x8545541749873b42c96e1699c2e62f0f4062ca57f3398270423c1089232ef7dd',
+        },
+      }
+
+      web3Service.on('error', e => {
+        errors.push(e)
+      })
+      web3Service.on('network.changed', () => {
+        done()
+      })
+
+      return web3Service.connect(Object.assign({}, defaultState))
+    })
+
+    it('should error if a user rejects access', () => {
+      expect(enable).toHaveBeenCalled()
+      expect(errors).toHaveLength(1)
+      expect(errors[0].message).toBe('User canceled access to ethereum wallet, cannot continue')
+    })
+  })
+  
   describe('once connected', () => {
     const lockAddress = '0x0d370b0974454d7b0e0e3b4512c0735a6489a71a'
     const netVersion = Math.floor(Math.random() * 100000)
@@ -226,7 +284,6 @@ describe('Web3Service', () => {
       web3Service.on('network.changed', () => {
         done()
       })
-
       return web3Service.connect(Object.assign({}, defaultState))
     })
 
--- a/unlock-app/src/__tests__/services/web3Service.test.js
+++ b/unlock-app/src/__tests__/services/web3Service.test.js
@@ -211,7 +211,7 @@ describe('Web3Service', () => {
         },
       }
 
-      web3Service.on('network.changed', () => {
+      web3Service.once('network.changed', () => {
         done()
       })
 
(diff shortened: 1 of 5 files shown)


unlock-app/src/pages/provider.js
```diff
@@ -17,7 +17,7 @@ export function Web3Provider({ setProvider, config, provider }) {
         <title>{pageTitle('Pick Web3 Provider')}</title>
       </Head>
       <NoSSR>
-        <p>
+        <div>
```
julien51: I'm ok with this change here because this page is not "user facing" but I think generally everytime we can use semantic elements we should ;) `div` are sad :(

unlock-app/src/services/web3Service.js
```diff
@@ -47,7 +47,7 @@ export default class Web3Service extends EventEmitter {
    * @param {object} account
    * @return
    */
-  connect({ provider }) {
+  async connect({ provider }) {
```
julien51: Will this have other side effect?

unlock-app/src/services/web3Service.js
```diff
@@ -65,6 +65,17 @@ export default class Web3Service extends EventEmitter {
 
     this.web3 = new Web3(providers[provider])
 
+    if (providers[provider].enable) {
```
julien51: I *think* this is the right approach BUT I want to make sure that the other pages (like `/` or `/about`) are still visible to users who do not have a Metamask wallet in their browser. Do you confirm that this works?

Also, do you mind writing a test for this? (probably mocking/stubbing `.enable()`?

cellog:
oops! forgot testing altogether. Highly unlike me :)

I'll split the other stuff into a separate PR

cellog:
I got it! fully tested without any restructuring. ready for review

unlock-app/src/__tests__/services/web3Service.test.js
```diff
@@ -6,6 +6,7 @@ import nock from 'nock'
 import Web3Service from '../../services/web3Service'
 import UnlockContract from '../../artifacts/contracts/Unlock.json'
 import LockContract from '../../artifacts/contracts/PublicLock.json'
+import configure from '../../config'
```
julien51: Since you're just using the `provider` property, maybe you can skip that and `configure(global)` and just create a fake `providers` object?

unlock-app/src/__tests__/services/web3Service.test.js
```diff
@@ -188,6 +189,81 @@ describe('Web3Service', () => {
...
+        [netVersion]: {
+          events: {},
+          links: {},
+          address: '0x3f313221a2af33fd8a430891291370632cb471bf',
+          transactionHash:
+            '0x8545541749873b42c96e1699c2e62f0f4062ca57f3398270423c1089232ef7dd',
+        },
+      }
+
+      web3Service.on('network.changed', () => {
```
julien51: Could you use `once` here? This would ensure that if one reason or another this is called twice this would fail.

unlock-app/src/__tests__/services/web3Service.test.js
```diff
@@ -188,6 +189,81 @@ describe('Web3Service', () => {
...
+  describe('fail while enabling access to acount', () => {
+    const netVersion = Math.floor(Math.random() * 100000)
+    let enable, errors
+
+    beforeEach(done => {
+      const { providers } = configure(global)
+      nock.cleanAll()
+
+      enable = providers.HTTP.enable = jest.fn(() => Promise.reject())
+      errors = []
```
julien51: why use an array here?

unlock-app/src/__tests__/services/web3Service.test.js
```diff
@@ -188,6 +189,81 @@ describe('Web3Service', () => {
...
+          events: {},
+          links: {},
+          address: '0x3f313221a2af33fd8a430891291370632cb471bf',
+          transactionHash:
+            '0x8545541749873b42c96e1699c2e62f0f4062ca57f3398270423c1089232ef7dd',
+        },
+      }
+
+      web3Service.on('error', e => {
+        errors.push(e)
```
julien51: Maybe easy enough to just use a single `error` and only trigger `once` to make sure only the one error we care about is triggered?

unlock-app/src/services/web3Service.js
```diff
@@ -19,8 +19,9 @@ export const keyId = (lock, owner) => [lock, owner].join('-')
  * upstream objects.
  */
 export default class Web3Service extends EventEmitter {
-  constructor() {
+  constructor(myproviders = providers) {
```
julien51: camelCase ;) `myProviders`

unlock-app/src/services/web3Service.js
```diff
@@ -19,8 +19,9 @@ export const keyId = (lock, owner) => [lock, owner].join('-')
...
     super()
+    this.providers = myproviders
```
julien51: Maybe rename is `this.availableProviders` as it gives a bit more context about what that variable holds?

unlock-app/src/__tests__/services/web3Service.test.js
cellog: I tried this, but we need other aspects of the provider internally for the tests to pass

cellog:
@julien51 ready for new review!

unlock-app/src/services/web3Service.js
```diff
@@ -19,8 +19,9 @@ export const keyId = (lock, owner) => [lock, owner].join('-')
  * upstream objects.
  */
 export default class Web3Service extends EventEmitter {
-  constructor() {
+  constructor(availableProviders = providers) {
```
julien51:
```suggestion
  constructor(availableProviders) {
```

unlock-app/src/services/web3Service.js
```diff
@@ -47,45 +48,54 @@ export default class Web3Service extends EventEmitter {
...
     // We fail: it appears that we are trying to connect but do not have a provider available...
-    if (!providers[provider]) {
+    const provider = this.providers[providerName]
+    if (!provider) {
       return this.emit('error', new Error('Provider does not exist'))
     }
 
-    this.web3 = new Web3(providers[provider])
+    try {
+      if (provider.enable) {
```
julien51: It would actually be worth testing if other types of wallet (on mobile for example) do implement a similar `enable` function.

Maybe worth trying if Brave and Opera (which both have crypto wallets in them) work?

cellog: Brave wallet doesn't even show up in https://localhost:3000/provider and neither does Opera. I also can't find any documentation - at all - about how to develop for these wallets, so I'm unclear on where to even begin here. We can talk tomorrow.

unlock-app/src/services/web3Service.js
cellog: This change will require every instance of constructing a Web3Service to pass in providers. The lockMiddleware would explicitly include it then, and call configure. Is that what we want to do? I can support this, just want to make sure that's the direction intended first.

julien51: Ohhh you're right. Then i am confused as to why we pass `availableProviders` at all. It seems weird that we include these both via `configure(global)` above and when instantiating a new `Web3Service`. What do you think?

unlock-app/src/services/web3Service.js
cellog: confirmed: all wallets we need to worry about provide "enable" which is basically metamask and opera. Brave uses metamask as its wallet in a chromium extension, mist and parity are still too difficult to install to have large market share. When that changes we can look at them again.

unlock-app/src/services/web3Service.js
cellog: The issue is that the test needs to inject an "enable" method into the HTTP provider to mock it. However, we need the rest of the provider code in order to get the stage where "connect" will actually work and call enable.  I considered injecting "enable" into the global HTTP provider, but this will literally affect all of the other unrelated tests. Since that is a dangerous anti-pattern (potentially affecting the execution of unrelated tests), I decided the best approach was to modify Web3Service to accept an injected dependency.

There are a couple of ways of doing this. The minimally invasive method is to provide a parameter, and then set its default value to the global providers, which is what I chose in order to reduce changes.

Another (and probably best) alternative is to find all the occurrences of `new Web3Service` and pass in a newly created configuration. Fortunately, in the current codebase, the only place that happens is in `src/middlewares/lockMiddleware.js`. This solution would febreze 2 code smells:

 1. use of global variable inside Web3Service
 2. hidden dependency inside lockMiddleware on the `config` global variable in Web3Service

As a consequence, this also allows us to easily mock the injected dependency for any future changes needed.

Ultimately, I think Web3Service has too many hidden dependencies and would benefit from a refactor, but that is well beyond the scope of this PR!

julien51: Ok, I think you're right, let's just put a BIG comment above `constructor(availableProviders = providers) {` indicating why we do that (for tests...) and why this is a bad idea but that we can hopefully get rid of that code very soon as part of a big web3Service refactor. 

After that, LG!

cellog: awesome. Thanks Julien!

PR #751, opened 2018-12-17
Display error in Dashboard when user account is not present, remove warnings, minor formatting

cellog:
removes warnings in console on viewing provider page, and on viewing dashboard. A comment is made more readable as well.

The dashboard now displays an error on user account not being present, after a 500ms delay (which is configurable)

screen shot 2018-12-18 at 3 40 09 pm

Component is fully tested and has a new story under dashboard to show the error state.


julien51:
I think the warning is actually legit and we should require an account to be passed to the Dashboard :(

The correct course of action here is probably to not render the Dashboard until an account is passed to it. What do you think?

cellog:
That is basically what happens now, but I guess you're suggesting that the bail out be in the dashboard page instead of the component?

julien51:
Yes exactly, we should not invoke `<Dashboard />` without passing it an actual account.

cellog:
@julien51 ready for final review! (I hope)

unlock-app/src/components/helpers/SuspendedError.js
```diff
@@ -0,0 +1,45 @@
...
+    delay: 200,
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      show: false,
+    }
+    this.mounted = false
+    this.clearTimeout = false
```
julien51: This will make typechecking upset later :/

unlock-app/src/pages/dashboard.js
```diff
@@ -30,13 +41,14 @@ export const Dashboard = ({ account, network, locks }) => {
...
 Dashboard.propTypes = {
-  account: UnlockPropTypes.account.isRequired,
+  account: UnlockPropTypes.account,
   network: UnlockPropTypes.network.isRequired,
   locks: UnlockPropTypes.locks,
 }
 
 Dashboard.defaultProps = {
   locks: {},
+  account: null,
```
julien51: no need, right?

unlock-app/src/pages/dashboard.js
```diff
@@ -4,14 +4,25 @@ import NoSSR from 'react-no-ssr'
...
 
 export const Dashboard = ({ account, network, locks }) => {
   if (!account) {
-    return null //loading
+    return (
+      <SuspendedError>
+        <DefaultError title="User account not initialized">
+          <p>
+            In order to display your Unlock dashboard, you need to connect a
+            crypto-wallet to your browser
```
julien51:
```suggestion
            crypto-wallet to your browser.
```

cellog:
@julien51 one more review and I think we are done. 😩

PR #756, opened 2018-12-18
Update index.js

vikaskyadav:
Fixes #

I think `with` is a more suitable choice. Makes sense. Statements can be further simplified so as to make them easily understandable by the end user.

### DIFF
--- a/unlock-app/src/pages/index.js
+++ b/unlock-app/src/pages/index.js
@@ -57,8 +57,8 @@ export const Home = ({ config }) => (
         <SubTitle>Simple Integration</SubTitle>
         <ImageWithHover base="code" />
         <Paragraph>
-          Unlock provides a simple snippet of code to integrate easily on your
-          website, as well as several other integration tools...
+          Unlock provides a simple snippet of code to integrate easily with your
+          website, as well as with several other integration tools...
         </Paragraph>
       </Column>
       <Column>


julien51:
I think @benwerd is our word jedi, with @cellog as our padawan ;)

cellog:
If we're going to wordsmith the front page, I think more extensive changes would be warranted. For instance, the structure of our 3 callouts could be more similar.

1. There are no middlemen, no fees and no gatekeeper who could shut you down or control your distribution.
2. Unlock provides a simple snippet of code to integrate easily on your website, as well as several other integration tools...
3. For example, Unlock comes with a points system to reward your most loyal fans when they share your content with their friends!

We can streamline this:

1. No middlemen, no fees, no gatekeeper. You control your content and distribution
2. Easily integrate Unlock's plugin into your existing website with one line of code
3. Single locks, subscriptions, rewards points for loyal fans, and much much more

These are just one possibility. By eliminating superfluous words, we pack much more punch. However, I think this is a larger discussion to be had. In my opinion, changing `on` to `with` is not a huge impact difference or a grammar issue (both styles are technically correct English grammar)

@julien51 ? @benwerd ?

julien51:
Agreed with the general sentiment that rather than micro-optimize we can eventually do a big change. I am not sure this is urgent though...

smombartz:
@cellog like the proposed changes. streamlines the information flow.

vikaskyadav:
yeah, I too think, the language should be very easy and to the point.

unlock-app/src/pages/index.js
```diff
@@ -57,8 +57,8 @@ export const Home = ({ config }) => (
         <SubTitle>Simple Integration</SubTitle>
         <ImageWithHover base="code" />
         <Paragraph>
-          Unlock provides a simple snippet of code to integrate easily on your
-          website, as well as several other integration tools...
+          Unlock provides a simple snippet of code to integrate easily with your
+          website, as well as with several other integration tools...
```
benwerd:
```suggestion
          website, as well as several other integration tools...
```
No need for a second _with_ here!

benwerd:
I think _with_ is clearer grammar (while both are technically correct). Here's my take: this was a good first change, and while more is certainly possible, I'm grateful to have @vikaskyadav jump in and take a look. Everything is an iteration!

PR #761, opened 2018-12-18
downgrading react-redux to 5.0.7 in an effort to fix #757

### LINKED ISSUE #757: Locks are not displayed on demo page
**Describe the bug**
When going to a lock's demo page, the lock is not loaded. For example, see [lock was deployed](https://rinkeby.etherscan.io/address/0xd8c88be5e8eb88e38e6ff5ce186d764676012b0b) yesterday on rinkeby, but it is not shown on [its demo page](https://staging.unlock-protocol.com/demo/0x973f8FDFD2dD6b5D64DACc0BD012eC6F191d6F4C/). 

screenshot 2018-12-18 at 10 24 47

**Expected behavior**
The lock should be displayed!

julien51:
Fixes #757


unlock-app/package.json
```diff
@@ -42,7 +42,7 @@
     "react-dom": "^16.6.3",
     "react-jazzicon": "^0.1.2",
     "react-no-ssr": "^1.1.0",
-    "react-redux": "^6.0.0",
+    "react-redux": "^5.0.7",
     "react-router-dom": "^4.2.2",
     "react-router-redux": "^5.0.0-alpha.9",
```
cellog:
```suggestion
    "react-router-redux": "^4.0.8",
```

cellog: also, and more importantly

connect-react-router must be at version 5.0.1 not 6.0 beta. 6.0 requires react-redux 6

cellog: github won't let me suggest that change

julien51:
Unfortunately that was not it :( Even after the downgrade, locks are still not loaded :(
cc @cellog

cellog:
That's good, actually. We can revert and go back to 6 and then find the actual problem

PR #770, opened 2018-12-18
Removing docs folder - all content is now present on the wiki

benwerd:
GitHub is now the source of truth for documentation.


docs/contributors/coc.md
```diff
@@ -1,74 +0,0 @@
-# Contributor Covenant Code of Conduct
```
julien51: Wait! Is this in the wiki?

julien51: It is! So cool!

PR #771, opened 2018-12-18
Implement erc165

### LINKED ISSUE #758: We must implement the ERC165 interface as part of achieving ERC721 compatibility.
We want unlock-protocols "keys" to be compliant rerc721 NFT's. This is one of the required steps. It allows our contract to be queried to determine if we're erc721 compliant. Obviously, there's no point in doing this in isolation, but it's one step that needs to be done. 

We need to inherit the erc165 interface and implement its single function.
```
interface ERC165 {
    /// @notice Query if a contract implements an interface
    /// @param interfaceID The interface identifier, as specified in ERC-165
    /// @dev Interface identification is specified in ERC-165. This function
    ///  uses less than 30,000 gas.
    /// @return `true` if the contract implements `interfaceID` and
    ///  `interfaceID` is not 0xffffffff, `false` otherwise
    function supportsInterface(bytes4 interfaceID) external view returns (bool);
}
```

**Additional context**
Depends on #148, #748, #759
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md

nfurfaro:
Inherited the oz eth contract in PublicLock.sol.
Added tests to make sure erc165 and erc721 are implemented.

Fixes #758
## Important!
Prior to merging (assuming we want to proceed), I would like to test that inheriting a new interface and contract will not present any problems with upgradeability.
## Proposed Changes

- Implement erc165 spec (required for erc721 compatibility)
- Add 2 new tests


nfurfaro:
oops. I need to do a little formatting.

smart-contracts/contracts/PublicLock.sol
```diff
@@ -2,6 +2,7 @@ pragma solidity 0.4.24;
 
 import "./interfaces/ILockPublic.sol";
 import "./interfaces/IUnlock.sol";
+import "openzeppelin-eth/contracts/introspection/ERC165.sol";
```
julien51: Hey do you mind adding details about what ERC165 is for?

nfurfaro: Will do.

nfurfaro: @julien51 I added a comment in the contract as well, but ERC165 allows our contract to be queried to determine whether it implements a given interface. It's part of the ERC721 spec.
One great benefit it brings is the ability to test our interfaces themselves, and protect them against unconsidered changes. 
Eg: As it stands, our `PublicLock.sol` inherits the `ILockPublic.sol` interface. The compiler enforces that we implement each function speced in the interface. But, if we remove or comment out a function in the interface, our contract will still compile. Which is fine, but the next version/implementation of the contract no longer has the safety-net of the full interface.
With erc165 we can test that our contract implements a given interface directly, so that even if the interface is changed, we still have a test making sure we're compliant. If the test fails but the contract compiles, it can only mean that the interface has been modified.

nfurfaro:
Closing due to problems in branch. replaced by

smart-contracts/contracts/PublicLock.sol
julien51: Very cool! Ok, please add a small comment in the code and/or link to the ERC165 page?

nfurfaro: @julien51 Done. Also, this is now closed and replaced by #778

PR #778, opened 2018-12-19
Implements erc165

### LINKED ISSUE #758: We must implement the ERC165 interface as part of achieving ERC721 compatibility.

nfurfaro:
Inherited the oz eth contract in PublicLock.sol.
Added tests to make sure erc165 and erc721 are implemented.

Fixes #758
## Important!
Prior to merging (assuming we want to proceed), I would like to test that inheriting a new interface and contract will not present any problems with upgradeability.
## Proposed Changes

- Implement erc165 spec (required for erc721 compatibility)
- Add 2 new tests


nfurfaro:
Note: I had problems with my branch, so this is a fresh PR. sorry about that.

smart-contracts/build/contracts/Unlock.json
```diff
@@ -4248,8 +4248,14 @@
       "events": {},
       "address": "0xd8c88be5e8eb88e38e6ff5ce186d764676012b0b",
```
julien51: Ok, this is preserved so I am tempted to say we're all good here ;p

nfurfaro:
@julien51 well well. finally.

package-lock.json
```diff
@@ -1092,8 +1092,7 @@
             "bindings": {
               "version": "1.3.0",
               "resolved": "https://registry.npmjs.org/bindings/-/bindings-1.3.0.tgz",
-              "integrity": "sha512-DpLh5EzMR2kzvX1KIlVC0VkC3iZtHKTgdtZ0a3pglBZdaQFjt5S9g9xd1lE+YvXyfd6mtCeRnrUfOLYiTMlNSw==",
-              "dev": true
```
nfurfaro: Any Idea why this would have changed?

julien51:
Yay!

package-lock.json
nfurfaro: cc @julien51

julien51: This is something I have been trying to understand for weeks. npm is weird. this has no impact.

nfurfaro: Ok, thx.

PR #785, opened 2018-12-19
Unlock.js autodetects domain it's hosted from

benwerd:
This is one of two PRs that will make it easier to test from staging, and also for third parties to self-host Unlock code and run code. Previously, unlock.js's hosted domain needed to be configured. Now it's autodetected - although it can still be hand-configured.


julien51:
Yeah, me neither. I am also a bit worried that this is not going to be minified and co...

julien51:
Ben, I think you can merge now, but please open an issue about the testing needs ;)

benwerd:
Will do.

PR #798, opened 2018-12-20
fix flash of lock on locked page after purchase of the lock

### LINKED ISSUE #792: Flash of lock visible on unlocked page
When viewing an unlocked page, there is a flash of the "purchase a lock" screen until the account and network have fully loaded. This happens for as little as a few milliseconds or as long as a quarter second.

A good solution would be to use `SuspendedError` to delay displaying the lock for 200ms, and then display a spinner until account is ready

cellog:
# Description

This fixes #792 

No visual change to content occurs, so no screenshots attached. The only change is that instead of a brief flash of the "purchase a lock" page, the content takes slightly longer to load and then displays only the content.

However, because there is no flash of the lock, it appears to load much faster! I am not sure if this is a mental trick or if the fact that we aren't rendering the entire iframe and then removing it is actually slower.

# Checklist:

- [ ] If my code adds or changes components, I have written correspoding stories with Storybook

### DIFF
--- a/unlock-app/src/__tests__/components/lock/ShowUnlessUserHasKeyToAnyLock.test.js
+++ b/unlock-app/src/__tests__/components/lock/ShowUnlessUserHasKeyToAnyLock.test.js
@@ -49,8 +49,11 @@ describe('ShowUnlessUserHasKeyToAnyLock', () => {
       expect(wrapper.queryByText('Show me')).toBeNull()
     })
 
-    it('should show the children if there is a modal', () => {
+    it('should show nothing if accounts are not loaded and 500ms has not elapsed (no flash)', () => {
       expect.assertions(1)
+
+      jest.useFakeTimers()
+
       const keys = [
         {
           id: 'keyId',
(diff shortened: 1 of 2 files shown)


julien51:
Hey, I don't think you need to add labels to PRs as long as there are on the issues ;)

unlock-app/src/__tests__/components/lock/ShowUnlessUserHasKeyToAnyLock.test.js
```diff
@@ -49,8 +49,33 @@ describe('ShowUnlessUserHasKeyToAnyLock', () => {
       expect(wrapper.queryByText('Show me')).toBeNull()
     })
 
+    it('should show nothing if accounts are not loaded and 500ms has not elapsed (no flash)', () => {
```
julien51: 500ms or 200?

cellog: yeah I was confused :) you can watch the confusion in action in the video I uploaded to google drive. Will fix

Issue #804, opened 2018-12-20
replace /provider with a developer overlay dropdown to choose metamask on localhost


cellog:
**Is your feature request related to a problem? Please describe.**
It is often inconvenient to go to /provider, and then click back, especially after a reload is necessary

**Describe the solution you'd like**
A simple overlay dropdown that allows choosing Metamask, and shows which network is selected, which would work on any page, whether it is dashboard or demo or another future page

**Describe alternatives you've considered**
Haven't thought of any alternatives

Issue #580, opened 2018-11-20
Snapshots for storybook


julien51:
Storybook can automatically create snapshots which would make testing very useful.
We tried that in the past but we had issues... we should try again!

julien51:
This would significantly increase our test coverage and avoid "broken" stories. cc @cellog

cellog:
closed by #827

PR #823, opened 2018-12-21
Add storybook knobs addon

cellog:
# Description

This is a sub-PR split off from #804 to enable configuration changes needed to test how the developer overlay can be manipulated inside storybook. It also paves the way for
the fix to #580 

# Checklist:

### DIFF
--- a/unlock-app/.storybook/addons.js
+++ b/unlock-app/.storybook/addons.js
@@ -1,0 +1,1 @@
+import '@storybook/addon-knobs/register'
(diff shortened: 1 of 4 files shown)


julien51:
Could you quickly explain why the new rule?

cellog:
yes. Storybook addons are not needed in the production build (for obvious reasons), but they do need to be imported into the stories. eslint by default warns if you try to import a devDependency because these will disappear in production. This rule removes that warning ONLY for stories (anything in the `stories/` directory.

cellog:
I will wait to merge until you've had a chance to respond @julien51 ! Just so we're on the same page

PR #837, opened 2018-12-22
Now showng a better error when declining to create lock and remove the pending lock

julien51:
This fixes issue #531.

### DIFF
--- a/unlock-app/src/__tests__/middlewares/lockMiddleware.test.js
+++ b/unlock-app/src/__tests__/middlewares/lockMiddleware.test.js
@@ -3,9 +3,10 @@ import { LOCATION_CHANGE } from 'react-router-redux'
 import lockMiddleware from '../../middlewares/lockMiddleware'
 import {
   ADD_LOCK,
-  UPDATE_LOCK,
   CREATE_LOCK,
+  DELETE_LOCK,
   LOCK_DEPLOYED,
+  UPDATE_LOCK,
   WITHDRAW_FROM_LOCK,
   UPDATE_LOCK_KEY_PRICE,
 } from '../../actions/lock'
@@ -16,7 +17,7 @@ import { SET_PROVIDER } from '../../actions/provider'
 import { ADD_TRANSACTION, UPDATE_TRANSACTION } from '../../actions/transaction'
 import { SET_ERROR } from '../../actions/error'
 import { SET_KEYS_ON_PAGE_FOR_LOCK } from '../../actions/keysPages'
-import { PGN_ITEMS_PER_PAGE } from '../../constants'
+import { PGN_ITEMS_PER_PAGE, TRANSACTION_TYPES } from '../../constants'
 
 /**
  * Fake state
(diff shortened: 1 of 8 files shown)


julien51:
screenshot 2018-12-21 at 23 23 14

unlock-app/src/middlewares/lockMiddleware.js
```diff
@@ -98,7 +99,14 @@ export default function lockMiddleware({ getState, dispatch }) {
     dispatch(updateTransaction(transaction.hash, update))
   })
 
-  web3Service.on('error', error => {
+  web3Service.on('error', (error, transaction) => {
+    if (transaction && transaction.type === TRANSACTION_TYPES.LOCK_CREATION) {
+      // delete the lock
+      dispatch(deleteLock(transaction.lock))
+      return dispatch(
+        setError(<p>Failed to create log. Did you decline the transaction?</p>)
```
cellog: Sending jsx to a reducer is usually an anti-pattern, because it disrupts things like the ability to time travel debug, and also to serialize state. Better is to define a set of limited possible values with some display metadata and construct the jsx when the error is rendered

unlock-app/src/reducers/locksReducer.js
```diff
@@ -81,6 +82,16 @@ const locksReducer = (state = initialState, action) => {
     }
   }
 
+  if (action.type === DELETE_LOCK) {
+    if (!state[action.address]) {
```
cellog: Errors should not be thrown in a reducer. The error condition should be detected either at user input, or if input could come asynchronously from web3, then in a middleware, which would then dispatch an error action.

unlock-app/src/__tests__/middlewares/lockMiddleware.test.js
```diff
@@ -334,15 +336,38 @@ describe('Lock middleware', () => {
...
+      )
+    })
+
+    it('it should handle error events triggered by the web3Service', () => {
+      const { store } = create()
+      mockWeb3Service.emit('error', { message: 'this was broken' })
+      expect(store.dispatch).toHaveBeenCalledWith(
+        expect.objectContaining({
+          type: SET_ERROR,
+          error: expect.anything(), // not sure how to test against jsx
```
cellog: This would be solved with comments below about how to store and to note error conditions

unlock-app/src/reducers/locksReducer.js
julien51: Hum. Do you have any more data on that?
This is also a pattern used all over the place and that worked fine?
My worry is that we obviously do not want to check the state for every action triggered (this would duplicate the effort... and, more importantly means that anyone triggering an action should know the state before which seems like an anti-pattern.

unlock-app/src/__tests__/middlewares/lockMiddleware.test.js
julien51: Good point!

unlock-app/src/reducers/locksReducer.js
julien51: Now that I think about it, I will actually delete all these errors because they are actually never triggered. They are more like `asserts` in a way. I'll submit a PR which removes them all.

julien51:
Yes! sorry for the delay on this one. https://github.com/unlock-protocol/unlock/issues/876

cellog:
I am happy to work on the errors stuff if it is the highest priority, I have made some progress, but nowhere near a PR for it

cellog:
see #978 #979 #982 to unblock this PR.

unlock-app/src/__tests__/middlewares/lockMiddleware.test.js
```diff
@@ -335,15 +337,38 @@ describe('Lock middleware', () => {
...
+      expect(store.dispatch).toHaveBeenNthCalledWith(1, {
+        type: DELETE_LOCK,
+        address: transaction.lock,
       })
-    )
+      expect(store.dispatch).toHaveBeenNthCalledWith(
+        2,
+        expect.objectContaining({
+          type: SET_ERROR,
+          error: 'Failed to create log. Did you decline the transaction?',
```
cellog: log or lock? probably lock but I don't know the code as well :)

julien51: lol! Typo! that is meant to be lock ;)

julien51:
Oh and let me deal with these conflicts!

PR #839, opened 2018-12-23
Implemented getKeysForLockOnPage to get keys by page on lock

julien51:
In order to show the list of keys, we implement a function which paginates over keys in a lock.
This first version of it is fairly ineficient but it is the best we can do until more changes are done
at the smart contract level.

### DIFF
--- a/unlock-app/src/services/web3Service.js
+++ b/unlock-app/src/services/web3Service.js
@@ -567,29 +567,66 @@ export default class Web3Service extends EventEmitter {
    */
   getKeyByLockForOwner(lock, owner) {
     const lockContract = new this.web3.eth.Contract(LockContract.abi, lock)
-
-    const getKeyExpirationPromise = lockContract.methods
-      .keyExpirationTimestampFor(owner)
-      .call()
-    const getKeyDataPromise = lockContract.methods.keyDataFor(owner).call()
-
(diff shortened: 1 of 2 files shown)


unlock-app/package.json
```diff
@@ -120,4 +120,4 @@
     "prettier-eslint-cli": "^4.7.1",
     "react-testing-library": "^5.2.3"
   }
-}
\ No newline at end of file
+}
```
cellog: ack!

unlock-app/src/services/web3Service.js
```diff
@@ -567,29 +567,66 @@ export default class Web3Service extends EventEmitter {
...
+        .catch(() => {
+          return resolve([0, null])
+        })
+    })
+  }
+
+  /**
+   * This loads and returns the keys for a lock per page
+   * This function is performing quite badly because it retrieves n owners one by one
+   * and then gets their keys. We need to implementation functions on the smart contract
```
cellog:
```suggestion
   * and then gets their keys. We need to implement functions on the smart contract
```

unlock-app/package.json
cellog: Just have this change to package.json, which I will assume you'll squash

julien51: Oups! Indeed!

PR #840, opened 2018-12-23
Not throwing errors in reducers but keep state unchanged

julien51:
This is after @cellog noted that we should not throw errors in reducers inside PR #837

### DIFF
--- a/unlock-app/src/reducers/keysReducer.js
+++ b/unlock-app/src/reducers/keysReducer.js
@@ -24,11 +24,13 @@ const keysReducer = (state = initialState, action) => {
 
   if (action.type === ADD_KEY) {
     if (action.key.id && action.key.id !== action.id) {
-      throw new Error('Could not add key with wrong id')
+      // 'Could not add key with wrong id' => Let's not change state
+      return state
     }
 
     if (state[action.id]) {
-      throw new Error('Could not add already existing key')
+      // 'Could not add already existing key' => Let's not change state
+      return state
     }
 
     return {
@@ -42,11 +44,13 @@ const keysReducer = (state = initialState, action) => {
 
   if (action.type === UPDATE_KEY) {
     if (action.update.id && action.update.id !== action.id) {
-      throw new Error('Could not change the key id')
+      // 'Could not change the key id' => Let's not change state
+      return state
     }
 
     if (!state[action.id]) {
-      throw new Error('Could not update missing key')
+      // 'Could not update missing key' => Let's not change state
+      return state
     }
 
     return {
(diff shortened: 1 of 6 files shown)


unlock-app/src/reducers/keysReducer.js
```diff
@@ -24,11 +24,13 @@ const keysReducer = (state = initialState, action) => {
 
   if (action.type === ADD_KEY) {
     if (action.key.id && action.key.id !== action.id) {
-      throw new Error('Could not add key with wrong id')
+      // 'Could not add key with wrong id' => Let's not change state
+      return { ...state }
```
cellog: I would suggest just returning state, and using toBe in the tests, since the state didn't change. What do you think?

cellog: I could go either way, but in this case, the state is not changing, is it? So our output should be unchanged, otherwise we get an unnecessary render?

Perhaps one answer to this is to add a new `errorReducer` which is a "higher-order reducer." This would be a top-level reducer that simply listens for every one of these edge cases and then sets up error that triggers a re-render with a modal or however we choose to render it.

But in this limited case of the keyReducer, nothing in the state is actually changing in the key state, so I wouldn't mutate the state and force a re-render. As a more limited alternative to the higher-order reducer, it would be very simple to just add an error key to the state, so we get:

```js
if (action.type === ADD_KEY) {
  if (action.key.id && action.key.id !== action.id) {
    return { ...state, error: { message: 'Could not add key with wrong id', action } }
  }
}
```

Then we can listen for this in a connected error component, which could display a modal on dev with details, or a generic error message on production (and this could be listened to with a middleware as well, to send errors back to us, for instance). This would be my personal preference, I find higher-order reducers are often confusing when there are bugs, as state is mutated in more than one logical place.

Anyways, I still wouldn't mutate state without actually changing anything.

This is one of the most common things that comes up on the discord channel for redux, I consider it to be a weak spot in the design, personally.

julien51: Ok, this makes sense. Thanks for researching this. I'll just return un-changed state.

julien51:
@cellog PTAL ;)

unlock-app/src/reducers/transactionsReducer.js
```diff
@@ -25,11 +25,13 @@ const transactionsReducer = (transactions = initialState, action) => {
   // Replace the transaction with the updated value
   if (action.type === UPDATE_TRANSACTION) {
     if (action.update.hash && action.update.hash !== action.hash) {
-      throw new Error('Could not change the transaction hash')
+      // 'Could not change the transaction hash' => Let's not change state
+      return { ...transactions }
```
cellog: Do same here?

julien51: Oups! Thanks!

PR #845, opened 2018-12-24
Replaced old with new chevron svg

smombartz:
Swapped out the mobile menu close triangle chevron with a line chevron (top right of screenshot)

image


smombartz:
created components

julien51:
@smombartz Good one! You probably want to update the "snapshots" (these are code snapshots) which are generated to ensure that the markup generated does not change too much.

You can do that by typing `npm test -- -u` in your terminal inside `unlock-app` and them commiting the changes again. You can push again and the tests should pass this time!

PR #863, opened 2018-12-29
Removed tests.sh and parallelize services

julien51:
This removes the tests.sh approach and now each running container has its own command to run tests. Each container now also runs in parallel which should make builds slightly faster but also less likely to grow as we add services.


docker/docker-compose.ci.yml
```diff
@@ -6,11 +6,13 @@ services:
     expose:
       - 8545
     command: ["-h=0.0.0.0", "-m=hello unlock save the web"]
-  unlock:
-    build:
-      context: ..
-      dockerfile: ./docker/unlock.dockerfile
+  smart-contracts:
+    image: unlock
```
akeem: This is more of a lack of understanding of docker on my part, but what occurs in this scenario of if the unlock image isn't currently available on this machine?

julien51: So, this is a good question. I just tested and it docker-compose refuses to build. I guess this is expected.

PR #868, opened 2018-12-31
Adding integration tests

julien51:
Introducing integration tests (with just a single dummy test for now!).
The PR adds a new script to travis which runs integration tests in a new docker image.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -7,4 +7,32 @@ This is useful to ensure that all blocks work "together" if they already work in
 We do not expect developers to necessarily run these tests for nost feature work but they will be
 run on every pull request as part of the CI/CD process.
 
-We use Puppeteer.
+We use Puppeteer and jest.
+
+## Writing tests
+
+We should test each "core" feature of the application. The approach is to write files in `/test`
+which perform these tests.
+
+## CI
+
+In CI, running these tests requires 3 docker images:
+
+- One with ganache, the local development ethereum node
+- One with the whole next.js application (connected to the ganache node)
+- One which actually runs puppeteer and executes the test suite.
+
+## Running locally
+
+It is possible to run the tests locally, without docker running. Additionnaly, running the test
+locally allows for a little more flexibility to debug. For example, even if the tests are running in
+a headless way by default, they can be run in an actual web browser which lets the developer see
+what is being rendered on screen.
+
+- start a local ganache node (at the root): `npm run start-ganache`
+- deploy the smart contracts (in `/smart-contracts`): `npm run deploy`
+- run the next application (in `/unlock-app`): `npm run start`
+- execute the tests (in `/tests`): `npm run test`
+
+There are a few interesting and useful debugging options [on this page](https://github.com/GoogleChrome/puppeteer#debugging-tips) including
+the ability to 'slow down' execution via `slowMo`...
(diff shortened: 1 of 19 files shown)


tests/README.md
```diff
@@ -4,7 +4,34 @@ This is run to test integration tests which test Unlock as a whole.
...
 
 This is useful to ensure that all blocks work "together" if they already work in isolation.
-We do not expect developers to necessarily run these tests for nost feature work but they will be
+We do not expect developers to necessarily run these tests for nost feature work but they will be 
 run on every pull request as part of the CI/CD process.
 
-We use Puppeteer.
+We use Puppeteer and jest.
+
+## Writting tests
```
cellog:
```suggestion
## Writing tests
```

tests/README.md
```diff
@@ -4,7 +4,34 @@ This is run to test integration tests which test Unlock as a whole.
...
+
+We should test each "core" feature of the application. The approach is to write files in `/test` 
+which perform these tests.
+
+## CI
+
+In CI, running these tests requires 3 docker images: 
+* One with ganache, the local development ethereum node
+* One with the whole next.js application (connected to the ganache node)
+* One which actually runs pupeteer and executes the test suite.
```
cellog:
```suggestion
* One which actually runs puppeteer and executes the test suite.
```

tests/README.md
cellog: oops force push killed this

tests/README.md
cellog: and this

tests/README.md
julien51: Ooops!

julien51:
Thanks for the feedback. Responses online:

> Could we make the test command in the global package.json call integration tests? 
I am actually not in favor of this. The `test` folder is for the integration tests. The `root` of the project does not necessarily mean that it will "cover" all of the sub folder. Think of it more as an organizational structure than a logical one.

Regarding your issue, this is actually intentional ;) If you look at the `travis.yml` file you will see that we build the docker images first. I believe we could/should at some point push the images to some docker repo, but for now we only rely on them being built (and hence defined) locally.
The reason I do not want to build the images *inside* the docker-compose script is because I want to be able to run things in parallel. If we build the images in the docker-compose file, my understanding is that the `smart-contract` and `unlock-app` tests will only run *after* the  `unlock-integration.dockerfile` has been built... which is a waste of time because these tests do not require that image to have been built.

Generally though, I think we should not expect to run docker locally to develop/run Unlock just yet. This would have a significant performance cost (building the images takes time) and would increase the requirements on the dev side (understanding docker, being able to debug it...) when we do not need it. At this point, the docker images are used to run CI and soon, to deploy.

tests/README.md
```diff
@@ -7,4 +7,32 @@ This is useful to ensure that all blocks work "together" if they already work in
...
+
+In CI, running these tests requires 3 docker images:
+
+- One with ganache, the local development ethereum node
+- One with the whole next.js application (connected to the ganache node)
+- One which actually runs puppeteer and executes the test suite.
+
+## Running locally
+
+It is possible to run the tests locally, without docker running. Additionnaly, running the test
```
cellog:
```suggestion
It is possible to run the tests locally, without docker running. Additionally, running the test
```

julien51:
Thanks!

PR #882, opened 2019-01-02
Triggering an event when the price of keys on a lock was changed

julien51:
We're now triggering an event when the price of keys was changed on a lock

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -26,6 +26,13 @@ contract PublicLock is ILockCore, ERC165, IERC721, Ownable {
     bytes data; // Note: This can be expensive?
   }
 
+  // Events
+  event PriceChanged(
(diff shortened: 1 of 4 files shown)


smart-contracts/contracts/PublicLock.sol
```diff
@@ -26,6 +26,12 @@ contract PublicLock is ILockCore, ERC165, IERC721, Ownable {
     bytes data; // Note: This can be expensive?
   }
 
+  // Events
+  event PriceChanged(
```
nfurfaro: Looks good. I typically use a pattern of logging both the old and new values in events, so `uint indexed oldKeyPrice, uint indexed newKeyPrice`. It's not strictly necessary, but it can help to achieve the ability to "reconstruct" the application state from the event logs only.

julien51: That's a good practice, I'll update the PR. Thanks!

PR #883, opened 2019-01-02
Adds the Locksmith application to unlock

akeem:
Introducing the Locksmith application

This is the first half of the changes required to introduce a storage service to the project. The abbreviated description of the functionality at this time is:

"An express application, utilizing a middleware that validates Ethereum signed JWTs upon request to store or update lock metadata"

The application provides a service allowing users of the unlock application with ther ability to store lock metadata supporting current and future needs of the platform

# Checklist:


docker/docker-compose.ci.yml
```diff
@@ -6,6 +6,14 @@ services:
     expose:
       - 8545
     command: ["-h=0.0.0.0", "-m=hello unlock save the web"]
+  db:
+    image: postgres
+    ports:
```
julien51: You probably only need to use `expose` here instead of `ports`. The difference IIRC is that `expose` only exposes this to other containers in the same "cluster" while `ports` will expose this outside of docker.
I think for security we should restrict this to `expose`.

docker/docker-compose.ci.yml
```diff
@@ -36,3 +44,11 @@ services:
       - ganache
       - unlock
     working_dir: /home/unlock/tests
+  locksmith:
+    image: unlock
+    user: node
+    working_dir: /home/unlock/locksmith
+    ports:
+      - 8080:8080
```
julien51: Same here as above.

locksmith/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:current
```
julien51: Is that still required since you use the same `unlock` image?

locksmith/config/config.json
```diff
@@ -0,0 +1,17 @@
+{
+  "development": {
+    "username": "locksmith_development",
```
julien51: Probably not for this PR but we should probably consider using env variables here, right?

locksmith/scripts/pending-changes.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
julien51: Now I am thinking we should do all that at the root level. (not in this PR tho)

locksmith/src/app.js
```diff
@@ -0,0 +1,66 @@
...
+const tokenMiddleware = require('./token_middleware')
+const Lock = require('./sequelize')
+
+const app = express()
+const router = express.Router()
+
+router
+  .route('/lock')
+  .put(function(req, res) {
+    let newAddress = req.body.address
```
julien51: I think a more restful approach would be to add `put` on `/lock/:lockAddress` rather than pass an `address` in the body?

locksmith/.eslintignore
```diff
@@ -0,0 +1,8 @@
+/build
+/coverage
+/__tests__/utils/withConfig/__snapshots
+/src/.next
+/src/out
+/.cache
+/models
```
julien51: Hum. You probably want `/models` to be eslint-ed? Also please cleanup the `/build`, ` /__tests__/utils/withConfig/__snapshots`, ` /src/*` since they're not applicable here?

locksmith/.eslintignore
```diff
@@ -0,0 +1,8 @@
...
+/migrations
```
julien51: same here. we actually want that linted, right?

locksmith/.eslintignore
akeem: Thanks for pointing this out, I added `models` and `migrations` here because the content was autogenerated. Will clean up the other bits

.dockerignore
```diff
@@ -1 +1,2 @@
-**/node_modules
\ No newline at end of file
+**/node_modules
+**/npmp-debug.log
```
cellog:
```suggestion
**/npm-debug.log
```

docker/docker-compose.ci.yml
```diff
@@ -6,6 +6,14 @@ services:
     expose:
       - 8545
     command: ["-h=0.0.0.0", "-m=hello unlock save the web"]
+  db:
+    image: postgres
+    expose:
+      - 5432
+    environment:
+      POSTGRES_USER: "locksmith_test"
```
cellog: I don't think this is a good idea, we can use Docker secrets and an environment variable for the user/password. I recognize this is a MVP user/password, but it's relatively trivial to set up Docker secret. We can talk tomorrow about it

locksmith/__tests__/app.test.js
```diff
@@ -0,0 +1,110 @@
+const request = require('supertest')
+const app = require('../src/app')
+const Lock = require('../src/sequelize')
+
+beforeEach((done) => {
```
cellog: nit - I think we should not use parents around single argument arrow functions

locksmith/__tests__/app.test.js
```diff
@@ -0,0 +1,110 @@
...
+  Lock.create({ name: 'Test Lock', address: '0xbac', owner: '0xdeadbed123' }).then(() => {
```
cellog: Can you use a proper length address for the test? I think it's 24 characters. You can copy from the CreatorLock storybook stories.

Also, use a constant for it and then reference the constant in the tests, will make refactoring easier if lock format changes

locksmith/__tests__/app.test.js
```diff
@@ -0,0 +1,110 @@
...
+  })
+
+  describe('Setting lock details', () => {
+    test('it saves the passed the information', (done) => {
+      Date.now = jest.fn(() => 1546130837000)
+      return request(app).post('/lock')
+        .set('Accept', /json/)
+        .set(
+          'Authorization',
+          'Bearer eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJsb2NrIjp7ImFkZHJlc3MiOiJqcWE2ZG5wMSIsIm5hbWUiOiJvaW9pb2kiLCJleHBpcmF0aW9uRHVyYXRpb24iOjI1OTIwMDAsImtleVByaWNlIjoiMTAwMDAwMDAwMDAwMDAwMDAiLCJtYXhOdW1iZXJPZktleXMiOjEwLCJvd25lciI6IjB4QWFBZEVFRDRjMEI4NjFjQjM2ZjRjRTAwNmE5QzkwQkEyRTQzZmRjMiJ9LCJpYXQiOjE1NDYxMzA4MzUsImV4cCI6MTU0NjEzMDgzOCwiaXNzIjoiMHhBYUFkRUVENGMwQjg2MWNCMzZmNGNFMDA2YTlDOTBCQTJFNDNmZGMyIn0.0xd36a7842eda1b2267c08b204407f9e71024a2152ae772882a1fa41a02696aee578688f02ef6c301451de95793bde669198359f0820fb35f6ba1303e89727a52601'
```
cellog: Again I would extract this into a constant and reuse

locksmith/__tests__/app.test.js
```diff
@@ -0,0 +1,110 @@
...
+        )
+        .send({ 'address': 'jqa6dnp1', 'name': 'oioioi', 'expirationDuration': 2592000, 'keyPrice': '10000000000000000', 'maxNumberOfKeys': 10, 'owner': '0xAaAdEED4c0B861cB36f4cE006a9C90BA2E43fdc2' }).then(() => {
+          Lock.findOne({ where: { address: 'jqa6dnp1' } }).then(record => {
+            expect(record.name).toBe('oioioi')
```
cellog: Oggyoggyoggy!

locksmith/__tests__/token_middleware.test.js
```diff
@@ -0,0 +1,163 @@
...
+var tokenMiddleware = require('../src/token_middleware')
+
+let request, response, validAuthorizationHeader
+
+beforeAll(() => {
+  request = httpMocks.createRequest()
+  response = httpMocks.createResponse()
+  validAuthorizationHeader = {
+    'Authorization':
+      'Bearer eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJsb2NrIjp7ImFkZHJlc3MiOiJqcWE2ZG5wMSIsIm5hbWUiOiJvaW9pb2kiLCJleHBpcmF0aW9uRHVyYXRpb24iOjI1OTIwMDAsImtleVByaWNlIjoiMTAwMDAwMDAwMDAwMDAwMDAiLCJtYXhOdW1iZXJPZktleXMiOjEwLCJvd25lciI6IjB4QWFBZEVFRDRjMEI4NjFjQjM2ZjRjRTAwNmE5QzkwQkEyRTQzZmRjMiJ9LCJpYXQiOjE1NDYxMzA4MzUsImV4cCI6MTU0NjEzMDgzOCwiaXNzIjoiMHhBYUFkRUVENGMwQjg2MWNCMzZmNGNFMDA2YTlDOTBCQTJFNDNmZGMyIn0.0xd36a7842eda1b2267c08b204407f9e71024a2152ae772882a1fa41a02696aee578688f02ef6c301451de95793bde669198359f0820fb35f6ba1303e89727a52601',
```
cellog: Maybe extract this mock jwt and lock addresses into a file in tests/ and import it into the tests? Since the same value is used in multiple tests

locksmith/__tests__/token_middleware.test.js
```diff
@@ -0,0 +1,163 @@
...
+describe('Locksmith JWT Validation Middleware', () => {
+  describe('when a valid token is received', () => {
+    it('moves the request to the application', (done) => {
+      Date.now = jest.fn(() => 1546130835000)
+      var request = httpMocks.createRequest({ headers: validAuthorizationHeader, 
+        body: {'address':'jqa6dnp1','name':'oioioi','expirationDuration':2592000,
+          'keyPrice':'10000000000000000','maxNumberOfKeys':10,'owner':'0xAaAdEED4c0B861cB36f4cE006a9C90BA2E43fdc2'}})
+
+      tokenMiddleware(request, response, function next() {
+        expect(request.owner).toEqual('0xAaAdEED4c0B861cB36f4cE006a9C90BA2E43fdc2')
```
cellog: nit - I prefer toBe over toEqual unless it is an array or object contents being compared since toEqual uses == and toBe uses ===. I always find it's toBe or not toBe. That is the question.

locksmith/config/config.json
cellog: And docker secrets in ci

locksmith/models/index.js
```diff
@@ -0,0 +1,38 @@
...
+  })
+
+Object.keys(db).forEach(modelName => {
+  if (db[modelName].associate) {
+    db[modelName].associate(db)
+  }
+})
+
+db.sequelize = sequelize
+db.Sequelize = Sequelize
```
cellog: Can we give the lower-cased one a less generic name? connection?

locksmith/src/sequelize.js
```diff
@@ -0,0 +1,4 @@
+const models = require('../models')
```
cellog: This seems the wrong name for this file. sequelizeLock.js? Also why is it needed? Used in several places this way?

locksmith/src/token_middleware.js
```diff
@@ -0,0 +1,109 @@
...
+    return null
+  }
+}
+
+var base64Decode = data => {
+  const buff = Buffer.from(data, 'base64')
+  return buff.toString('utf-8')
+}
+
+const validateHaders = headers => {
```
cellog:
```suggestion
const validateHeaders = headers => {
```

Typo, need to correct validateHaders everywhere. validateHaters works if you don't like validateHeaders ;)

locksmith/src/token_middleware.js
```diff
@@ -0,0 +1,109 @@
...
+    return
+  }
+
+  const signingAddress = web3.eth.accounts.recover(
+    `${header}.${payload}`,
+    signature
+  )
+
+  if (
+    validateHaders(parsedHeaders) &&
```
cellog:
```suggestion
    validateHeaders(parsedHeaders) &&
```

locksmith/src/sequelize.js
akeem: Good point, initially that was written with the idea that multiple models could be loaded from this. I will rename this to better reflect the scenario that exists right now; we have a single model

locksmith/models/index.js
akeem: This file was generated by Sequelize as part of the initial model creation process. If possible, i'd like to hold off on making change here until we have some certainty that this wont change dramatically

locksmith/src/app.js
```diff
@@ -0,0 +1,66 @@
...
+  let newAddress = req.body.address
+  let tempAddress = req.params.lockAddress
+
+  if (tempAddress && newAddress) {
+    Lock.findOne({
+      where: { address: tempAddress, owner: req.owner },
+      raw: true,
+    })
+      .then(result => {
+        result.address = newAddress
```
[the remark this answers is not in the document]
akeem: @julien51 I believe in a scenario like this the potential attacker would receive a 412 back. Our validation middleware injects the signee/owner to the express application. If the message signee is not the owner of the object they are attempting to modify then the lookup will return an empty result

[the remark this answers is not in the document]
julien51: Ok, I did not see a test for this, maybe I am missing it?
Also I still think it would be a quick check here too and fail if the lock we're moving to exists? Better safe than sorry ;p

docker/docker-compose.ci.yml
akeem: yeah i'd like to punt on this until after this is merged in as the deployment bits are currently shifting

locksmith/src/app.js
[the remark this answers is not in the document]
akeem: All good, now addressed in coverage

locksmith/__tests__/app.test.js
```diff
@@ -0,0 +1,151 @@
...
+      return request(app)
+        .get('/lock/0xdeadbeef')
+        .then(response => {
+          expect(response.statusCode).toBe(404)
+        })
+    })
+  })
+
+  describe('Setting lock details', () => {
+    test('it saves the passed the information', done => {
```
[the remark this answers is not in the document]
akeem: now covered!

locksmith/__tests__/app.test.js
```diff
@@ -0,0 +1,151 @@
...
+        .send(validLockDetails)
+        .then(response => {
+          expect(response.statusCode).toBe(200)
+          done()
+        })
+    })
+  })
+
+  describe('Updating lock details', () => {
+    describe('when the lock exists', () => {
```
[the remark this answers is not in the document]
akeem: now covered as well

akeem:
@julien51 & @cellog with the test coverage updated to reflect highlighted scenarios, are we comfortable with the state of this sans the configuration adjustments (pending changes, lint settings, test docker config) to be made a follow up branch shortly after this has been merged

akeem:
Closing as this pull request has been succeeded by #920

PR #900, opened 2019-01-04
Update zos-lib for Solidity 0.5 support

hardlydifficult:
# Description

This latest release of zos-lib has updated the .sol files to allow for compiling with Solidity 0.5.  This update is a pre-req for resolving https://github.com/unlock-protocol/unlock/pull/892

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [x] This PR only contains configuration changes (package.json, etc.)
  - [ ] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [x] My code follows the style guidelines of this project, including naming conventions
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] If my code adds or changes components, I have written corresponding stories with Storybook
- [x] I have performed a self-review of my own code
- [x] If my code involves visual changes, I am adding applicable screenshots to this thread

PS: [Read how to write the perfect pull request](https://blog.github.com/2015-01-21-how-to-write-the-perfect-pull-request/)

PR #904, opened 2019-01-04
Removing .toNumber from potential BigNumbers

hardlydifficult:
# Description

Any data which is uint256 in Solidity cannot safely be stored as a number in Javascript.  This change removes `.toNumber` usage and instead uses BigNumbers (specifically https://github.com/indutny/bn.js/).

This is basically a minor refactor of existing tests, no new tests should be required.

This becomes a blocking issue when we update Truffle for Solidity 0.5 support (https://github.com/unlock-protocol/unlock/pull/892).  With the update, converting a BigNumber using `.toNumber` throws `Error: Number can only safely store up to 53 bits`.

Travis CI may fail due to the Netlify configuration, following up on that concern here: https://github.com/unlock-protocol/unlock/pull/900

# Checklist:


nfurfaro:
I can review this later tonight, but I know this needs to happen.

cellog:
assert simply requires a Boolean value, whereas assert.equals takes 2 values and compares them

nfurfaro:
@cellog Ok thanks. I'm guessing that `.eq` is just stock javascript then?

hardlydifficult:
I do not use `assert.equal` when comparing BigNumbers because it does not work consistently in Javascript.  Sometimes it does work, e.g. when comparing to 0.  However, an example from `transferFrom.js`:
```
assert(keyExpiration.eq(expirationTimestamp)) // pass
assert.equal(keyExpiration, expirationTimestamp) // fail
```

`.isEqualTo` is not supported by this BigNumber implementation, however we could choose to use another.  They all convert to/from other BigNumber implementations without trouble.

nfurfaro:
Sounds good. I don't think there's a need to switch BigNumber implementations, but good to know it's not an issue if we did have to.

PR #903, opened 2019-01-04
Move unlock.js to paywall.min.js

cellog:
# Description

Fixes #885 rework snippet to be testable/versionable and still easy to paste into site

This creates a framework to auto-generate a minimized version of the paywall. The new code is based on /static/unlock.js, and lives in src/paywall-builder.

The paywall loads /paywall page using the first lock found in a meta tag with name attribute `unlock`. The content attribute is used as the lock address. The hostname for the paywall is either implicitly in the url of the paywall.min.js script tag, or explicitly specified with `data-unlock-uri` attribute on the script tag.

All of these things are specified in the lock code snippet that users will paste into their site.

~~This PR also enables building the paywall.min.js where needed in package.json, which probably needs to be split out but I have a train to catch so consider this a code-only PR, but I won't merge it until the config is split out in a separate PR.~~ *update: split off into #906 and #910*

screen shot 2019-01-04 at 5 04 29 pm
screen shot 2019-01-04 at 5 00 55 pm

# Checklist:

- [X] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [X] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [X] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] If my code adds or changes components, I have written corresponding stories with Storybook
- [X] I have performed a self-review of my own code
- [X] If my code involves visual changes, I am adding applicable screenshots to this thread

PS: [Read how to write the perfect pull request](https://blog.github.com/2015-01-21-how-to-write-the-perfect-pull-request/)

PR #906, opened 2019-01-05
prepare configuration changes for paywall

cellog:
# Description

This is a PR for configuration changes in preparation for #903 

# Checklist:


unlock-app/package.json
```diff
@@ -82,6 +82,7 @@
     "set-link:darwin:freebsd:linux:sunos": "pushd src && (test -e artifacts && rm -f artifacts) && ln -s ../../smart-contracts/build artifacts",
     "set-link:win32": "cd src && if exist artifacts (rmdir artifacts /q /s || del artifacts) && cmd /c mklink /d artifacts ..\\..\\smart-contracts\\build",
     "build": "next build src",
+    "build-paywall": "NODE_ENV=production rollup -c rollup.paywall.config.js -o ./src/static/paywall.min.js",
```
julien51: Are we only building this in production?

cellog: Yes, the minimized paywall is built using production settings because it generates a minimized file. I think it will be important as a next step to create a Dev build with sourcemaps as a next step. Currently the paywall is so simple that it is not absolutely necessary for v0. As soon as it starts being the paywall instead of simply loading the iframe for it then yes, we will want to be able to debug it :). That's my take, what do you think?

And to be explicit, there is nothing in this PR that would make it difficult to add on a Dev build. Thanks for pushing on this, it's an important point!!

PR #938, opened 2019-01-07
Moved dependencies to dev because they are not needed for build

julien51:
The rule of thumb to add dependencies vs. devDependencies is whether they are required to build or
compile the project.

### DIFF
--- a/smart-contracts/package.json
+++ b/smart-contracts/package.json
@@ -17,12 +17,9 @@
     "eslint": "^5.7.0",
     "ethereumjs-units": "^0.2.0",
     "openzeppelin-eth": "^2.0.2",
-    "shelljs": "^0.8.3",
     "truffle": "^4.1.14",
     "truffle-contract": "^3.0.6",
-    "web3-utils": "^1.0.0-beta.36",
-    "zos": "^v2.1.0-rc.0",
-    "zos-lib": "^2.0.2"
+    "web3-utils": "^1.0.0-beta.36"
   },
   "devDependencies": {
     "eslint-config-standard": "^12.0.0",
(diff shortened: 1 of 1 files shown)


smart-contracts/package.json
```diff
@@ -33,7 +30,10 @@
     "eslint-plugin-standard": "^4.0.0",
     "remix-ide": "^0.7.5",
     "remixd": "^0.1.8-alpha.6",
-    "solhint": "^1.4.0"
+    "shelljs": "^0.8.3",
+    "solhint": "^1.4.0",
+    "zos": "^v2.1.0-rc.0",
```
nfurfaro: Zos is not required to build, but will be required for automated deploymen, so I'm not sure where that puts it in the current CI setup...

smart-contracts/package.json
```diff
@@ -17,12 +17,9 @@
     "eslint": "^5.7.0",
     "ethereumjs-units": "^0.2.0",
     "openzeppelin-eth": "^2.0.2",
-    "shelljs": "^0.8.3",
     "truffle": "^4.1.14",
     "truffle-contract": "^3.0.6",
-    "web3-utils": "^1.0.0-beta.36",
-    "zos": "^v2.1.0-rc.0",
-    "zos-lib": "^2.0.2"
```
nfurfaro: Zos-lib is where we import `Initializable.sol`, so I think we need it to compile.

smart-contracts/package.json
julien51: For now, let's keep our out and add it back when we automate deployments..

smart-contracts/package.json
julien51: Something fishy is going on then because build passed :(

julien51: Something fishy is going on then because build passed :(

julien51:
Ok, I know what that is... since we're not using`--production` on `npm install` or `npm ci`, all dependencies are being included. I'll submit another PR first then ;)

PR #949, opened 2019-01-08
Added env shim to travis.ci for Locksmith DB  details

akeem:
GH forks are unable to access Travis env variables but they can set their own via the `.travis.yml`.  The docker compose file has been updated to start the test postgres instance with matching credentials

# Description

*(Delete this, but please include a summary of the change and which issue is fixed -include its number-. Please also include relevant motivation and context.)*

# Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
 language: generic
+env:
+  - DB_USERNAME='locksmith_test'
+  - DB_PASSWORD='password'
+  - DB_NAME='locksmith_test'
+  - DB_HOSTNAME='db'
 services:
   - docker
 before_script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,9 @@
 language: generic
+env:
+  - DB_USERNAME='locksmith_test'
```
cellog: this runs the test suite in parallel, setting each env variable for one of the builds, so another solution needs to be found to set the env variables all at once in each build

cellog: my research suggests using `global:` as a sub-thingy of env should do what you're trying to do

PR #952, opened 2019-01-08
Fix ierc721Receiver pragma version

nfurfaro:
This PR corrects a solidity pragma version which was not fixed.

### DIFF
--- a/smart-contracts/contracts/interfaces/IERC721Receiver.sol
+++ b/smart-contracts/contracts/interfaces/IERC721Receiver.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.24;
+pragma solidity 0.4.24;
 
 
 /**


hardlydifficult:
Curious, why not include the `^`?  I know it's consistent with other files, but it seems they should all use `pragma solidity ^0.4.24;` until we upgrade to 0.5

nfurfaro:
I agree that it should be consistent across all solidity contracts either way. That said, I'm of the opinion that using a fixed compiler version is a tradeoff; a minor inconvenience when we do upgrade the compiler for a bit of added security, knowing that our code works as expected with the current compiler version and won't change unexpectedly if compiled with a different version. I believe it's consistent with "best practices" for Solidity, but I'm aware that even so-called best practices are not unanimously agreed upon in the community😉.

julien51:
LG @nfurfaro

PR #974, opened 2019-01-10
Deploy contract in dev using ZOS part_1

nfurfaro:
This is just the config portion of a 2-part PR. The second part is #975, which depends on this one to work, so I will rebase it after this one is merged.

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "start-ganache": "ganache-cli -i 1984 --mnemonic 'hello unlock save the web'",
     "postinstall": "./scripts/postinstall.sh",
-    "deploy-lock": "cd smart-contracts && npm run deploy",
+    "deploy-lock": "cd smart-contracts && npm run push-dev",
     "build": "cd smart-contracts && npm run build && cd ../unlock-app && npm run build && cd ../locksmith && npm run build",
     "start": "cd unlock-app && npm run start",
     "deploy-netlify": "./scripts/deploy-netlify.sh"
--- a/smart-contracts/package.json
+++ b/smart-contracts/package.json
@@ -41,6 +41,7 @@
     "lint": "solhint contracts/**/*.sol",
     "ci": "npm run lint && npm run test",
     "build": "truffle compile --reset",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -18,8 +18,8 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "start-ganache": "ganache-cli -i 1984 --mnemonic 'hello unlock save the web'",
     "postinstall": "./scripts/postinstall.sh",
-    "deploy-lock": "cd smart-contracts && npm run deploy",
-    "build": "cd smart-contracts && npm run build && cd ../unlock-app && npm run build && cd ../locksmith && npm run build",
```
akeem: Is this intentional?

nfurfaro: @akeem looking now..

nfurfaro: Nope. Fixing now, thanks for the catch!

nfurfaro:
Will do Ben.

smart-contracts/package.json
```diff
@@ -41,6 +41,7 @@
     "lint": "solhint contracts/**/*.sol",
     "ci": "npm run lint && npm run test",
     "build": "truffle compile --reset",
+    "push-dev": "zos push --reset --network development --from 0xe29ec42f0b620b1c9a716f79a02e9dc5a5f5f98a && npm run deploy",
```
julien51: Hey, not sure this is a good approach because it means we can only push in dev, when we actually need to do that on other occasions, such as CI.

nfurfaro: I agree. I originally though I could use a variable here, but now I think a cleaner way to do this would be to move this command to a migrations script,
maybe named `2_push_unlock.js`, followed by `3_create_proxy.js`. 
• On any given network, each these scripts will only be run once unless we pass the `--reset` flag to truffle migrate. 
• this keeps zos in migrations and removes it from use in package.json
• we can have fine-grained control over what happens on a network-by-network basis.

julien51: This is a great approach. Do you want to take that on?

nfurfaro: I've got it implemented already. At the ski hill with my kids, but will push when home. 2 parts: config and code, so will push config first, then rebase part 2 after part 1 is merged of that sounds good.

PR #1023, opened 2019-01-15
Handling missing keys on page

julien51:
If the number of keys does not "complete" the page the EVM triggered an error which prevented us from 
retrieving the whole page. Not anymore!

  - X ] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)

### DIFF
--- a/unlock-app/src/__tests__/services/web3Service.test.js
+++ b/unlock-app/src/__tests__/services/web3Service.test.js
@@ -1282,6 +1282,44 @@ describe('Web3Service', () => {
 
         web3Service.getKeysForLockOnPage(lockAddress, onPage, byPage)
       })
+
+      it('should get as many owners as there are per page, and not fail on missing keys', done => {
+        expect.assertions(3)
+
+        web3Service._getKeyByLockForOwner = jest.fn(() => {
+          return new Promise(resolve => {
+            return resolve([100, 'hello'])
+          })
+        })
+
+        const onPage = 3
+        const byPage = 5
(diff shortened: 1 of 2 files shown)


unlock-app/src/__tests__/services/web3Service.test.js
```diff
@@ -1282,6 +1282,44 @@ describe('Web3Service', () => {
 
         web3Service.getKeysForLockOnPage(lockAddress, onPage, byPage)
       })
+
+      it('should get as many owners as there are per page, and not fail on missing keys', done => {
+        expect.assertions(3)
+
+        web3Service._getKeyByLockForOwner = jest.fn(() => {
+          return new Promise(resolve => {
```
cellog: not important, but this is a good candidate for [`Promise.resolve`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve)

unlock-app/src/services/web3Service.js
```diff
@@ -680,9 +680,13 @@ export default class Web3Service extends EventEmitter {
             }
           )
         })
+        .catch(() => {
+          // Triggered when there are missing keys
+          return null
```
cellog: is there a way to log this error, just in case something more serious happens?

julien51: I think this is actually not an error per se. This is hos solidity tells us that there is no such key at that address.

PR #1027, opened 2019-01-15
Implement form validation

### LINKED ISSUE #797: Show an error when trying to submit a form with missing/wrong data
When creating a lock with missing or wrong data, we do not let the user submit... (nothing happens), however I believe it is a good design  principal to *always* have something happening on user explicit actions (like clicking on a button). We should probably show an error with "Please make sure you complete all from fields" for now.

cellog:
# Description

Fixes #797 

This PR is a master PR, and may need to be split into a couple of PRs. However, I am unsure how best to do this, so that will be part of the review.

screen shot 2019-01-15 at 12 12 26 pm

# Checklist:

### DIFF
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -48,7 +48,7 @@ export class CreatorLockForm extends React.Component {
       name: props.name,
       address: props.address,
     }
-    const { validityState: valid, errors } = this.valid(this.state)
+    const { validityState: valid, errors } = this.formValidity(this.state)
     this.state.valid = valid
     this.state.errors = errors
     this.handleSubmit = this.handleSubmit.bind(this)
@@ -56,10 +56,16 @@ export class CreatorLockForm extends React.Component {
     this.handleChange = this.handleChange.bind(this)
     this.handleUnlimitedClick = this.handleUnlimitedClick.bind(this)
     this.saveLock = this.saveLock.bind(this)
-    this.processFormErrors = this.processFormErrors.bind(this)
+    this.processFormErrors = this.sendErrorsToRedux.bind(this)
   }
 
-  valid(state) {
+  /**
+   * Traverses each form field and verifies its validity.
+   * returns a hash of fields to error message and all errors triggered.
+   * valid fields hash to the value false
+   * invalid fields hash to the error constant (a string) that represents the error condition
+   */
+  formValidity(state) {
     // the list of errors we will pass to setError
     const errors = []
     // for each field, retrieve the error triggered by current state
@@ -87,6 +93,9 @@ export class CreatorLockForm extends React.Component {
     return { validityState, errors }
   }
 
+  /**
+   * validate an individual form field
+   */
   validate(name, value) {
     switch (name) {
       case 'name':
@@ -136,8 +145,11 @@ export class CreatorLockForm extends React.Component {
     createLock(lock)
   }
 
-  processFormErrors(state) {
-    const { validityState: valid, errors } = this.valid(state)
+  /**
+   * calculate form errors, and propagate them to redux
+   */
+  sendErrorsToRedux(state) {
+    const { validityState: valid, errors } = this.formValidity(state)
     const { setError, resetError } = this.props
     const allFormErrors = [
       FORM_EXPIRATION_DURATION_INVALID,
@@ -160,7 +172,7 @@ export class CreatorLockForm extends React.Component {
       ...state,
       unlimitedKeys: true,
       maxNumberOfKeys: '∞',
-      valid: this.valid({ ...state, [name]: '∞' }),
+      valid: this.formValidity({ ...state, [name]: '∞' }),
     }))
   }
 
@@ -169,13 +181,13 @@ export class CreatorLockForm extends React.Component {
       unlimitedKeys:
         name === 'maxNumberOfKeys' ? value === '∞' : state.unlimitedKeys,
       [name]: value,
-      valid: this.valid({ ...state.valid, [name]: value }),
+      valid: this.formValidity({ ...state.valid, [name]: value }),
     }))
   }
 
   handleSubmit() {
     this.setState(state => {
-      const { valid, errors } = this.processFormErrors(state)
+      const { valid, errors } = this.sendErrorsToRedux(state)
       if (!valid.formValid) return { valid, errors }
       const { hideAction } = this.props
       if (hideAction) hideAction()
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -16,7 +16,6 @@ import {
   LockKeys,
 } from './LockStyles'
 import { LockStatus } from './lock/CreatorLockStatus'
-import { createLock } from '../../actions/lock'
 import { setError, resetError } from '../../actions/error'
 import {
   FORM_LOCK_NAME_MISSING,
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -16,6 +16,19 @@ import {
   LockKeys,
 } from './LockStyles'
 import { LockStatus } from './lock/CreatorLockStatus'
+import { createLock } from '../../actions/lock'
+import { setError, resetError } from '../../actions/error'
+import {
+  FORM_LOCK_NAME_MISSING,
+  FORM_EXPIRATION_DURATION_INVALID,
+  FORM_MAX_KEYS_INVALID,
+  FORM_KEY_PRICE_INVALID,
+} from '../../errors'
+import {
+  isNotEmpty,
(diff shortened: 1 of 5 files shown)


unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -14,86 +14,157 @@ import {
...
     super(props, context)
     this.state = {
-      expirationDuration: 30,
-      expirationDurationUnit: 86400, // Days
-      keyPrice: '0.01',
-      keyPriceCurrency: 'ether',
-      maxNumberOfKeys: 10,
-      unlimitedKeys: false,
-      name: 'New Lock',
+      expirationDuration: props.expirationDuration,
```
julien51: This part can be in its own PR.

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -14,86 +14,157 @@ import {
...
-    let { unlimitedKeys } = this.state
-    return (
-      (name === 'name' && typeof value === 'string' && value.length > 0) ||
-      ((name === 'expirationDuration' || name === 'maxNumberOfKeys') &&
-        !isNaN(value) &&
-        value > 0) ||
-      (name === 'maxNumberOfKeys' && value === '∞' && unlimitedKeys) ||
-      (name === 'keyPrice' && !isNaN(value) && value >= 0)
-    )
+    const missing = val => val === undefined || val === null || val === ''
```
julien51:
```suggestion
    const missing = val => (val === undefined || val === null || val === '')
```

julien51: We can probably extract all of these functions from this one. Either at the top of the file or (even better), in helper file (`validations.js`?) which can be done in its own PR.

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -14,86 +14,157 @@ import {
...
+        if (missing(value)) {
+          return FORM_LOCK_NAME_MISSING
+        }
+        break
+      case 'expirationDuration':
+      case 'maxNumberOfKeys': {
+        const error = {
+          expirationDuration: FORM_EXPIRATION_DURATION_INVALID,
+          maxNumberOfKeys: FORM_MAX_KEYS_INVALID,
+        }[name]
```
julien51: I see what you're trying to do here by not using `break` but I would recommend that you either use `break` and duplicate some code, or that you use an if statement. This would make the code easier to read and less likely to confuse people.

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -14,86 +14,157 @@ import {
...
+      errors.push(invalidError)
+      return fieldValidity
+    }, {})
+
+    // the form can be submitted if and only if there are no errors triggered by any field
+    validityState.formValid = errors.length === 0
+    return { validityState, errors }
   }
 
   validate(name, value) {
```
julien51: I like that refactor of `validate` which can probably be done in its own PR ;)

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -14,86 +14,157 @@ import {
...
+      [name]: value,
+      valid: this.valid({ ...state.valid, [name]: value }),
+    }))
   }
 
   handleSubmit() {
-    if (document.querySelector('[data-valid="false"]')) return false
+    this.setState(state => {
+      const { validityState: valid, errors } = this.valid(state)
+      const { setError, resetError } = this.props
```
julien51: Pull that out of the `setState` callback (I think we should do the assignments at the beginning of the function)

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -14,86 +14,157 @@ import {
...
-    const {
-      expirationDuration,
-      expirationDurationUnit,
-      keyPriceCurrency,
-      maxNumberOfKeys,
-      unlimitedKeys,
-      keyPrice,
-      name,
-    } = this.state
+      const { account, createLock, hideAction } = this.props
```
julien51: Combine with `const { setError, resetError } = this.props`?

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -14,86 +14,157 @@ import {
...
+      const {
```
julien51: Maybe move the part which does create the lock from `state` to its own function to make `handleSubmit` easier to read?

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -119,24 +191,30 @@ export class CreatorLockForm extends React.Component {
             name="name"
             onChange={this.handleChange}
             defaultValue={name}
+            data-valid={valid.name}
           />
         </FormLockName>
         <FormLockDuration>
           <input
-            type="text"
+            type="number"
```
julien51: This is good! We could start with a PR which does just that for example (and does not change any of the logic)

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -169,19 +250,34 @@ CreatorLockForm.propTypes = {
...
+  expirationDurationUnit: PropTypes.number,
+  keyPrice: PropTypes.string,
+  keyPriceCurrency: PropTypes.string,
+  maxNumberOfKeys: PropTypes.oneOfType([PropTypes.number, PropTypes.string]), // string is for '∞'
+  unlimitedKeys: PropTypes.bool,
+  name: PropTypes.string,
 }
 
-CreatorLockForm.defaultProps = {}
+CreatorLockForm.defaultProps = {
```
julien51: Same as above: this is its own PR ;)

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -195,15 +291,30 @@ const LockLabelUnlimited = styled(LockLabel)`
...
 
 const FormLockRow = styled(LockRow)`
-  ${LockRowGrid} input[type='text'] {
-    background-color: var(--lightgrey);
-    border: 0;
-    padding: 5px;
-    font-family: 'IBM Plex Sans', sans-serif;
-    font-size: 13px;
+  input[type='number'] {
+    -moz-appearance: textfield;
```
julien51: Meh. Are you sure this is a good idea to use prefixed-CSS stuff? I thought this would be handled by our build tools?

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -219,22 +330,25 @@ const FormLockName = styled(LockName)`
 `
 
 const FormLockDuration = styled(LockDuration)`
-  input[type='text'] {
+  input[type='text'],
+  input[type='number'] {
```
julien51: This stuff belongs in the PR which just set the field types ;)

unlock-app/src/components/helpers/ErrorMessage.js
```diff
@@ -23,6 +23,17 @@ const defaultMessages = {
   NON_DEPLOYED_CONTRACT: (
     <p>Unlock has not been deployed on the current network.</p>
   ),
+
+  FORM_LOCK_NAME_MISSING: (
```
julien51: This could be in its own PR, along with the stuff in `src/errors.js`

unlock-app/src/components/creator/CreatorLockForm.js
cellog: in this case, the appearance thing is only non-standard, and this is what was recommended to remove the up-down spinner for numbers on all browsers. Perhaps in a year or two, the browsers will standardize something, but for now, we are in 2010 when it comes to number input styling :)

unlock-app/src/components/creator/CreatorLockForm.js
cellog: the only issue is that setState callbacks are async, I am not sure if we can rely on props in handleSubmit being the same when we finish. Although it may be OK, because it doesn't save any processor cycles or add other benefits, I don't think we should take this risk. What do you think?

unlock-app/src/components/creator/CreatorLockForm.js
cellog: not only that, this allows us to use infinity in expirationDuration, which is probably not an intended feature :)

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -36,31 +48,67 @@ export class CreatorLockForm extends React.Component {
...
+    // for each field, retrieve the error triggered by current state
+    // and then make sure we set it as existing.
+    const validityState = [
+      'expirationDuration',
+      //'expirationDurationUnit',
+      'keyPrice',
+      //'keyPriceCurrency',
+      'maxNumberOfKeys',
+      'name',
+    ].reduce((fieldValidity, field) => {
```
julien51: nit: would `formValidity` be a better name to indicate that this is about all the fields?

Issue #1025, opened 2019-01-15
Partial withdraws


julien51:
Right now, when the lock owner withdraws, they withdraw all the funds on the lock. 
Since we want to implement a mechanism to cancel a key, we need to allow the lock creator to only withdraw funds partially. 

For this feature to exist we need to things:
1. Support on the smart contract. Assigning this to @nfurfaro .
2. Support on the creator dashboard. Assigning this to @smombartz for the design and @cellog for implementation.

nfurfaro:
@julien51 off the top of my head...there could be 2 withdraw functions, 1 with 0 args and 1 with one uint arg. Both would end up calling an internal function `_withdraw(uint amount)` similar to how `purchaseFor` and `purchaseForFrom` are implemented. The uint amount could be in ETH or a percentage (such as `withdraw(50)` to take half.)

nfurfaro:
RE: TODOS in withdraw function
>    * TODO: consider allowing anybody to trigger this as long as it goes to owner anyway?
>   * TODO: check for re-entrency?
As for number 1, Not sure we want to let anyone call this even if it can only go to the owner...
As for number 2, this is easy to implement, but as only the owner can currently call the function is reentrancy a concern in this case?

smombartz:
want to make sure i understand fully. is this to partially withdraw funds or cancel a key (which essentially is a partial withdraw)?

julien51:
No, this is not about canceling keys. It's for the lock owner. Once people have purchased keys, their lock receives the funds. Right now the lock owner can withdraw the funds when they want, but we want to allow them to do that only partially.

For example, I have a lock whose keys are $10. If 2 people purchase keys, I receive $20. Right now I can only withdraw $20, but we want the ability to withdraw only $15 (that is leaving $5 on the lock until I withdraw again).

Makes sense?

julien51:
@nfurfaro For the use case 1. I actually think this could be a nice feature to build for some organizations where we trigger withdraws automatically on behalf of creators (I see this as a way to secure the funds, because we never know 100% that a smart contract could not be abused, or as a bug which would allow malicious people to steal the funds)

nfurfaro:
@julien51 so for now I'll just leave the TODOs in place.

smombartz:
@julien51 thats was my hunch. thanks for clarifying. so in the UI we need them to enter the amount they want to withdraw + confirmation button (which then gets sent to the wallet). correct?

cellog:
[...]

Issue #637, opened 2018-11-30
Allow for cancellation


julien51:
This is a feature which I thinks public locks could have: the ability for their owner to define terms of a cancellation.
In practice I think cancelling a key would just mean "selling it back" to the lock by  the key owner. In that case the key owner would get a (partial) refund. I believe the easiest way to do that would be to decide on the amount of the refund based on the remaining validity of the key with an optional penalty by the lock owner.

What do you think @nfurfaro ?

nfurfaro:
It seems reasonable on a technical level. It's probably preferable to have a key owner sell their key to another user rather than back to unlock ( from a revenue perspective). But if this is something you want to  implement, it will be relatively easy if done after some of the ERC721 changes discussed in #148, specifically the ability to " top-up a key's time with another key's time." If this were implemented, and the lock retained one key, cancellation would simply be selling the time remaining on a key to the lock via it's key.

julien51:
I think this is right: cancelling a key would be transferring back the remaining time on it back to the lock owner.

nfurfaro:
@julien51 Does it make sense to you to do this after #148 is complete?

hardlydifficult:
How much of a refund would be offered?  e.g. maybe the Lock owner defines a rate per unit of time, where `refundRate * expirationDuration <= keyPrice`.  And/or a fixed penalty (selected by the Lock owner) could be applied as you suggested above.  Note that if both a refundRate and a fixed penalty are defined cancel won't make sense for keys which are near their cancellation date (which seems to make sense).  Options seem to be:
 - Cancel for a refund of `keyPrice - fixedPenalty`
 - Cancel for a refund of `costPerUnitTime * timeRemaining - fixedPenalty`
 - Cancel for a refund of `refundRate * timeRemaining`
 - Cancel for a refund of `refundRate * timeRemaining - fixedPenalty`

Cancel would fail if the Lock owner withdraws too much funds.  Do we accept that cancel may fail if this happens?  Or do we require a minimum amount of reserve funds remain in the contract? (if we implement a min, we will need to consider when all funds may be withdrawn so we don't effectively burn ETH) 

On a related note, `withdraw` has a TODO to consider partial withdraws.  It seems if this feature is added we need partial withdraws implemented as well.

julien51:
I think all of these are good questions. Thanks @hardlydifficult !
I think the `refundRate` is fairly simple to compute `price * keyDuration/(expiration-now)`. I think the penalty is fair game. 

[...]

PR #1043, opened 2019-01-15
Partial withdrawals - smart-contract

nfurfaro:
This PR is a WIP based on Issue #1025. I still expect to modify a little, but wouldn't mind feedback early.

# Checklist:

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -33,6 +33,11 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     uint keyPrice
   );
 
+  event Withdrawal(
+    address indexed _sender,
+    uint _amount
+    );
+
 
   // Fields
   // Unlock Protocol address
@@ -227,20 +232,42 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
   }
 
   /**
-   * @dev Called by owner to wiwthdraw all funds from the lock.
+   * @dev Called by owner to withdraw all funds from the lock.
    * TODO: consider allowing anybody to trigger this as long as it goes to owner anyway?
-   * TODO: consider partial withdraws?
    * TODO: check for re-entrency?
    */
-  function withdraw(
-  )
+  function withdraw()
     external
     onlyOwner
   {
     uint balance = address(this).balance;
     require(balance > 0, "Not enough funds");
-    address owner = Ownable.owner();
-    owner.transfer(balance);
+    _withdraw(balance);
+  }
+
+  /**
+   * @dev Called by owner to partially withdraw funds from the lock.
+   * TODO: consider allowing anybody to trigger this as long as it goes to owner anyway?
+   * TODO: check for re-entrency?
+   */
+  function partialWithdraw(uint _amount)
+    external
+    onlyOwner
+  {
+    uint256 balance = address(this).balance;
+    require(balance > 0 && balance >= _amount, "Not enough funds");
+    _withdraw(_amount);
+  }
+
+  /**
+   * @dev private version of the withdraw function which handles all withdrawals from the lock.
+   * TODO: check for re-entrency?
+   */
+  function _withdraw(uint _amount)
+    private
+  {
+    Ownable.owner().transfer(_amount);
+    emit Withdrawal(msg.sender, _amount);
   }
 
   /**
--- a/smart-contracts/contracts/interfaces/ILockCore.sol
+++ b/smart-contracts/contracts/interfaces/ILockCore.sol
@@ -43,7 +43,8 @@ interface ILockCore {
   /**
    * @dev Called by owner to partially withdraw funds from the lock.
    */
-  function partialWithdraw(uint
+  function partialWithdraw(
+    uint _amount
   )
     external;
 
--- a/smart-contracts/contracts/interfaces/ILockCore.sol
+++ b/smart-contracts/contracts/interfaces/ILockCore.sol
@@ -34,9 +34,16 @@ interface ILockCore {
     payable;
 
   /**
-   * @dev Called by owner to wiwthdraw all funds from the lock.
+   * @dev Called by owner to withdraw all funds from the lock.
    */
   function withdraw(
+  )
+    external;
+
+  /**
+   * @dev Called by owner to partially withdraw funds from the lock.
+   */
+  function partialWithdraw(uint
   )
     external;
 
--- a/smart-contracts/test/Lock/partialWithdraw.js
+++ b/smart-contracts/test/Lock/partialWithdraw.js
@@ -1,0 +1,100 @@
+
+const Units = require('ethereumjs-units')
+
+const deployLocks = require('../helpers/deployLocks')
+const Unlock = artifacts.require('../Unlock.sol')
+let unlock, locks
+
+contract('Lock', (accounts) => {
+  before(async () => {
+    unlock = await Unlock.deployed()
+    locks = await deployLocks(unlock)
+  })
+
+  describe('partialWithdraw', () => {
+    let owner, price, initialLockBalance, withdrawalAmount
+    price = Units.convert('0.01', 'eth', 'wei')
+    withdrawalAmount = Units.convert('0.005', 'eth', 'wei')
+
+    before(() => {
+      const purchases = [accounts[1], accounts[2]].map((account) => {
+        return locks['OWNED'].purchaseFor(account, '', {
+          value: price,
+          from: account
+        })
+      })
+      return Promise.all(purchases)
+        .then(() => {
+          return locks['OWNED'].owner.call()
+        })
+        .then((_owner) => {
+          owner = _owner
+        })
+    })
+
+    it('should fail if called by address other than owner', async () => {
+      try {
+        assert.notEqual(owner, accounts[1]) // Making sure
+        await locks['OWNED'].partialWithdraw(withdrawalAmount, {
+          from: accounts[1]
+        })
+      } catch (error) {
+        assert.equal(error.message, 'VM Exception while processing transaction: revert')
+        return
+      }
+      assert.fail()
+    })
+
+    it('should fail if too much is withdrawn', async () => {
+      try {
+        initialLockBalance = web3.eth.getBalance(locks['OWNED'].address)
+        await locks['OWNED'].partialWithdraw(initialLockBalance.add(withdrawalAmount), {
+          from: owner
+        })
+      } catch (error) {
+        assert.equal(error.message, 'VM Exception while processing transaction: revert Not enough funds')
+        return
+      }
+      assert.fail()
+    })
+
+    it('should fail if requesting partial withdraw of 0', async () => {
+      try {
+        initialLockBalance = web3.eth.getBalance(locks['OWNED'].address)
+        await locks['OWNED'].partialWithdraw(0, {
+          from: owner
+        })
+        return
+      } catch (error) {
+        assert.equal(error.message, 'VM Exception while processing transaction: revert Not enough funds')
+      }
+      assert.fail()
+    })
+
+    describe('when the owner withdraws some funds', () => {
+      let initialOwnerBalance, expectedLockBalance, finalLockBalance, finalOwnerBalance, gasPrice, gasUsed, txObj, txHash, txFee
+
+      before(async () => {
+        expectedLockBalance = initialLockBalance.sub(withdrawalAmount)
+        initialOwnerBalance = web3.eth.getBalance(owner)
+        txObj = await locks['OWNED'].partialWithdraw(withdrawalAmount, {
+          from: owner
+        })
+      })
+
+      it('should increase the owner\'s balance with the amount of funds withdrawn from the lock', async () => {
+        txHash = await web3.eth.getTransaction(txObj.tx)
+        gasUsed = txObj.receipt.gasUsed
+        gasPrice = txHash.gasPrice
+        txFee = gasPrice.mul(gasUsed)
+        finalOwnerBalance = web3.eth.getBalance(owner)
+        assert(finalOwnerBalance.eq(initialOwnerBalance.add(withdrawalAmount).sub(txFee)))
+      })
+
+      it('should decrease the lock\'s balance by the amount of funds withdrawn from the lock', () => {
+        finalLockBalance = (web3.eth.getBalance(locks['OWNED'].address))
+        assert(finalLockBalance.eq(expectedLockBalance))
+      })
+    })
+  })
+})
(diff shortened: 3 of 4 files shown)


smart-contracts/contracts/PublicLock.sol
```diff
@@ -221,20 +221,42 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
...
     require(balance > 0, "Not enough funds");
+    _withdraw(balance);
+  }
+
+  /**
+   * @dev Called by owner to partially withdraw funds from the lock.
+   * TODO: consider allowing anybody to trigger this as long as it goes to owner anyway?
+   * TODO: check for re-entrency?
+   */
+  function partialWithdraw(uint _amount)
```
nfurfaro: Should we set a lower bound on withdrawal `_amount`? it could be less than tx fee...

nfurfaro: And/or a limit on the minimum lock balance which allows partial withdrawals?

smart-contracts/test/Lock/partialWithdraw.js
```diff
@@ -0,0 +1,81 @@
+
+const Units = require('ethereumjs-units')
+
+const deployLocks = require('../helpers/deployLocks')
+const Unlock = artifacts.require('../Unlock.sol')
+let unlock, locks
+
+const Promise = require('bluebird')
+if (typeof web3.eth.getBlockPromise !== 'function') {
+  Promise.promisifyAll(web3.eth, { suffix: 'Promise' })
```
nfurfaro: Thoughts?

smart-contracts/contracts/PublicLock.sol
```diff
@@ -221,20 +221,42 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
...
   {
     uint256 balance = address(this).balance;
     require(balance > 0, "Not enough funds");
+    _withdraw(balance);
+  }
+
+  /**
+   * @dev Called by owner to partially withdraw funds from the lock.
+   * TODO: consider allowing anybody to trigger this as long as it goes to owner anyway?
+   * TODO: check for re-entrency?
```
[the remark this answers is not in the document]
julien51: See my comments on the other thread.

smart-contracts/contracts/PublicLock.sol
julien51: I guess we should not do that at the contract level but we can do that in the UI?

smart-contracts/test/Lock/partialWithdraw.js
julien51: What is `promisifyAll`?

smart-contracts/contracts/PublicLock.sol
nfurfaro: yes, I think that's fine here.

smart-contracts/test/Lock/partialWithdraw.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+contract('Lock', (accounts) => {
+  before(() => {
+    return Unlock.deployed()
+      .then(_unlock => {
```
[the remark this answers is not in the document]
nfurfaro: no problem. I'll convert it to use async/await. What do you think about (in a separate PR) extracting the zos deployment process into a helper, and reusing in cases like this where we currently deploy Unlock without a proxy? It would help to keep our use of a proxy consistent in our tests.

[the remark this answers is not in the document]
julien51: That is a good idea for another PR!

smart-contracts/test/Lock/partialWithdraw.js
nfurfaro: I've removed this as I got it working without it. My understanding is that older versions of web3.eth didn't return promises, so this was a workaround to promisify all of them. It seems that's no longer the case with web3, so this is unneeded.

nfurfaro:
@julien51 I believe I've addressed all the changes you requested on this one.

smart-contracts/contracts/interfaces/ILockCore.sol
```diff
@@ -34,12 +34,19 @@ interface ILockCore {
...
+   * @dev Called by owner to withdraw all funds from the lock.
    */
   function withdraw(
   )
     external;
 
+  /**
+   * @dev Called by owner to partially withdraw funds from the lock.
+   */
+  function partialWithdraw(uint
```
hardlydifficult: We may want to include a variable name here.  It is option in the interface, but including a variable name offers a bit more data for people viewing the interface, and is consistent with the rest of the interface.

nfurfaro: yes, good call.

smart-contracts/contracts/PublicLock.sol
```diff
@@ -221,20 +221,42 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
...
+   */
+  function partialWithdraw(uint _amount)
+    external
+    onlyOwner
+  {
+    uint256 balance = address(this).balance;
+    require(balance > 0 && balance > _amount, "Not enough funds");
```
hardlydifficult: balance >= _amount (not just greater than)... unless you prefer forcing them to use the other method when withdrawing all.

smart-contracts/contracts/PublicLock.sol
```diff
@@ -221,20 +221,42 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
...
+  /**
+   * @dev private version of the withdraw function which handles all withdrawals from the lock.
+   * TODO: check for re-entrency?
+   */
+  function _withdraw(uint _amount)
+    private
+  {
     address owner = Ownable.owner();
-    owner.transfer(balance);
+    owner.transfer(_amount);
```
hardlydifficult: Should we include an event here?

nfurfaro: Yes! Actually, events, in general, is on my list of things to look at. I think this should have one, and there may be other areas where we want to add events as well. For now I'll focus on this one

smart-contracts/test/Lock/partialWithdraw.js
```diff
@@ -0,0 +1,79 @@
...
+        })
+    })
+
+    it('should only allow the owner to withdraw', () => {
+      assert.notEqual(owner, accounts[1]) // Making sure
+      return locks['OWNED'].partialWithdraw(withdrawalAmount, {
+        from: accounts[1]
+      }).catch(error => {
+        assert.equal(error.message, 'VM Exception while processing transaction: revert')
+      })
```
hardlydifficult: This test will not fail if it did not revert, as expected.  It will only fail if it reverts for the wrong reason.

There are various ways to handle this.  Longer term, I suggest we add a `shouldFail` helper method.  However a quick fix may be:

```
}).catch(error => {
    assert.equal(error.message, 'VM Exception while processing transaction: revert')
    return
})
assert.fail()
```

hardlydifficult: Same comment for should fail below.

smart-contracts/test/Lock/partialWithdraw.js
```diff
@@ -0,0 +1,79 @@
...
+    })
+
+    it('should fail if too much is withdrawn', () => {
```
hardlydifficult: For completeness, maybe add `should fail if requesting partial withdraw of 0`

smart-contracts/contracts/PublicLock.sol
nfurfaro: well, semantically it's a bit weird to use the `partialWithdraw` function to withdraw all the funds. I'd rather let each withdraw function do one thing, but am open to modifying this if you or @julien51 feel strongly otherwise...

smart-contracts/test/Lock/partialWithdraw.js
nfurfaro: yep.

smart-contracts/test/Lock/partialWithdraw.js
nfurfaro: I like the idea of a helper method, something I've used in the past as well. As for your quick fix suggestion above, eslint strips out the `return` and says the `assert.fail()` is unreachable

smart-contracts/contracts/interfaces/ILockCore.sol
nfurfaro: done.

smart-contracts/test/Lock/partialWithdraw.js
hardlydifficult: Remove the previous instance of `return`

smart-contracts/contracts/PublicLock.sol
nfurfaro: Julien had mentioned possible opening this function up to be called be addresses other than the owner in the future, and we know the transfer can only go to the owner so I'm thinking of this:
```
event Withdrawal(
    address _sender,
    uint _amount
    );
```

nfurfaro: with `address _sender` possibly indexed?

hardlydifficult: Having other people trigger a withdraw seems a bit strange to me.  And if we allow cancels (https://github.com/unlock-protocol/unlock/issues/637) others could troll the system by calling withdraw, leaving not enough money in the contract to allow refunds.  If we include this feature, we may want to consider permissions for it.

But yea, if we include that feature, the event looks good.  Yes include `indexed` here.

nfurfaro: Yes, you echoed my concerns around this exactly. I thought it would require a registry of whitelisted addresses or some other such permissions to prevent that exact attack.

nfurfaro: done.

smart-contracts/contracts/PublicLock.sol
```diff
@@ -221,20 +221,42 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
...
    */
-  function withdraw(
-  )
+  function withdraw()
     external
     onlyOwner
   {
     uint256 balance = address(this).balance;
     require(balance > 0, "Not enough funds");
+    _withdraw(balance);
```
hardlydifficult: I understand we are not too focused on gas costs ATM, however a possible minor optimization here.  We are currently checking the balance twice and technically it could be done only once.

nfurfaro: Yeah, there was another spot in `_purchaseFor` where we were reading `keyprice` from storage 3 or 4 times...I modified to read it into memory once and then use that instead.
Can you elaborate on how we can eliminate 1 check here? I think it's worth it, and may as well do it now while I'm in there tinkering:)

hardlydifficult: Oh sorry, I misread.  I thought _withdraw had a require line as well.

But on the topic of gas, a common practice in our .sol files today is to use temp variables.  I agree this is good for readability, but the Solidity compiler does not optimize that away.

This implementation is cheaper:
```
  function _withdraw(uint _amount)
    public
  {
    msg.sender.transfer(_amount);
  }
```

than the current implementation
```
  function _withdraw(uint _amount)
    public
  {
    address owner = msg.sender;
    owner.transfer(_amount);
  }
```

Removing the temp saves 13 gas.  Very little, but there are a few instances where we could do the same.

nfurfaro: Right, I see what you're saying. In our`_withdraw` function, as we're not transferring to `msg.sender` but to `owner`,  I think we could do this:
```
function _withdraw(uint _amount)
    private
  {
    Ownable.owner().transfer(_amount);
    emit Withdrawal(msg.sender, _amount);
  }
```

nfurfaro: cool. Done.

smart-contracts/test/Lock/partialWithdraw.js
nfurfaro: done.

smart-contracts/contracts/PublicLock.sol
julien51: I agree that it is weird but it may be confusing to people to return an error when in practice I think it does not change much for us? I'd be liberal in what we expect from people and conservative in what we do ;)

smart-contracts/contracts/PublicLock.sol
julien51: I think you both are right. I just wanted a way for "us" to trigger withdraws in order to make sure we keep the locks "safe" (funds are safer on a user's address than inside a smart contract IMO)... but maybe this is not the right approach. Maybe insisting on them withdrawing is the way forward. Let's push that to the backlog and see how 'disciplined' people are.

smart-contracts/test/Lock/partialWithdraw.js
nfurfaro: So, stripping out the previous instance of "return" has no effect here. This seems to work, but I need another set of eyes to take a look. 
```
it('should fail if called by address other than owner', async function () {
      try {
        assert.notEqual(owner, accounts[1]) // Making sure
        await locks['OWNED'].partialWithdraw(withdrawalAmount, {
          from: accounts[1]
        })
      } catch (error) {
        assert.equal(error.message, 'VM Exception while processing transaction: revert')
        return
      }
      assert.fail()
    })
```

There are probably many places in our tests where this needs to be updated, so I'm proposing that we move ahead with a separate PR to implement a helper for testing failing transactions, possible based on this approach here which I've used in the past:
https://gist.github.com/xavierlepretre/d5583222fde52ddfbc58b7cfa0d2d0a9

smart-contracts/contracts/PublicLock.sol
nfurfaro: sounds good, I'll modify now.

nfurfaro: done.

smart-contracts/test/Lock/partialWithdraw.js
hardlydifficult: Yea that looks good now.  

I'd love to see the helper added.  I've previously noticed other 'should fail' tests which don't fail when no revert occurs (e.g. all the `approve` should fail tests).  The helper along with async style could make our tests much more readable as well!

smart-contracts/contracts/PublicLock.sol
nfurfaro: @julien51 Would you like me to remove `address _sender` from the event for now, and only add it in if we end up implementing that feature?

julien51: That is probably a good idea! Thanks!

PR #1053, opened 2019-01-16
Enable edit button

cellog:
# Description

This enables the edit button on the LockIconBar. The actual display of the edit form will be done in a separate PR, this just wires up the walls, so to speak.

# Checklist:

### DIFF
--- a/unlock-app/src/components/creator/CreatorLocks.js
+++ b/unlock-app/src/components/creator/CreatorLocks.js
@@ -27,6 +27,7 @@ export class CreatorLocks extends React.Component {
     const { showDashboardForm } = this.state
     let lockFeed = Object.values(locks).reverse() // We want to display newer locks first
 
+    // key must be JSON.stringify(lock) to guarantee re-render when props change beneath
     return (
       <Locks>
         <LockHeaderRow>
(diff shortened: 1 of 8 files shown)


unlock-app/src/components/creator/CreatorLocks.js
```diff
@@ -43,9 +43,9 @@ export class CreatorLocks extends React.Component {
         </LockHeaderRow>
         <Errors />
         {showDashboardForm && <CreatorLockForm hideAction={this.toggleForm} />}
-        {lockFeed.map(lock => {
-          return <CreatorLock key={JSON.stringify(lock)} lock={lock} />
-        })}
+        {lockFeed.map(lock => (
+          <CreatorLock key={lock.address} lock={lock} edit={() => {}} />
```
julien51: So, this is interesting. We previously had this `key={lock.address}` but apparently React uses the keys to identify if the component needs to be re-rendered. However it only checks using shallow verification which means that the component was actually not re-rendered even though some of its properties have changed. This is the reason I used `JSON.stringify(lock)`.

cellog: ah. Another way to solve this is to do {...lock} instead of lock={lock}, then it will re-render on any change to the lock, but that is a future PR ;)

cellog:
I'm actually going to split this one up. closing

PR #1061, opened 2019-01-16
lift createLock up to CreatorLocks from CreatorLockForm

cellog:
# Description

The `createLock` callback in CreatorLockForms is currently hard-coded to the `createLock` action. Now that the form will be used for both creating and editing locks, this needs to be more flexible. This PR lifts the property up to CreatorLocks, so that CreatorLocks has the flexibility to pass in a different callback for new locks and editing existing locks

# Checklist:

- [X] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [X] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [ ] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [X] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PS: [Read how to write the perfect pull request](https://blog.github.com/2015-01-21-how-to-write-the-perfect-pull-request/)

PR #1069, opened 2019-01-16
Enable edit form for lock

cellog:
# Description

This enables the edit form on the dashboard. saving edits is a no-op.

out

# Checklist:

### DIFF
--- a/unlock-app/src/components/creator/CreatorLocks.js
+++ b/unlock-app/src/components/creator/CreatorLocks.js
@@ -16,8 +16,45 @@ export class CreatorLocks extends React.Component {
     const { showForm } = this.props
     this.state = {
       showDashboardForm: !!showForm,
+      editingLocks: {},
     }
     this.toggleForm = this.toggleForm.bind(this)
+  }
+
+  getLockElement(lock) {
+    const key = JSON.stringify(lock)
+    const { editingLocks } = this.state
+    if (editingLocks[lock.address]) {
+      return (
+        <CreatorLockForm
+          key={key}
+          createLock={() => {}}
+          hideAction={() => this.stopEditingLock(lock.address)}
+          {...lock}
+        />
+      )
+    }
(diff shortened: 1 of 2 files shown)


unlock-app/src/components/creator/CreatorLocks.js
```diff
@@ -16,10 +16,47 @@ export class CreatorLocks extends React.Component {
     const { showForm } = this.props
     this.state = {
       showDashboardForm: !!showForm,
+      editingLocks: {},
```
julien51: This works, but I wonder if there would not be a way to isolate the lock "state" (editing/not editing) in its state? (basically only the lock knows whether it is being edited by having some kind of property on it). I think this would be better separation of concerns?

unlock-app/src/components/creator/CreatorLocks.js
```diff
@@ -16,10 +16,47 @@ export class CreatorLocks extends React.Component {
...
+        <CreatorLockForm
+          key={key}
+          createLock={() => {}}
+          hideAction={() => this.stopEditingLock(lock.address)}
+          {...lock}
+        />
+      )
+    }
+    return (
+      <CreatorLock
```
julien51: If the lock knows of its state the conditional of whether to display the form or the non form version would happen inside `CreatorLock`.

cellog:
> I am not sure I fully understand why you went down this route which exposes lock state at the level of the list when I believe the lock state should only be known inside the `<CreatorLock>` component. Do you mind clarifying?

Sure. It was my understanding in our conversation this morning that you wanted this done outside the `CreatorLock` component. Since that is not the case, I can re-do this inside `CreatorLock`, no problem. Can we back out one of the earlier PRs, #1061? It was a precursor to this one, and has to be reverted.

cellog:
closing in preparation for a new implementation

PR #1070, opened 2019-01-17
disable non-editable fields when editing a lock

cellog:
# Description

When editing a lock, we only want key price to be editable (currently). This PR disables all other fields and dims the text as a visual indicator.

screen shot 2019-01-16 at 6 59 45 pm
screen shot 2019-01-16 at 6 59 39 pm

# Checklist:

### DIFF
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -136,7 +136,6 @@ export class CreatorLockForm extends React.Component {
 
   render() {
     const { pending } = this.props
-    const disableEdit = !pending
     const {
       expirationDuration,
       maxNumberOfKeys,
@@ -157,8 +156,8 @@ export class CreatorLockForm extends React.Component {
             onChange={this.handleChange}
             defaultValue={name}
             data-valid={valid.name}
-            required={!disableEdit}
-            disabled={disableEdit}
+            required={pending}
+            disabled={!pending}
           />
         </FormLockName>
         <FormLockDuration>
@@ -170,8 +169,8 @@ export class CreatorLockForm extends React.Component {
             onChange={this.handleChange}
             defaultValue={expirationDuration}
             data-valid={valid.expirationDuration}
-            required={!disableEdit}
-            disabled={disableEdit}
+            required={pending}
+            disabled={!pending}
           />{' '}
           days
         </FormLockDuration>
@@ -182,10 +181,10 @@ export class CreatorLockForm extends React.Component {
             onChange={this.handleChange}
             value={maxNumberOfKeys}
             data-valid={valid.maxNumberOfKeys}
-            required={!disableEdit}
-            disabled={disableEdit}
+            required={pending}
+            disabled={!pending}
           />
-          {!disableEdit && !unlimitedKeys && (
+          {pending && !unlimitedKeys && (
             <LockLabelUnlimited onClick={this.handleUnlimitedClick}>
               Unlimited
             </LockLabelUnlimited>
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -135,6 +135,8 @@ export class CreatorLockForm extends React.Component {
   }
 
   render() {
+    const { address } = this.props
+    const disableEdit = !!address
     const {
       expirationDuration,
       maxNumberOfKeys,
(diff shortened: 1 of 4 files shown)


unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -135,6 +135,8 @@ export class CreatorLockForm extends React.Component {
   }
 
   render() {
+    const { address } = this.props
+    const disableEdit = address.length === 42
```
julien51: Do you mind clarify what is 42 here and why this matters?

unlock-app/src/stories/creator/CreatorLockForm.stories.js
```diff
@@ -22,17 +22,13 @@ storiesOf('CreatorLockForm', module)
     // TODO: implement this
     const lock = {
       keyPrice: '10000000000000000000',
-      expirationDuration: '172800',
+      expirationDuration: 172800,
```
julien51: Not sure why this change is part of the PR? Could you please clarify?

unlock-app/src/stories/creator/CreatorLockForm.stories.js
```diff
@@ -22,17 +22,13 @@ storiesOf('CreatorLockForm', module)
...
       address: '0xab7c74abc0c4d48d1bdad5dcb26153fc8780f83e',
       name: 'Existing Lock',
     }
     return (
-      <CreatorLockForm
-        lock={lock}
-        hideAction={() => {}}
-        createLock={() => {}}
-      />
+      <CreatorLockForm hideAction={() => {}} createLock={() => {}} {...lock} />
```
julien51: I am not sure why you change the `lock={lock}` into `{...lock}`? I think it is probably cleaner to keep the object "as-is" rather than pass its value in order to differentiate between the actual values and the other props which are not lock properties, such as `hideAction` or `createLock`.

unlock-app/src/components/creator/CreatorLockForm.js
cellog: the difference between a new lock and a full lock is the new lock uses a `uniqid()` as its address, which is length 18. A standard lock address is 42. It may be better to explicitly pass in a "editing" prop for this.

unlock-app/src/stories/creator/CreatorLockForm.stories.js
cellog: it generated a console warning, which I assumed was due to code changes in this PR since I hadn't seen it before

unlock-app/src/stories/creator/CreatorLockForm.stories.js
cellog: this has been part of the API for CreatorLockForm since #1029 

If you'd like, it would be relatively trivial to change this post-this merge in another PR, I'm happy to open an issue and self-assign.

unlock-app/src/components/creator/CreatorLockForm.js
julien51: I think this is too brittle and we should not use the address to determine that a lock is `pending` because we recently discovered that we could get the address of the lock even though the transaction was not mined previously.

There is actually a `pending` value on the lock object which you can re-use here?

cellog: excellent! That is what I was looking for, thanks

unlock-app/src/stories/creator/CreatorLockForm.stories.js
julien51: Let's remove this from this PR then.

unlock-app/src/stories/creator/CreatorLockForm.stories.js
julien51: Ha indeed, it would be much cleaned to use the `lock` object. Please open an issue and asisgn it to yourself and work on that after all is done on this front.

cellog: #1075

unlock-app/src/stories/creator/CreatorLockForm.stories.js
cellog: ok

cellog:
@julien51 ready for final (hopefully) review :)

cellog:
hang on - the console error I referred to is introduced by this PR, so I need to add that back in

unlock-app/src/stories/creator/CreatorLockForm.stories.js
cellog: this is needed because we are using `{...lock}` whereas prior, the lock was not being passed in at all, and props require a number. It was an error that was present but never triggered, so we do need it after all

julien51: ok that works.

Issue #748, opened 2018-12-17
Each key should have a unique and immutable ID for ERC721 compatibility.


nfurfaro:
We currently use the owner's address as the tokenId which means that when ownership is changed, the tokenId itself is changed, which is not how the ERC721spec works.

From the official spec: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md
"Every NFT is identified by a unique uint256 ID inside the ERC-721 smart contract. This identifying number SHALL NOT change for the life of the contract. "

**Additional context**
#148, #758, #759
 @ChrisCates has done some work on this.

ChrisCates:
@nfurfaro, hey, sorry I've been AWOL on that PR. Fell behind on other work last week, I have been catching up this weekend, and, will make sure to get a proper PR for you guys to review.

Chat soon!~

nfurfaro:
@ChrisCates No worries. I just thought it would make sense to break down "ERC721" compatibility into some smaller elements to make for simpler PR's/reviews. This is one such element.

nfurfaro:
@hardlydifficult I was thinking the tokenID could be as simple as a uint that gets incremented, starting with 1 and leaving 0 unused (so we can always check a tokens' existence by checking if its ID == 0)
Is this more or less what you're thinking here?

nfurfaro:
I think once we've defined here what the strategy is for assigning tokenID's we should get it implemented. Not sure that we want to merge changes before v0 launch or not, see my concerns here:
https://unlock-protocol.slack.com/archives/CB59A879U/p1547665104386300

hardlydifficult:
I'm going to take a stab at this.  Let me know if it would be better to wait for our ERC721 conversation to continue, otherwise we can review in more detail once there's a PR.

nfurfaro:
I think it's reasonable to go for it and discuss more on the PR review @hardlydifficult.

PR #1073, opened 2019-01-17
Add numberOfKeysSold

hardlydifficult:
# Description

Add a counter to track each unique key purchased.  And update the notSoldOut check to use this new counter instead of the owner count.

To-date we have been using the number of owners as a proxy for the number of keys sold.  It's up for debate on if purchasing a key second hand should count towards the total number of keys available or not... however it's a concern to use the owner count because one bad actor could move a single key across accounts they control and cause the Lock to appear as sold out.

This is a pre-req for #748 as we will use this counter to define the tokenID in a future PR.

# Checklist:

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -51,6 +51,9 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
   // Max number of keys sold if the keyReleaseMechanism is public
   uint public maxNumberOfKeys;
 
+  // A count of how many new key purchases there have been
+  uint public numberOfKeysSold;
+
   // Keys
   // Each owner can have at most exactly one key
   // TODO: could we use public here? (this could be confusing though because it getter will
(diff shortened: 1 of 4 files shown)


nfurfaro:
I think this is good approach for now, unless we think we'll need more functionality than a uint can give us. For example, if we stuff tokenID's into an array, we get the length, plus the ability to iterate and know what the existing tokenID's are, but it's a tradeoff. We made not need the added functionality/complexity, and to some degree, we can already access tokenID's via the `uint tokenID` field on the `Key` struct, but it's harder/more expensive to access this way. I guess the question is,
Are we going to need to do more with tokenID's than just know the number of them?

hardlydifficult:
I agree, if we start to track more information we may be able to replace this counter with a different data structure.  I think we should hold off on implementing something more until we have a clear requirement driving it, as every new piece of data raises tx fees.  It'll be easy to swap the count for a data.length in the future.

nfurfaro:
Sounds good. Gets the immediate job done while leaving options open.

PR #1076, opened 2019-01-17
Style update - uint vs uint256

hardlydifficult:
# Description

This is just a style change, no functional impact.

At the moment we are inconsistent about how `uint256` is declared.  Sometimes we have `uint` and other times `uint256`.  In Solidity, both these keywords mean the same thing.  i.e. `uint` is a `uint256`.  I believe we should pick one or the other and use it consistently.

I propose we use `uint`.  The reason for this is because that is the uint256 is the native type for operations in Solidity.  This means if you use a different sized int, it must first be casted to uint256 in order to perform an operation.  For example:

```
  // cost 172
  function a()
    public
  {
      uint a = 5;
      uint b = 2;
      uint c = a + b;
  }
  
  // cost 194
  function b()
    public
  {
      uint8 a = 5;
      uint8 b = 2;
      uint8 c = a + b;
  }
```

There may be benefit in using smaller sized ints when storing data, but not for normal usage.  That's why I prefer using `uint` as the standard, calling attention to anytime we use a smaller sized int as we should be more cautious there.

But ultimately it's just nice to be consistent.  So if you would rather use `uint256` instead, I can update this PR.

# Checklist:

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -97,7 +97,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
 
   // Ensure that the caller owns the key
   modifier onlyKeyOwner(
-    uint256 _tokenId
+    uint _tokenId
   ) {
     require(
       address(_tokenId) == msg.sender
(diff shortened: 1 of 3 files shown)


nfurfaro:
I'm all for this. I agree this should generally be the size of uint we use unless there's an opportunity to pack them.

PR #1082, opened 2019-01-17
Using a template for svgr which includes propTypes

julien51:
SVG has an option to add a title to the SVG files, but this option does not set the propTypes or
default prop values. By using our own template we can actually achieve that.


unlock-app/src/components/interface/svg/template.js
```diff
@@ -0,0 +1,30 @@
...
+    ${imports}
+    import PropTypes from 'prop-types'
+    const ${componentName} = ${props} => ${jsx}
+
+    ${componentName}.propTypes = {
+      title: PropTypes.string
+    }
+
+    ${componentName}.defaultProps = {
+      title: ''
```
cellog: have you tried using null? This might keep the snapshots the same (using `<title />`)

julien51: I have not... but I remember that when we use type checking in another project it would complain about `null` because null is not a string and we had to replace all of the types with '<type> or null' which was painful.

PR #1087, opened 2019-01-17
Add a requiredNetworkId property to the configuration

### LINKED ISSUE #966: Adjust default network value to be the same as the default network for the NODE_ENV
in the reducer that handles setting network, currently we do:

```js
export const initialState = {
  name: 0,
}
```

It should instead be set to the network that is currently in use, to remove a flash screen on startup in `npm run dev` and to make integration testing simpler

cnasc:
# Description

In order to handle #966, we need to know the numeric ID of the required network so that we can set it as the initial state for the network reducer. This pull request adds a property to the configuration containing the appropriate network ID for the current environment.

Let me know if it would be better to create an `ETHEREUM_NETWORKS_IDS` constant and refer to that instead.

# Checklist:

### DIFF
--- a/unlock-app/src/config.js
+++ b/unlock-app/src/config.js
@@ -58,6 +58,7 @@ export default function configure(
   let providers = {}
   let isRequiredNetwork = () => false
   let requiredNetwork = 'Dev'
+  let requiredNetworkId = 101
   let requiredConfirmations = 12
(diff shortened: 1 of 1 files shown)


cellog:
LG to me, I'll wait for Julien to review

PR #1091, opened 2019-01-17
Add tokenId to key and a method to getTokenIdFor(account)

hardlydifficult:
# Description

This is part of ERC721 support, allowing us to map from tokenId to owner address and then owner address to key.  When we do this, having the tokenId in the key structure will allow us to discover the tokenId for an owner by their address (i.e. it enables the getTokenIdFor method when tokenId changes to a sequenceID).

We could have flipped this around, adding the owner to the key struct instead of the tokenId.  With this approach we would map address to tokenId and tokenId to key.  However since we support 1 key per address it seems we would want to continue primarily referring to key's by address.

The new tests are basically a copy paste of ownerOf slightly modified for checking getTokenIdFor.  I also switched to an async style for readability.

# Checklist:

### DIFF
--- a/smart-contracts/test/Lock/erc721/getTokenIdFor.js
+++ b/smart-contracts/test/Lock/erc721/getTokenIdFor.js
@@ -1,0 +1,37 @@
+
+const Units = require('ethereumjs-units')
+const BigNumber = require('bignumber.js')
+
+const deployLocks = require('../../helpers/deployLocks')
+const Unlock = artifacts.require('../../Unlock.sol')
+
+let unlock, locks
+
+contract('Lock ERC721', (accounts) => {
+  before(async () => {
+    unlock = await Unlock.deployed()
+    locks = await deployLocks(unlock)
+  })
+
+  describe('getTokenIdFor', () => {
+    it('should abort when the key has no owner', async () => {
+      try {
+        await locks['FIRST'].getTokenIdFor(accounts[3])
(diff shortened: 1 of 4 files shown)


smart-contracts/test/Lock/erc721/getTokenIdFor.js
```diff
@@ -0,0 +1,37 @@
...
+    it('should return the tokenId for the owner\'s key', async () => {
+      await locks['FIRST'].purchaseFor(accounts[1], 'Satoshi', {
+        value: Units.convert('0.01', 'eth', 'wei'),
+        from: accounts[1]
+      })
+      let address = await locks['FIRST'].getTokenIdFor(accounts[1])
+      // Note that as we implement ERC721 support, the tokenId will no longer
+      // be the same as the user's address
+      assert(address.eq(new BigNumber(accounts[1])))
+    })
```
nfurfaro: Maybe add a test checking that `numberOfKeysSold` gets incremented correctly.

hardlydifficult: I believe that's covered here: https://github.com/unlock-protocol/unlock/blob/master/smart-contracts/test/Lock/purchaseFor.js#L181

nfurfaro: so it is. Looks good.

PR #1096, opened 2019-01-17
Add mapping tokenId -> ownerAddress

hardlydifficult:
# Description

This mapping allows us to discover an owner, and therefore a key, from a tokenId.  This gets us one step closer to decoupling tokenId from ownerAddress.

Each tokenId can have at most exactly one owner at a time.  At the moment tokenId is still defined by the owner's address, but eventually this will change to a sequence Id.  When that happens we can consider using an array of token -> address instead (and possibly replace numberOfKeysSold with that array.length).

Existing tests should confirm these changes do not cause regressions.

# Checklist:

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -60,6 +60,12 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
   // return 0 values when missing a key)
   mapping (address => Key) internal keyByOwner;
 
+  // Each tokenId can have at most exactly one owner at a time.
+  // Returns 0 if the token does not exist
+  // TODO: once we decouple tokenId from owner address (incl in js), then we can consider
+  // merging this with numberOfKeysSold into an array instead.
+  mapping (uint => address) internal ownerByTokenId;
+
   // Addresses of owners are also stored in an array.
   // Addresses are never removed by design to avoid abuses around referals
   address[] public owners;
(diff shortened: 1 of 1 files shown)


nfurfaro:
I think this mapping makes perfect sense in the current context. 
If tracking past owners becomes necessary, this could become `mapping(uint => address[]  ) allOwnersByTokenID;` where the last address pushed is always the current owner. But I don't know yet that is going to be needed.

nfurfaro:
One thing I noticed is that we ned to be careful about Where in the contract we declare our new storage, keeping in mind that, while this is going to be v0 on mainnet, it's a zos upgrade on rinkeby and we don't want a storage-collision. I always try to keep uints, mappings, etc grouped together, but we're going to have to move them all  to the end, just above our modifiers. Pretty sure zos will yell at us otherwise (if we're lucky)

julien51:
HA! I was raising the point that @nfurfaro raised but I think we will have to keep `owners` because if I understanding this correctly we will not be able to use `allOwnersByTokenID` to find easily if an owner previously owned a key?

nfurfaro:
@julien you're right, `allOwnersByTokenID` would not be useful in this case, unless you only wanted to know whether an owner had previously owned a specific key. BUt this would assume more knowledege up front.

hardlydifficult:
Yea for the storage order, we will need to be careful about that in the future.  ZOS is supposed to explicitly check for this when we are testing upgrades, and we may be able to add some tests of our own.

In order to track previous owners, we could keep a non-zero expiration date in keyByOwner (as we do today).

nfurfaro:
I'm assuming we'll have to modify our storage before upgrade, but that can be done in a cleanup PR I think. I think we're good to merge.

PR #1107, opened 2019-01-17
Update truffle-config.js with mainnet network

nfurfaro:
This PR fixes#896 by adding a network configuration for mainnet in our `truffle-config.js` file. It's a WIP and likely needs modification depending on the details of how we'll connect to a mainnet node for deployment.
cc @julien51 

# Checklist:

### DIFF
--- a/smart-contracts/truffle-config.js
+++ b/smart-contracts/truffle-config.js
@@ -37,8 +37,14 @@ module.exports = {
     rinkeby: {
       provider: new Web3.providers.HttpProvider(rinkebyProviderUrl),
       network_id: '4' // Network Id for Rinkeby
+    },
+    mainnet: {
+      host: 'localhost',
+      port: 8545,
+      network_id: 1,
+      gas: 7000000,
+      gasPrice: 35000000000
     }
-
   },
   solc: {
     optimizer: {


hardlydifficult:
Will localhost work for this? Maybe it should point to an Infura node

nfurfaro:
We currently have a node running on AWS, but I want to check in with @julien51 ajout this.

nfurfaro:
Update: After discussing with  @julien51, I think these setting should be fine, at least as a starting point. As far as the `gas` and `gasPrice` I think these are reasonable figures but would appreciate a second opinion @hardlydifficult . I've checked against what zOS is using to deploy their own demo projects to mainnet, as well as Aragon.. The current EthGasStation "standard"  gas price is 2.5 Gwei, so 3.5 seems reasoinable to ensure we don't wait too long. One thing we can double check though is to run `eth.getBlock("latest").gasLimit` on our AWS mainnet node to see where it's at. We may be able to bump ours up a bit depending on what this returns.

hardlydifficult:
Looks good.

One thing we could consider, but is probably not necessary, is using an Oracle to dynamically assign the gasPrice.  That can't be in the config directly, but could be added to the migrations logic, e.g. https://web3js.readthedocs.io/en/1.0/web3-eth.html#getgasprice

nfurfaro:
@hardlydifficult Yes, dynamically assigning the gasPrice could come in handy when we progress towards automated upgrades on rinkeby. I think the idea for mainnet was to keep it as a somewhat manual process, so probably not necessary for this deployment.

PR #1114, opened 2019-01-18
Moved eslint package to root of project

julien51:
Since all sub-projects are using eslint we can save a couple bytes on the image and a couple seconds on CI by only installing once and sharing the binary between each sub-repo.


package.json
```diff
@@ -7,14 +7,30 @@
...
+    "prettier-eslint-cli": "^4.7.1"
   },
   "devDependencies": {
-    "husky": "^1.2.0",
-    "ganache-cli": "^6.2.3"
+    "ganache-cli": "^6.2.3",
+    "husky": "^1.2.0"
   },
   "scripts": {
+    "link-parent-bin": "link-parent-bin -c . -s true",
```
julien51: This copies the binaries from the root into the subfolder's node_modules too.

julien51: And right now it does it in a dumb way which means we have a couple danging `node_modules` (in wiki in particular...) but I have [commented](https://github.com/nicojs/node-link-parent-bin/issues/7#issuecomment-454042779) on an issue with that project to allow us to specify which things folder we need to have this copied to.

julien51:
@cnasc Great point! Added the issue...

PR #1125, opened 2019-01-18
adding a function which loads the past transactions on a lock

julien51:
This will be used to display pending lock price updates


unlock-app/src/services/web3Service.js
```diff
@@ -299,6 +299,36 @@ export default class Web3Service extends EventEmitter {
     )
   }
 
+  /**
+   * This function is able to retrieve the past transaction on a lock as long as these transactions
+   * triggered events.
+   * @param {*} lockAddress
+   * @param {*} userAddress
```
akeem: Would we be using `userAddress` in this method in the future?

unlock-app/src/services/web3Service.js
```diff
@@ -299,6 +299,36 @@ export default class Web3Service extends EventEmitter {
...
+   */
+  getPastLockTransactions(lockAddress) {
+    const lockContract = new this.web3.eth.Contract(
+      LockContract.abi,
+      lockAddress
+    )
+    lockContract.getPastEvents(
+      'allevents',
+      {
+        fromBlock: 0, // TODO start only when the smart contract was deployed?
```
akeem: Is it taxing to request transactions from that far back?  Should we consider adding a recent-ish block number maybe something from the fall of last year?

julien51: I did a bunch of tests and that did not seem to actually matter so not sure what to think.

PR #1143, opened 2019-01-21
Adds unique tokenID's for Keys

### LINKED ISSUE #748: Each key should have a unique and immutable ID for ERC721 compatibility.

nfurfaro:
This PR fixes #748 by introducing the use of sequential, unique, immutable tokenID's, with the first key purchased assigned the `tokenId` of `1`, and so on.
Testing for tokenID's was already done, so this was mostly a matter of modifying tests to pass the correct ID (rather than the address of the owner) where required. There is much work to do in refactoring tests to use async/await syntax, but this PR is big enough already. 

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -119,7 +119,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     uint _tokenId
   ) {
     require(
-      ownerByTokenId[_tokenId] == msg.sender
+      ownerByTokenId[_tokenId] == msg.sender, "Not the key owner"
     );
     _;
   }
@@ -220,7 +220,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     hasValidKey(_from)
     onlyKeyOwnerOrApproved(_tokenId)
   {
-    require(_recipient != address(0));
+    require(_recipient != address(0), "Can't Transfer to 0x Address");
 
     uint previousExpiration = keyByOwner[_recipient].expirationTimestamp;
 
@@ -228,6 +228,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
       // The recipient did not have a key previously
       owners.push(_recipient);
       ownerByTokenId[_tokenId] = _recipient;
+      keyByOwner[_recipient].tokenId = _tokenId;
     }
 
     if (previousExpiration <= now) {
@@ -539,7 +540,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     internal
     notSoldOut()
   { // solhint-disable-line function-max-lines
-    require(_recipient != address(0));
+    require(_recipient != address(0), "Can't Purchase For 0x Address");
 
     // Let's get the actual price for the key from the Unlock smart contract
     IUnlock unlock = IUnlock(unlockProtocol);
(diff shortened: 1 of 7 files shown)


smart-contracts/contracts/PublicLock.sol
```diff
@@ -59,6 +59,9 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
   // A count of how many new key purchases there have been
   uint public numberOfKeysSold;
 
+  // The tokenID counter keeps track of the last ID # that was assigned
+  uint public currentTokenID;
```
hardlydifficult: I created numberOfKeysSold for this purpose.

smart-contracts/contracts/PublicLock.sol
```diff
@@ -219,12 +222,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     if (previousExpiration == 0) {
       // The recipient did not have a key previously
       owners.push(_recipient);
-      // Note: not using the tokenId above since ATM we do not transfer tokenId's
-      // this will change when we decouple tokenId from address
-      ownerByTokenId[uint(_recipient)] = _recipient;
-      // At the moment tokenId is the user's address, but as we work towards ERC721
-      // support this will change to a sequenceId assigned at purchase.
-      keyByOwner[_recipient].tokenId = uint(_recipient);
```
hardlydifficult: We do still need to set the tokenId in the struct here.

smart-contracts/test/Lock/erc721/approve.js
```diff
@@ -23,7 +23,7 @@ contract('Lock ERC721', (accounts) => {
     describe('when the token does not exist', () => {
       it('should fail', () => {
         return locks['FIRST']
-          .approve(accounts[2], accounts[1], {
+          .approve(accounts[2], 42, {
```
hardlydifficult: Instead of hardcoding the tokenId in tests, maybe use `await locks['FIRST'].getTokenIdFor("ownerAddress")`

hardlydifficult: lol I added this comment to the one fake tokenId example, but the point applies to other examples :)

nfurfaro:
@hardlydifficult thanks for the review, will look at this tonight.

smart-contracts/test/Lock/erc721/approve.js
nfurfaro: yes, got it. I started doing that, but never went back and modified tests I had already fixed. Will do that.

smart-contracts/contracts/PublicLock.sol
hardlydifficult: Good point, I think we are missing a test case here.  To test, we could add to getTokenIdFor.js - just confirm that the result for the recipient of a transferFrom is not 0.

smart-contracts/test/Lock/erc721/approve.js
```diff
@@ -41,53 +41,55 @@ contract('Lock ERC721', (accounts) => {
...
-            })
+        it('should fail', async () => {
+          try {
+            let ID = await locks['FIRST'].getTokenIdFor(accounts[1])
+            await locks['FIRST']
+              .approve(accounts[2], ID, {
+                from: accounts[2]
+              })
+          } catch (error) {
+            assert.equal(error.message, 'VM Exception while processing transaction: revert')
```
hardlydifficult: Hmm, why is there no revert reason in this error?

hardlydifficult:
Looking good.  I added one more comment about the missing test case. 

BTW the async tests are much easier to read.  Happy to see us slowly migrate :)

smart-contracts/test/Lock/erc721/approve.js
nfurfaro: Fixed. Added reason to contract and updated test cases (x2)

hardlydifficult:
FYI I'm not an authorized reviewer, so someone else will need to provide feedback here as well.

PR #1170, opened 2019-01-23
GlobalErrorConsumer component

cellog:
# Description

This component implements the consumer portion of the `GlobalErrorProvider`. It will be rebased on #1169 when that is approved.

The `GlobalErrorConsumer` component implements a simple idea: We want to be able to define the components that wrap our error component, but do not want to deal with the implementation details of the error components themselves. To do this, the consumer takes as a prop `displayError`: a function that receives as its parameters the error element (not component) and children element(s). Because these are passed as elements, they can be used directly in jsx (as `{error}` and `{children}` for example).

Thus, the `withConfig` implementation of this would be:

```js
function displayErrorWithConfig(error, children) {
  const { router } = props
  if (router && !Component.skipConstraints && error) {
    return (
      <Layout title="">{error}</Layout>
    )
  }
  return <>{children}</>
}
```

and withConfig would simply call:

```js
  return (
    <GlobalErrorConsumer displayError={displayErrorWithConfig}>
      <ConfigContext.Consumer>
        {config => <Component {...props} config={config} />}
      </ConfigContext.Consumer>
    </GlobalErrorConsumer>
```

For the paywall, we might use

```js
displayErrorForPaywall(error, children) {
  return (
    <>
      {children}
      {error}
    </>
  )
}
```

and then we can pass the locks as children to the GlobalErrorConsumer, so that we always display the locks even if there is an error, but provide the error message context for those who are misconfigured in some way.

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/GlobalErrorConsumer.js
+++ b/unlock-app/src/components/interface/GlobalErrorConsumer.js
@@ -1,0 +1,38 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { GlobalErrorContext } from '../../utils/GlobalErrorProvider'
+import { mapping } from '../creator/FatalError'
+import Layout from './Layout'
+
+const Consumer = GlobalErrorContext.Consumer
+
+export const displayError = (error, children) => {
+  if (error) return <Layout title="">{error}</Layout>
+  return <>{children}</>
+}
+
+export default function GlobalErrorConsumer({ displayError, children }) {
+  return (
+    <Consumer>
+      {({ error, errorMetadata }) => {
+        // if the error condition exists, set it to the mapped fatal error component
+        // or to the fallback
+        // if no error exists, set it to false
+        const Error = error ? mapping[error] || mapping['*'] : false
+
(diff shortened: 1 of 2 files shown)


cellog:
@julien51 rebased and ready for review

unlock-app/src/components/interface/GlobalErrorConsumer.js
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from 'prop-types'
+
+import { GlobalErrorContext } from '../../utils/GlobalErrorProvider'
+import { mapping } from '../creator/FatalError'
+import Layout from './Layout'
+
+const Consumer = GlobalErrorContext.Consumer
+
+export const displayError = (error, children) => {
+  if (error) return <Layout title="">{error}</Layout>
```
julien51: Do you mind putting the return statement on a new line with `{}`? I think this would be more consistence with out style guide (I thought there was an eslint rule for that?)

cellog: oops saw this post-merge, I'll do it in a separate PR

PR #1182, opened 2019-01-23
Require expirationDuration to be < 100 years.

hardlydifficult:
# Description

When creating a lock, the expirationDuration must be < 100 years.

This PR addresses part of issue #1110.  It's a slight alternative to https://github.com/unlock-protocol/unlock/pull/1179 putting the require statement inside PublicLock instead. This is because:

 - Longer term I believe we should remove parameters from Unlock.sol and accept arbitrary data instead. This will free us up to define Lock types which require a completely different set of configuration information.
 - We may also want to add the ability for a Lock owner to change the expirationDuration, and if we did it's a little cleaner to have the requirement for creation and updates in the same place.

# Checklist:

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -156,7 +156,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     uint _maxNumberOfKeys
   )
   public {
-    require(_expirationDuration < 100 * 365 * 24 * 60 * 60, "Expiration duration must be less than 100 years");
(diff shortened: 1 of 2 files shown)


smart-contracts/contracts/PublicLock.sol
```diff
@@ -156,6 +156,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     uint _maxNumberOfKeys
   )
   public {
+    require(_expirationDuration < 100 * 365 * 24 * 60 * 60, "Expiration duration must be less than 100 years");
```
nfurfaro: Minor, but maybe we could use <= here? Not too consequential either way.

hardlydifficult: done

PR #1200, opened 2019-01-24
prefix all redux actions

### LINKED ISSUE #1052: redux action types can collide
**Describe the bug**
Currently, redux action constants are co-located with the actions. Although this provides a convenience, it is dangerous.

For instance, I just tried to define a constant named UPDATE_LOCK, which already exists.

There is nothing that would have caught this, except that upon dispatching the actions, it would trigger in 2 different reducers with unpredictable results

There are 2 possible solutions

1) prefix each action type with its filename:

```js
const UPDATE_LOCK = 'locks/UPDATE_LOCK'
```

Because we use constants, there is no usability issue

2) move all constants to a single "types.js" file

This is a lot more work initially, but is a possible solution

Because I just ran into this issue, it is something that will come up again, and when it happens it may not be obvious for a while depending on how similar the actions are.

cellog:
# Description

Fixes #1052 by adding a prefix based on the file name to every redux action constant, and fixing the few locations where constants were not being used.

screen shot 2019-01-23 at 11 03 46 pm

# Checklist:

### DIFF
--- a/unlock-app/src/actions/currencyconvert.js
+++ b/unlock-app/src/actions/currencyconvert.js
@@ -1,4 +1,5 @@
-export const SET_ETHER_CONVERSION_RATE = 'SET_ETHER_CONVERSION_RATE'
+export const SET_ETHER_CONVERSION_RATE =
+  'currencyconvert/SET_ETHER_CONVERSION_RATE'
 
 export function setConversionRate(currency, rateFor1Eth) {
   return {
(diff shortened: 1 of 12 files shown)


unlock-app/src/actions/currencyconvert.js
```diff
@@ -1,4 +1,5 @@
-export const SET_ETHER_CONVERSION_RATE = 'SET_ETHER_CONVERSION_RATE'
+export const SET_ETHER_CONVERSION_RATE =
+  'currencyconvert/SET_ETHER_CONVERSION_RATE'
```
julien51: Damned. I wish I did not miss the case issue on that one :(

PR #1201, opened 2019-01-24
remove convert from CreatorLockForm

### LINKED ISSUE #1068: refactor CreatorLockForm tests to not need the convert prop to be false
Currently, the `CreatorLockForm` component has a `convert` prop, which is used to tell the constructor not to convert `keyPrice` and `expirationDuration` to form-friendly values from lock values.

In the app, `keyPrice` is represented in wei, and in the form, it is represented in eth.
In the app, `expirationDuration` is represented in seconds, and in the form, it is represented in days.

The unit tests for `CreatorLockForm` test several error conditions that cannot be tested simply by using `react-testing-library`'s `fireEvent.change` event trigger, because jsdom's implementation of the number input will not allow anything to be fired on it that is not a number. This is the behavior of opera, but unfortunately, both chrome and firefox allow users to input non-number characters.

To remove the convert prop, we will need to find another way to test the form field validation edge cases, and that is the scope of this issue.

ref: #1064

cellog:
# Description

Fixes #1068 by removing convert. In order to make it possible to test against invalid values, we are extremely permissive on input. So, when invalid input is encountered, it is simply left untouched. This way, we leave the error conditions to the form itself to display. In other words, if passed a cockamamie lock, the edit form will simply refuse to allow you to save it until the errors are fixed. Screenshots to show I didn't break anything :)

screen shot 2019-01-23 at 11 50 39 pm
screen shot 2019-01-23 at 11 50 28 pm
screen shot 2019-01-23 at 11 49 44 pm
screen shot 2019-01-23 at 11 49 37 pm
screen shot 2019-01-23 at 11 50 54 pm

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/components/creator/CreatorLockForm.test.js
+++ b/unlock-app/src/__tests__/components/creator/CreatorLockForm.test.js
@@ -33,6 +33,7 @@ describe('CreatorLockForm', () => {
     )
     return ret
   }
+  const secondsInADay = 86400
   const allFormErrors = [
     FORM_EXPIRATION_DURATION_INVALID,
     FORM_KEY_PRICE_INVALID,
@@ -93,7 +94,7 @@ describe('CreatorLockForm', () => {
       }
     })
     it('key expiration is a negative number', () => {
-      const wrapper = makeLockForm({ expirationDuration: -2 * 86400 })
+      const wrapper = makeLockForm({ expirationDuration: -2 * secondsInADay })
 
       expect(wrapper.getByValue('-2').dataset.valid).toBe('false')
     })
@@ -244,7 +245,7 @@ describe('CreatorLockForm', () => {
       )
     })
     it('key expiration is a positive number', () => {
-      const wrapper = makeLockForm({ expirationDuration: 35 * 86400 })
+      const wrapper = makeLockForm({ expirationDuration: 35 * secondsInADay })
 
       expect(wrapper.getByValue('35').dataset.valid).toBe('true')
     })
--- a/unlock-app/src/__tests__/components/creator/CreatorLockForm.test.js
+++ b/unlock-app/src/__tests__/components/creator/CreatorLockForm.test.js
@@ -9,7 +9,7 @@ import {
   FORM_KEY_PRICE_INVALID,
 } from '../../../errors'
 
-import { INFINITY } from '../../../constants'
+import { INFINITY, UNLIMITED_KEYS_COUNT } from '../../../constants'
 
 describe('CreatorLockForm', () => {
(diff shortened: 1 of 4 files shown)


unlock-app/src/__tests__/components/creator/CreatorLockForm.test.js
```diff
@@ -99,9 +93,9 @@ describe('CreatorLockForm', () => {
       }
     })
     it('key expiration is a negative number', () => {
-      const wrapper = makeLockForm({ expirationDuration: -1 })
+      const wrapper = makeLockForm({ expirationDuration: -2 * 86400 })
```
julien51: Nit: when using numeric values like this, feel free to put the corresponding "human" format like "2 days" as comments. This makes it easier to skip the math ;)

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -37,15 +37,31 @@ export class CreatorLockForm extends React.Component {
...
     let maxNumberOfKeys = props.maxNumberOfKeys
-    if (props.convert) {
-      keyPrice = Web3Utils.fromWei(keyPrice, props.keyPriceCurrency)
-      expirationDuration = expirationDuration / props.expirationDurationUnit
-      // unlimited keys is represented differently in the frontend and the backend,
-      // so when a lock has `maxNumberOfKeys` set to UNLIMITED_KEYS_COUNT,
-      // convert it to infinity symbol for frontend display
-      if (maxNumberOfKeys === UNLIMITED_KEYS_COUNT) {
-        maxNumberOfKeys = INFINITY
+    try {
```
julien51: Please add a comment to explain the rationale for the try catch thing. IIf I understanding things correctly this is because we invoke this with different formats which is probably something we should fix  (I k now I have an issue assigned to me for that!)

unlock-app/src/__tests__/components/creator/CreatorLockForm.test.js
cnasc: Just for the sake of argument, when I do things like this I like to have a named variable like `secondsInADay` so that you end up with `2 * secondsInADay` which is a bit more self-documenting and resilient against change (e.g., if we change it in the future to being 3 days, there won't be a comment that can be out of sync with the code).

cellog: I like that @cnasc thanks

unlock-app/src/components/creator/CreatorLockForm.js
cellog: 👍

PR #1237, opened 2019-01-25
pulling remote branches before actually checking that this is a tag build

julien51:
After debugging further I found that travis does not actually checkout all of the branches, just the
commit that it needs to test.
So this fix will actually fetch all branches and will check that the commit is actually on master
before proceeding to a full deploy.

### DIFF
--- a/scripts/deploy-netlify.sh
+++ b/scripts/deploy-netlify.sh
@@ -18,8 +18,11 @@ echo ""
 if [ -n "$TRAVIS_TAG" ] &&
    [ -n "$TRAVIS_COMMIT" ] &&
    [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
-  # Check that the TRAVIS_COMMIT is actually on the master branch [to avoid deploying tags which are not on master]
-  git branch --contains $TRAVIS_COMMIT | grep 'master' >> /dev/null
+  # Check that the TRAVIS_COMMIT is actually on the master branch in
+  # origin to avoid deploying tags which are not on master.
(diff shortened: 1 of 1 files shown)


scripts/deploy-netlify.sh
```diff
@@ -18,8 +18,11 @@ echo ""
 if [ -n "$TRAVIS_TAG" ] &&
    [ -n "$TRAVIS_COMMIT" ] &&
    [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
-  # Check that the TRAVIS_COMMIT is actually on the master branch [to avoid deploying tags which are not on master]
-  git branch --contains $TRAVIS_COMMIT | grep 'master' >> /dev/null
+  # Check that the TRAVIS_COMMIT is actually on the master branch in
+  # origin to avoid deploying tags which are not on master.
+  git config --replace-all remote.origin.fetch +refs/heads/*:refs/remotes/origin/*
+  git fetch
+  git branch -r --contains $TRAVIS_COMMIT | grep 'master' >> /dev/null
```
akeem: So I get what is happening here, but I am not sure why this would be dramatically different from what was here before in this case

julien51: So in practice, before that we were not fetching the remote branches and only checking against the "local" master. The local master actually does not exist. Travis only has the one commit that it is building against.

julien51: The important line is 23rd actually because it "links" again.
See this https://stackoverflow.com/questions/34405047/how-do-you-merge-into-another-branch-using-travis-with-git-commands/44036486 for details.

Issue #1066, opened 2019-01-16
Refactor CreatorLockForm to extract ethereum-specific bits into web3service


cellog:
Currently, `CreatorLockForm` converts locks to form-friendly format in its constructor, and then back out to unlock-app-friendly format in `saveLock()`. We need to extract this functionality out into the web3service in order to contain our ethereum-specific code in there. This will make it possible to swap out for other blockchains in the future if necessary while preserving the same UI.

Ref: #1064

PR #1240, opened 2019-01-26
Using a string of eth as the only format for all kinds of prices and balances

julien51:
We are now using eth as the way to represent prices and balances. We use Eth as a string for now.
All conversions to wei are done *inside* the web3 provider.

### DIFF
--- a/unlock-app/src/propTypes.js
+++ b/unlock-app/src/propTypes.js
@@ -6,17 +6,18 @@ export const address = PropTypes.string
 export const account = PropTypes.shape({
   address: address,
   privateKey: PropTypes.string,
-  balance: PropTypes.string, // Too big to be a number
+  balance: PropTypes.string, // Must be expressed in Eth!
 })
 
 export const lock = PropTypes.shape({
   address,
   name: PropTypes.string,
   expirationDuration: PropTypes.number,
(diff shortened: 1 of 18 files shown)


julien51:
Fixes https://github.com/unlock-protocol/unlock/issues/1066

julien51:
This also makes the `unit` attribute irrelevant... (and we were actually not using it anywhere but tests).

unlock-app/src/propTypes.js
```diff
@@ -6,17 +6,18 @@ export const address = PropTypes.string
...
 export const lock = PropTypes.shape({
   address,
   name: PropTypes.string,
   expirationDuration: PropTypes.number,
-  keyPrice: PropTypes.string,
+  keyPrice: PropTypes.string, // Must be expressed in Eth!
   maxNumberOfKeys: PropTypes.number,
   owner: PropTypes.string,
   outstandingKeys: PropTypes.number,
+  balance: PropTypes.string, // Must be expressed in Eth!
```
cellog: this adds a field to the lock propType. I don't have an issue with that, but is it doing 2 things in this PR?

julien51: Well, I think this is still about the type/format of balances and prices?

Issue #1225, opened 2019-01-25
Use a different (faster!) provider to retrieve data from the chain


julien51:
We currently use the same provider to get data from the chain and send transactions. 
We could use a different provider when loading data from the chain (which does not require an unlocked user!) and when sending transactions.

This should make loading data from the chain significantly faster!

julien51:
cc @cellog because I think this is relevant to your current work on the wallet front ;)

julien51:
One of the key elements is that all transactions sent from `walletService` are then handled by `web3Service` since all data has to be able to be rebuilt from the smart contract/chain. A consequence of this is that `walletService` is very minimalist in its implementation.

julien51:
The `walletService` has now been deployed. The next step is to "watch" transactions from inside the web3Service instead of when we trigger the transactions.

julien51:
The next step is to add a `walletMiddleware` which will be able to invoke the `walletService` methods as well as listens from the event it may trigger.

The bulk of this middleware will be about triggering new transactions and listening to `transaction.new` events triggered when a transaction has been sent.

julien51:
We now have 2 disctinct services: WalletService which is for anything that requires a user signature (transaction sending... etc) and Web3Service wich is technically a read only provider and does not require to go thru a wallet. The next (and final) step is to actually set the right configuration variables to use that!

julien51:
This can be closed after https://github.com/unlock-protocol/unlock/pull/1316 and https://github.com/unlock-protocol/unlock/pull/1317 have been merged!

PR #1257, opened 2019-01-28
Watching transactions

julien51:
Since the piece triggering transactions and the piece actually loading data from the chain will be
distinct, we should not rely on the `sendTransaction` flow to "watch" on-going transaction.
Most of the web3 providers do not support using `filters`, so we implement a lightweight version
of this using `setTimeout` on transactions which are pulled from the chain but for which we
want to get updates.

### DIFF
--- a/unlock-app/src/__tests__/services/web3Service.test.js
+++ b/unlock-app/src/__tests__/services/web3Service.test.js
@@ -9,7 +9,6 @@ import LockContract from '../../artifacts/contracts/PublicLock.json'
 import configure from '../../config'
 import { TRANSACTION_TYPES } from '../../constants'
 import {
-  MISSING_TRANSACTION,
   NOT_ENABLED_IN_PROVIDER,
   MISSING_PROVIDER,
   NON_DEPLOYED_CONTRACT,
@@ -120,6 +119,8 @@ nock.emitter.on('no match', function(clientRequestObject, options, body) {
 
 let web3Service
 
+const { providers, blockTime } = configure()
+
 describe('Web3Service', () => {
   beforeEach(() => {
     nock.cleanAll()
@@ -205,7 +206,6 @@ describe('Web3Service', () => {
     let enable
 
     beforeEach(done => {
-      const { providers } = configure()
       nock.cleanAll()
 
       enable = providers.HTTP.enable = jest.fn(() => Promise.resolve())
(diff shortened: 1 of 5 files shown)


unlock-app/src/services/web3Service.js
```diff
@@ -516,8 +507,16 @@ export default class Web3Service extends EventEmitter {
...
     ]).then(([blockNumber, blockTransaction]) => {
       if (!blockTransaction) {
-        return this.emit('error', new Error(MISSING_TRANSACTION))
+        // The transaction is still pending: it has been sent to the network but not
+        // necessarily received by the node we're asking it (and not mined...)
+        this._watchTransaction(transactionHash)
+        return this.emit('transaction.updated', transactionHash, {
+          status: 'submitted',
+          confirmations: 0,
+          blockNumber: Number.MAX_SAFE_INTEGER,
```
akeem: Would it be better to emit this without a blockNumber? Is there a downstream consumer with an expectation that this would be set?

unlock-app/src/__tests__/services/web3Service.test.js
```diff
@@ -1454,5 +1482,22 @@ describe('Web3Service', () => {
         })
       })
     })
+
+    describe('_watchTransaction', () => {
+      it('should set a timeout based on half the block time', () => {
+        expect.assertions(1)
+        jest.useFakeTimers()
+        web3Service._watchTransaction('0x')
+        expect(setTimeout).toHaveBeenCalledTimes(1)
```
akeem: How would you feel about making this assertion a bit more specific to ensure that the timing is close to our expected duration( half of the block time)?

unlock-app/src/__tests__/services/web3Service.test.js
```diff
@@ -473,55 +472,138 @@ describe('Web3Service', () => {
...
+              '0xdc7c95899e030f3104871a597866767ec296e948a24e99b12e0b251011d11c36',
+            blockNumber: `0x${(14).toString('16')}`,
+            transactionIndex: '0x0d',
+            from: '0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1',
+            to: '0xcfeb869f69431e42cdb54a4f4f105c19c080a601',
+            value: '0x0',
+            gas: '0x16e360',
+            gasPrice: '0x04a817c800',
+            input:
+              '0x2bc888bf00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000278d00000000000000000000000000000000000000000000000000002386f26fc10000000000000000000000000000000000000000000000000000000000000000000a',
```
akeem: Would you be able to adjust the formatting of this input value?

unlock-app/src/services/web3Service.js
julien51: So, the goal of adding a very large number is because we sort things (like locks) based on their recency. If we put no blockNumber I guess that would be at the bottom... but maybe it is worth changing the sorting function?

unlock-app/src/__tests__/services/web3Service.test.js
julien51: That's a good one. Adding it.

unlock-app/src/__tests__/services/web3Service.test.js
julien51: Let me try but I am actually not sure how this is built :/ (it was taken from a nock spy)

julien51: huh, this is a mess :/ I'll push that to another PR if you don't mind.

julien51:
Changes pushed! PTAL ;)

julien51:
BTW, this is more progress toward https://github.com/unlock-protocol/unlock/issues/1225

akeem:
LG 👍

PR #1261, opened 2019-01-28
Implements versioning locks

### LINKED ISSUE #1116: Versionning the Lock Smart Contract
Earlier in the week we added the ability for the lock owner to list the keys sold on their lock.
The initial approach was to retrieve each owner one by one and then pull the corresponding key one by one. This was fairly low performance but "worked".

@akeem implemented a function on the publicLock to make that faster via pagination https://github.com/unlock-protocol/unlock/pull/1018 (in the form of getting X owners at once and then getting all of their keys). Not 100x better but still, an improvement.

Later, he introduced a new function inside the web3Service which used that new smart contract function. https://github.com/unlock-protocol/unlock/pull/1062

Unfortunately, all of the previously deployed locks on Rinkeby do not support this new smart contract method, which means that listing keys do not work for them anymore.

As a way to move forward, I suggest the following:

* Update our rinkeby deployed contracts so that any new lock will now support the new API. This means that our mainnet version also supports this new API.
* All of the previously deployed lock will keep failing on rinkeby (but I guess that is why we have rinkeby). 
* We should implement version numbers in the Locks which are updated when adding new features so that the web app (or other clients) may know what kind of features are available to them.
* We should either find a way to upgrade a lock, or (probably easier), to migrate them (deploy a new version of them with the data "copied" from the old lock to the new lock.

nfurfaro:
This PR adds a version number to each lock contract, and fixes #1116 
The version is set in the contructor, and is hard-coded into the `Unlock.createLock()` function, as that contract must be updated at the same time as the Lock version is bumped.

# Checklist:

### DIFF
--- a/smart-contracts/contracts/Old_Locks/V0_PublicLock
+++ b/smart-contracts/contracts/Old_Locks/V0_PublicLock
@@ -62,9 +62,6 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
   // A count of how many new key purchases there have been
   uint public numberOfKeysSold;
 
-  // The version number for this lock contract,
-  uint public publicLockVersion; // 1
-
   // Keys
   // Each owner can have at most exactly one key
   // TODO: could we use public here? (this could be confusing though because it getter will
@@ -159,8 +156,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     address _owner,
     uint _expirationDuration,
     uint _keyPrice,
-    uint _maxNumberOfKeys,
-    uint _version
+    uint _maxNumberOfKeys
   )
   public {
     require(_expirationDuration <= 100 * 365 * 24 * 60 * 60, "Expiration duration exceeds 100 years");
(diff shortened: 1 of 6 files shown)


nfurfaro:
This will require us to upgrade Unlock to start deploying the new version of Locks. This approach means that each new version of Lock needs an upgrade to Unlock.

hardlydifficult:
Hmm, could you maybe rename `PublicLock.sol` to `PublicLockV1.sol` in a future PR?  That way we can see the diff to the contract itself.

nfurfaro:
Hmm. I see what you're saying, way easier to review the changes that way. I would like to keep the old lock versions around for reference as well.
Maybe the way forward is to copy/paste the Old lock into a new file with a different name (and .gitignore it) and then make the changes to the lock contract without changing its name till after review.

hardlydifficult:
Could do it the other way around to, rename the file first and then this PR.  Not sure if this is overkill, but it would give you a full history and easy to review PRs:
 - PR 1: Rename to PublicUnlockV0.sol
 - PR 2: Copy paste to PublicUnlockV1.sol
 - PR 3: Modify PublicUnlockV1.sol to including versioning

nfurfaro:
Closing as this PR is superceded by #1276

PR #1276, opened 2019-01-29
Lock version numbers

nfurfaro:
# Description
This PR supercedes #1261 which I've closed, as it was bumping the lock version in addition to laying the groundwork for version numbers. 
I've added a markdown doc in the new directory `smart-contracts/old_lock_versions` with details on bumping the lock version number.

###Still to do on Issue #1116:
- [ ] deal with burning/killing old locks
- [ ] deal with migrating old lock to new versions

# Checklist:

This is the `lockVersions.MD` file:
### Lock Versions

As the lock contract evolves, there will likely be some changes which break the front-end integration. When this occurs, we will bump the version number on the lock contract itself. For ease of reviewing changes, the process might look like this:

#### PR 1/2

- the old lock version (ie: `PublicLock.sol`) is copied & pasted into a new file (`oldPublicLock.sol`) in the `smart-contracts/old_lock_versions` directory so we keep it for reference.
- the old lock version has its name changed to reflect the new version we're about to update to (ie: `PublicLockV1.sol`).

#### PR 2/2

- the version number for the new Lock contract is edited in `Unlock.sol.createLock()`.
- Unlock is Upgraded using zos so that it will start to deploy the new lock version moving forward.

### DIFF
--- a/smart-contracts/test/Lock/Lock.js
+++ b/smart-contracts/test/Lock/Lock.js
@@ -6,7 +6,7 @@ const Unlock = artifacts.require('../Unlock.sol')
 
 let unlock, locks
 
-contract('Lock', (accounts) => {
+contract('Lock', accounts => {
   before(() => {
     return Unlock.deployed()
       .then(_unlock => {
@@ -26,7 +26,8 @@ contract('Lock', (accounts) => {
       lock.keyPrice.call(),
       lock.maxNumberOfKeys.call(),
       lock.outstandingKeys.call(),
-      lock.numberOfOwners.call()
+      lock.numberOfOwners.call(),
+      lock.publicLockVersion.call()
     ]).then(
       ([
         owner,
@@ -34,7 +35,8 @@ contract('Lock', (accounts) => {
         keyPrice,
         maxNumberOfKeys,
         outstandingKeys,
-        numberOfOwners
+        numberOfOwners,
+        publicLockVersion
       ]) => {
         expirationDuration = new BigNumber(expirationDuration)
         keyPrice = new BigNumber(keyPrice)
@@ -43,13 +45,11 @@ contract('Lock', (accounts) => {
         numberOfOwners = new BigNumber(numberOfOwners)
         assert.strictEqual(owner, accounts[0])
         assert.equal(expirationDuration.toFixed(), 60 * 60 * 24 * 30)
-        assert.strictEqual(
-          Units.convert(keyPrice, 'wei', 'eth'),
-          '0.01'
-        )
+        assert.strictEqual(Units.convert(keyPrice, 'wei', 'eth'), '0.01')
         assert.equal(maxNumberOfKeys.toFixed(), 10)
         assert.equal(outstandingKeys.toFixed(), 0)
         assert.equal(numberOfOwners.toFixed(), 0)
+        assert.equal(publicLockVersion.toFixed(), 1) // needs updating each lock-version change
       }
     )
   })
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -62,6 +62,9 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
   // A count of how many new key purchases there have been
   uint public numberOfKeysSold;
 
+  // The version number for this lock contract,
+  uint public publicLockVersion;
+
   // Keys
   // Each owner can have at most exactly one key
   // TODO: could we use public here? (this could be confusing though because it getter will
@@ -121,7 +124,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     );
     _;
   }
-  
+
   // Ensures that a key has an owner
   modifier isKey(
     uint _tokenId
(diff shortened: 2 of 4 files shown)


nfurfaro:
I will add details to the doc if we're ok with this approach. For instance, everywhere we import PublicLock.sol, we'll need to update the name to the new version name as well, so in Unlock.sol, in our tests, in the front-end(?)

hardlydifficult:
Some thoughts:
 - I think the first version with a Lock version defined should be 1.  
 - The Locks which were deployed with out v0 release will not have that variable, which means calling it should fallback and then return 0.  By starting with `1` we can identify the unversioned Lock from v0 vs a newer Lock implementation.
 - As for when we should increase the Lock version, maybe anytime something changes in the .sol file between mainnet releases.  e.g. if we make 5 more PRs to Lock before Unlock releases v1, the Lock could remain version `1`.

smart-contracts/test/Lock/Lock.js
```diff
@@ -43,13 +45,11 @@ contract('Lock', (accounts) => {
...
         assert.equal(expirationDuration.toFixed(), 60 * 60 * 24 * 30)
-        assert.strictEqual(
-          Units.convert(keyPrice, 'wei', 'eth'),
-          '0.01'
-        )
+        assert.strictEqual(Units.convert(keyPrice, 'wei', 'eth'), '0.01')
         assert.equal(maxNumberOfKeys.toFixed(), 10)
         assert.equal(outstandingKeys.toFixed(), 0)
         assert.equal(numberOfOwners.toFixed(), 0)
+        assert(publicLockVersion.eq(0)) // needs updating each lock-version change
```
hardlydifficult: Minor style:
Maybe use the `assert.equal` style from the line above.  I had previously implemented asserts using `assert(var.eq(other))` but the trouble is when the assert fails the error message is not helpful.  Using `toFixed()` makes a string out of the number so when the `assert.equal` fails it prints the actual values, which is helpful while debugging.

nfurfaro: sounds good.

julien51:
Aslo, there may actually be no need to keep the old ABI in a separate folder. After all we could use git to get them back ;)

nfurfaro:
@julien51 Regarding your points above:

> "...we could use git to get them back " 

 I'm fine with this approach.
> "...how do to deal with "old" versions"

I need to be clear on how/where we're using the ABI's, so please correct me if I have this wrong:
- We're not commiting artifacts to git anymore, but rather building them fresh as needed.
- We have 2 copies of the ABI's (and the rest of the artifacts) one in `/smart-contracts/build` and one in `/unlock-app/src/artifacts/`.
- the front-end uses the artifacts/ABI's in `unlock-app/`
- We want to support "old" versions of locks.
- We are currently using a "conditional" approach in `web3Service`.
- with the addition of version numbers on the locks, we have a way to call the lock and know which ABI to use to interact with it based on the version.
- In order for this to work, we'd need to keep the ABI accessible to the front end, and use the correct one for the version, so roughly:
```
switch(lockVersionNumber) {
  case 0:
    PublicLock = PublicLockV0.at(lockAddress)
    break;
  case 1:
    PublicLock = PublicLockV1.at(lockAddress)
    break;
  default:
    // load most recent version of ABI
}

```
As creators migrate (another related discussion) to the newest version we track this, and when all V0 Locks have been migrated to V1, we no longer need to keep the ABI for V0 around.
I'm may still be missing part of the picture here...

julien51:
I think this is the right big picture.
My concern is around keep track of the old ABIs. Is there a way that we actually don't need to? (for example if we only *add*  stuff to the PublicLocks then I think this is fine?)

nfurfaro:
@julien51 A couple thoughts here. 
First, if we only _add_ functionality to new locks and never remove old functions (even if we decide we no longer need/want them) then I _don't think_ we would need to keep the old ABI around, as the old version would be default be a subset of the new  version ABI. That said, I don't love the idea of restricting ourselves to only adding functionality on the Locks, but maybe that's just me wanting too much;) (I know we already have this restriction on the Unlock contract for different reasons)
There's another possibility we could explore, which is the ability to extract the abi from a deployed contract: https://ethereum.stackexchange.com/questions/37931/is-there-a-way-to-extract-abi-from-a-deployed-contract
Either way, I think we will find a way forward which does not involve manually tracking/storing the old ABI's

julien51:
As I was thinking more about this, I was wondering if we could not package our ABI for a given lock version and ship them to `npm`. This way we can easily include the ABI on any application that uses it, we also know that the package has the right version we look for.

We would create as many modules as we have lock versions. For example `publick-lock-abi-0`, `publick-lock-abi-1`... etc.

nfurfaro:
This is an approach I hadn't considered... I like it. 
I looked into fetching the ABI from the deployed contract. It's pretty straightforward, but has a prerequisite that the contract has to be verified first. This is fine for something like the Unlock contract, but would need to be done separately for every single lock deployed, so not very practical.

nfurfaro:
@julien51 So, steps to push this through:
- [ ] Package and publish public-lock-abi-0 (pre-version numbers) ?
- [ ] clean up pr to remove uneeded directory for old abis
- [ ] merge and move on!

Sound ok?

nfurfaro:
This is what I have so far:
```
{
  "name": "public-lock-abi-0",
  "version": "1.0.0",
  "description": "The ABI(application binary interface) for version 0 of Unlock-protocol's publicLock contract.",
  "main": "abi_V0.json",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "unlock-protocol",
  "license": "ISC"
}
```
And, in `abi_V0` it's just the contract name and ABI, not the bytecode, deployed addresses, etc...

PR #1283, opened 2019-01-29
Plugging the walletMiddleware and services to replace web3Service for transactions

julien51:
This is the final unplugging of web3Service and using walletService (thru their respective middlewares).
This is part of #1225 (and probably a lot of other things...)

### DIFF
--- a/unlock-app/src/middlewares/web3Middleware.js
+++ b/unlock-app/src/middlewares/web3Middleware.js
@@ -3,56 +3,31 @@
 import { LOCATION_CHANGE } from 'react-router-redux'
 import {
   ADD_LOCK,
-  CREATE_LOCK,
-  WITHDRAW_FROM_LOCK,
   UPDATE_LOCK,
   addLock,
-  deleteLock,
   lockDeployed,
   updateLock,
-  UPDATE_LOCK_KEY_PRICE,
-  LOCK_DEPLOYED,
 } from '../actions/lock'
-import { PURCHASE_KEY, updateKey, addKey } from '../actions/key'
-import { setAccount, updateAccount, SET_ACCOUNT } from '../actions/accounts'
-import { setNetwork, SET_NETWORK } from '../actions/network'
+import { updateKey, addKey } from '../actions/key'
+import { updateAccount, SET_ACCOUNT } from '../actions/accounts'
 import { setError } from '../actions/error'
-import { SET_PROVIDER } from '../actions/provider'
-import { addTransaction, updateTransaction } from '../actions/transaction'
 import {
-  LOCK_PATH_NAME_REGEXP,
-  PGN_ITEMS_PER_PAGE,
-  TRANSACTION_TYPES,
-} from '../constants'
+  addTransaction,
+  updateTransaction,
+  ADD_TRANSACTION,
+} from '../actions/transaction'
+import { LOCK_PATH_NAME_REGEXP, PGN_ITEMS_PER_PAGE } from '../constants'
 
-import generateJWTToken from '../utils/signature'
 import Web3Service from '../services/web3Service'
 import {
   SET_KEYS_ON_PAGE_FOR_LOCK,
   setKeysOnPageForLock,
 } from '../actions/keysPages'
-import { signatureError } from '../actions/signature'
-import { storeLockCreation, storeLockUpdate } from '../actions/storage'
-import configure from '../config'
-
-const config = configure()
 
 // This middleware listen to redux events and invokes the web3Service API.
 // It also listen to events from web3Service and dispatches corresponding actions
 export default function web3Middleware({ getState, dispatch }) {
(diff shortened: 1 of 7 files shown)


julien51:
Something fishy is going on here. All of these tests pass locally :/ Debugging further.

unlock-app/src/middlewares/web3Middleware.js
```diff
@@ -64,14 +39,9 @@ export default function web3Middleware({ getState, dispatch }) {
...
     dispatch(lockDeployed(lock, address))
-    dispatch(
-      updateTransaction(lock.transaction, {
-        lock: address,
-      })
-    )
     web3Service.refreshAccountBalance(getState().account)
     web3Service.getLock(address)
-    web3Service.getPastLockTransactions(address)
+    web3Service.getPastLockTransactions(address) // This is costly and not useful for the paywall app...
```
cellog: just to check - if we have price changes and the like, is the lock retrieved by `lock.saved` the most current state for the lock?

unlock-app/src/pages/_app.js
```diff
@@ -13,13 +13,18 @@ import { ConfigContext } from '../utils/withConfig'
...
 const config = configure()
 
 const __NEXT_REDUX_STORE__ = '__NEXT_REDUX_STORE__'
 
 function getOrCreateStore(initialState, history) {
-  const middlewares = [web3Middleware, currencyConversionMiddleware]
+  const middlewares = [
+    web3Middleware,
+    currencyConversionMiddleware,
+    walletMiddleware,
```
cellog: can I assume that a future PR will recognize context and load only the walletMiddleware for the paywall route?

unlock-app/src/services/web3Service.js
```diff
@@ -6,43 +6,34 @@ import LockContract from '../artifacts/contracts/PublicLock.json'
...
 export const keyId = (lock, owner) => [lock, owner].join('-')
 
 /**
- * This service interacts with the web3 RPC endpoint.
- * All the methods return promises.
- * Some methods (createLock, withdrawFromLock, purchaseKey) are performing transactions. They return
- * a promise but also accept a callback which lets the user 'listen' to changes on the pending
- * transaction Web3Service is an event emitter and will trigger events which can be handled by
- * upstream objects.
+ * This service readsd data from the RPC endpoint.
```
cellog:
```suggestion
 * This service reads data from the RPC endpoint.
```

unlock-app/src/middlewares/web3Middleware.js
julien51: Yes, `getLock` will always get the most up to date version of the lock... but there may be flickering. This is actually something that I've been thinking about about earlier and forgot to open as an issue. We should probably keep track of an indication of how up to date a given object is.

unlock-app/src/pages/_app.js
julien51: Not sure what you mean here?
At this point the `_app` will load all middlewares at all points because it does not know which route the user is currently seeing/might see? 
We also need the `walletMiddleware` for both the paywall and daashboard because this is the piece that interracts with the user wallet to send transactions (create lock or purchase key for example)

cellog: In other (unrelated) PRs I am working on, I created a middleware that should only be loaded on paywall.js to do the inter-window communication. Since the web3Middleware won't be needed on the paywall, at some point, we can use this idea to not load it if the route is `'/paywall'`. That's what I was alluding to.

unlock-app/src/middlewares/web3Middleware.js
cellog: 👍

unlock-app/src/pages/_app.js
julien51: Actually we need the web3Middleware on the paywall.

julien51:
@cellog Thanks for trying this out locally! The bugs you're seeing with the "Need account" message is actually from a different commit (and now fixed https://github.com/unlock-protocol/unlock/pull/1294)

I think the storybook are also from something else?

unlock-app/src/services/web3Service.js
```diff
@@ -279,11 +152,7 @@ export default class Web3Service extends EventEmitter {
       },
       (error, events = []) => {
         events.forEach(event => {
-          const transaction = {
-            hash: event.transactionHash,
-          }
-          this.emit('transaction.new', transaction)
-          this.getTransaction(transaction.hash)
+          this.emit('transaction.new', event.transactionHash)
```
akeem: After the dust settles, it might worthwhile for to extract this method for uniform handling of events. `getPastUnlockTransactionsForUser` & ` getPastLockTransactions` are doing the same thing here

julien51: You are right there is little difference between the 2. I'll refactor in a later PR1

Issue #630, opened 2018-11-29
Better tests with expect.assertions()


julien51:
We should add a lint rule which [enforces](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/prefer-expect-assertions.md) the use of `expect.assertions(_)` on each test to make sure the right number of assertions is executed.

Additionally we should ensure that the `only.` modifier is never used (we can also consider forbidding `.skip`)

julien51:
The second part (about `.only`) is done via eslint.

The first part still needs to be done. cc @akeem with whom I had a discussion about that yesterday!

julien51:
Closing because the eslint rules have been added and the `unlock-app` is fully asserted ;)

PR #1292, opened 2019-01-30
Updating lint rules to aid in consistent development experience

akeem:
* Suggestion to use await async over then/catch for async code
* Jest for calling out disabled tests, tests with identical tests, and tests using `.only`
* Locksmith linting rules updated to be more inline with those usedin unlock-app
* Explicit expect assertion check to Locksmith

**Note**

This PR introduces ~75 linter warnings for the project, around the use of the vs async/await. The strategy is to introduce the linter warnings here, fix either in a subsequent PR, or have the team address them slowly, and adjust the setting from warning to error once resolved.  

The explicit assertion check will be added to unlock-app once the other issues have been resolved, adding at this time would add too much noise as there are roughly ~300 instance of this being raised currently

# Checklist:


akeem:
This should help to address #630 as well as a few other items previously raised

unlock-app/.eslintrc.js
```diff
@@ -47,5 +47,9 @@ module.exports = {
     'standard/computed-property-even-spacing': 0,
     'standard/object-curly-even-spacing': 0,
     'standard/array-bracket-even-spacing': 0,
+    'promise/prefer-await-to-then': 'warn',
+    'jest/no-disabled-tests': 'warn',
+    'jest/no-identical-title': 'error',
+    'jest/no-focused-tests': 'warn',
```
julien51: I'd make that an error because a warning is too easy to miss here...

akeem:
Once these changes have made their way into master i'll open the follow up cleanup tasks

PR #1299, opened 2019-01-30
Recompile contracts on `npm run dev`

hardlydifficult:
# Description

Changing the `ci` command to first build the contracts.  This should not have any real impact to the Travis CI environment, since it always works from a clean environment and when `truffle test` is run it uses the artifacts from `truffle compile` instead of compiling again.

Also updating `truffle compile --reset` to `truffle compile --all`.  The reset param is used for deploy only.

The benefit of this change is when testing locally.  Today both Nick and I hit an unexpected issue which was caused by a stale artifact.  

...alternatively we could make a `npm run dev` command

# Checklist:

### DIFF
--- a/smart-contracts/package.json
+++ b/smart-contracts/package.json
@@ -32,7 +32,7 @@
     "remix": "remixd -s contracts --remix-ide http://localhost:8080 --read-only",
     "test": "truffle test --network test",
     "lint": "solhint \"contracts/**/*.sol\"",
-    "ci": "npm run lint && npm run deploy && npm run test",
+    "ci": "npm run lint && npm run build && npm run test",
     "build": "truffle compile --all",
     "deploy": "truffle deploy",
     "zos": "zos"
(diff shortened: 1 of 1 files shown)


nfurfaro:
@hardlydifficult I like the general idea here. Personally I never run `npm run ci` to test...(but maybe I should start). I just do `npm run test`. 
The switch from `truffle compile --reset` to `truffle compilee --all` is good.

smart-contracts/package.json
```diff
@@ -32,8 +32,8 @@
     "remix": "remixd -s contracts --remix-ide http://localhost:8080 --read-only",
     "test": "truffle test --network test",
     "lint": "solhint \"contracts/**/*.sol\"",
-    "ci": "npm run lint && npm run test",
-    "build": "truffle compile --reset",
+    "ci": "npm run lint && npm run build && npm run test",
```
nfurfaro: I'm just not sure if this is the right place to add `npm run build`. I could be mistaken here though.

hardlydifficult: How about a dev command which does `lint` `build` then `test`?  That way we can still use just `test` when rerunning tests again to save time.

That may be better than CI anyways, once we have test coverage that will probably be part of the `ci` command but not something we would run locally.

nfurfaro: I think that would work, and though we do have a dev command already it's in the Unlock-app directory, so no conflict there.

Issue #1314, opened 2019-01-31
Redirect to the content after closing the paywall if it was in a new window


cellog:
If the paywall is in a new window, the final button after purchasing a key needs to redirect back to the original URL. This issue is dependent on #1313 being implemented.

This is a step on the path to implementing #1287

cellog:
this is implemented

cellog:
not fully implemented yet!

cellog:
fixed!

Issue #1287, opened 2019-01-29
Send transaction in new tab


julien51:
We have several issues with some web3 wallets/dapp browsers which all come down to the fact that we are trying to access the web3 provider from *inside* an iframe. 

The work around which was suggested was to use a new tab to actually perform the web3 transaction.
Basically, from inside the iframe, if we detect that no web3 provider is available, clicking on the lock should actually open a new tab/frame in which the user is asked to perform the key purchase.

Once the transaction has been initiated we can send the user back to the original creator page and show the transaction status (or the full content if the consumer already has a key) since we can use our "read only" web3 provider for this.

This will provide workarounds for https://github.com/unlock-protocol/unlock/issues/1154 https://github.com/unlock-protocol/unlock/issues/1241 and https://github.com/unlock-protocol/unlock/issues/1158

Comments @akeem @cellog @cnasc @benwerd ?

julien51:
This does require https://github.com/unlock-protocol/unlock/pull/1283 to be merged and https://github.com/unlock-protocol/unlock/issues/1225 to be fixed (working on both!)

cellog:
I'll work up a short document to describe the workflow before I begin work, and get the thing ironed out that way

cellog:
The 4 issues above are the steps needed to implement this, plus the new read-only provider in #1225

julien51:
closing per @cellog

PR #1343, opened 2019-02-01
add postMessage middleware

cellog:
# Description

This adds a new middleware, `postMessageMiddleware`. This middleware listens for redux actions, and then posts messages from the iframe to the parent window. In its first guise, it only listens for the `OPEN_MODAL_IN_NEW_WINDOW` action and posts a message to the parent window which will prompt it to redirect to the paywall in the main window. In subsequent PRs, this will extract the `showModal`/`hideModal` events and post those to the parent as well. This PR also enables the middleware in `_app.js`

One important feature of the middleware: if the current page is not executing in an iframe, it is disabled and just passes actions through to the next middleware in the chain. This will avoid exceptions when trying to post to a non-existent parent window.

This is the final step for implementing #1314 and a step on the path to #1287 

# Checklist:

### DIFF
--- a/unlock-app/src/middlewares/postMessageMiddleware.js
+++ b/unlock-app/src/middlewares/postMessageMiddleware.js
@@ -1,11 +1,21 @@
 import { OPEN_MODAL_IN_NEW_WINDOW } from '../actions/modal'
 
+// cribbed from https://stackoverflow.com/questions/326069/how-to-identify-if-a-webpage-is-being-loaded-inside-an-iframe-or-directly-into-t
+export function inIframe(window) {
+  try {
+    return window.self !== window.top
+  } catch (e) {
+    return true
+  }
+}
+
 // store is unused in this middleware, it is only for listening for actions
 // and converting them into postMessage
 const postMessageMiddleware = window => () => {
-  let iframe = window.parent
+  const iframe = window.parent
+  const enabled = inIframe(window)
   return next => action => {
-    if (!iframe) return next(action)
+    if (!enabled) return next(action)
     if (action.type === OPEN_MODAL_IN_NEW_WINDOW) {
       iframe.contentWindow.postMessage('redirect', iframe.contentWindow.origin)
     }
--- a/unlock-app/src/middlewares/postMessageMiddleware.js
+++ b/unlock-app/src/middlewares/postMessageMiddleware.js
@@ -1,0 +1,16 @@
+import { OPEN_MODAL_IN_NEW_WINDOW } from '../actions/modal'
+
+// store is unused in this middleware, it is only for listening for actions
+// and converting them into postMessage
+const postMessageMiddleware = window => () => {
+  let iframe = window.parent
+  return next => action => {
+    if (!iframe) return next(action)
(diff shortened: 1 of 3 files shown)


unlock-app/src/pages/_app.js
```diff
@@ -14,13 +14,15 @@ import web3Middleware from '../middlewares/web3Middleware'
...
 import walletMiddleware from '../middlewares/walletMiddleware'
+import postMessageMiddleware from '../middlewares/postMessageMiddleware'
 
 const config = configure()
 
 const __NEXT_REDUX_STORE__ = '__NEXT_REDUX_STORE__'
 
 function getOrCreateStore(initialState, history) {
   const middlewares = [
+    postMessageMiddleware(global),
```
julien51: I think we need to have a discussion about the order of middlewares because it matters, right?
Also why pass `global` here?

unlock-app/src/middlewares/postMessageMiddleware.js
```diff
@@ -0,0 +1,26 @@
+import { OPEN_MODAL_IN_NEW_WINDOW } from '../actions/modal'
+
+// cribbed from https://stackoverflow.com/questions/326069/how-to-identify-if-a-webpage-is-being-loaded-inside-an-iframe-or-directly-into-t
+export function inIframe(window) {
```
julien51: Probably worth putting in `config` (along with `isServer` and co) just in case we want to re-use it?

unlock-app/src/pages/_app.js
cellog: This allows the middleware to only use local variables, making it pure. If we access window directly, it makes it much harder to test, much easier to screw up and so on.

cellog: I forgot to answer your first question. The order matters if/only if middlewares are swallowing actions (not returning `next(action)`. Since this one may cause a url change, I figured having it first will improve performance in that case (user clicks or taps, and the page changes immediately), but it still passes the actions on transparently. This was very intentional so that the order this is inserted in the chain will not matter to the other middlewares

unlock-app/src/middlewares/postMessageMiddleware.js
cellog: I see what you mean. Perhaps "config" is a misnomer, it's more like we're setting up environment. Food for thought in a later PR. I'll slide this over there. One question is how I should access it then? import just that function for this middleware? In any case, this seems to be a separate PR since it will be doing 2 things (adding config and making a new middleware) so I'll extract that out first

cellog: #1350 implements what you're describing, I'll rebase when that's merged

unlock-app/src/pages/_app.js
julien51: Thanks!

unlock-app/src/middlewares/postMessageMiddleware.js
julien51: ;) Great! Thanks, good point about the name too!

cellog:
@julien51 rebased and ready for review

PR #1356, opened 2019-02-01
only loading the lock from the chain when on a paywall page

julien51:
only loading the lock from the chain when on a paywall page

### DIFF
--- a/unlock-app/src/constants.js
+++ b/unlock-app/src/constants.js
@@ -37,7 +37,7 @@ export const TRANSACTION_TYPES = {
 /**
  * Matches /lock /demo or /paywall
  */
-export const LOCK_PATH_NAME_REGEXP = /\/[a-z0-9]+\/(0x[a-fA-F0-9]{40})(\/(.+))?/
+export const LOCK_PATH_NAME_REGEXP = /\/([a-z0-9]+)\/(0x[a-fA-F0-9]{40})(\/(.+))?/
 
 export const PAGE_DESCRIPTION =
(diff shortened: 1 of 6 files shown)


unlock-app/src/constants.js
```diff
@@ -37,7 +37,7 @@ export const TRANSACTION_TYPES = {
 /**
  * Matches /lock /demo or /paywall
  */
-export const LOCK_PATH_NAME_REGEXP = /\/[a-z0-9]+\/(0x[a-fA-F0-9]{40})(\/(.+))?/
+export const LOCK_PATH_NAME_REGEXP = /\/([a-z0-9]+)\/(0x[a-fA-F0-9]{40})(\/(.+))?/
```
cellog: this is getting complex enough that I think we should use constants to define the matches, so something like

```suggestion
export const LOCK_PATH_NAME_REGEXP = /\/([a-z0-9]+)\/(0x[a-fA-F0-9]{40})(\/(.+))?/
export const LOCK_REGEXP_MATCH_PATH = 1
export const LOCK_REGEXP_MATCH_LOCK = 2
export const LOCK_REGEXP_MATCH_REDIRECT = 3
```

I'll need to update my PR to use the new constant, but I think this frees us up from "oops missed a spot" errors in the future

julien51: Yikes. Ok, let me work on something better then.

julien51:
Changed this by adding a util library for this! PTAL!

PR #1357, opened 2019-02-01
Now detecting pending transaction as ones which are not in a block yet

julien51:
Now detecting pending transaction as ones which are not in a block yet

### DIFF
--- a/unlock-app/src/services/web3Service.js
+++ b/unlock-app/src/services/web3Service.js
@@ -325,7 +325,7 @@ export default class Web3Service extends EventEmitter {
         blockTransaction.input
       )
 
-      if (blockTransaction.transactionIndex === null) {
+      if (blockTransaction.blockNumber === null) {
         // This means the transaction is not in a block yet (ie. not mined), but has been propagated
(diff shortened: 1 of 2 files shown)


unlock-app/src/services/web3Service.js
```diff
@@ -352,21 +352,23 @@ export default class Web3Service extends EventEmitter {
...
+            if (
+              !transactionReceipt.status ||
+              transactionReceipt.status == '0x0'
+            ) {
+              return this.emit('transaction.updated', transactionHash, {
+                status: 'failed',
+              })
+            }
+
+            return this.parseTransactionLogsFromReceipt(
```
akeem: This is a stylistic thing, but it may be worthwhile to adjust this so that we explicitly check for the positive case and have the failures be the fall through 

`transactionReceipt && transactionReceipt.status &&  transactionReceipt.status !== '0x0' `

julien51: Ho but this is a different thing right? `transactionReceipt.status && transactionReceipt.status !== '0x0' ` means the transaction actually failed...

PR #1361, opened 2019-02-02
Detect account changes

### LINKED ISSUE #1242: watch for account change to prevent 'invalid from address' error
**Describe the bug**
Switching web3 accounts after the lock has loaded causes an error in web3

**To Reproduce**
1. Visit https://www.ouvre-boite.com/members/ with an account that doesn't have a key 
2. Switch accounts in MetaMask
3. Click Purchase
4. See error in console: `MetaMask - RPC Error: Error: WalletMiddleware - Invalid "from" address.`

**Expected behavior**
Watch for changes and update the selected account

 - OS: OSX
 - Browser: Chrome 
 - Version: 71.0.3578.98 (Official Build) (64-bit)
 - MetaMask: v5.3.1

**Additional context**
FYI, I'm doing it like this atm but I'm not sure how well `publicConfigStore`    is supported...

```
    // watch changes to current account
    web3.currentProvider.publicConfigStore.on('update', data => {
      if (data.selectedAddress) {
        store.dispatch({
          type: 'ACCOUNT_CHANGE',
          payload: {
            currentAccount: data.selectedAddress
          }
        });
      } else {
        store.dispatch({
          type: 'ACCOUNT_LOGOUT'
        });
      }
    });
```

cellog:
# Description

This fixes #1242 

~~It depends on #1360 to be merged and will need a rebase (and replace of all references to `500` ms with the polling constant). Otherwise, it can be reviewed as is.~~

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/services/walletService.test.js
+++ b/unlock-app/src/__tests__/services/walletService.test.js
@@ -21,7 +21,7 @@ const nockScope = nock('http://127.0.0.1:8545', { encodedQueryParams: true })
 
 let rpcRequestId = 0
 
-let debug = false // set to true to see more logging statements
+let debug = true // set to true to see more logging statements
 
 function logNock(...args) {
   if (debug) {
@@ -61,7 +61,28 @@ describe('WalletService', () => {
   beforeEach(() => {
     nock.cleanAll()
     providers = configure().providers
-    walletService = new WalletService(providers)
+    walletService = new WalletService(providers, () => {})
+  })
+
+  describe('pollAccounts', () => {
+    it('calls pollAccounts when ready', done => {
+      expect.assertions(2)
+      const timeout = jest.fn()
+      const unlockAccountsOnNode = [
+        '0xaaadeed4c0b861cb36f4ce006a9c90ba2e43fdc2',
+      ]
+
+      accountsAndYield(unlockAccountsOnNode)
+
+      walletService = new WalletService(providers, timeout)
+
+      walletService.pollAccount = jest.fn()
+      walletService.once('ready', () => {
+        expect(walletService.ready).toBe(true)
+        expect(walletService.pollAccount).toHaveBeenCalled()
+        done()
+      })
+    })
   })
 
   describe('connect', () => {
--- a/unlock-app/src/__tests__/services/walletService.test.js
+++ b/unlock-app/src/__tests__/services/walletService.test.js
@@ -21,7 +21,7 @@ const nockScope = nock('http://127.0.0.1:8545', { encodedQueryParams: true })
 
 let rpcRequestId = 0
 
-let debug = true // set to true to see more logging statements
+let debug = false // set to true to see more logging statements
 
 function logNock(...args) {
   if (debug) {
@@ -63,28 +63,54 @@ describe('WalletService', () => {
     providers = configure().providers
     walletService = new WalletService(providers, () => {})
   })
-
-  describe('pollAccounts', () => {
-    it('calls pollAccounts when ready', done => {
+  describe('pollAccount', () => {
+    it('constructor calls timeout with pollAccount when ready', done => {
       expect.assertions(2)
       const timeout = jest.fn()
-      const unlockAccountsOnNode = [
-        '0xaaadeed4c0b861cb36f4ce006a9c90ba2e43fdc2',
-      ]
+      walletService = new WalletService(providers, timeout)
+      walletService.getAccount = jest.fn(() => Promise.resolve('account'))
 
-      accountsAndYield(unlockAccountsOnNode)
+      walletService.emit('ready')
+      expect(walletService.ready).toBe(true)
+      expect(timeout).toHaveBeenCalledWith(
+        walletService.pollForAccountChange,
+        500
+      )
+      done()
+    })
+
+    it('pollAccount calls getAccount and timeout', async () => {
+      expect.assertions(3)
+
+      const timeout = jest.fn()
 
       walletService = new WalletService(providers, timeout)
+      walletService.getAccount = jest.fn(() => Promise.resolve('thank u, next'))
+      walletService.account = 'old news'
 
-      walletService.pollAccount = jest.fn()
-      walletService.once('ready', () => {
-        expect(walletService.ready).toBe(true)
-        expect(walletService.pollAccount).toHaveBeenCalled()
-        done()
-      })
+      await walletService.pollForAccountChange()
+
+      expect(walletService.getAccount).toHaveBeenCalledTimes(1)
+      expect(walletService.getAccount).toHaveBeenCalledWith(true, 'old news')
+      expect(timeout).toHaveBeenCalledWith(
+        walletService.pollForAccountChange,
+        500
+      )
+    })
+    it('pollAccount does not run on the server', () => {
+      expect.assertions(2)
+
+      const timeout = jest.fn()
+      walletService = new WalletService(providers, timeout, true)
+      walletService.getAccount = jest.fn(() => Promise.resolve('thank u, next'))
+      walletService.account = 'old news'
+
+      walletService.pollForAccountChange(timeout, true)
+
+      expect(walletService.getAccount).not.toHaveBeenCalled()
+      expect(timeout).not.toHaveBeenCalled()
     })
   })
-
   describe('connect', () => {
     it('should get the network id', done => {
       expect.assertions(1)
--- a/unlock-app/src/__tests__/services/walletService.test.js
+++ b/unlock-app/src/__tests__/services/walletService.test.js
@@ -61,13 +61,13 @@ describe('WalletService', () => {
   beforeEach(() => {
     nock.cleanAll()
     providers = configure().providers
-    walletService = new WalletService(providers, () => {})
+    walletService = new WalletService(providers, () => {}, false)
   })
   describe('pollAccount', () => {
     it('constructor calls timeout with pollAccount when ready', done => {
       expect.assertions(2)
       const timeout = jest.fn()
-      walletService = new WalletService(providers, timeout)
+      walletService = new WalletService(providers, timeout, false)
       walletService.getAccount = jest.fn(() => Promise.resolve('account'))
 
       walletService.emit('ready')
@@ -84,14 +84,19 @@ describe('WalletService', () => {
 
       const timeout = jest.fn()
 
-      walletService = new WalletService(providers, timeout)
-      walletService.getAccount = jest.fn(() => Promise.resolve('thank u, next'))
+      walletService = new WalletService(providers, timeout, false)
+      walletService.checkForAccountChange = jest.fn(() =>
+        Promise.resolve('thank u, next')
+      )
       walletService.account = 'old news'
 
       await walletService.pollForAccountChange()
 
-      expect(walletService.getAccount).toHaveBeenCalledTimes(1)
-      expect(walletService.getAccount).toHaveBeenCalledWith(true, 'old news')
+      expect(walletService.checkForAccountChange).toHaveBeenCalledTimes(1)
+      expect(walletService.checkForAccountChange).toHaveBeenCalledWith(
+        true,
+        'old news'
+      )
       expect(timeout).toHaveBeenCalledWith(
         walletService.pollForAccountChange,
         500
--- a/unlock-app/src/__tests__/services/walletService.test.js
+++ b/unlock-app/src/__tests__/services/walletService.test.js
@@ -16,6 +16,7 @@ import {
   FAILED_TO_UPDATE_KEY_PRICE,
   FAILED_TO_WITHDRAW_FROM_LOCK,
 } from '../../errors'
+import { POLLING_INTERVAL } from '../../constants'
 
 const nockScope = nock('http://127.0.0.1:8545', { encodedQueryParams: true })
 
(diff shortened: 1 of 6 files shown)


unlock-app/src/services/walletService.js
```diff
@@ -26,15 +26,25 @@ export const keyId = (lock, owner) => [lock, owner].join('-')
  * actually retrieving the data from the chain/smart contracts
  */
 export default class WalletService extends EventEmitter {
-  constructor(availableProviders = providers) {
+  constructor(
+    availableProviders = providers,
+    timeout = setTimeout,
```
julien51: This is very very strange. I am not sure this is the right approach. Are jest time mocks working?

unlock-app/src/__tests__/services/walletService.test.js
```diff
@@ -61,9 +61,56 @@ describe('WalletService', () => {
   beforeEach(() => {
     nock.cleanAll()
     providers = configure().providers
-    walletService = new WalletService(providers)
+    walletService = new WalletService(providers, () => {})
```
julien51: I see you pass the second argument just to "mock" the setTimeout, right?
I really think this is not the right approach :/

unlock-app/src/services/walletService.js
```diff
@@ -91,12 +102,22 @@ export default class WalletService extends EventEmitter {
...
+    if (isServer) return
+    this.account = await this.getAccount(true, this.account)
+    timeout(this.pollForAccountChange, POLLING_INTERVAL)
+  }
+
   /**
    * Function which yields the address of the account on the provider or creates a key pair.
    */
-  getAccount() {
+  getAccount(polling = false, priorAccount) {
```
julien51: Hum, I think we should keep that function "clean" and put the logic of checking wether the account changed inside the `pollForAccountChange`? `getAccount` should just get the account while `pollForAccountChange` is actually the function which should know/care about whether the account changed.

unlock-app/src/services/walletService.js
cellog: no, jest time mocks do not work. at all.

cellog:
note: the current changes depend on #1370 being merged and rebased. Currently, the PR does 2 things

cellog:
ready for a 2nd pass, @julien51

unlock-app/src/middlewares/walletMiddleware.js
```diff
@@ -31,7 +31,7 @@ export default function walletMiddleware({ getState, dispatch }) {
   // Buffer of actions waiting for connection
   const actions = []
 
-  const walletService = new WalletService()
+  const walletService = new WalletService(config.providers, config.isServer)
```
julien51: Great!
I was actually thinking about your complaints that we inject the configuration everywhere and I was wondering if there was just a way to pass the *full* config object by default everywhere and then use named assignments in the constructor signatures... but that's for another PR!

unlock-app/src/services/walletService.js
```diff
@@ -26,7 +28,7 @@ export const keyId = (lock, owner) => [lock, owner].join('-')
  * actually retrieving the data from the chain/smart contracts
  */
 export default class WalletService extends EventEmitter {
-  constructor(availableProviders = providers) {
+  constructor(availableProviders, runningOnServer, wait = delayPromise) {
```
julien51: Still not a huge fan of the `wait` thing here, especially because I worry this will create weird problems when we add params to this function.
What do you think of passing a object like this:
```suggestion
  constructor({availableProviders: config.providers, runningOnServer: config.isServer, runningOnServer, wait:  delayPromise} = configure()) {
```

This lets us just call this with no object and would use the config by default, making "pure" function while still avoiding the hassle of drilling down the config all over?

unlock-app/src/services/walletService.js
```diff
@@ -91,6 +94,35 @@ export default class WalletService extends EventEmitter {
...
+      try {
+        this.account = await this.checkForAccountChange(this.account)
+      } catch (e) {
+        continue // if the account poll fails, silently ignore it
+      }
+    }
+  }
+
+  async checkForAccountChange(priorAddress) {
+    const accounts = await this.web3.eth.getAccounts()
```
julien51: I think there is some logic duplicated here with `getAccount`. Any way to DRY this up?

julien51: [that probably means splitting up `getAccount` and/or `getOrCreateAccount` which is fine.

unlock-app/src/services/walletService.js
```diff
@@ -91,6 +94,35 @@ export default class WalletService extends EventEmitter {
     }
   }
 
+  /**
+   * Poll to see if account has changed
+   */
+  async pollForAccountChange(timeout, isServer) {
+    if (isServer) return
+    while (true) {
```
julien51: Not a huge fan of `while (true)`.

Couldn't we just call `pollForAccountChange` after we have waited for the stuff we care about?

cellog: Perhaps a little clever `await this.pollAccountChange()` at the end of the function? Although this will kill the call stack eventually. Hmm... Let think on this

unlock-app/src/services/walletService.js
cellog: Sure so perhaps a new PR to include the promise into the configuration? Then we extract it like this?

cellog: Sure so perhaps a new PR to include the promise into the configuration? Then we extract it like this?

unlock-app/src/middlewares/walletMiddleware.js
cellog: Yes this would be a great solution to that particular issue. Would you like that PR to be a follow-up or a precursor to this one?

cellog: I'll do it as a precursor because why not :)

unlock-app/src/services/walletService.js
cellog: Something feels odd about including a promise-that-never-changes in configuration. Perhaps a 2nd parameter for utils so we get:

```js
constructor({availableProviders: config.providers, runningOnServer: config.isServer, runningOnServer}, { delayPromise }}) {
```

Also, I removed the default value of `configure()` from my example, primarily because we can easily do that inside `walletMiddleware`, which is currently the only place `walletService` is initialized in any case. What do you think?

unlock-app/src/services/walletService.js
cellog: sure I'll see what I can wrangle up

unlock-app/src/__tests__/services/walletService.test.js
```diff
@@ -55,15 +57,75 @@ nock.emitter.on('no match', function(clientRequestObject, options, body) {
 })
 
 let walletService
+let config
 let providers
 
 describe('WalletService', () => {
+  const WalletService = require('../../services/walletService').default
```
julien51: I think this really belongs in the top of the file?

unlock-app/src/services/walletService.js
```diff
@@ -73,8 +75,10 @@ export default class WalletService extends EventEmitter {
     this.web3 = new Web3(provider)
 
     const networkId = await this.web3.eth.net.getId()
-    if (unlockAddress) {
-      this.unlockContractAddress = Web3Utils.toChecksumAddress(unlockAddress)
+    if (this.unlockAddress) {
```
julien51: I think it is strange that we have both `this.unlockAddress` and `this.unlockContractAddress`? I think we only use the latter one. Could we just maybe move this part to the constructor and not use `this.unlockAddress` anymore?

unlock-app/src/__tests__/services/walletService.test.js
cellog: The loading of promises.js has to happen after the `jest.mock('../../utils/promises')` so we can't use import, but have to use require. It's awkward but that's the price of module mocks :)

unlock-app/src/services/walletService.js
```diff
@@ -25,15 +25,17 @@ export const keyId = (lock, owner) => [lock, owner].join('-')
  * actually retrieving the data from the chain/smart contracts
  */
 export default class WalletService extends EventEmitter {
-  constructor(availableProviders = providers) {
+  constructor({ providers, runningOnServer, unlockAddress } = configure()) {
```
julien51: 👍

unlock-app/src/services/walletService.js
```diff
@@ -25,15 +25,17 @@ export const keyId = (lock, owner) => [lock, owner].join('-')
...
     super()
-    this.providers = availableProviders
+    this.unlockAddress = unlockAddress
```
julien51: Move the `Web3Utils.toChecksumAddress` part from below up here so we do not have both `this.unlockAddress` and `this.unlockContractAddress`.

unlock-app/src/services/walletService.js
```diff
@@ -91,23 +95,58 @@ export default class WalletService extends EventEmitter {
...
-      if (accounts.length === 0) {
-        return this._createAccount()
-      } else {
-        return Promise.resolve(accounts[0]) // defaults to the first account for now
-      }
-    })
+  async pollForAccountChange(isServer) {
+    if (isServer) return
+    try {
+      await delayPromise(POLLING_INTERVAL)
```
julien51: Is there any reason why the `delayPromise` would throw? IIRC it does not based on what you sent earlier?

unlock-app/src/services/walletService.js
```diff
@@ -91,23 +95,58 @@ export default class WalletService extends EventEmitter {
...
+    } catch (e) {
+      return // if the timeout throws, we don't start the loop, this is primarily for testing
+    }
+    try {
+      this.account = await this.checkForAccountChange(this.account)
```
julien51: Same here, is there any reason it would throw?

unlock-app/src/services/walletService.js
```diff
@@ -91,23 +95,58 @@ export default class WalletService extends EventEmitter {
...
+    } catch (e) {
+      // if the account poll fails, silently ignore it
+    }
+    // because this is async, the call stack is not affected, and does not grow
+    // for a non-async function this would quickly fill all available memory
```
julien51: Great comment!

unlock-app/src/services/walletService.js
```diff
@@ -91,23 +95,58 @@ export default class WalletService extends EventEmitter {
...
+    this.pollForAccountChange()
+  }
+
+  /**
+   * given the prior account address, determine if the account has changed and return the
+   * current address regardless of any change
+   */
+  async checkForAccountChange(priorAddress) {
+    const nextAddress = await this._getAccountAddress(Promise.resolve(null))
```
julien51: What is this promise passed as argument?

julien51: Maybe name the function in a way that makes it more obvious what the argument is? `_getAccountAddressOrFallback`  for example ;)

unlock-app/src/services/walletService.js
```diff
@@ -91,23 +95,58 @@ export default class WalletService extends EventEmitter {
...
-    })
+    const address = await this._getAccountAddress(this._createAccount)
+    this.emit('account.changed', address)
+    this.emit('ready')
+    return Promise.resolve(address)
+  }
+
+  /**
+   * Get the current user account address, calling a fallback if there is none
+   * @param fallback async function
```
julien51: Make it clear that the fallback should also return/yield an address.

unlock-app/src/utils/__mocks__/promises.js
```diff
@@ -0,0 +1,17 @@
+let x = 0
```
julien51: What's the `x`? Maybe use a better variable name?

unlock-app/src/utils/__mocks__/promises.js
```diff
@@ -0,0 +1,17 @@
...
+
+const delayPromise = jest.fn(ms => {
+  if (x--) {
```
julien51: This is not pure ;p
Not sure what `delayPromise` does actually?

cellog: first of all, my main takeaway from your feedback is that the PR is not clear enough from a devel perspective, and this needs fixin!

To answer this specific point, the main issue that makes this PR so dang hard to do well is that the account is polled every half second.

The simplest solution is to use a `setInterval` but it turns out that this can wreak havoc on mobile browsers, according to the coinbase doc that Akeem and I used as a cheat sheet on our beginning exploration of this problem.

The next-most-simplest solution (setTimeout and infinitely looping) caused the tests to flail about like a fish and slowly croak when I tried to mock it with `jest.useFakeTimers()`. So, I moved to the delay promise, because it allows us to mock it out without using fake timers. A few commits ago, when it was passed as a parameter to the `walletService`, I could then directly pass in what this mock does, which is provide a delay promise that neither delays, nor loops.

In other words, I need a promise that will only successfully resolve once, and then reject. The rejection is the thing that kicks the account polling out of its infinite looping.

This mock module allows us also to specify the number of times that our delay will be called, in case we want to use it in the future and need more calls to succeed. Note that by default, the delay promise never resolves, and only rejects. This allows us to not generate an infinite loop in the other `walletService` tests, which prevents all kinds of weird side effects in those tests, mostly failing because the test doesn't mock a call to `web3.eth.getAccounts()`.

If this sounds incredibly complex, it's because it is. It took me 10 minutes to write the first working version of this, and has taken about 5 hours of work and 4+ days to get to this stage :). I'm not complaining, it's been a fun challenge, but it's a doozy, and I'm not surprised my code is not easy to read as a result.

Maybe we can chat briefly and look over the PR together for 10 minutes tomorrow, finalize how to solve/name/comment things and then I can implement any changes?

unlock-app/src/__tests__/services/walletService.test.js
julien51: What? let's debug this tomorrow because I have been using mocks all over the place and never had to do that ;)
jest will actually hoist the mock for modules at the beginning of files no matter what.

cellog:
OK, @julien51 PTAL, I think this is going to be much better

unlock-app/src/services/walletService.js
cellog: yes, if web3 throws.

cellog: (which it does inside an iframe in coinbase wallet, getAccount() freaks out)

unlock-app/src/__tests__/services/walletService.test.js
```diff
@@ -55,15 +58,64 @@ nock.emitter.on('no match', function(clientRequestObject, options, body) {
 })
 
 let walletService
+let config
 let providers
 
 describe('WalletService', () => {
+  let delay
   beforeEach(() => {
+    delay = require('../../utils/promises').delayPromise
```
julien51: I still don't like with that ;)

unlock-app/src/services/walletService.js
julien51: Is there a way to only catch that error? I m not a huge fan of blanket `catch (e)` which catches and swallows everything.

unlock-app/src/services/walletService.js
```diff
@@ -91,23 +98,64 @@ export default class WalletService extends EventEmitter {
...
+      this.emit('account.changed', nextAccount)
+    }
+    return nextAccount
+  }
+
+  /**
+   * Function which yields the address of the account on the provider or creates a key pair.
+   */
+  async getAccount() {
     // Once we have the account, let's refresh it!
```
julien51: Remove this?

unlock-app/src/__tests__/services/walletService.test.js
cellog: The mock for delayPromise is a `jest.fn()` so we need a new one for every test. That's all this does, is give us something to assert on. Is that kosher? :)

unlock-app/src/services/walletService.js
cellog: remove what?

unlock-app/src/__tests__/services/walletService.test.js
cellog: This allows the tests to assert on the mock surrounding the promise function. Because we don't pass it as a parameter to the service, this is the only way to get a new one for every test.

unlock-app/src/services/walletService.js
cellog: I'm pretty sure it's a generic Error, so I don't think so. Also, there may be new errors introduced later. The solution is a TODO that says "log error here" for when we implement Akeem's idea of logging

unlock-app/src/services/walletService.js
cellog: sorry, remove what?

unlock-app/src/__tests__/services/walletService.test.js
```diff
@@ -16,6 +15,11 @@ import {
   FAILED_TO_UPDATE_KEY_PRICE,
   FAILED_TO_WITHDRAW_FROM_LOCK,
 } from '../../errors'
+import { POLLING_INTERVAL } from '../../constants'
+
+jest.mock('../../utils/promises')
+const WalletService = require('../../services/walletService').default
```
julien51: Huuuhuhuh I wish you'd use a regular import here. Please fix this or add a comment about why you need to use `require` here? This is going to be extremely confusing in 2 weeks.

unlock-app/src/__tests__/services/walletService.test.js
```diff
@@ -16,6 +15,11 @@ import {
...
+const delay = require('../../utils/promises').delayPromise
```
julien51: Same here?

PR #1364, opened 2019-02-02
do not retrieve key without a valid account address.

### LINKED ISSUE #1362: web3Middleware should not try to retrieve when no the redux store does not have an account
**Describe the bug**
When loading the paywall with no account, it fails with an exception in `web3Middleware`

The lines in question are:

```js
      if (action.type === ADD_LOCK || action.type == UPDATE_LOCK) {
        const lock = getState().locks[action.address]
        web3Service.getKeyByLockForOwner(
          lock.address,
          getState().account.address
        )
```

When pausing on the `getKeyByLockForOwner` when logged in, the lock has an address. When pausing while logged out, the lock has no address, and does not exist in state, so `lock` is undefined. In addition, a casual look shows that we are assuming that the `account` in state has an address, where it is in fact `null` when there is no account.

**To Reproduce**
1. run `UNLOCK_ENV=staging npm run dev`
2. go to `http://localhost:3000/paywall/0x42dbdc4CdBda8dc99c82D66d97B264386E41c0E9`
3. See error

**Expected behavior**
lock displays under a "missing wallet" error

**Screenshots**
screen shot 2019-02-01 at 9 28 23 pm

 - OS: Mac
 - Browser chrome with Metamask logged out

julien51:
fixes #1362

### DIFF
--- a/unlock-app/src/middlewares/web3Middleware.js
+++ b/unlock-app/src/middlewares/web3Middleware.js
@@ -123,10 +123,12 @@ export default function web3Middleware({ getState, dispatch }) {
 
       if (action.type === ADD_LOCK || action.type == UPDATE_LOCK) {
         const lock = getState().locks[action.address]
-        web3Service.getKeyByLockForOwner(
-          lock.address,
-          getState().account.address
-        )
+        if (getState().account) {
+          web3Service.getKeyByLockForOwner(
+            lock.address,
+            getState().account.address
+          )
+        }
       } else if (
         action.type === LOCATION_CHANGE &&
         action.payload.location &&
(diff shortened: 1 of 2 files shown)


unlock-app/src/middlewares/web3Middleware.js
```diff
@@ -123,10 +123,12 @@ export default function web3Middleware({ getState, dispatch }) {
 
       if (action.type === ADD_LOCK || action.type == UPDATE_LOCK) {
         const lock = getState().locks[action.address]
-        web3Service.getKeyByLockForOwner(
-          lock.address,
-          getState().account.address
-        )
+        if (getState().account) {
```
cellog: can probably move this up 1 line, since lock is unused otherwise

PR #1367, opened 2019-02-02
pass the error code to displayError

cellog:
# Description

This adds the error code as a third parameter to `displayError` for the `GlobalErrorConsumer`. We can use this information to not display an error in the paywall if the error is user account missing.

This is a step on the path to implementing #1287 

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/GlobalErrorConsumer.js
+++ b/unlock-app/src/components/interface/GlobalErrorConsumer.js
@@ -30,7 +30,11 @@ export default function GlobalErrorConsumer({
           : false
 
         // call displayError with either false or the error element, and our child elements
-        return displayError(Error && <Error {...errorMetadata} />, children)
+        return displayError(
+          Error && <Error {...errorMetadata} />,
+          children,
+          error
+        )
       }}
     </Consumer>
   )
--- a/unlock-app/src/__tests__/components/interface/GlobalErrorConsumer.test.js
+++ b/unlock-app/src/__tests__/components/interface/GlobalErrorConsumer.test.js
@@ -15,7 +15,7 @@ const Provider = GlobalErrorContext.Provider
 
 describe('GlobalErrorConsumer', () => {
   it('passes error initialized with metadata to displayError prop', () => {
-    expect.assertions(4)
+    expect.assertions(5)
 
     const listen = jest.fn(() => <div>internal</div>)
     const wrapper = rtl.render(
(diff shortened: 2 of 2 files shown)


unlock-app/src/components/interface/GlobalErrorConsumer.js
```diff
@@ -30,7 +30,11 @@ export default function GlobalErrorConsumer({
           : false
 
         // call displayError with either false or the error element, and our child elements
-        return displayError(Error && <Error {...errorMetadata} />, children)
+        return displayError(
+          Error && <Error {...errorMetadata} />,
+          children,
+          error
```
julien51: Since w're passing error down, wouldn't it then be cleaner to move the logic above 
```
const Error = error
          ? overrideMapping[error] || mapping[error] || mapping['*']
          : false`
```

down inside `displayError` and only then pass `errorMetadata`, `children` and `error`?

cellog: I am not sure I follow what you're saying. Can you clarify "down inside `displayError`" and where you mean by "only then pass"? Thanks

cellog: I understand. `displayError` is a callback that can be modified in any context that it is called, so if we do what you're proposing, it will result in duplicated code, in `Overlay.js` and here in `GlobalErrorConsumer.js`.

The `displayError` here in `GlobalErrorConsumer.js` is only a default value, it is assumed to be modifiable in different contexts. If we were to make this change, we would also need to pass down `mapping`, so that `displayError`'s API would be `displayError(mapping, error, errorMetadata, children)`. The goal with `displayError` was to allow components to customize the placement of the error and its children. By adding the error code, it also makes it possible to customize whether an error is shown at all.

I think, ultimately, what you are seeing is that the API isn't quite the right one. If we can accept this PR for now, and open an issue to refactor `displayError` to be more flexible, this would be a good idea, I think. For instance, I think we will want the ability to display a custom error component in different contexts. Our paywall error message for "missing wallet" may want to direct consumers to purchase our handy-dandy Unlock wallet with their credit card, whereas on the dashboard we would just say "you need a wallet" as we do now. The current design of `displayError` does not allow this, and we would indeed have to pass down both the error and the errorMetadata to make that possible.

I don't have a good answer, we can talk about this Tuesday?

cellog: #1376 is the first attempt to answer this properly

cellog:
closing in favor of #1377

PR #1390, opened 2019-02-04
moving lock address generation to web3Service

julien51:
Since this does not require the user's wallet information this should go thru the web3Service instead
of the walletService.
This also adds the added benefit of having both web3Middleware and walletMiddleware not breaking the middleware chain anymore.
They both act on the CREATE_LOCK action, but only if it respectively has no address or an address for the lock.

### DIFF
--- a/unlock-app/src/__tests__/middlewares/web3Middleware.test.js
+++ b/unlock-app/src/__tests__/middlewares/web3Middleware.test.js
@@ -1,7 +1,7 @@
 import EventEmitter from 'events'
 import { LOCATION_CHANGE } from 'react-router-redux'
 import web3Middleware from '../../middlewares/web3Middleware'
-import { ADD_LOCK, UPDATE_LOCK } from '../../actions/lock'
+import { ADD_LOCK, UPDATE_LOCK, CREATE_LOCK } from '../../actions/lock'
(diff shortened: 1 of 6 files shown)


unlock-app/src/__tests__/middlewares/web3Middleware.test.js
```diff
@@ -378,6 +378,41 @@ describe('Lock middleware', () => {
...
+        name: 'my lock',
+      }
+      const address = '0x123'
+      const action = { type: CREATE_LOCK, lock }
+      mockWeb3Service.generateLockAddress = jest.fn(() => {
+        return Promise.resolve(address)
+      })
+
+      invoke(action)
+      expect(mockWeb3Service.generateLockAddress).toHaveBeenCalledWith()
```
akeem: How do we feel about updating the assertion to `toHaveBeenCalled` as we aren't validating the passed value?

akeem: How do we feel about updating the assertion to `toHaveBeenCalled` as we aren't validating the passed value?

julien51: Oh actually, i think it tests that's is called with nothing? (Which it should?)

PR #1392, opened 2019-02-04
Never blocking the middlware chain even when the user is not connected

julien51:
I believe blocking the middlware chain is an anti-pattern.
This PR now removes the blocking from `walletMiddleware` but will trigger errors when any of
the action is invoked while no user is available and/or the `walletService` is not ready.

I believe this will make is easy to display things (like the paywall!) even to users who are
not using a wallet (metamask and co).

### DIFF
--- a/unlock-app/src/__tests__/middlewares/walletMiddleware.test.js
+++ b/unlock-app/src/__tests__/middlewares/walletMiddleware.test.js
@@ -14,6 +14,7 @@ import { SET_PROVIDER } from '../../actions/provider'
 import { ADD_TRANSACTION } from '../../actions/transaction'
 import { SET_ERROR } from '../../actions/error'
 import { TRANSACTION_TYPES } from '../../constants'
+import { NO_USER_ACCOUNT } from '../../errors'
 
 /**
  * Fake state
@@ -69,6 +70,7 @@ class MockWalletService extends EventEmitter {
     super()
     this.ready = true
   }
+  connect() {}
 }
 
 let mockWalletService = new MockWalletService()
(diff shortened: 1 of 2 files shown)


cellog:
Akeem is right, I think our decision to put fatal errors in `errors.js` has come back to bite us. Fatal errors are not intended to be used in the store, they are detected and passed down by the `GlobalErrorProvider` and are a separate error process from the redux errors, which are intended for display on the dashboard page(s) only. So there is currently no error handler for any of the fatal errors because they pre-empt the UI for that. We may need to quickly chat about this tomorrow morning, I don't think it's a huge issue for this PR, but I would not recycle a fatal error for this, we need a new error for the normal app cycle, or we could (for example) disable the UI on the dashboard or paywall when the wallet is not ready yet so the error is no longer possible. In other words a loading spinner of sorts :)

cellog:
think of "user account is initialized" as "user has a crypto wallet and web3 reports an account in `getAccount()`" as in this state either exists or it doesn't, whereas there is a slight delay between the start of the UI being ready and the point when `walletService.ready` is true. It is that small window of time where the user can (for example) try to click on "withdraw funds" on the dashboard, or "purchase" on the paywall. During that time frame, this PR simply allows the user to click the button, and triggers an error. I'm suggesting that instead of this, it would be more accurate to not allow the user to click these buttons, or to not display them at all and display a loading state.

When I refer to "invalid UI" what I mean is that we present something that the user can interact with, but the program is not able to handle the interaction. This encompasses these possibilities:

1. the user clicks on the button, and nothing happens because the `walletService` stores the action for when it is ready
2. the user clicks on the button, and it displays an error informing them that the wallet service isn't ready (what this PR is designed to do)

I'm suggesting that a "valid UI" is one where the user cannot generate an error condition from reasonable behavior, and that the UI gracefully informs the user why it is not yet ready for their input.

julien51:
Ok, @cellog rebased and pushed, PTAL ;)

julien51:
Well, I initially started with the giant wrapper, but than that broke `walletService.connect(action.provider)` which obviously cannot require things to be connected before ;)

PR #1400, opened 2019-02-05
extract dependency on global config variable from services, extract unused config from LockIconBar

cellog:
# Description

This PR extracts the global variable configuration from the `web3Service` and `walletService` in order to make them closer to pure, and make testing less brittle as a result. It also removes dead code in `LockIconBar` (it already uses withConfig, the call to `configure` was unnecessary as we can just use the currently unused `config` prop. Linting missed this because of the name collision)

# Checklist:

### DIFF
--- a/unlock-app/src/services/walletService.js
+++ b/unlock-app/src/services/walletService.js
@@ -3,7 +3,6 @@ import Web3 from 'web3'
 import Web3Utils from 'web3-utils'
 import LockContract from '../artifacts/contracts/PublicLock.json'
 import UnlockContract from '../artifacts/contracts/Unlock.json'
-import configure from '../config'
 import {
   MISSING_PROVIDER,
   NON_DEPLOYED_CONTRACT,
@@ -14,8 +13,6 @@ import {
   FAILED_TO_WITHDRAW_FROM_LOCK,
 } from '../errors'
 
-const { providers, unlockAddress } = configure()
-
 export const keyId = (lock, owner) => [lock, owner].join('-')
 
 /**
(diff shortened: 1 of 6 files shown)


cellog:
FYI:

screen shot 2019-02-05 at 10 38 08 am

Why would this fail remotely and pass locally?

julien51:
This is because you have local artifacts files which are not deployed on travis.
The error is 
```
 Unhandled "error" event. (Error: NON_DEPLOYED_CONTRACT)
```

Which means that somewhere in your code you are calling the `connect` method?

julien51:
We will soon use the npm package which should make that behavior obsolete!

unlock-app/src/services/walletService.js
```diff
@@ -25,9 +22,10 @@ export const keyId = (lock, owner) => [lock, owner].join('-')
  * actually retrieving the data from the chain/smart contracts
  */
 export default class WalletService extends EventEmitter {
-  constructor(availableProviders = providers) {
+  constructor({ providers, unlockAddress }) {
     super()
-    this.providers = availableProviders
+    this.providers = providers
+    this.unlockAddress = unlockAddress
```
cnasc: Fun destructuring tip (not necessary to change, but just for the trivia):
```
Object.assign(this, { providers, unlockAddress })
```
or
```
// If you haven't already destructured the props (personally I think this one is kind of ugly)
({ providers: this.providers, unlockAddress: this.unlockAddress } = props)
```

Just in case you ever need to set a whole bunch of properties and want to save yourself some typing.

PR #1414, opened 2019-02-05
Add `versions` directory and doc

nfurfaro:
# Description

This PR adds a new directory (`versions`)to the root of the project, as well as a `versions.md` doc. The purpose is to provide a location for saving the artifacts for old `Unlock.sol` versions, as well as a reference (in the doc) to the commits for each old version of both `Unlock.sol` and `PublicLock.sol`.
Note that when we upgrade Unlock next time, we'll need to overwrite the `Unlock_V0.json` file with the final state of `Unlock.json` prior to the upgrade to V1, as well as add the commit hash for Unlock to the doc.

# Checklist:


versions/versions.md
```diff
@@ -0,0 +1,18 @@
...
+
+Use the command `$ git checkout <commit_hash_here>`to view the unlock repository at this point in time.
+
+After the upgrade to `Unlock.sol` V1, we'll need to modify our test suite to use the artifacts from this directory when testing Unlock V0, so: `const Unlock = artifacts.require('../../versions/Unlock_V0.json')`
+
+Note: It's possible that the artifacts.require syntax won't work for this use case. If this is the case, we'll need to look at using truffle-contract instead: https://github.com/trufflesuite/truffle/tree/master/packages/truffle-contract
+
+## Unlock.sol
+
+- `Unlock.sol` (V0) commit: <add commit# prior to upgrading to UnlockV1.sol>
```
hardlydifficult: Would it make sense to list the mainnet and testnet addresses here as well?

versions/versions.md
```diff
@@ -0,0 +1,18 @@
...
+
+## PublicLock.sol
+
+- `PublicLock.sol` (V0) commit: 9ee8a2dcd037cbab747bfb0b5cf28e2b7b38360e
```
hardlydifficult: This commit seems to be from after the v0 tag.  Shouldn't we use exactly what was there when deployed?

versions/versions.md
nfurfaro: Hmm. We already have the address for each network in our zos network files (`zos.rinkeby.json`, etc) for the current version. This would be the logical place to record that info for old versions if you have a use case in mind where we would need it? 
Our test suite deploys a new version each run, and as for rinkeby/mainnet, having the old version address would mean we could interact with it directly, but not through the proxy (without setting the implementation to the old version). 
Perhaps I'll add the address for V0 now just in case...we can always remove if not needed.

versions/versions.md
nfurfaro: You are right. Fixing this now.

versions/versions.md
nfurfaro: Done.

versions/versions.md
nfurfaro: Done.

versions/versions.md
hardlydifficult: This is not important, I was just thinking that it may be convenient.  For example, knowing the mainnet address of the Unlock v0 implementation (i.e. the code not just the proxy address) would allow us to easily confirm the bytecode is correct (e.g. check etherscan).

PR #1419, opened 2019-02-06
Cancel key and get a partial refund

### LINKED ISSUE #637: Allow for cancellation

hardlydifficult:
# Description

Allows a key to be canceled for a partial refund of the purchase price.  The refund is based on how much time has passed (effectively the % remaining on the Key) as well as a configurable penalty, defaulting to 10% of the original purchase price.  

Fixes #637

# Checklist:


smart-contracts/contracts/PublicLock.sol
```diff
@@ -159,7 +168,8 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
     uint _expirationDuration,
     uint _keyPrice,
     uint _maxNumberOfKeys,
-    uint _version
+    uint _version,
+    uint _cancelRefundPenaltyDenominator
```
julien51: Will this kill compatibility? 
I think it would be ok to only have a small number of "customizable" constructor arguments and then have setters/getters for the things which are less likely to be changed on the constructor, such as `cancelRefundPenaltyPercentage`.

smart-contracts/contracts/Unlock.sol
```diff
@@ -86,7 +86,8 @@ contract Unlock is IUnlock, Initializable, Ownable {
   function createLock(
     uint _expirationDuration,
     uint _keyPrice,
-    uint _maxNumberOfKeys
+    uint _maxNumberOfKeys,
+    uint _cancelRefundPenaltyDenominator
```
julien51: Same comment as above? I think it is fine to set the default `cancelRefundPenaltyDenominator` and then have a setter for it so that we do not need to change the signature of `createLock` everywhere?

smart-contracts/contracts/PublicLock.sol
hardlydifficult: Re compatibility: it requires JS updates but otherwise no, this won't break our upgrade process.  (this was the PR I was working on when I thought to bring up the JS wrapper, it makes changes like this and adding a default value easier).

Adding a setter for something like this may be concerning to the Key owners, unless we add some business logic on what the Lock owner may do here.  This is a tough one for me - it's always a trade off - how much control/flexibility do you offer the Lock owner vs how much guarantees/expectations can we keep for users?  A Lock owner changing the penalty after any Key has been purchased, is kind of changing the terms of that previous purchase without any warning.

hardlydifficult: One option for 'business logic' that would make a setter okay is to say the owner can only LOWER the penalty and once it has been lowered, it may never go up again.  We start with a default of say 20% and then they can choose to lower it to 10%, 5%, or 0.  Thoughts?

hardlydifficult: Oh another option may be to allow the Lock owner to make changes up until the first Key is purchased.  So basically `advanced` type options would all require follow up transactions to the initial lock creation.

julien51: I think this idea of only allow lowering is interesting...

Generally though, as much as I understand the worry about making it "unfair" for the key owner, we should remember that in practice the lock owner may just decide to stop honoring the lock altogether inside their application :/ Making all keys invalid by definition.

So, in the end, I am not fully sure that limiting the ability to change the cancellation terms is a very useful thing which may backfire in something much worse.

smart-contracts/contracts/PublicLock.sol
```diff
@@ -599,6 +649,68 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
...
+    );
+    emit CancelKey(_owner, refund);
+    // expirationTimestamp is a proxy for hasKey, setting this to `now` instead
+    // of 0 so that we can still differentiate hasKey from hasValidKey.
+    key.expirationTimestamp = now;
+
+    // 3) Issue Refund
+    if (refund > 0) {
+      // Security: doing this last to avoid re-entrancy concerns
+      _owner.transfer(refund);
```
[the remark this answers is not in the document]
julien51: That's another point of "trust" where making the `cancelRefundPenaltyPercentage` less flexible can be bypassed by the lock owner who would never keep a balance on their lock? Preventing anyone from canceling in practice?

smart-contracts/contracts/PublicLock.sol
```diff
@@ -599,6 +649,68 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
...
+    refund = keyPrice.mul(timeRemaining) / expirationDuration;
+    if (cancelRefundPenaltyDenominator > 0) {
+      uint penalty = keyPrice / cancelRefundPenaltyDenominator;
+      if (refund > penalty) {
+        // Math: safeSub is not required since the if confirms this won't underflow
+        refund -= penalty;
+      } else {
+        refund = 0;
+      }
+    }
```
julien51: My solidity knowledge is limited but I guess that the return is implicit based on `returns (uint refund)`?

smart-contracts/contracts/PublicLock.sol
```diff
@@ -205,6 +216,29 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
...
+  {
+    _cancelAndRefund(msg.sender);
+  }
+  
+  /**
+   * @dev Destroys the user's key and sends a refund based on the amount of time remaining.
+   * The caller must first be approved to modify the owner's key.
+   * @param _owner The owner of the key to be destroyed.
+   */
+  function cancelAndRefundFor(
```
julien51: Is that a realistic use case?
I mean if the user has to approve someone else first, why can't they cancel at the same time?

smart-contracts/contracts/PublicLock.sol
nfurfaro: > in practice the lock owner may just decide to stop honoring the lock altogether inside their application

I think this is the clincher. Despite our best efforts at programming fairness here, we end up just adding complexity without preventing this sort of action by the lock owner.

smart-contracts/contracts/PublicLock.sol
nfurfaro: @julien51 your guess is correct, `refund` is returned implicitly.

smart-contracts/contracts/interfaces/IUnlock.sol
```diff
@@ -27,7 +27,8 @@ interface IUnlock {
   function createLock(
     uint _expirationDuration,
     uint _keyPrice,
-    uint _maxNumberOfKeys
+    uint _maxNumberOfKeys,
+    uint _cancelRefundPenaltyDenominator
```
nfurfaro: this could be removed _IF_ we decide to just set a default value in the lock constructor and provide a setter.

smart-contracts/contracts/PublicLock.sol
nfurfaro: I'm in favour of removing this for now for the sake of simplicity', and only adding it if it becomes necessary down the road.

smart-contracts/contracts/PublicLock.sol
```diff
@@ -599,6 +649,68 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
...
+  )
+    internal
+  {
+    Key storage key = keyByOwner[_owner];
+
+    // 1) Calc refund
+    uint refund = _getCancelAndRefundValue(_owner);
+
+    // 2) Destroy token
+    emit Transfer(
```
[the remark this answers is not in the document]
nfurfaro: Hmm. I'm not seeing the value in leaving the owner holding an invalid key...unless this leads into a use-case I'm not seeing?
 I would be in favour of making the assumption here that someone cancelling their key doesn't want it anymore, so lets clean house a bit. If there's an exception to this, they can always buy another (but I do think that would be the exception).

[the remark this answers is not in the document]
nfurfaro: _If_ cancelling a key means"destroying" it and we transfer to `address(0)`, we should probably modify `ownerByTokenId` for the token as well...?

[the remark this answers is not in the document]
nfurfaro: Also, (and I may be missing something here) while this emits the `Transfer` event, it doesn't actually transfer the token. Our `transferFrom` function won't allow this because of `require(_recipient != address(0), "Can't Transfer to 0x Address");`
Should we consider adding a `burn` function to achieve this?

smart-contracts/contracts/PublicLock.sol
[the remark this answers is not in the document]
nfurfaro: > what happens if the lock's balance is too low?

Hmm. need to think about this one.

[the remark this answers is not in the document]
hardlydifficult: If the balance is too low this tx fails and the Key remains valid.
If the refund is <= 0, no refund and the Key becomes invalid.

smart-contracts/contracts/PublicLock.sol
hardlydifficult: I can remove it.
The thinking was if someone is already approved to trade away my Key, they could choose to cancel instead which is effectively another way of trading the Key for ETH.

smart-contracts/contracts/PublicLock.sol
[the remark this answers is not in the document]
hardlydifficult: The Transfer event to address(0) is the ERC721 standard way of destroying a Key.  I'm interpreting making a Key invalid to be the same as a destroy.  We could remove this and just lean on the cancel event.

Interesting point on cleanup.  If we do clear the data and ownerByTokeId, we should do the same for transferFrom.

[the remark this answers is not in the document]
hardlydifficult: From ERC721:

```
    /// @dev This emits when ownership of any NFT changes by any mechanism.
    ///  This event emits when NFTs are created (`from` == 0) and destroyed
    ///  (`to` == 0). Exception: during contract creation, any number of NFTs
    ///  may be created and assigned without emitting Transfer. At the time of
    ///  any transfer, the approved address for that NFT (if any) is reset to none.
    event Transfer(address indexed _from, address indexed _to, uint256 indexed _tokenId);
```
"Also note that a NFTs MAY become invalid (be destroyed)."
"Creating of NFTs ("minting") and destruction NFTs ("burning") is not included in the specification. Your contract may implement these by other means. Please see the event documentation for your responsibilities when creating or destroying NFTs."

smart-contracts/contracts/PublicLock.sol
hardlydifficult: I'll go with a default of 20% and allow only lowering for now.  Let me know if you'd rather something else

julien51: Could we just allow the lock owner to change either up or down?

smart-contracts/contracts/PublicLock.sol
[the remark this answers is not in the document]
julien51: So I am thinking we should have both a "cancel" (basically get a refund) and a "destroy" distinctly.

 I know I sound like a broken record but it will be very important down the line to know that someone previsouly owned a key (even though that this weak on the sybil attack side)

smart-contracts/contracts/PublicLock.sol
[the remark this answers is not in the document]
julien51: Works for me!

smart-contracts/contracts/PublicLock.sol
julien51: Ah, indeed, but again, if I am willing to get a refund I can just cancel myself rather than authorize someone else, right?

smart-contracts/contracts/interfaces/IUnlock.sol
julien51: Yes!

smart-contracts/contracts/PublicLock.sol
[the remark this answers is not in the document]
nfurfaro: I understand that the `transfer` event must emit "when ownership of any NFT changes by any mechanism". I was under the impression that we still had to set the new owner of the token to `address(0)` in the lock contract for this event to reflect the actual state? That is, I don't think that emitting the event is enough on its own. I will take another look, as you may have already done this and I missed it.
I'm in favour of cleaning up the lock contract where possible, but I understand that we don't want to remove any addresses from the `owners` array.

smart-contracts/contracts/PublicLock.sol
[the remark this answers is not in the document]
nfurfaro: Sounds good.
( I was getting sidetracked by the "should we allow the balance get too low?" question.)

hardlydifficult:
This build broke when addressing feedback.  I'm not sure why yet..

smart-contracts/contracts/PublicLock.sol
hardlydifficult: Done.  Default 10% with owner option to change.

smart-contracts/contracts/PublicLock.sol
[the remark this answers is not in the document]
hardlydifficult: Done. Removed emit Transfer

smart-contracts/contracts/interfaces/IUnlock.sol
hardlydifficult: Done

smart-contracts/contracts/PublicLock.sol
hardlydifficult: Removed

hardlydifficult:
Fixed.  @nfurfaro could you take another look?

smart-contracts/contracts/PublicLock.sol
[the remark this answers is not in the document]
nfurfaro: Don't we still need the `Transfer` event to be emitted though? I was more looking at what happens after we emit `Transfer`...
In our contract, `ownerByTokenId` for this token would still return the old owner (but should actually now return `address(0)` right?)

nfurfaro:
Looks good!

PR #1421, opened 2019-02-06
Parallelizing Travis for faster CI builds

hardlydifficult:
# Description

Reduces the build time to 8.5 minutes vs the usual 10.  I believe we can make this much faster if we run with this approach - at the moment each the the parallel runs do redundant steps.  For example, the longest at the moment is smart-contracts but it runs a lot of other Travis commands which are not related to smart contracts (and that suite is going to take another couple mins once solidity-coverage is working).

If we want to continue this direction, I believe the next step may be to add build stages: https://docs.travis-ci.com/user/build-stages/

# Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,14 +14,16 @@
 # Both of these steps could be run in parallel
 language: generic
 env:
-  - TEST_SUITE=smart-contracts
-  - TEST_SUITE=unlock-app
-  - TEST_SUITE=locksmith
   global:
     - DB_USERNAME='locksmith_test'
     - DB_PASSWORD='password'
     - DB_NAME='locksmith_test'
     - DB_HOSTNAME='db'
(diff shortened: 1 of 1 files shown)


hardlydifficult:
Ah thought I'd give it a shot.

hardlydifficult:
@cellog oh a quick process thought to add.  If you create the branch on a fork from your own account, and then PR from there then it counts to Travis as another public repo and therefore has its own 3 runs available.  So if everyone used a fork, we would be pushing half as many builds to our Travis limit and still get all the same coverage.  The only difference is (and its unfortunate) is when viewing a PR you only see the one build (the PR build), you have to return to the fork to see the other (the branch build).  But that should be fine as the PR build is the one that matters, the branch build simply offers more information in case of failure.

PR #1480, opened 2019-02-08
Add wallet check overlay

### LINKED ISSUE #1080: Support full screen web browser
When browsing in full screen, Firefox does not visibly open a popup with Metamask. The user has to click on the Metamask icon to actually approve the transaction (the Metamask icon will actually show a very small number overlay to indicate that something is expected but that's fairly hard to see!)

A solution to that issue is to actually show something like "Check your web3 provider" when a transaction is being initiated so that the user knows that something is expected from them.
We could either do that globally across the app (by applying a window wide lightbox which says "check your web3Provider" or something custom for each transaction but I think the first approach would be easier for now.

What do you think @smombartz ?

cnasc:
# Description
This PR fixes #1080. It adds a `WalletCheckOverlay` component to the top-level of the app. When the `walletStatus` in the state is set to `{ waiting: true }`, it renders an overlay that instructs the user to interact with their wallet to complete the transaction. 

**This isn't quite ready to go yet, but I would appreciate some review.**

- I think the appearance of this component on mobile is still an open question.
- At the moment the `ModalOverlay` file is named as such in case we want to create other descendants of the overlay. Maybe it would be a good idea to rename it to only have the specific focus of blocking the page for wallet interaction.

image

# Checklist:


unlock-app/src/components/interface/ModalOverlay.js
```diff
@@ -0,0 +1,91 @@
...
+  const {
+    walletStatus: { waiting },
+  } = state
+  return { waiting }
+}
+
+const Dismiss = styled.button`
+  height: 24px;
+  font-size: 20px;
+  font-family: Roboto;
```
cellog: provide a fallback font?

unlock-app/src/components/interface/ModalOverlay.js
```diff
@@ -0,0 +1,91 @@
...
+  background: rgba(0, 0, 0, 0.5);
+  position: fixed;
+  height: 100%;
+  width: 100%;
+  left: 0;
+  top: 0;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
```
cellog: add `z-index: var(--alwaysontop)`?

unlock-app/src/components/interface/ModalOverlay.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { dismissWalletCheck } from '../../actions/walletStatus'
+
+const ModalOverlay = ({ children }) => (
```
cellog: perhaps rename to `FullScreenModal`

cnasc: Rename the file, right?

cellog: yes. Since it contains multiple components, I might go for `FullScreenModals.js`. Up to you.

PR #1481, opened 2019-02-08
using docker.com instead of aws to store our docker image

julien51:
This seems faster than AWS and faster than caching locally (which is surprising) so for now we'll use
this. Since this is still network-dependent I do expect latency to be a problem which is why I want
to keep exploring a local cache.


julien51:
We can probably tag the versions with the commit id for example?

PR #1488, opened 2019-02-08
fix regexp for detecting account

cellog:
# Description

The lock regexp did not allow a trailing slash before the `#` and so was failing to detect the hash in the paywall. This fixes that and adds a regression test

# Checklist:

### DIFF
--- a/unlock-app/src/constants.js
+++ b/unlock-app/src/constants.js
@@ -46,7 +46,7 @@ export const TRANSACTION_TYPES = {
  *
  * You should not use this directly, instead use the utils/routes.js lockRoute function
  */
-export const LOCK_PATH_NAME_REGEXP = /\/([a-z0-9]+)\/(0x[a-fA-F0-9]{40})(\/([^#]+))?(#(0x[a-fA-F0-9]{40}))?/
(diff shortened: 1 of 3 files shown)


unlock-app/src/constants.js
```diff
@@ -46,7 +46,7 @@ export const TRANSACTION_TYPES = {
  *
  * You should not use this directly, instead use the utils/routes.js lockRoute function
  */
-export const LOCK_PATH_NAME_REGEXP = /\/([a-z0-9]+)\/(0x[a-fA-F0-9]{40})(\/([^#]+))?(#(0x[a-fA-F0-9]{40}))?/
+export const LOCK_PATH_NAME_REGEXP = /\/([a-z0-9]+)\/(0x[a-fA-F0-9]{40})(\/([^#]+))?(\/?#(0x[a-fA-F0-9]{40}))?/
```
benwerd: Jumping into a call but this doesn't quite look right to me / it looks like it could be simplified - is the second clause from the end correct?

cellog:
Actually I had an idea for a follow-up to refactor the thing but I just
realized this PR is the right place for that. Stay tuned.

unlock-app/src/constants.js
cellog: Hey, I just completely refactored the whole thing to use `RegExp` so we can use variable interpolation to make the thing comprehensible as well as correct. PTAL

PR #1497, opened 2019-02-11
Enabling integration tests

julien51:
It took a little more time than I wish but integration tests are back!
The core problem was that we did not deploy the smart contract before actually running the tests.
To fix this we had to do 2 things:
* make sure we deploy the smart contracts before running the application when in CI
* we "wait" before running the integration tests that the app is listening on port 3000.


scripts/start.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# This script starts the application.
+# Based on the environment it may also deploy the smart contracts.
+
+if [ -n "$CI" ]; then
+  # Before running the application, when in CI, we first need to deploy the smart contracts
+  # The artifact needs to be re-written before the application starts
+  cd smart-contracts
```
akeem: I think we can leverage  bash subshell here and down below. https://www.tldp.org/LDP/abs/html/subshells.html

julien51:
That's an interesting idea! Definitely useful for another PR!

Issue #1478, opened 2019-02-07
QA Dashboard Lock


smombartz:
A couple of QA items for the dashboard locks. let's chat about them briefly

- [x] add Key to header column labels for Duration Quantity, Price
![image](https://user-images.githubusercontent.com/1377389/52448352-a255bb80-2b01-11e9-9bd8-651dec3189bc.png)

- [x] Shorten Button Text (Withdraw, Edit, Embed)
![image](https://user-images.githubusercontent.com/1377389/52448348-9c5fda80-2b01-11e9-935b-6d9ee2fdb897.png)
- [x] Lock Form Field Fixes (lemniscate, spacing)
![image](https://user-images.githubusercontent.com/1377389/52448358-a71a6f80-2b01-11e9-9690-ebefa3e2967a.png)

- [x] Lock UX improvements (hover states and transition animation)
![image](https://user-images.githubusercontent.com/1377389/52448365-aa156000-2b01-11e9-8d9a-9dad299d10fd.png)

julien51:
I just checked the lemniscate (which was the last on the list!) and it is fixed:
<img width="123" alt="image" src="https://user-images.githubusercontent.com/17735/54454712-7fd24600-4730-11e9-83c2-9c9acae8e4cb.png">

PR #1500, opened 2019-02-11
Design Lock QA

smombartz:
# Description
related to #1478 

- fixed header label names
- fixed round button names and spacing
- added hover over state for lock cards with transition
- right aligned lock from text
- fixed form font size, padding and font height, added border on select, removed blue outline
- added hover over state for create button

image

image

image

Lock Card Animation
https://cl.ly/76e14979897c


cnasc:
Good work on getting the storyshots updated. For the remaining tests, it's easiest if you run `npm run test -- --watch`. That will rerun the tests every time you save. So you can just go one failing test at a time until they're all fixed. It looks like these are due to changed button text (it's looking for the older, longer labels)

PR #1511, opened 2019-02-12
Add paywall/ to Travis

cnasc:
# Description

This PR adds the new split-off paywall to Travis. Travis will now lint (implemented, with a lint error fixed in this PR so we can merge) and test (once there are actual tests) the paywall as it does with all other Unlock code.

# Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ script:
   - scripts/tests.sh smart-contracts
   - scripts/tests.sh unlock-app
   - scripts/tests.sh locksmith
+  - scripts/tests.sh paywall
   - scripts/integration-tests.sh
 after_success:
   - scripts/deploy.sh netlify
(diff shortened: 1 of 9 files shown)


cellog:
I think the issue is that the copying of eslint to paywall isn't happening. I forgot the name of the thing that does it, but check in the scripts dir for that. I think it belongs in this PR, as it's all about config.

cnasc:
Right, `link-parent-bin` should take care of that. Odd, it seems like it should do it for all subdirs already...

cnasc:
> We shouldn't commit artifacts, but should gitignore them and rebuild like unlock-app does, otherwise they'll show modifications every time we run dev

So in this case, `artifacts` is just a symlink not the actual artifacts themselves, which is the same as what `unlock-app` has at the moment

cellog:
Right, but isn't that in gitignore in unlock-app or did that change when I
wasn't looking? :)

Issue #1533, opened 2019-02-14
Move paywall from redux to React hooks


cellog:
To manage state, we need to create React hooks that can then be used in the components. These hooks need to manage:

- [ ] communication with the iframe through postMessage
- [ ] setting up web3
- [ ] configuration
- [ ] localStorage
- [ ] user account retrieval and balance retrieval
- [ ] key retrieval and purchase
- [ ] transaction monitoring
- [ ] lock retrieval

cellog:
closing, byebye hooks... for now

PR #1540, opened 2019-02-15
new Web3 component for passing down the read-only web3 instance to children in the paywall

cellog:
# Description

This PR is a step to #1533 and depends on #1536 to go green, and #1535 to remove `useConfig.js` from the diff

The component pulls in configuration and uses it to create a `Web3` object that can be used in any hook that needs to query the read-only provider. As an example, both the `useKeys` and `useLocks` hooks (in a future PR) depend on the `useWeb3` hook (also for a future PR), which pulls in the web3 instance from the context this component creates. This component requires the `ConfigContext` provider be initialized as a parent with the configuration, which it pulls in using the `useConfig` hook.

In order to simplify testing, it introduces a dead-simple automock for `web3`, which requires telling the existing `web3Service`/`walletService` tests not to mock using `jest.unmock('web3')`

# Checklist:

### DIFF
--- a/paywall/src/__tests__/services/walletService.test.js
+++ b/paywall/src/__tests__/services/walletService.test.js
@@ -16,6 +16,7 @@ import {
 import { POLLING_INTERVAL } from '../../constants'
 
 jest.mock('../../utils/promises')
+jest.unmock('web3')
 
 const nockScope = nock('http://127.0.0.1:8545', { encodedQueryParams: true })
 
--- a/paywall/__mocks__/web3.js
+++ b/paywall/__mocks__/web3.js
@@ -1,0 +1,3 @@
+const web3 = jest.genMockFromModule('web3')
+
+module.exports = web3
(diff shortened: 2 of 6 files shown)


paywall/src/__tests__/services/walletService.test.js
```diff
@@ -16,6 +16,7 @@ import {
 import { POLLING_INTERVAL } from '../../constants'
 
 jest.mock('../../utils/promises')
+jest.unmock('web3')
```
cnasc: I'm a bit curious about these, is this because `web3` is mocked elsewhere and jest stores that between tests?

LG overall!

cellog: if you have 3 minutes, try checking out the branch and removing that line. The difference is that I made a global mock for the web3 node module, and we enabled automocking, so it is instantly mocked in every test. I spent HOURS trying to find a way to mock the damn thing only in 1 test, and jest does not seem to make it possible. If there is a way, I will do that instead in a heartbeat because this way sucks

PR #1556, opened 2019-02-16
new useAccount hook for the paywall

cellog:
# Description

This is a step on the path to #1533 

**comparison to the current model**

The `useAccount` hook is a perfect example of the difference in state management strategies between hooks and redux. In redux, account state management was spread across several middleware and services:

- `walletService` handled retrieving the account, and polling for changes
- `interWindowCommunicationMiddleware` handled retrieving account information from `localStorage`, the URL on redirect from key purchase, and saving the account info back to `localStorage`

All of these account-related activities now happen in the same hook.

**how it is used**

```js
const { account, balance, localStorageAccount } = useAccount(window)
```

The implementation is hidden from users of the hook. All we need to act upon the account as an external user of the hook are the account address, the account balance, and whether it was set from localStorage or not. So, this is what the hook returns. Any change to any of the 3 values will trigger a re-render of the component using the hook.

**implementation detail: handling asyncrony and testing it in hooks**

Hooks are designed to replace side effects, and in particular asynchronous ones, although Suspense will probably be a better place for most of this once it is released.

There is one important logical separation in the `useAccount` hook. While researching the best way to test asynchrony in `useEffect`, I discovered that the React team [hasn't quite figured this out either](https://github.com/facebook/react/issues/14769#issuecomment-462528230).

Ultimately, it looks like using asynchrony directly inside the hook is probably an anti-pattern. It couples the implementation far too tightly, so prying it apart for testing is virtually impossible.

For this reason, the internals of `useAccount` are abstracted into `asyncActions/accounts.js`. In order to take advantage of the state changes, the async functions need to close over the values from the hook. If the function was declared directly in the hook, this would happen naturally, but makes testing impossibly complex. To achieve this same functionality, the 2 async functions are instead created using factory functions. The factories pass in the variables that the async functions need to close over.

With this change, testing is super simple, at the expense of arguably slightly less readability. However, it also yields much shorter functions, and clearer separation of concerns, so this may actually be an advantage. It does increase the surface area for bugs, as forgetting to pass in a variable or changing a signature can result in bugs. However, because this is simply a split of a single hook, the async actions are not called anywhere else, which reduces the chances for a bug.

After considering the above trade-offs, I chose the current design.

**improved polling using hooks**

One massive advantage over the old redux-based/*service design is apparent in the polling for account change. Because polling is abstracted out into a separate hook, the function that does the polling does not need to have any awareness that it is being called repeatedly. This vastly simplifies testing.

# Checklist:

### DIFF
--- a/paywall/src/__tests__/hooks/asyncActions/accounts.test.js
+++ b/paywall/src/__tests__/hooks/asyncActions/accounts.test.js
@@ -19,6 +19,8 @@ describe('useAccount async account actions', () => {
       },
     }
 
+    // this helper function allows us to easily create a mock getAccount
+    // to test the code works as expected
     function testMockGetAccount() {
       return makeGetAccount({
         window: fakeWindow,
@@ -62,7 +64,7 @@ describe('useAccount async account actions', () => {
         }
       })
       describe('paywall in iframe', () => {
-        describe('account does not exist', () => {
+        describe('account does not exist (like in coinbase wallet)', () => {
           beforeEach(() => {
             accounts = []
           })
@@ -124,6 +126,7 @@ describe('useAccount async account actions', () => {
     let setAccount
     let setBalance
 
+    // easy mock pollForAccountChange creator for testing
     function testMockPollForAccount() {
       return makePollForAccountChange({
         web3,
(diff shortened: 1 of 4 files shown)


cellog:
figured out a better way, closing to do a new PR

PR #1557, opened 2019-02-16
new useLock hook for the paywall

cellog:
# Description

This is a step on the path to #1533 

the `useLock` hook retrieves lock data from the read-only web3 provider, and forces a re-render when it is done retrieving it.  It is used simply by:

```js
const lock = useLock(address)
```

Presumably the lock address will be parsed from the URL and passed in (that's how the paywall uses it).

Currently there is no polling, and the lock is only retrieved once. If desired, adding polling for price changes/key availability would be trivial to add, just would need to do a usePoll instead of useEffect

# Checklist:


cellog:
closing in order to refactor after learning a better way forward for handling async

cellog:
(incorrect, this is as good as it gets for this particular PR, reopening)

PR #1563, opened 2019-02-18
new useNetwork hook for the paywall

cellog:
# Description

A step on the path to #1533 

This is the new `useNetwork` hook for the paywall. It queries the user wallet to determine which network we are currently on. The output will be used in the paywall to trigger an error if the wallet is on a different network than the read-only web3. Unlike the old way, this one polls the network so that when the user changes the network, it will trigger an error as well. Note that metamask currently triggers a page refresh when the network changes, so that behavior is not currently necessary. However, it is possible that other wallets like coinbase wallet do not trigger this behavior, so this is a nice failsafe.

This PR perfectly exhibits the way to isolate async stuff while preserving most of the logic in the hook. This allows us to mock out the asynchronous stuff and still fully test the business logic using a component. This detail-hiding both allows us to test the user/dev experience and to ensure we aren't testing implementation details. Refactoring will be much easier in the future with this style of testing.

# Checklist:

### DIFF
--- a/paywall/src/__tests__/hooks/web3/useNetwork.test.js
+++ b/paywall/src/__tests__/hooks/web3/useNetwork.test.js
@@ -1,0 +1,101 @@
+import React from 'react'
+import * as rtl from 'react-testing-library'
+import PropTypes from 'prop-types'
+
+import useNetwork from '../../../hooks/web3/useNetwork'
+import { wrapperMaker } from '../helpers'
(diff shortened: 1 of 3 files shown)


julien51:
Hey so I am still a bit lost on all of these PRs. What is the status? I see this one re-implements a lot of the logic in walletService which I thought we agreed to use here? Please let me know the status of things.

PR #1565, opened 2019-02-18
new useKeyPurchaseTransaction hook for the paywall

cellog:
# Description

This is a BIG step on the path to #1533. Requesting the review of all 3 of you, because the transaction stuff took me a really long time to extract, and I want to be sure I fully understood the key purchase flow and implemented it correctly here. Note that I elected not to pull in the code for purchasing a key for someone else, but left comments where it would go, in order to get the most simple MVP. Adding that feature in should be trivial.

This PR introduces a new hook to manage key purchase transactions. A follow-up PR will introduce the use of the read-only web3 to pull in key data.

This PR is a big momma, and a huge change from the way the `walletService` and `web3Service` handle key purchasing.

The basic steps needed to implement this were scattered all over both services, primarily to support generic reading of web3 transactions regardless of source (lock contract creation, key purchase, key price update of lock...)

This PR streamlines the process. There are 3 important locations that handle the key creation and listening for confirmations.

- `sendNewKeyPurchaseTransaction` in `asyncActions/keyPurchaseTransactions.js`
- `getTransactionInfo` in `asyncActions/keyPurchaseTransactions.js`
- `handleTransactionUpdates` in `useKeyPurchaseTransaction.js`

`sendNewKeyPurchaseTransaction` triggers the actual purchase, pulls in the key hash when it's ready, and triggers errors as necessary. `getTransactionInfo` is called on a polling interval to retrieve confirmations and blockNumber. `handleTransactionUpdates` converts these events into state.

**"hook"ing it all up**

Because hooks declare state management rather than use logic, everything has to exist in the top of the function. Thus, the polling for transaction happens continuously, but we only call `getTransactionInfo` when a key is actively being purchased, and is not yet fully confirmed.

This PR also introduces a new state management hook. Instead of using `useState`, we use `useReducer`, which is a better fit for handling a complex state object like `transaction`, where the new state depends heavily on the prior state. This way, we can implement a simple state machine for updating the transaction over time in a declarative manner, just like a redux action/reducer chain. Because everything is contained in a single hook, we can use simple strings as the action type.

I elected to make action creators to pass into `getTransactionInfo` in order to hide implementation details.

This is by far the most complex hook, and the last major one. After this, `useKey` is left, and then the components that actually use the hooks!

# Checklist:


cellog:
What did you have in mind for other ways it can be used? Or more specifically what should I change?

akeem:
@cellog At first glance, I'd probably shift the code out of the hooks directory.  I could be interpreting the code wrong, but I feel like methods like `getTransactionInfo` , `handleTransactionUpdates`, etc provide value that isn't tied to our current use of hooks

cnasc:
Based on the idea of using `useLocalStorage` back in the dashboard, I wonder if the hooks are foundational infrastructure such as to justify their own spot in the monorepo, from which the various apps can pull them in (rather than dipping into the `paywall` directory to use them). The difference is minor, but I think indicates intent better and just feels cleaner.

PR #1568, opened 2019-02-18
QA on Create Lock Form

smombartz:
fixed unsightly balance cut off due to padding and made form field slightly bigger.

image


julien51:
@smombartz I think you need to regenerate the snapshots here and then tests should pass just fine ;)

smombartz:
merge conflict bs – redoing....

PR #1577, opened 2019-02-18
new Web3Service component for paywall

### LINKED ISSUE #1569: Create a component that passes web3Service down through context
similar to the `Web3` component, this `Web3Service` component will pass the service down through context

cellog:
# Description

Fixes #1569 by adding a new `Web3Service` component to pass the `web3Service` instance down through context.

This will pair with the `useWeb3Service` hook to pull the service into other custom hooks

# Checklist:

### DIFF
--- a/paywall/src/hooks/components/Web3Service.js
+++ b/paywall/src/hooks/components/Web3Service.js
@@ -1,0 +1,16 @@
+import PropTypes from 'prop-types'
+import React, { createContext } from 'react'
+
+import web3Service from '../../services/web3Service'
+
+export const Web3ServiceContext = createContext()
+
+export default function Web3Service({ children }) {
+  const { Provider } = Web3ServiceContext
+  const web3 = new web3Service()
+  return <Provider value={web3}>{children}</Provider>
+}
+
+Web3Service.propTypes = {
+  children: PropTypes.node.isRequired,
+}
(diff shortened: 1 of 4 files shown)


paywall/src/hooks/components/Web3Service.js
```diff
@@ -0,0 +1,16 @@
+import PropTypes from 'prop-types'
+import React, { createContext } from 'react'
+
+import web3Service from '../../services/web3Service'
+
+export const Web3ServiceContext = createContext()
+
+export default function Web3Service({ children }) {
```
julien51: I think it is going to be confusing to have many `Web3Service`. Since this is not a service but a component, could we just name this `Web3Component`?

paywall/src/hooks/components/Web3Service.js
```diff
@@ -0,0 +1,16 @@
+import PropTypes from 'prop-types'
+import React, { createContext } from 'react'
+
+import web3Service from '../../services/web3Service'
+
+export const Web3ServiceContext = createContext()
```
julien51:
```suggestion
export const Web3Context = createContext()
```

cellog:
Sure. FYI there is already a Web3.js which was based on my prior work. It literally creates a `new Web3` as opposed to a `new web3Service`, hence the name. How about `Web3ServiceComponent`?

paywall/src/hooks/components/Web3Service.js
cellog: renamed to `Web3ServiceProvider`. The suffix `Provider` should be used for all context providers, and the prefix should describe what it is passing down through context. So, for example, the existing `ConfigProvider` component is a great example of this naming convention. This component passes down the `web3Service` through its context. PTAL

PR #1585, opened 2019-02-18
Set up CI with Azure Pipelines

julien51:


julien51:
Ha! so far so good ;) trying to see what it takes to move away from travis...

PR #1587, opened 2019-02-18
Experiment: Refactor mixins and add approveForAll (ERC721 standard)

hardlydifficult:
# Description

Adding approveForAll, which is required to meet the ERC721 standard.  Since an owner can have up to 1 Key, the impact of this method is very similar to the standard approve - the difference being the approval is not lost when a transfer occurs.

This change only adds a few functions to the contract.  However I have included a refactor with these changes for consideration.  

The goal of this refactor is to simplify our large PublicLock contract.  I have separated a few distinct capabilities into their own file.  e.g. all approval related data and functions are in `MixinERC721Approval.sol`.  I find this makes it easier to read/review functionality (and may encourage re-use as we implement new types of Locks).

Inheritance in Solidity is effectively a copy/paste then compile process - so the impact of this refactor is 100% non-functional (i.e. the bytecode diff from the refactor alone should be null). 

This approach was inspired by 0x, see https://github.com/0xProject/0x-monorepo/tree/development/contracts/exchange/contracts/src

If you are not onboard with this "mixin" approach, I can roll that back while keeping the new approveForAll implementation.

# Checklist:

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -7,6 +7,9 @@ import "openzeppelin-eth/contracts/ownership/Ownable.sol";
 import "./interfaces/IERC721Receiver.sol";
 import "openzeppelin-eth/contracts/introspection/ERC165.sol";
 import "openzeppelin-eth/contracts/math/SafeMath.sol";
+import "./mixins/MixinERC721Approval.sol";
+import "./mixins/MixinKeyOwner.sol";
+import "./mixins/MixinDeprecateAndDestroy.sol";
 
 /**
  * TODO: consider error codes rather than strings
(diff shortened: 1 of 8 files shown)


hardlydifficult:
@julien51 @nfurfaro could you both take a look at this one, curious how you feel about the refactor.

nfurfaro:
Will review tonight, but I'm  in favor of the approach in general.

hardlydifficult:
To clarify, if we go with this "mixin" design pattern the long term goal is the PublicLock contract is nothing more than a collection of mixin components.  e.g. https://github.com/0xProject/0x-monorepo/blob/development/contracts/exchange/contracts/src/Exchange.sol

nfurfaro:
Would you change the ILockCore interface at all to reflect the mixin pattern, or leave it as is (monolithic)? Functionally it probably doesn't matter, and it still serves its purpose without a big refactor.

hardlydifficult:
That's a good question.  It may be easier to keep one massive ILockCore file - but we should definitively revisit that when we start to add new lock types.

hardlydifficult:
This is an experiment (but I don't think I can downgrade it to a draft PR)

I will be closing this PR down and follow up with a few separate/more targeted PRs.  But do let me know if you have more comments on this approach as I plan on running with it.

nfurfaro:
@hardlydifficult on the subject of interfaces in the context of the mixin pattern:
 I think making use of both interfaces and abstract contracts will likely make sense here. 
For example, `ILockCore` could be implemented as an abstract contract, allowing it to inherit (and compose) the individual pure interfaces behind each mixin...It could also possibly form the backbone of a new type of lock (Private?), but if that didn't make sense we just use a new abstract contract `IPrivateLock.sol` which composes the interfaces it needs (including any new ones we may need to write).

hardlydifficult:
Yea I agree - creating abstract contracts like IPublicLock which bring a few lock interfaces together is a good approach.

hardlydifficult:
Last PR related to this coming shortly.

Issue #1175, opened 2019-01-23
Find lock address before deployment


julien51:
We currently assign a new (temporary) address to locks while we wait for them to have been deployed.
It turns out that we can know the address which will be used for the lock before the transaction has actually been mined. 
This will remove the need for the temporary address.

hardlydifficult:
This creates a race condition to consider.  Addresses are based on the account `nonce`.  When a contract is deployed by another contract, the `nonce` is from the original contract (Unlock for us) and not the user's `nonce`.  

When someone creates a new Lock, we can determine the address of the next Lock.  However if someone else is creating a Lock around the same time - the first to be mined will have correctly determined the Lock address however the second to be mined would be pointing to a valid Lock, but not the Lock they expected.

hardlydifficult:
I believe we could address this once EIP-1014 is merged using CREATE2. https://eips.ethereum.org/EIPS/eip-1014

akeem:
^ is super cool. We should shift to this when it become available. @julien51 since we shipped the first pass of this. How you feel about us adjusting this issues to update when available with a weekly check?

hardlydifficult:
Labeling `Blocked` since we have an implementation today and the next step, switching to `CREATE2` can only happen after an Ethereum hard fork.

hardlydifficult:
eta this week. Fingers crossed

julien51:
I believe this was implemented by @akeem!
Closing!

PR #1596, opened 2019-02-19
Add transaction log page

### LINKED ISSUE #1444: Create transaction log page UI
Requirements for this feature mostly live in the [Figma](https://www.figma.com/file/gmhVq6Pqeu2gH7e3FC2lr0ax/Unlock-Dashboard?node-id=0%3A67)

Per discussion with @smombartz, the plan right now is to show only the first 3 columns until a good design for showing disparate kinds of data in the table can be devised. So this issue is really a v1 of the transaction log, and it will be enhanced over time.

I anticipate the following steps:

- [x] create `transactionLog` page
- [x] format the page into columns
  - [ ] make that formatting mobile-friendly
- [x] determine how transaction data will be fetched (it seems that right now only lock creation is always available from the Redux state, so we have to figure out key purchase, price update, and withdrawal)
- [ ] set up navigation
- [ ] date time of each transaction rather than blockNumber (and probably "pending" for ones that have yet to be mined)

cnasc:
# Description

This PR fixes #1444. It creates the initial transaction log page. This will remain as a draft PR until some items are sorted out.

image

A future PR will change the navigation so that this page is actually accessible. Until then, it can be reached at `/log`.

TODOs for future PRs:

- Update the navigation so that this page is actually accessible. Until then, it can be reached at `/log`.
- Right now, lock creations show the wrong address in the `lock` field, I believe that when #1175 is fixed this should not be a problem.
- Coloring of various events is now limited to green for lock creation and dark grey for everything else. There may be room for more contextual coloring (pending transactions in yellow? ...)
- Message when there are no transactions, like the message on the dashboard when there are no locks.
- Translate `blockNumber` to date (this should be handled elsewhere in the stack)

# Checklist:

### DIFF
--- a/unlock-app/src/pages/log.js
+++ b/unlock-app/src/pages/log.js
@@ -95,7 +95,7 @@ Log.propTypes = {
 const Body = styled.div`
   display: grid;
   grid-template-columns: min-content min-content min-content;
-  grid-template-rows: 20px;
+  grid-auto-rows: 20px;
   grid-column-gap: 20px;
 `
 
--- a/unlock-app/src/pages/log.js
+++ b/unlock-app/src/pages/log.js
@@ -11,6 +11,14 @@ import GlobalErrorConsumer from '../components/interface/GlobalErrorConsumer'
 import GlobalErrorProvider from '../utils/GlobalErrorProvider'
 import { pageTitle } from '../constants'
 
+const etherScanUrlFor = address => `https://etherscan.io/address/${address}`
+
+const humanize = type => {
+  if (!type) return ''
+  let parts = type.split('_')
+  return parts.map(part => part[0] + part.slice(1).toLowerCase()).join(' ')
+}
+
 export const Log = ({ account, network, transactionFeed }) => {
   return (
     <GlobalErrorProvider>
@@ -31,9 +39,14 @@ export const Log = ({ account, network, transactionFeed }) => {
                     <LogElement key={tx.id + '__blockNumber'}>
                       {tx.blockNumber}
                     </LogElement>
-                    <Address key={tx.id + '__address'}>{tx.lock}</Address>
-                    <Type key={tx.id + '__type'} type={tx.event}>
-                      {tx.type}
+                    <Address
+                      href={etherScanUrlFor(tx.lock)}
+                      key={tx.id + '__address'}
+                    >
+                      {tx.lock}
+                    </Address>
+                    <Type key={tx.id + '__type'} type={tx.type}>
+                      {humanize(tx.type)}
                     </Type>
                   </>
                 ))}
@@ -80,9 +93,13 @@ const typeColors = {
 
 const Type = styled(LogElement)`
   color: ${props => 'var(--' + typeColors[props.type] || 'darkgrey'});
+  white-space: nowrap;
 `
 
-const Address = styled(LogElement)``
+const Address = styled.a`
+  font-size: 11px;
+  font-weight: 300;
+`
 
 export const mapStateToProps = ({ account, network, transactions }) => {
   const transactionFeed = Object.values(transactions)
--- a/unlock-app/src/pages/log.js
+++ b/unlock-app/src/pages/log.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React from 'react'
 import { connect } from 'react-redux'
 import Head from 'next/head'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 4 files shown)


julien51:
> the effect hook seems to run many more times than I think should be necessary.

I thought for now we were **not** using hooks for the unlock-app but for paywall only?

unlock-app/src/pages/log.js
```diff
@@ -0,0 +1,132 @@
...
+import CreatorAccount from '../components/creator/CreatorAccount'
+import BrowserOnly from '../components/helpers/BrowserOnly'
+import GlobalErrorConsumer from '../components/interface/GlobalErrorConsumer'
+import GlobalErrorProvider from '../utils/GlobalErrorProvider'
+import { pageTitle } from '../constants'
+import Web3Service from '../services/web3Service'
+
+export const Log = ({ account, network, lockAddresses }) => {
+  const [transactions, setTransactions] = useState([])
+  const w3s = new Web3Service()
```
julien51: Huh... I do not think we should access `Web3Service` from here especially since all transactions are already in the redux store?

cnasc:
Per discussion with Julien, this will be refactored to no longer use a hook. This is blocked until I go and add all transactions to the Redux state

cnasc:
At long last, the first iteration on the transaction log is done. See the main PR comment up top for a review of where we stand on this. @smombartz I would definitely appreciate a design review on this!

unlock-app/src/pages/log.js
```diff
@@ -0,0 +1,116 @@
...
+  return (
+    <GlobalErrorProvider>
+      <GlobalErrorConsumer>
+        <Layout title="Creator Log">
+          <Head>
+            <title>{pageTitle('Log')}</title>
+          </Head>
+          <BrowserOnly>
+            <CreatorAccount network={network} account={account} />
+            <Body>
```
cellog: "Body" is a confusing name since we have "Head" right above, perhaps "Content"?

unlock-app/src/pages/log.js
```diff
@@ -0,0 +1,116 @@
...
+              <LogHeader>Block Number</LogHeader>
+              <LogHeader>Lock Name/Address</LogHeader>
+              <LogHeader>Type</LogHeader>
+              {transactionFeed.length > 0 &&
```
cellog: if length is 0 it will not display anything, so this line is unnecessary

unlock-app/src/pages/log.js
```diff
@@ -0,0 +1,116 @@
...
+                transactionFeed.map(tx => (
+                  <React.Fragment key={tx.hash}>
+                    <LogElement>{tx.blockNumber}</LogElement>
+                    <Address
+                      href={
+                        chainExplorerUrlBuilders.etherScan(tx.lock) || undefined
```
cellog: Do this in the mapStateToProps, we should only have dumb data in the component (this will make converting to anything non-redux in the future easier)

unlock-app/src/pages/log.js
```diff
@@ -0,0 +1,116 @@
...
+                      }
+                      target="_blank"
+                    >
+                      {tx.lock}
+                    </Address>
+                    <Type type={tx.type}>{humanize(tx.type)}</Type>
```
cellog: same for this, let's assume the data arrives formatted for the component

unlock-app/src/pages/log.js
```diff
@@ -0,0 +1,117 @@
...
+import UnlockPropTypes from '../propTypes'
+import Layout from '../components/interface/Layout'
+import CreatorAccount from '../components/creator/CreatorAccount'
+import BrowserOnly from '../components/helpers/BrowserOnly'
+import GlobalErrorConsumer from '../components/interface/GlobalErrorConsumer'
+import GlobalErrorProvider from '../utils/GlobalErrorProvider'
+import { pageTitle } from '../constants'
+import configure from '../config'
+
+const { chainExplorerUrlBuilders } = configure()
```
cellog: It's better if you use the ConfigContext to get config values

unlock-app/src/pages/log.js
```diff
@@ -0,0 +1,117 @@
...
+
+export const humanize = type => {
+  if (!type) return ''
+  let parts = type.split('_')
+  return parts.map(part => part[0] + part.slice(1).toLowerCase()).join(' ')
+}
+
+export const Log = ({ account, network, transactionFeed }) => {
+  return (
+    <GlobalErrorProvider>
```
cellog: I could be wrong but I believe the GlobalErrorProvider is passed down from the _app.js or _document.js

cnasc: It isn't but maybe it should be, at the _app.js level would make sense.

unlock-app/src/pages/log.js
cnasc: In this case I need the value from the config in `mapStateToProps`. I'm not super familiar with HOCs; what's the best way to connect those?

unlock-app/src/pages/log.js
cellog: ok

unlock-app/src/pages/log.js
cellog: use `withConfig` around `connect` and it will pass it in as a prop named `config`. Then you don't need to wrangle with the context either (and I had forgotten about that too because I'm old)

PR #1599, opened 2019-02-19
When we set the account, also get all the transactions for the locks...

cnasc:
…associated with that account

# Description

This PR establishes the state needed for #1596 to proceed without hooks. Now, when the account is set and the locks created by that account are added to the state, we also get all the transactions on those locks and add them to the state. For example, now key price updates and key purchases are present in Redux.

# Checklist:

### DIFF
--- a/unlock-app/src/middlewares/web3Middleware.js
+++ b/unlock-app/src/middlewares/web3Middleware.js
@@ -144,7 +144,11 @@ export default function web3Middleware({ getState, dispatch }) {
         web3Service
           .getPastLockCreationsTransactionsForUser(action.account.address)
           .then(lockCreations => {
-            // For each lock this user created, go get the history of interactions with that lock.
+            // For each lock this user created, go get the history of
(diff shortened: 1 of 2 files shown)


unlock-app/src/middlewares/web3Middleware.js
```diff
@@ -141,9 +141,16 @@ export default function web3Middleware({ getState, dispatch }) {
...
         web3Service.refreshAccountBalance(action.account)
-        web3Service.getPastLockCreationsTransactionsForUser(
-          action.account.address
-        )
+        web3Service
+          .getPastLockCreationsTransactionsForUser(action.account.address)
+          .then(lockCreations => {
+            // For each lock this user created, go get the history of interactions with that lock.
+            lockCreations.forEach(lockCreation => {
+              web3Service.getPastLockTransactions(
```
julien51: We may want to only get the transactions on the lock for *that user* because I worry that for popular locks this maybe thousands of transactions... (but for now just add a TODO because we do not have that problem!)

PR #1604, opened 2019-02-19
No fallback function

hardlydifficult:
# Description

Blocks any attempts to call a contract function which does not exist.  

This came up in slack chat, suggesting that throwing an error here will aid debugging.

Depends on #1603

# Checklist:


nfurfaro:
Am I right in thinking that after `selfdestruct` is called, this no longer applies? That is, a default fallback becomes available again after `selfdestruct`?

hardlydifficult:
Yes that's correct, after `selfdestruct` this won't apply.

PR #1607, opened 2019-02-19
Add openzeppelin-solidity package

hardlydifficult:
# Description

Adding the `openzeppelin-solidity` package.  

This is similar to `openzeppelin-eth`, however `-eth` contracts have been modified to support upgrading.  We need those for `Unlock` but not for the `PublicLock` contract.  One of the features used for upgrading is a "gap" variable (reserving space for future use) - which increases the gas cost.  

Using the version that's most similar to the version of `openzeppelin-eth` we are using.  Later versions require upgrading to Solidity 0.5

I'll follow up this PR with one switching some of our dependencies in PublicLock over.

# Checklist:

### DIFF
--- a/smart-contracts/package-lock.json
+++ b/smart-contracts/package-lock.json
@@ -9780,9 +9780,9 @@
       "integrity": "sha512-gUdT24ozD5xXeB9tdTzHOYTpz7R1s0YVXhrYe5TSusiqP0YMqh97GB7Zugee2D6446cKxaE94BblgfgPNGaXRQ=="
     },
     "openzeppelin-solidity": {
-      "version": "1.10.0",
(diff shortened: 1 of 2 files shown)


nfurfaro:
@hardlydifficult We already have `openzeppelin-solidity` in the project, as it's a dependencuy of `zos-lib`. Not sure if its the same version though. Also , I understand about not needing the upgradeability features (and extra gas costs) of oz-eth, but eventually it would be great if we could only depend on one of these packages as far as reducing dependency bloat. I realize that can't happen as long as we need both `oz-eth` and `zos-lib` though...

hardlydifficult:
Yes, we already had openzeppelin-solidity as a dependency.  That was a slightly older version of the contracts though.  Net-net this should have little impact but allows us to be explicit about versions instead.

PR #720, opened 2018-12-14
Migrate from openzeppelin-solidity to openzeppelin-eth.

nfurfaro:
## Proposed Changes

- migrate from openzeppelin-solidity to openzeppelin-eth.
- Small modifications to `UnlockTestV2.sol` and `UnlockTestV3.sol`s initialize functions.
- Updates tests to match the workflow when using zos-cli.
- Updated artifacts to reflect changes in tests.
- Added shell.js as a dep. For now this is used to run the zos-cli directly, but hopefully, it can be removed if we find a way to use 'zos add' programmatically.

PR #1618, opened 2019-02-20
Switch to openzeppelin-solidity (from -eth) for the lock

hardlydifficult:
# Description

`openzeppelin-solidity` is similar to `openzeppelin-eth`, however `-eth` contracts have been modified to support upgrading. We need those for `Unlock` but not for the `PublicLock` contract. One of the features used for upgrading is a "gap" variable (reserving space for future use) - which increases the gas cost.

Switching the PublicLock imports to use `-solidity` with one exception, `Ownable`.  There is a limitation in the Solidity compiler which allows for only one contract per name and since we need the initializable version of Ownable for Unlock we cannot import the normal version insider PublicLock without causing an error.

# Checklist:


julien51:
Wait, I thought we *needed* to use `openzeppelin-eth`?
https://github.com/unlock-protocol/unlock/pull/720

julien51:
I'll let @nfurfaro weigh in on that because he worked on the migration in the other direction in https://github.com/unlock-protocol/unlock/pull/720

hardlydifficult:
Correct, we need it for `Unlock` since it's upgradable but when it comes to the `PublicLock`, using `-eth` costs more without any benefit.

nfurfaro:
As I mentioned earlier, The only (potential) downside to this PR is that it explicitly required us to have both oz-eth and oz-solidity as dependencies...but, as `zos-lib` requires oz-solidity anyway, this is functionally no different from what we already have going on.

hardlydifficult:
Yes, that is a downside.  The conflict with Ownership is an example of that concern.  I think we strongly prefer `-eth` anytime a conflict arises, otherwise use `-solidity` where we can for gas optimizations.

PR #1625, opened 2019-02-21
Add builders for chain explorer URLs to config

cnasc:
# Description

This PR adds a property `chainExplorerUrlBuilders` to the config. Since the chains are different between dev, staging, and prod, this allows us to set which URL to use to point to the explorers. Right now only EtherScan is in there, but it's possible that we'll add more in the future.

# Checklist:

### DIFF
--- a/unlock-app/src/config.js
+++ b/unlock-app/src/config.js
@@ -75,6 +75,9 @@ export default function configure(
   let services = {}
   let supportedProviders = []
   let blockTime = 8000 // in mseconds.
+  let chainExplorerUrlBuilders = {
+    etherScan: () => false,
+  }
   const readOnlyProviderUrl = runtimeConfig.readOnlyProvider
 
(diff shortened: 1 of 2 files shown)


unlock-app/src/config.js
```diff
@@ -75,6 +75,9 @@ export default function configure(
   let services = {}
   let supportedProviders = []
   let blockTime = 8000 // in mseconds.
+  let chainExplorerUrlBuilders = {
+    etherScan: () => false,
```
cellog: how will this be used in the app? to disable the button?

cnasc: Yeah, basically to have a do-nothing link. Open to other ideas on how to handle this

cnasc:
I think we'll leave it as is for now, and if we ever get around to adding more explorers we can re-evaluate.

PR #1627, opened 2019-02-21
account polling is happening at the middleware level, not inside the service

julien51:
Polling for account changes is more app logic then business logic... so we are moving this
away from walletService!


julien51:
It does not call `setTimeout`. It calls it every time...

cellog:
I just realized that the PR does not actually do what it says - the polling is still inside walletService (setTimeout). Can you do another PR that removes all setTimeout from walletService and puts it in the middleware?

The reason is that we can then turn it off. As written now, it will always poll, and it keeps too much implementation of polling inside the walletService

julien51:
No it is not :/ `setTimeout` is inside the middleware... please make sure you read PR fully when reviewing them ;)

PR #1630, opened 2019-02-21
Refactor mixin for key owner related functions

hardlydifficult:
# Description

Part of the refactor described in #1587

Specifically for key owner related functions.  With this I changed `ownerByTokenId` to private (vs internal) and added a couple internal functions with a goal of keeping data itself private to the Mixin -- to ease readability.

No functionality should change with this PR.

# Checklist:


nfurfaro:
@hardlydifficult The one thing I'm not sure of is why we would import `IERC721.sol` here instead of in `PublicLock.sol` ?

hardlydifficult:
Ah that would cause a circular dependency - PublicLock requires MixinKeyOwner which requires PublicLock...

The compiler will confirm that all methods from the interface are implemented - so a mixin can inherit from an interface and only implement a portion of it.  Then the final contract, in our case PublicLock, needs to have all methods from interfaces included in order to compile.

nfurfaro:
Ah, I see it now. I forgot that ownerOf was from erc721.

PR #1645, opened 2019-02-22
Fix pre-push for edge cases

### LINKED ISSUE #1636: Loosen items evaluated by GIT pre push hook
At the moment the current pre push hook appears to be linting deleted files in addition to non-javascript files (json files). 

Ideally we would only pass valid javascript files to eslint for evaluation

cellog:
# Description

Fixes #1636

The existing pre-push fails on an change to any non-javascript file, including changes to .json files. It also attempts to lint files that have been deleted, resulting in an error.

This PR changes the pre-push script to use `--name-status` instead of `--name-only` to retrieve a list of files and their statuses. Each line looks like:

```
D    paywall/oldassfile.js
A    paywall/newhotness.js
M   unlock-app/somerandom.gif
```

The grep filter starts with `^[^D]` which removes all file deletions, then matches the whitespace with `.*` (note: this is somewhat dangerous. I wasn't sure how to do whitespace matching in grep, but this works for our purposes. the danger is simply that if we have a file named `unlock-app/somepaywall.js` it will match when linting unlock-app, which is probably not the end of the world). Then it matches the path, and finally the filename (`.*`) and then the javascript extension `js$` which must be at the end of the expression.

The sed match strips off the A or the M, matches the whitespace more accurately, the path, and the leading slash to match the expected behavior with the new stuff at the start of the line from git status.

The rest is the same, so we get blazing fast push without errors!

# Checklist:

### DIFF
--- a/.huskyrc.js
+++ b/.huskyrc.js
@@ -22,7 +22,7 @@ const lintStaged = path => {
 // It will try to amend the latest commit if possible to fix.
 const eslint = path => {
   return {
-    command: `git diff-tree --no-commit-id --name-only -r HEAD . | sed 's/${path}\\///' | xargs eslint --fix`,
+    command: `git diff-tree --no-commit-id --name-only -r HEAD . | grep js$ | sed 's/${path}\\///' | xargs eslint --fix`,
     path: path
   }
 }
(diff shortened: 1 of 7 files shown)


cellog:
I think that if we decide to go that route we can update it then. Currently we don't have any so I didn't add the `x?` at the end, what do you think?

cnasc:
Does this still only lint files changed in the last commit? Tried the `diff-tree` command in my sandbox branch and it looks like that's what happens.

PR #1658, opened 2019-02-22
Updates registerServiceWorker to async/await paradigm

akeem:
Updates the library to leverage async/await. We may want to consider removing this file from the code base if it isn't currently being utilized.

# Checklist:

### DIFF
--- a/unlock-app/src/registerServiceWorker.js
+++ b/unlock-app/src/registerServiceWorker.js
@@ -31,7 +31,7 @@ export default function register() {
       return
     }
 
-    window.addEventListener('load', () => {
+    window.addEventListener('load', async () => {
       const swUrl = `${process.env.PUBLIC_URL}/service-worker.js`
 
       if (isLocalhost) {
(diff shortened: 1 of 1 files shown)


unlock-app/src/registerServiceWorker.js
```diff
@@ -40,12 +40,11 @@ export default function register() {
 
         // Add some additional logging to localhost, pointing developers to the
         // service worker/PWA documentation.
-        navigator.serviceWorker.ready.then(() => {
-          console.log(
-            'This web app is being served cache-first by a service ' +
-              'worker. To learn more, visit https://goo.gl/SC7cgQ'
-          )
-        })
+        await navigator.serviceWorker.ready
```
cellog: can you wrap in a try/catch that throws away the error to match the promise behavior?

unlock-app/src/registerServiceWorker.js
```diff
@@ -54,66 +53,61 @@ export default function register() {
...
+  }
 }
 
-export function unregister() {
+export async function unregister() {
   if ('serviceWorker' in navigator) {
-    navigator.serviceWorker.ready.then(registration => {
-      registration.unregister()
-    })
+    let registration = await navigator.serviceWorker.ready
```
cellog: wrap in throw-away try/catch to match the original code

unlock-app/src/registerServiceWorker.js
akeem: The existing behavior didn't have a catch here, i'd like to ensure that the replacements here are as they were to reduce the likelihood of unintended side-effects

PR #1672, opened 2019-02-22
Adding a Lock name field

hardlydifficult:
# Description

Adding a `name` field for the Lock.  This field follows the IERC721 optional metadata standard: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md

The name defaults to empty string and may be set anytime by the Lock owner.

I put this in a new `MixinMetadata` file since 1) it's completely independent logic and 2) we are likely to add more Lock metadata in the future.

# Checklist:

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -12,6 +12,7 @@ import './mixins/MixinDisableAndDestroy.sol';
 import './mixins/MixinKeys.sol';
 import './mixins/MixinLockCore.sol';
 import './mixins/MixinRefunds.sol';
+import './mixins/MixinLockMetadata.sol';
 import './mixins/MixinNoFallback.sol';
 import './mixins/MixinTransfer.sol';
 
@@ -34,6 +35,7 @@ contract PublicLock is
   MixinKeys,
   MixinApproval,
   MixinLockCore,
+  MixinLockMetadata
   MixinRefunds,
   MixinTransfer
 {
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -35,7 +35,7 @@ contract PublicLock is
   MixinKeys,
   MixinApproval,
   MixinLockCore,
-  MixinLockMetadata
+  MixinLockMetadata,
   MixinRefunds,
   MixinTransfer
 {
--- a/smart-contracts/contracts/interfaces/IERC721.sol
+++ b/smart-contracts/contracts/interfaces/IERC721.sol
@@ -99,4 +99,7 @@ interface IERC721 {
   /// @param _operator The address that acts on behalf of the owner
   /// @return True if `_operator` is an approved operator for `_owner`, false otherwise
   function isApprovedForAll(address _owner, address _operator) external view returns (bool);
+
+  /// @notice A descriptive name for a collection of NFTs in this contract
+  function name() external view returns (string memory _name); 
 }
--- a/smart-contracts/contracts/mixins/MixinLockMetadata.sol
+++ b/smart-contracts/contracts/mixins/MixinLockMetadata.sol
@@ -1,0 +1,41 @@
+pragma solidity 0.4.25;
+
+import 'openzeppelin-eth/contracts/ownership/Ownable.sol';
+import '../interfaces/IERC721.sol';
+
+
+/**
+ * @title Mixin for metadata about the Lock.
+ * @author HardlyDifficult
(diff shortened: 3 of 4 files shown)


akeem:
This makes sense but I thought we initially made a conscious decision not to store this information on the contract. This was the impetus for spinning up the backend service. Did something change?

hardlydifficult:
I was not aware.  

I just ran with this since we were discussing metadata a bit and I thought this would be an uncontroversial way to kick that off.  What was the reason for leaving this out of the contract?  It seems making the contract self-descriptive would be a nice-to-have.

akeem:
From my understanding, we've opted to not store non-essential data within the contract. I believe one of the drivers was around the additional gas associated. cc'ing @julien51 here as he we be able to add more context here

hardlydifficult:
No problem rolling this back it was a quick change, but it should be very cheap and has 0 impact for Key users (only Lock creators).  I could run a few tests to put exact numbers on it.  

The Key specific data is much more impactful it seems, if we want to consider moving that data over instead.

julien51:
I am indeed against storing non essential data inside the contracts as much as possible. Now I was not aware that this was required for ERC721 metadata so now I am not 100% sure.

I also want to highlight that the lock name stored inside the lock contract (if we do it), the one shown on the dashboard, the one shown inside the consuming app (paywall...) may all be different because they have different purposes?

It would be helpful to know the actual cost of this?

hardlydifficult:
I believe this should be included on chain, however to be clear it is not required by the ERC721 spec.  ERC721 includes an optional field, `name`, which can optionally store a descriptive name.

From the spec:
"The metadata extension is OPTIONAL for ERC-721 smart contracts. This allows your smart contract to be interrogated for its name and for details about the assets which your NFTs represent." 

This matters for any 3rd party integrations.  Consider support on a 0x marketplace.  Ideally the Keys listed on the marketplace are self-descriptive - otherwise 3rd parties such as 0x marketplaces would need to do the extra work of integrating with our (centralized) services in order to surface the relevant information.

Here are gas costs.  Note that gas is not 100% consistent since it depends on so many things so I tested two different ways which is why some of the deltas have a small range.  I added the current gas cost and the impact of the Key data field for a point of reference.
 - createLock costs 2,106,954 today. Adding lock data adds 73,373 - 104,241; removing key data saves 161,359 - 192,227.
 - purchaseFor costs 195,662 today. Adding lock data adds 22-44; removing key data saves 21,129 - 21,107.
 - transferFrom costs 126,857 today. Adding lock data adds 22; removing key data saves 20,954.
 - updateName costs 44,256 - 44,219.

nfurfaro:
Just wanted to contribute my 2 cents...(1 cent USD):
I agree that, while this is optional, it is worthwhile adding it to the Lock contract, even if only for 3rd-party integration. It's worth highlighting that this is basically 1 write to storage (SSTORE) per lock (barring any name changes) vs the key data which is writing to storage at least once per individual key, potentially a very large number.

julien51:
Thanks for your work on this @hardlydifficult .

> This matters for any 3rd party integrations. Consider support on a 0x marketplace. Ideally the Keys listed on the marketplace are self-descriptive - otherwise 3rd parties such as 0x marketplaces would need to do the extra work of integrating with our (centralized) services in order to surface the relevant information.
This resonates with me a lot. We want to be "good citizens" and be integrated in platforms such as 0x.

 > purchaseFor costs 195,662 today. Adding lock data adds 22-44;

Can you explain why this *adds*? I would expect this to be unchanged? now 23 gas is probably just a compilation change?

What are the constraints regarding the size of the name?

hardlydifficult:
Good question.  I had thought it was an optimization different but a bit of guess and check suggests it's related to `function name`... and the strange part is if I copy paste that function the cost to `purchaseFor` grows with each instance.  I'll continue investigation tomorrow.

There are no constraints on name other than the gas cost to publish it.  So 5,000 characters costs about 3,500,000 gas.  And reading the name is free of course.

The first purchaseFor costs a bit more (35,000) so I skipped it this time, here are updated results:

```
  Contract: w/o name
2,106,954
    √ ^ gas used to createLock (124ms)
150,662
    √ ^ gas used to purchaseFor (119ms)
126,857
    √ ^ gas used to transferFrom (289ms)

  Contract: w/ name
2,180,327
    √ ^ gas used to createLock (124ms)
150,684
    √ ^ gas used to purchaseFor (123ms)
126,879
    √ ^ gas used to transferFrom (278ms)
```

hardlydifficult:
I created an issue to track the gas issue #1757 ... I have no idea what's causing that ATM..

PR #1674, opened 2019-02-23
fixing syntax for IS_PULL_REQUEST en variable

julien51:
The approach I took to set up env variables for CircleCI was not right.
This fixes it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,22 +106,30 @@ jobs:
     steps:
       - checkout
       - run:
+          name: Set IS_PULL_REQUEST
+          command: scripts/circleci/set-is-pull-request.sh >> $BASH_ENV
+      - run:
           name: Pull Unlock Image
           command: scripts/pull-image.sh unlock "$CIRCLE_SHA1"
       - run:
           name: Deploy to Netlify
-          command: scripts/deploy.sh netlify $CIRCLE_SHA1 $CIRCLE_TAG $CIRCLE_BRANCH ($CIRCLE_PULL_REQUEST || "false")
+          command: scripts/deploy.sh netlify $CIRCLE_SHA1 $CIRCLE_TAG $CIRCLE_BRANCH $IS_PULL_REQUEST
 
   deploy-locksmith-beanstalk:
     machine: true
     environment:
       DOCKER_REPOSITORY: unlockprotocol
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -104,24 +104,32 @@ jobs:
     environment:
       DOCKER_REPOSITORY: unlockprotocol
     steps:
+      - run:
+          name: Set IS_PULL_REQUEST
+          command: echo 'export IS_PULL_REQUEST=$(if [ -z "$CIRCLE_PULL_REQUEST" ]; then echo false; else echo $CIRCLE_PULL_REQUEST; fi;)' >> $BASH_ENV
```
akeem: I think we should move the pieces echoing out setup to their own scripts. Its a bit of extra work, but without it this super hard to test in isolation if the occasion arises

julien51: Well I actually think the stuff that is specific to each CI platform should not be in our scripts so that we can move easily from a platform to the next?

julien51: Basically here, the only CircleCI specific code is in `.circleci/config.yml`. Similarly the only Travis specific code is in `.travis`... etc.

akeem: We can utilize our own variables in of our script, reassigning their values into their CI vendor specific counter parts into a file similar to this.

We would like our build to be aware of pull request in general, the vendor specific components of is this a PR will change, but our need to know if it is a PR will  not

julien51: I think I am confused :(

We do want to know if a given build is a PR (and we use the `IS_PULL_REQUEST` in the script for this. It is just that CircleCI does not set that value, so we have to set it ourselves.

I think it would be a bad idea to put that logic into one of the deploy script, because if we move to another CI provider we would have to change the deploy script (like you recently had to do to change from the travis 'plugin' to deploy to AWS).

akeem: Yeah I believe we are experiencing miscommunication. I'm suggesting have a script that sets `IS_PULL_REQUEST` rather than doing it here & passing the vendor specific PR_ENV as $1 . Approving to unblock this

julien51: Ohhhh ! Now I understand :) will fix!

julien51: I have just submitted a fix for this.
One caveat is that if we're using a script we need to checkout the code *first*.

PR #1677, opened 2019-02-23
Unifies the launching of our component applications

akeem:
This explores an approach of launching our existing applications: unlock-app & locksmith, with space for launching others if/when the time comes. 

The current approach is not a large deviation from what currently exists and should fulfill an immediate need that currently exists. 

Developers would launch the components from our root directory with `npm run dev`

# Checklist:

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,2 @@
+unlock-app: cd unlock-app && npm run dev
+locksmith: cd locksmith && npm run dev
--- a/locksmith/npm-shrinkwrap.json
+++ b/locksmith/npm-shrinkwrap.json
@@ -84,11 +84,6 @@
         }
       }
     },
-    "acorn-jsx": {
-      "version": "5.0.1",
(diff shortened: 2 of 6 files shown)


julien51:
Not sure this is the rioght approach because it does not run te other servers we need (ganache for now and posgres)? I think eventually we need a docker based system. The team at tilt.dev is comig to show us their thing some time next week!

akeem:
@julien51 When I was initially working on locksmith you pushed back hard on not running locksmith and postgres in docker. I think we should commit to that for the time being and move forward in order to ensure that product features are prioritized. 

Ganache is run as part of the existing development script for running unlock-app in dev, locksmith is run utilizing Sqllite in the development context.

Excited to see what they show us, but don't think that we should halt for a perfect solution as this is a solution that can be reversed or revisited at anytime

akeem:
shelving

PR #1683, opened 2019-02-23
EXPERIMENT: add polling utility and implement it for cancelable account polling

cellog:
This is an experiment to demonstrate the utility that a polling abstraction can provide. There are no new tests or changes to existing ones, so the build fails.

The `pollWithConditions` function allows declaring a function to poll and the interval to poll it at. Simple enough, but the difference from a regular setTimeout loop is the new features

Basically, the new features are:
- polling can be temporarily paused, and then resume
- polling can be permanently stopped

This can be used in a few helpful ways:

1. account polling can be stopped on the server or when we are in a coinbase wallet iframe
2. currency conversion polling can be paused when offline
3. transaction polling could start when a transaction is sent, and stop when it is confirmed or failed
4. we can finally detect network changes by using polling

Because the polling is independent from the thing being polled, the thing being polled need not be aware of its internal state, which simplifies implementation.

In addition, by thinking of middleware as redux-specific and hooks as React 16.8-specific, we can do polling in a hooks-native way, which differs subtly from a redux-specific way. This separation of side effects from the business logic makes it simpler to migrate code to novel environments.

It took about 50 minutes to design this, it's not complex, but it provides a lot more flexibility in the API usage.

### DIFF
--- a/paywall/src/middlewares/currencyConversionMiddleware.js
+++ b/paywall/src/middlewares/currencyConversionMiddleware.js
@@ -4,6 +4,10 @@
 import axios from 'axios'
 
 import { setConversionRate } from '../actions/currencyConvert'
+import pollWithConditions from '../utils/polling'
+import configure from '../config'
+
+const config = configure()
 
 export default store => {
   axios
@@ -13,18 +17,28 @@ export default store => {
         setConversionRate(info.data.data.currency, info.data.data.amount)
       )
     )
-  setInterval(() => {
-    axios.get('https://api.coinbase.com/v2/prices/ETH-USD/buy').then(info => {
-      const {
-        data: {
-          data: { currency, amount },
-        },
-      } = info
-      const current = store.getState().currency[currency]
-      if (current === +amount) return
+  pollWithConditions(
+    () => {
+      axios.get('https://api.coinbase.com/v2/prices/ETH-USD/buy').then(info => {
+        const {
+          data: {
+            data: { currency, amount },
+          },
+        } = info
+        const current = store.getState().currency[currency]
+        if (current === +amount) return
 
-      store.dispatch(setConversionRate(currency, amount))
-    })
-  }, 10000)
+        store.dispatch(setConversionRate(currency, amount))
+      })
+    },
+    10000,
+    () => {
+      if (config.isServer) {
+        throw new Error('currency conversion polling is only on the client')
+      }
+      // temporarily stop polling when offline
+      return !window.navigator.isOnline
+    }
+  )
   return next => action => next(action)
 }
--- a/paywall/src/middlewares/walletMiddleware.js
+++ b/paywall/src/middlewares/walletMiddleware.js
@@ -25,6 +25,7 @@ import { storeLockCreation } from '../actions/storage'
 import configure from '../config'
 import { NO_USER_ACCOUNT } from '../errors'
 import generateSignature from '../utils/signature'
+import pollWithConditions from '../utils/polling'
 
 const config = configure()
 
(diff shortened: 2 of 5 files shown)


paywall/src/utils/polling.js
```diff
@@ -0,0 +1,22 @@
...
+ */
+export default function pollWithConditions(
+  pollingCallback,
+  intervalMs,
+  disablePolling,
+  handleError = () => {}
+) {
+  async function poll() {
+    try {
+      if (disablePolling()) return
```
julien51: Why use a function for this? And if we need to, why assume that this will always be sync?

paywall/src/utils/polling.js
```diff
@@ -0,0 +1,22 @@
...
+      await pollingCallback()
+      setTimeout(poll, intervalMs)
+    } catch (e) {
```
julien51: I worry about wrapping the whole block inside a `try/catch` statement because many things can fail... and would be trace back to this (arguably the `pollingCallback` can trickle down to many things)

paywall/src/middlewares/currencyConversionMiddleware.js
```diff
@@ -13,18 +17,28 @@ export default store => {
         setConversionRate(info.data.data.currency, info.data.data.amount)
       )
     )
-  setInterval(() => {
```
julien51: Note that the way `pollWithConditions` is not thru `setInterval`. This may not be a big deal, but I wonder about un-anticipated side-effects here.

paywall/src/middlewares/currencyConversionMiddleware.js
```diff
@@ -4,6 +4,10 @@
 import axios from 'axios'
```
julien51: I am confused because I thought you wanted to remove the polling from services to make it easier to re-use the services inside the pawyall application. Why change the middleware inside the experiment?

paywall/src/utils/polling.js
```diff
@@ -0,0 +1,22 @@
...
+export default function pollWithConditions(
+  pollingCallback,
+  intervalMs,
+  disablePolling,
+  handleError = () => {}
+) {
+  async function poll() {
+    try {
+      if (disablePolling()) return
+      await pollingCallback()
```
julien51: This implied that the polling callback must be returning a promise. Is that assuming too much? I worry about coupling here.

paywall/src/middlewares/walletMiddleware.js
```diff
@@ -46,23 +47,43 @@ export default function walletMiddleware({ getState, dispatch }) {
...
+    checkDisableAccountPolling,
+    config.env === 'dev' ? handlePollError : () => {} // log errors in dev only
+  )
+
+  async function pollForAccountChange() {
+    const state = getState()
+    const currentAccount = state.account && state.account.address
+    const newAccounts = await walletService.getAccount() // errors fall through to the polling handler
+    const newAccount = newAccounts && newAccounts[0]
+    if (newAccount === currentAccount) return
```
julien51: Since the line `await walletService.getAccount()` is async, isn't there a risk that `currentAccount` might be out of date?

paywall/src/middlewares/walletMiddleware.js
```diff
@@ -46,23 +47,43 @@ export default function walletMiddleware({ getState, dispatch }) {
...
+     */
+    dispatch(
+      setAccount({
+        address: newAccount,
+      })
+    )
+  }
+
+  function checkDisableAccountPolling() {
+    if (config.isServer) return true // true disables polling temporarily
```
julien51: Wouldn't that be permanent? If `checkDisableAccountPolling` returns `true`, then `setTimeout(poll, intervalMs)` in your util function will never be called anymore?

paywall/src/utils/polling.js
julien51: And if any error happens deep in the stack, the `setTimeout` function will not be called anymore which means that polling will be halted forever.

paywall/src/middlewares/web3Middleware.js
```diff
@@ -116,12 +120,31 @@ export default function web3Middleware({ getState, dispatch }) {
         }
       }
 
-      if (action.type === ADD_TRANSACTION) {
```
julien51: I would keep  `ADD_TRANSACTION` and `NEW_TRANSACTION` distinct because the behavior is not strictly the same...

paywall/src/middlewares/web3Middleware.js
```diff
@@ -116,12 +120,31 @@ export default function web3Middleware({ getState, dispatch }) {
...
-        web3Service.getTransaction(action.transaction.hash)
-      }
-
-      if (action.type === NEW_TRANSACTION) {
-        web3Service.getTransaction(action.transaction.hash, action.transaction)
+      if (action.type === ADD_TRANSACTION || action.type === NEW_TRANSACTION) {
+        const defaults =
+          action.type === NEW_TRANSACTION ? action.transaction : undefined
+        web3Service.getTransaction(action.transaction.hash, defaults)
```
julien51: I am not sure why you call `web3Service.getTransaction` twice? here and below?

paywall/src/middlewares/web3Middleware.js
```diff
@@ -116,12 +120,31 @@ export default function web3Middleware({ getState, dispatch }) {
...
+        const { hash } = action
+        pollWithConditions(
+          // the polling function simply returns the transaction with the closed over transaction hash
+          // note that a unique poll starts for each NEW_TRANSACTION call
+          () => {
+            web3Service.getTransaction(hash)
```
julien51: We certainly would want to keep the `defaults` value here because nothing guarantees that on the second call to `web3Service.getTransaction` we would have seen a submitted transaction...

paywall/src/middlewares/web3Middleware.js
```diff
@@ -116,12 +120,31 @@ export default function web3Middleware({ getState, dispatch }) {
...
+          },
+          // we poll every half block
+          config.blockTime / 2,
+          // When a transaction has failed, or reached full confirmations, we stop polling
+          () => {
+            const currentTransaction = getState().transactions[hash]
+            if (!currentTransaction) return
```
julien51: Why would this be true?

Issue #1690, opened 2019-02-25
Translate blockNumber to date/time


cnasc:
For the transaction log (and perhaps other places in the app), we may want to show when a transaction occurred using a date rather than a block number. Ideally we'll do this in a somewhat efficient way, probably making use of localStorage so we don't make tons of repeated little requests every time we have a list of blockNumbers.

julien51:
Given that we know the average block time and that the blocks are "Sequential" we can probably get good estimates by just getting a few block's actual mining time? (not sure the log needs the actual second-precise value?)

cnasc:
It looks like average block time is somewhere around 20 seconds, I'd just be concerned about the possibility of drift over longer time intervals

akeem:
closing as this code has shipped

PR #1703, opened 2019-02-25
Initial pass at an API providing block timestamps

akeem:
As part of addressing #1690

This pull request introduces an addition to Locksmith that provides the following:
* a service providing block timestamps; supporting our transaction page as well as further use cases.

known knows:

* There are currently no indexes for the newly created data. 
* We are currently storing all data in the same table; this is intentional as it is relatively easy to extract the we reach the point of needed an instance of the service for each chain.

Once we are comfortable with the direction of the change, the change will be partitioned to separate the addition of dependencies for the code
# Checklist:

### DIFF
--- a/locksmith/src/app.js
+++ b/locksmith/src/app.js
@@ -8,6 +8,7 @@ const router = express.Router()
 
 var transactionRouter = require('./routes/transaction')
 var lockRouter = require('./routes/lock')
+var blockRouter = require('./routes/block')
 
 app.use(cors())
 app.use(bodyParser.json())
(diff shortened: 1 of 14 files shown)


cnasc:
This looks reasonable to me

locksmith/src/app.js
```diff
@@ -16,5 +17,6 @@ app.post(/^\/lock$/i, signatureValidationMiddleware)
 app.use('/', router)
 app.use('/', transactionRouter)
 app.use('/', lockRouter)
+app.use('/block', blockRouter)
```
cnasc: Is this route too specific? I.e., will we wish in the future that we reserved `/block` for something other than just the timestamp?

akeem: The general thinking right now is that we will keep the same structure and expand the output to support more of our block centric/related use cases

PR #1707, opened 2019-02-26
KEY_PURCHASE transactions now have lock address metadata on transaction log page

### LINKED ISSUE #1704: KEY_PURCHASE transactions should include the lock they're for
Right now KEY_PURCHASE transactions don't indicate which lock they're for. This makes the transaction log page a little less useful.

Include the target lock in the transaction in Redux.

cnasc:
# Description

This PR fixes #1704. It refactors the `mapStateToProps` of the transaction log page to create a metadata object that holds the calculated values associated with each transaction, now including a lock address for key purchases.

# Checklist:

### DIFF
--- a/unlock-app/src/pages/log.js
+++ b/unlock-app/src/pages/log.js
@@ -58,7 +58,15 @@ Log.propTypes = {
   account: UnlockPropTypes.account.isRequired,
   network: UnlockPropTypes.network.isRequired,
   transactionFeed: PropTypes.arrayOf(UnlockPropTypes.transaction).isRequired,
-  metadata: PropTypes.objectOf(PropTypes.objectOf(PropTypes.string)).isRequired,
+  metadata: PropTypes.objectOf(
+    PropTypes.objectOf(
+      PropTypes.shape({
+        href: PropTypes.string,
+        readableName: PropTypes.string,
+        lock: PropTypes.string,
+      })
+    )
+  ).isRequired,
 }
 
 const Content = styled.div`
--- a/unlock-app/src/pages/log.js
+++ b/unlock-app/src/pages/log.js
@@ -58,7 +58,7 @@ Log.propTypes = {
   account: UnlockPropTypes.account.isRequired,
   network: UnlockPropTypes.network.isRequired,
   transactionFeed: PropTypes.arrayOf(UnlockPropTypes.transaction).isRequired,
-  metadata: PropTypes.object.isRequired,
+  metadata: PropTypes.objectOf(PropTypes.objectOf(PropTypes.string)).isRequired,
 }
 
 const Content = styled.div`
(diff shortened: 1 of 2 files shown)


unlock-app/src/pages/log.js
```diff
@@ -55,6 +58,15 @@ Log.propTypes = {
   account: UnlockPropTypes.account.isRequired,
   network: UnlockPropTypes.network.isRequired,
   transactionFeed: PropTypes.arrayOf(UnlockPropTypes.transaction).isRequired,
+  metadata: PropTypes.objectOf(
```
julien51: Do you mind putting propType definition in `proptypes` with the other ones?

unlock-app/src/pages/log.js
```diff
@@ -102,16 +114,34 @@ export const mapStateToProps = (
...
+  // mutating the state by assigning to the objects
+  let metadata = {}
+
+  transactionFeed.forEach(tx => {
+    const href = chainExplorerUrlBuilders.etherScan(tx.lock) || undefined
+    const readableName = humanize(tx.type)
+    let lock = tx.lock
+
+    // Key purchases don't have a lock field, but they do encode the lock the
+    // key is for as part of the key field. So here we create the necessary lock
```
julien51: We should not do that because the key field shgould be considered an arbitrary string... (it will soon be replaced by the actual TOKEN_ID from the smart contract)

julien51: You an add the `lock` to the transaction object if needed ;)

unlock-app/src/pages/log.js
```diff
@@ -32,15 +32,18 @@ export const Log = ({ account, network, transactionFeed }) => {
...
-                  <Address href={tx.href} target="_blank">
-                    {tx.lock}
-                  </Address>
-                  <Type type={tx.type}>{tx.readableName}</Type>
-                </React.Fragment>
-              ))}
+              {transactionFeed.map(tx => {
+                const { href, readableName, lock } = metadata[tx.hash]
+                return (
+                  <React.Fragment key={tx.hash}>
```
cellog: just FYI, because it could be a consistency issue, other places in the code that use fragment use the new jsx shorthand `<></>`. I am thinking probably we should refactor those in favor of `<React.Fragment>` but import it as `{ Fragment } from 'react'`. @julien51 what do you think?

unlock-app/src/pages/log.js
cellog: agree this would be a great fit for `src/propTypes.js` as a new `transactionMetadata` prop type

unlock-app/src/pages/log.js
cellog: Right, can we get that in the transaction retriever in web3Service somehow?

unlock-app/src/pages/log.js
cnasc: A factor to consider is that the shorthand syntax does not allow for a `key` prop. So if we do want to refactor to maintain consistency, we should definitely go with your suggestion.

cnasc:
Closing in favor of #1870

PR #1543, opened 2019-02-15
Remove flash of MAX_UINT from unlimited key locks

cnasc:
# Description

This PR fixes #1476. By adding an additional boolean property to the lock indicating whether or not it has unlimited keys, we sidestep some issues around handling different numerical representations of that concept.

I think a future PR should be to refactor `CreatorLockForm` (the way that unlimited keys are handled from a user perspective is kind of confusing, and by improving that we can also probably simplify what we pass around and validate) and the rest of the code involved in lock creation. 
# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [x] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [x] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [x] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #1711, opened 2019-02-26
Refactored creatorLockForm which now takes a full lock object as props rather than individual lock properties

julien51:
Some cleeanup before dealing with propTypes and implementing name changes

### DIFF
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -29,59 +29,81 @@ import {
 
 import { INFINITY, UNLIMITED_KEYS_COUNT } from '../../constants'
 
+/**
+ * Converts the lock values into form values
+ * @param {*} lockValues
+ */
+export const lockToFormValues = lockValues => {
+  const formValues = Object.assign({}, lockValues)
+
+  // In the form, duration is shown in days, vs seconds in the lock object
+  formValues.expirationDuration =
+    lockValues.expirationDuration / lockValues.expirationDurationUnit
+
+  // Unlimited keys
+  if (lockValues.maxNumberOfKeys === UNLIMITED_KEYS_COUNT) {
+    formValues.unlimitedKeys = true
+    formValues.maxNumberOfKeys = INFINITY
+  }
+
+  return formValues
+}
+
+/**
+ * Converts the form values into lock values
+ * @param {*} lockValues
+ */
+export const formValuesToLock = formValues => {
+  const lockValues = {}
+
+  lockValues.keyPrice = formValues.keyPrice
+  lockValues.maxNumberOfKeys = formValues.maxNumberOfKeys
+  lockValues.name = formValues.name
+  lockValues.address = formValues.address
+
+  // In the form, duration is shown in days, vs seconds in the lock object
+  lockValues.expirationDuration =
+    formValues.expirationDuration * formValues.expirationDurationUnit
+
+  // Unlimited keys
+  if (formValues.unlimitedKeys) {
+    lockValues.maxNumberOfKeys = UNLIMITED_KEYS_COUNT
+  }
+
+  return lockValues
+}
+
 export class CreatorLockForm extends React.Component {
+  /**
+   *
+   * @param {*} props
+   * @param {*} context
+   */
   constructor(props, context) {
     super(props, context)
-    let expirationDuration = props.expirationDuration
-    let keyPrice = props.keyPrice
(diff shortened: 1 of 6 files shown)


julien51:
Screenshots to make sure that it is not broken:

image

julien51:
With an unlimited lock:

image

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -29,59 +29,81 @@ import {
...
+ */
+export const lockToFormValues = lockValues => {
+  const formValues = Object.assign({}, lockValues)
+
+  // In the form, duration is shown in days, vs seconds in the lock object
+  formValues.expirationDuration =
+    lockValues.expirationDuration / lockValues.expirationDurationUnit
+
+  // Unlimited keys
+  if (lockValues.maxNumberOfKeys === UNLIMITED_KEYS_COUNT) {
```
cellog: Chris added `unlimitedKeys` to locks in the web3Service, so you can use that directly as in:

```js
  if (lockValues.unlimitedKeys) {
```

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -29,59 +29,81 @@ import {
...
+  lockValues.name = formValues.name
+  lockValues.address = formValues.address
+
+  // In the form, duration is shown in days, vs seconds in the lock object
+  lockValues.expirationDuration =
+    formValues.expirationDuration * formValues.expirationDurationUnit
+
+  // Unlimited keys
+  if (formValues.unlimitedKeys) {
+    lockValues.maxNumberOfKeys = UNLIMITED_KEYS_COUNT
```
cellog: this conversion should also be done in walletService.createLock, the idea being that in unlock-app, infinite keys is represented on the lock object, and the conversion to chain-specific representation is in the services.

unlock-app/src/components/creator/CreatorLockForm.js
julien51: So... I am not sure this move was right :/ We discussed (very vocally if you recall on @cnasc 's first day) that the services should not be defining the "format" of the data being exchanged in the application... the `lock` propTypes as defined in `propTypes.js` does not have `unlimitedKeys`.

unlock-app/src/components/creator/CreatorLockForm.js
julien51: so... I think this is a different issue. The goal of this PR is to make sure that the form component's input and output are consistent with the data format used by the unlock-app. 
Another PR can/will isolate the walletService logic/format from the unlock-app logic/format in the same way.

unlock-app/src/components/creator/CreatorLockForm.js
julien51: To rephrase this I think it is not right that code used inside of the react components has a format that is defined by the web3Service.

julien51:
Thanks for the comment. Please see my responses and let me know if you'd like to discuss further.

unlock-app/src/components/creator/CreatorLockForm.js
cellog: Oh! My mistake if I misunderstood, I was working with Chris to do something similar to this:

```js
    const attributes = {
      keyPrice: x => Web3Utils.fromWei(x, 'ether'),
      expirationDuration: parseInt,
      maxNumberOfKeys: value => {
        if (value === MAX_UINT) {
          return UNLIMITED_KEYS_COUNT
        }
        return parseInt(value)
      },
      owner: x => x,
      outstandingKeys: parseInt,
    }
```

Which was a change to convert keyPrice to eth, as the standard representation of keyPrice throughout unlock-app.

The idea to represent unlimited keys as the boolean `unlimitedKeys` was a similar rationale, which was to say "inside unlock-app, unlimited keys will be represented by this bool" and then to convert to and from the ethereum-specific representation of it inside the web3Service.

I see the missed spot in propTypes, it should definitely be an issue opened to fix if once (and if) we agree that this idea is consistent.

The original issue, for reference is in this PR: #1526 and you can see in my comment the source of my (potential) misunderstanding:

> I would like to take the same approach Julien used for keyPrice, which is to do the conversion in the walletService, and use the same representation throughout the frontend. This is probably more lines of code to touch (especially in the tests) but is a cleaner way to do it, and will ultimately make it a lot easier to move to hooks too. If you want to chat later this afternoon about this, I am around

The final PR was #1543

PR #1749, opened 2019-02-27
Registering the erc721 interface

### LINKED ISSUE #148: ERC721 compatibility
The current implementation of the `Lock` smart contract is not fully compatible with the ERC721 spec.
The main reason for this is that we use the owner's address as the `tokenId` which means that when ownership is changed, the `tokenId` itself is changed, which is not how the spec works.

One approach would be to actually have unique `tokenId` for each key, but that would incure more gas fees. Is there any other option to achieve this?

### LINKED ISSUE #838: We need to register the interfaceID for ERC721.
`ERC165.sol` acts as a registry for interface ID's. This allows other contracts to query ours to determine whether it implements a given interface. Currently, only the interfaceID for erc165 is registered.

Once `PublicLock.sol` is modified to implement the complete basic ERC721 interface, we need to register the interfaceID for ERC721, which is `0x80ac58cd`.
 This would mean adding `_registerInterface(0x80ac58cd);` to the constructor for `PublicLock.sol`. 

Documentation for erc165 can be found [here](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-165.md).

nfurfaro:
# Description

This PR fixes #148 and fixes #838 
- Adds one line to the constructor for `PublicLock`, and is the final step in making our lock contracts fully erc721-compliant.
- Modifies 1 test in `compliance.js` as per the comment found there:
```
describe('Supports ERC721 Interface', () => {
    it('should support the erc721 interface()', async function () {
      // Note: the ERC-165 identifier for the erc721 interface is "0x80ac58cd"
      const result = await locks['FIRST'].supportsInterface.call('0x80ac58cd')
      // when ERC721 interfaceID is registered, this will break and we change to `assert.equal(result, true)`
      assert.equal(result, false)
```

# Checklist:

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -52,6 +52,7 @@ contract PublicLock is
     public
     MixinLockCore(_owner, _expirationDuration, _keyPrice, _maxNumberOfKeys, _version)
   {
+    _registerInterface(0x80ac58cd);
   }
 
   /**
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -52,6 +52,8 @@ contract PublicLock is
     public
     MixinLockCore(_owner, _expirationDuration, _keyPrice, _maxNumberOfKeys, _version)
   {
+    // registering the interface for erc721 with ERC165.sol using
+    // the ID specified in the standard: https://eips.ethereum.org/EIPS/eip-721
     _registerInterface(0x80ac58cd);
   }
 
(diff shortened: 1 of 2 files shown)


smart-contracts/contracts/PublicLock.sol
```diff
@@ -52,6 +52,8 @@ contract PublicLock is
     public
     MixinLockCore(_owner, _expirationDuration, _keyPrice, _maxNumberOfKeys, _version)
   {
+    // registering the interface for erc721 with ERC165.sol
```
hardlydifficult: Another magic number.  Not a fan, but that's how it goes =p

Could you maybe change to something like
`registering the interface for erc721 with ERC165.sol using the ID specified in the standard: https://eips.ethereum.org/EIPS/eip-721`

nfurfaro: Sure, will do.

Issue #1731, opened 2019-02-27
paywall unlock flag cannot access contentWindow unless origin is the same


cellog:
The paywall.min.js code tries to access contentWindow of its child iframe to set the body css, but this can only be done in the context of the child's domain. The solution is for the paywall to set its own body css on hiding of the iframe.

cnasc:
Good catch!

PR #1751, opened 2019-02-27
Refactor Paywall component to use hooks

### LINKED ISSUE #1744: Paywall should use hooks instead of React class
The Paywall uses several complex items that would be better served in a hooks context. These include:

- subscribing to postMessage events
- sending postMessage events

and after #1731 is fixed, it will also include messing with the CSS for the window body.

Hooks make abstracting this behavior in a testable way simpler, and make the source simpler to reason through.

cellog:
# Description

This PR refactors the Paywall component to take advantage of the cleaner approach of hooks.

The paywall does several local state things:

1. retrieve the scroll position of the parent window by listening for postMessage
2. send the locked status to the parent window by sending postMessage
3. (in the near future) modify the document.body CSS directly on hiding the overlay in order to position the flag correctly

Currently, we don't quite do any of these correctly.

1. the paywall always assumes it is in an iframe. This is not necessarily true
2. the postMessage calls do not send a specific origin, this is a major security hole and opens us up to both XSS and CSRF.

By abstracting these items out into the existing `usePostMessage` and `useListenForPostMessage` hooks, we can easily focus on the behavior inside `Paywall` that is relevant to the paywall, and not the implementation details of `postMessage`.

*I have elected not to merge this into unlock-app, as the split of paywall is imminent*

Screenshots to show same behavior (note: #1731 is why the unlocked flag is way up there)

screen shot 2019-02-27 at 12 16 12 pm
screen shot 2019-02-27 at 12 16 02 pm
screen shot 2019-02-27 at 12 15 11 pm
screen shot 2019-02-27 at 12 14 58 pm

# Issues

Fixes #1744 

# Checklist:

### DIFF
--- a/paywall/src/components/Paywall.js
+++ b/paywall/src/components/Paywall.js
@@ -12,7 +12,7 @@ import { lockRoute } from '../utils/routes'
 import useListenForPostMessage from '../hooks/browser/useListenForPostMessage'
 import usePostMessage from '../hooks/browser/usePostMessage'
 
-function Paywall({ locks, locked, redirect }) {
+export function Paywall({ locks, locked, redirect, window }) {
   const data = useListenForPostMessage(window) || { scrollPosition: 0 }
   const scrollPosition = data.scrollPosition || 0
   const { postMessage } = usePostMessage(window)
@@ -48,6 +48,7 @@ Paywall.propTypes = {
   locks: PropTypes.arrayOf(UnlockPropTypes.lock).isRequired,
   locked: PropTypes.bool.isRequired,
   redirect: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
+  window: PropTypes.shape({ postMessage: PropTypes.func }).isRequired, // for ease of unit testing
 }
 
 Paywall.defaultProps = {
@@ -76,7 +77,7 @@ export const mapStateToProps = ({ locks, keys, modals, router }) => {
 
   const modalShown = !!modals[locksFromUri.map(l => l.address).join('-')]
   const locked = validKeys.length === 0 || modalShown
-  return { locked, locks: locksFromUri, redirect }
+  return { locked, locks: locksFromUri, redirect, window }
 }
 
 export default connect(mapStateToProps)(Paywall)
--- a/paywall/src/components/Paywall.js
+++ b/paywall/src/components/Paywall.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import UnlockPropTypes from '../propTypes'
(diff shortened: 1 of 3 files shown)


cellog:
just discovered my understanding of how to retrieve the parent window origin was incorrect. Will need to do a pre-PR that has only the fixes to usePostMessage

cnasc:
Is this ready for another look?

cellog:
Not yet, 4 more PRs in the love train first :)

cellog:
once #1766 is reviewed and merged, this will be rebased. It is ready for review after that

cellog:
rebased and ready, PTAL @julien51 and @cnasc

paywall/src/components/Paywall.js
```diff
@@ -7,69 +7,48 @@ import DeveloperOverlay from './developer/DeveloperOverlay'
...
-    this.handleScrollPosition = this.handleScrollPosition.bind(this)
-    this.handleIframe = this.handleIframe.bind(this)
-  }
-  componentDidMount() {
-    window.addEventListener('message', event => {
-      if (event.data.scrollPosition) {
-        this.handleScrollPosition(event.data.scrollPosition)
+export function Paywall({ locks, locked, redirect, window }) {
+  const data = useListenForPostMessage(window) || { scrollPosition: 0 }
+  const scrollPosition = data.scrollPosition || 0
```
cnasc: Can lines 16 and 17 be joined? Or is it possible that `useListenForPostMessage` on `window` will return a truthy value that doesn't contain `scrollPosition`? More a question of curiosity, other than that looks good!

cellog: `useListenForPostMessage` returns whatever the data value is. If it is not an object, the scrollPosition will not be present, and we'll error. Most commonly, on startup it is `null` prior to receiving a message.

The other thing is perhaps that it should be validated, this is a poor man's validation. We actually don't know for sure if the message will come from `paywall.min.js` or from some other script running in the main page, so we have to be careful about any assumptions. Do you have a better idea with this info?

paywall/src/components/Paywall.js
```diff
@@ -7,69 +7,48 @@ import DeveloperOverlay from './developer/DeveloperOverlay'
...
+  const { postMessage } = usePostMessage(window)
+  useEffect(
+    () => {
+      if (locked) {
+        postMessage('locked')
```
julien51: Nit: could we use constants  for `locked` and `unlocked`?

paywall/src/components/Paywall.js
```diff
@@ -7,69 +7,48 @@ import DeveloperOverlay from './developer/DeveloperOverlay'
...
+      </ShowWhenUnlocked>
+    </GlobalErrorProvider>
+  )
 }
 
 Paywall.propTypes = {
   locks: PropTypes.arrayOf(UnlockPropTypes.lock).isRequired,
   locked: PropTypes.bool.isRequired,
   redirect: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
+  window: PropTypes.shape({ postMessage: PropTypes.func }).isRequired, // for ease of unit testing
```
julien51: :(

julien51: This still feels very strange.

julien51: I'd love it if you could some research on actual good practices for this kind of dependency injection.

paywall/src/components/Paywall.js
```diff
@@ -98,7 +77,7 @@ export const mapStateToProps = ({ locks, keys, modals, router }) => {
 
   const modalShown = !!modals[locksFromUri.map(l => l.address).join('-')]
   const locked = validKeys.length === 0 || modalShown
-  return { locked, locks: locksFromUri, redirect }
+  return { locked, locks: locksFromUri, redirect, window }
```
julien51: I would actually be arguing that having the `mapStateToProps` function pure (no global js object) is more important than having the react component themselves being pure, but again, I think we would need some research on this.

paywall/src/components/Paywall.js
cnasc: I have an idea, not sure if it's reasonable. If we tell `useListenForPostMessage` what data we're expecting, maybe it can always return an object that has default values for the data we want.

Just to spitball:
```
useListenForPostMessage = (window, dataAndDefaults) => {
  // Will always return an object containing the keys of `dataAndDefaults`,
  // if a key is not present in the data value, it takes the default value
  ...
}

// call it like this:
const { scrollPosition } = useListenForPostMessage(window, { scrollPosition: 0 })
```

I don't know if that's the ideal abstraction for this, but it might clean up the code a bit by hoisting the `thing || defaultForThing` into the hook so that the call sites are a little simpler

I don't think this is a critical change right now though, maybe just something to keep in mind if our use of this hook (or others like it) grows to use more properties.

cellog:
I'm going to close this, it's close, but needs some new work to improve the DI of window and also the `useListenForPostMessage` stuff Chris and I talked about. Stay tuned

PR #1754, opened 2019-02-27
Removed Earning

smombartz:
# Description
removed earning column from dashboard header and cleaned up unused header buttons (etherscan, upload, export) since we're not using them.

image

image

### DIFF
--- a/unlock-app/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
+++ b/unlock-app/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
@@ -15595,11 +15595,6 @@ Object {
             Balance
           </div>
           <div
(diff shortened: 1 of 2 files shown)


smombartz:
having issues with git pushing a cleanup change and updated test...

julien51:
do you mind also posting a screenshot from mobile just to make sure things look good?

smombartz:
@benwerd mobile shot added

cellog:
unrelated issue - the column headers are wrapping, we probably need a completely different look for mobile, one that does not have column headers at all, but has small locks that when clicked expand down into vertical form stuff

smombartz:
@cellog they are actually not wrapping but stacked :)

image

Issue #1752, opened 2019-02-27
usage of postMessage in Paywall does not work properly cross-origin


cellog:
The way the paywall communicates with the parent window through postMessage in the `usePostMessage` hook references `window.parent.origin` directly. This is not possible cross-origin, the entire thing is shut down.

In the `unlock-app` we are actually using `'*'` as the domain, which opens up a security hole for XSS/CSRF that is unacceptable. Instead we need to specify the specific domain.

Because `window.parent.origin` is inaccessible, I see 2 ways to handle this:

1. use document.referer
2. use postMessage to send the origin to the iframe from the script.

I'm leaning towards (1) if it will work.

cellog:
so document.referrer works, but is unreliable. We're going to pass the origin as a query parameter to the iframe and pull it from the `window.location`

cellog:
hmm zenhub issues here...

PR #1755, opened 2019-02-27
rework lock parsing to use URL, add origin url parameter

cellog:
# Description

In preparation for the fix of #1752, this PR introduces a new query parameter to the paywall route parsing, `origin`. This is used to pass the `window.origin` of the 3rd party site loading our paywall through the `paywall.min.js` script, so that we can securely `postMessage` to the parent window and be confident that only it will receive the message.

Follow-up PRs will enable passing origin to the iframe from `paywall.min.js` and then using the parameter in the `usePostMessage` and `useListenForPostMessage` hooks.

# Issues

Refs #1752

# Checklist:

### DIFF
--- a/paywall/src/constants.js
+++ b/paywall/src/constants.js
@@ -44,8 +44,8 @@ export const ACCOUNT_REGEXP = new RegExp(accountRegex)
 
 // helpers for the LOCK_PATH_NAME_REGEXP
 const prefix = '[a-z0-9]+'
-const urlEncodedRedirectUrl = '[^#?]+'
-const userAccount = accountRegex
+const urlEncodedUrl = '[^#?]+'
+const urlEncodedRedirectUrl = urlEncodedUrl
 const lockAddress = accountRegex
 
 /**
@@ -61,10 +61,7 @@ const lockAddress = accountRegex
  * You should not use this directly, instead use the utils/routes.js lockRoute function
  */
 export const LOCK_PATH_NAME_REGEXP = new RegExp(
-  `(?:/(${prefix}))?/(${lockAddress})` +
-    // either "/urlEncodedRedirectUrl/#account" or just "#account" and these are all optional
-    // note that "/#account" as in "/paywall/<lockaddress>/#<useraccount>" is also matched
-    `(?:/(${urlEncodedRedirectUrl}))?(?:/?#(${userAccount}))?`
+  `(?:/(${prefix}))?/(${lockAddress})(?:/(${urlEncodedRedirectUrl})/?)?`
 )
 
 export const PAGE_DESCRIPTION =
--- a/paywall/src/constants.js
+++ b/paywall/src/constants.js
@@ -42,26 +42,24 @@ const accountRegex = '0x[a-fA-F0-9]{40}'
  */
 export const ACCOUNT_REGEXP = new RegExp(accountRegex)
 
-// helpers for the LOCK_PATH_NAME_REGEXP
+// private helpers for the LOCK_PATH_NAME_REGEXP
 const prefix = '[a-z0-9]+'
 const urlEncodedUrl = '[^#?]+'
-const urlEncodedRedirectUrl = urlEncodedUrl
 const lockAddress = accountRegex
 
 /**
  * This regexp matches several important parameters passed in the url for the demo and paywall pages.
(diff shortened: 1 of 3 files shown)


paywall/src/utils/routes.js
```diff
@@ -1,11 +1,12 @@
-import { LOCK_PATH_NAME_REGEXP } from '../constants'
+import { LOCK_PATH_NAME_REGEXP, ACCOUNT_REGEXP } from '../constants'
 
 /**
  * Returns a hash of lockAddress and prefix based on a path.
  * @param {*} path
  */
 export const lockRoute = path => {
-  const match = path.match(LOCK_PATH_NAME_REGEXP)
+  const url = new URL(path, 'http://dummy.com')
```
akeem: Is this url here as a placeholder as part of the refactor step?

cellog: I'll add a comment, this is because path is a relative path, and so `new URL` throws with "invalid URL" unless you pass in a base path. Since lockRoute doesn't use base path, I'm using a dummy to prevent the error

paywall/src/constants.js
```diff
@@ -44,8 +44,8 @@ export const ACCOUNT_REGEXP = new RegExp(accountRegex)
 
 // helpers for the LOCK_PATH_NAME_REGEXP
 const prefix = '[a-z0-9]+'
-const urlEncodedRedirectUrl = '[^#?]+'
-const userAccount = accountRegex
+const urlEncodedUrl = '[^#?]+'
```
julien51: Consistency alert! Use upper case for CONSTANTS!
Also maybe something that says it's a regex: URL_ENCODED_REGEX?

paywall/src/constants.js
```diff
@@ -44,8 +44,8 @@ export const ACCOUNT_REGEXP = new RegExp(accountRegex)
...
+const urlEncodedRedirectUrl = urlEncodedUrl
```
julien51: I think you do not need that? The constant name should describe what it is, not what it can be used for?

paywall/src/utils/routes.js
julien51: Let's use at least a URL we like: https://unlock-protocol.com and with a comment  to explain the purpose!

paywall/src/utils/routes.js
```diff
@@ -16,17 +17,26 @@ export const lockRoute = path => {
     }
   }
 
+  let account = url.hash && url.hash.substring(1)
+  const matchAccount = account.match(ACCOUNT_REGEXP)
+  account = matchAccount ? matchAccount[0] : undefined
```
julien51: `account = matchAccount ? matchAccount[0] : undefined` maybe we can just use `account = matchAccount[0]` (I know `null` is not `undefined` but then you need to explain why you need `undefined` because future-us will not know what you're doing here.

paywall/src/utils/routes.js
```diff
@@ -16,17 +17,26 @@ export const lockRoute = path => {
...
+
   return {
     lockAddress: match[2],
     prefix: match[1],
     redirect: match[3] && decodeURIComponent(match[3]),
-    account: match[4],
+    account,
+    origin: url.searchParams.has('origin')
+      ? url.searchParams.get('origin')
+      : undefined,
```
julien51: I think you can skip the `: undefined` part?

paywall/src/constants.js
cellog: per our conversation, as these are private, I'm going to keep them lower-cased, but add "private" to the comment above them to help clarify that.

paywall/src/constants.js
cellog: 👍

paywall/src/utils/routes.js
cellog: this is actually an attempt to unify the output from `lockRoute`, but I think I will make the lines below return `null` instead if falsy instead of the current `undefined`. That will require changing the `baseRoute` in tests, but I think it is more consistent and less weird. Because I agree, explicitly using `undefined` smells bad :).

paywall/src/utils/routes.js
cellog: decided on https://paywall.unlock-protocol.com

cellog:
this is ready for final pass, @julien51 PTAL

Issue #492, opened 2018-11-07
Get right gas amount


julien51:
When issuing transactions the web3Service needs to provide a maximum amount of gas required for the transaction to succeed. Right now the values are hard coded. We could actually provide that number "automatically".

cc @nfurfaro

nfurfaro:
@julien51 I just had a look at `Web3Service.js` and will look into this.

iworkforthem:
@julien51 is this issue fixed? i can push the fix to services/web3Service.js

julien51:
@iworkforthem I am so sorry for the delay. I missed your message earlier.
I believe this is not fixed yet.

iworkforthem:
@julien51 .. i created a function estimateGas(address) in services/web3Service.js...

1. Do I applied this to the purchaseKey & createLock methods?
2. For sendTransaction, I see that it take in gas as a parameter, I will skip this method, right?

cnasc:
What's the failure mode when the gas is underestimated? The transaction (likely) never goes through, right?

hardlydifficult:
Hi, not sure if someone had attempted this task previously - but turns out it was broken on Ganache.  The most recent Ganache updates claim to correct the problem.

So with this fix, gas can be underestimated only if there is another transaction mined between the time you estimated gas and when your transaction is mined.  And that other transaction changes something relevant to you.  A common example of this is when a value is set to 0, the gas for that state change is significantly cheaper than setting the value to anything else.

When gas is underestimated, it's possible for nodes to purge it from the mempool but it is more likely to be mined.  When that happens the user pays the max gas specified on the transaction, and that transaction fails / i.e. no state change.

hth

julien51:
This issue needs to be bumped up as it bit us today: https://github.com/unlock-protocol/unlock/pull/3399

@cellog suggested that this should be fairly simple to achieve (even if it requires a lot of test boilerplate cleanup):

> oh! if you want to use dynamic values, all we have to do is not pass in gas limit overrides in unlock-js
> it will require updating all the unit tests for the contract functions, because there will be a new call to eth_estimateGas (or whatever it's called)
> we might be able to squeak this out by simply adding to `walletServiceHelper.js`
but every one of the `v0/` `v01` etc. files will need to be modified
> so, simple, easy, but a lot of changing boilerplate

julien51:
Let's close this as stale for now.

NickCuso:
[...]

PR #1787, opened 2019-02-28
Add partialWithdrawFromLock method

### LINKED ISSUE #1769: Partial Withdrawal needs to be implemented in WalletService
In order to merge the partial withdrawal UI, it needs to do something. For that we need to give `WalletService` access to the smart contract method.

cnasc:
# Description

This PR connects the `walletService` with the smart contract implementation of partial withdrawal, providing a path for the frontend implementation of partial withdrawals.

Future PRs may remove the regular `withdrawFromLock` method, since we can implement a total withdrawal in terms of this new method.
Another future PR will remove the `mockUnlockContract` declarations from the test file, since they are vestigial.

# Issues

Fixes #1769 

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/services/walletService.test.js
+++ b/unlock-app/src/__tests__/services/walletService.test.js
@@ -743,11 +743,21 @@ describe('WalletService', () => {
         account = '0xdeadbeef'
       })
 
-      it('should invoke sendTransaction with the right params', () => {
-        expect.assertions(3)
+      it('should invoke sendTransaction with the right params', done => {
+        expect.assertions(4)
         const data = '' // mock abi data for partialWithdraw
-
-        walletService._sendTransaction = jest.fn()
+        const cb = (args, cb) => {
+          expect(args).toEqual({
+            to: lock,
+            from: account,
+            data,
+            gas: 1000000,
+            contract: LockContract,
+          })
+          expect(cb).toBeInstanceOf(Function)
+          done()
+        }
+        walletService._sendTransaction = cb
 
         const ContractClass = class {
           constructor(abi, address) {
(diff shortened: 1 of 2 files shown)


unlock-app/src/__tests__/services/walletService.test.js
```diff
@@ -734,6 +734,67 @@ describe('WalletService', () => {
...
+        account = '0xdeadbeef'
+      })
+
+      it('should invoke sendTransaction with the right params', () => {
+        expect.assertions(3)
+        const data = '' // mock abi data for partialWithdraw
+
+        walletService._sendTransaction = jest.fn()
+
+        const ContractClass = class {
```
cellog: nit: I prefer using something in the name that indicates this is a test helper, either `Mock` or `Test` or whatever makes sense.

unlock-app/src/__tests__/services/walletService.test.js
```diff
@@ -734,6 +734,67 @@ describe('WalletService', () => {
...
+        walletService.web3.eth.Contract = ContractClass
+
+        walletService.partialWithdrawFromLock(lock, account, '3')
+
+        expect(walletService._sendTransaction).toHaveBeenCalledWith(
+          {
+            to: lock,
+            from: account,
+            data,
+            gas: 1000000,
```
cellog: do we want this to fail if the gas cost changes? If so, keep it, otherwise I would use the value from `walletService.gasAmountConstants()`

cnasc: Good question! So we've actually added a line item for partial withdrawal costs in the constants (which is certainly wrong and may need to be adjusted, @hardlydifficult feel free to let me know if that needs a change). As for the hardcoded number, I did it based on the test for the full withdrawal which used a hardcoded number. I agree that the intent is not entirely clear there. I do plan on doing a refactoring pass on this file in a separate PR, maybe that can be included?

cellog: sure. Your call on process, but let's get that sorted eventually. This PR is good to go if it isn't a quick fix

hardlydifficult: `partialWithdraw` should require slightly more gas than `withdraw`, however they are very similar and should be much cheaper than the 1,000,000 limit you have ATM.

On a related note, #492 (use estimate gas) may change the solution here.

cnasc: Thanks for the feedback! I will do a follow-on test refactor PR to keep this one focused, and I'll keep my eye on #492

Issue #1720, opened 2019-02-26
Set up deployment for paywall sub-directory on netlify


cellog:
After #1649 is complete, we will need a place for those URLs to point to. The deployment scripts in `package.json` on the `paywall/` are all set up, and consistent with how `unlock-app` starts (build, start, etc.) so it just needs the deployment scripts to be updated properly and so on.

PR #1800, opened 2019-03-01
the deploy.sh script can now be used for any service

julien51:
This will be useful to deploy the paywall application.
Once the jobs have run, I'll double check that the unlock-app has been deployed as previously!

Refs https://github.com/unlock-protocol/unlock/issues/1720


julien51:
This will be useful to deploy the paywall app to netlify!

julien51:
I think we *could* but in the end I want each app to know how/where it is deployed and we should expect them to be deployed independently... so as much as I also do mot like the duplicate work I think it is the exception more than the rule here.

PR #1794, opened 2019-02-28
Refactored the signData method on walletService

julien51:
In the process of refactoring the way we send signed data to locksmith I refactored the signData method
on walletService. The logic is actually taken from the TypedDataSignature and signature utils.

- [X] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [X] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [X] My code follows the style guidelines of this project, including naming conventions
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [X] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

Issue #1776, opened 2019-02-28
Lock names


julien51:
Lock names shown in the dashboard are stored in locksmith. We should allow a lock creator to name  lock when created, update it and show the creator picked name on the dashboard.

julien51:
All implemented!

PR #1808, opened 2019-03-01
Refactored how lock names are signed

julien51:
In order to simplify how signatures are done, I refactored the signData method in a [previous PR](https://github.com/unlock-protocol/unlock/pull/1794).
This PR now uses this new signData method by using redux actions passing a message to be signed and a signed message.

After this a cleanup PR will remove all of the previous signature mechanisms

Refs # https://github.com/unlock-protocol/unlock/issues/1776

### DIFF
--- a/unlock-app/src/middlewares/walletMiddleware.js
+++ b/unlock-app/src/middlewares/walletMiddleware.js
@@ -20,13 +20,8 @@ import {
 import { TRANSACTION_TYPES, POLLING_INTERVAL } from '../constants' // TODO change POLLING_INTERVAL into ACCOUNT_POLLING_INTERVAL
 
 import WalletService from '../services/walletService'
-import { signatureError } from '../actions/signature'
-import { storeLockName } from '../actions/storage'
-import configure from '../config'
 import { NO_USER_ACCOUNT } from '../errors'
-import generateSignature from '../utils/signature'
-
-const config = configure()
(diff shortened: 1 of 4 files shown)


unlock-app/src/middlewares/walletMiddleware.js
```diff
@@ -129,29 +124,7 @@ export default function walletMiddleware({ getState, dispatch }) {
...
-                        token.result
-                      )
-                    )
-                  })
-                  .catch(error => {
-                    dispatch(signatureError(error))
-                  })
-              }
-            })
+          walletService.createLock(action.lock, getState().account.address)
```
akeem: I'll have to look at the current implementation of createLock but I think this might introduce a change in behavior we may not be intending. 

I previously stumbled upon this, a rejected promise is returned from createLock in a few scenarios, one of which is something like a person canceling the the lock creation in metamask. To cover this case we'll need to ensure that request to sign is dispatched only upon success

julien51: Good point. I think though that the request to sign now only happens when/if the lock has been created? (it's been moved to the `storageMiddleware` (see `action.type === CREATE_LOCK && action.lock.address`).

akeem: Ah that makes sense 👍

PR #1821, opened 2019-03-01
Refactor useListenForPostMessage

### LINKED ISSUE #1816: refactor useListenForPostMessage to be stricter about inputs and easier to use declaratively
the event data field from a `'message'` event (triggered by postMessage from the main window) can be simpler to use.

Example:
```js
const scrollPosition = useListenForPostMessage(window, { type: SCROLL_POSITION, validator: isPositiveNumber, defaultValue: 0 })
```

The value will only update if the posted message's data has this format: `{ type: SCROLL_POSITION, payload: <any positive integer> }` and will use `0` as the default value for the scrollPosition.

The rest of the code can use `scrollPosition` declaratively, and it will automatically re-render on receiving `postmessage` events

cellog:
# Description

This refactors `useListenForPostMessage` to do several new things:

1. restrict `event.data` values to be similar to a redux action. They are an object with indices `type` and `payload`.
2. the hook accepts as parameters the type, an optional validator for the payload, and an optional default value
3. it can now simply be used in a single line as in:

```js
const scrollPosition = useListenForPostMessage(window, {
  type: POST_MESSAGE_SCROLL_POSITION,
  validator: isPositiveInteger,
  default: 0
})
```

The value will trigger a re-render only if the received message is secure, and matches the expected message format, AND the value has actually changed. This will result in less jank for the scrolling stuff (probably, can't test because no iphone).

In addition, this test completely refactors the test for `useListenForPostMessage` to follow best practices. Instead of using `testHook`, the test creates a simple functional component that actually uses the hook, and uses its output to populate a text field that it asserts against. It also takes advantage of Jest's new ability to run a test against multiple inputs (https://github.com/sapegin/jest-cheat-sheet#data-driven-tests-jest-23) to greatly simplify the actual test file.

A small update to the paywall is necessary to move to the new format for scrollPosition, and this PR also takes advantage of the chance to make the postMessage more secure and only send it to the paywall's iframe.

# Issues

Fixes #1816 

# Checklist:


paywall/src/hooks/browser/useListenForPostMessage.js
```diff
@@ -2,21 +2,32 @@ import { useEffect, useState } from 'react'
...
     // ignore messages that do not come from our parent window
     if (event.source !== parent || event.origin !== origin) return
-    setData(event.data)
+    // data must be of shape { type: 'type', payload: <value> }
+    if (!event.data || !event.data.type) return
+    if (!event.data.hasOwnProperty('payload')) return
+    if (typeof event.data.type !== 'string') return
+    if (event.data.type !== type) return
+    // optional validator
+    if (!validator || (validator && validator(event.data.payload))) {
```
cnasc: `validator` is a function that returns truthy for data we want? Do you have an example of how that would be used in practice? Just curious, it seems like a good thing to have but I can't come up with anything concrete at the moment.

LG!

cellog: yep. The most immediate use is ensuring that scrollPosition is always a positive integer, so the existing `unlock-app/utils/validators.js` function `isPositiveInteger` can be used as a validator in that instance. That's why I chose this specific example in the description:

```js
const scrollPosition = useListenForPostMessage(window, {
  type: POST_MESSAGE_SCROLL_POSITION,
  validator: isPositiveInteger, // this function only validates numbers >= 0
  default: 0
})
```

julien51:
Ha! Do you mind using a better PR name please ;)

paywall/src/__tests__/hooks/browser/useListenForPostMessage.test.js
```diff
@@ -28,164 +32,145 @@ describe('useListenForPostMessage hook', () => {
     config = { isServer: false, isInIframe: true }
   })
 
-  it('sets up the event listener', () => {
-    expect.assertions(1)
-
-    rtl.testHook(() => useListenForPostMessage(fakeWindow), {
-      wrapper,
+  function MockListener({
+    /* eslint-disable react/prop-types */
```
julien51: Any reason you need to disable these?

cellog:
> Ha! Do you mind using a better PR name please ;)

Whatever do you mean? ;)

paywall/src/__tests__/hooks/browser/useListenForPostMessage.test.js
```diff
@@ -28,164 +32,145 @@ describe('useListenForPostMessage hook', () => {
...
-    )
-  })
+  function getListener() {
+    return fakeWindow.addEventListener.mock.calls[0][1]
+  }
 
-  it('ignores calls on the server', () => {
-    expect.assertions(1)
+  describe('does nothing in invalid contexts', () => {
+    it.each([
```
julien51: 👍

paywall/src/hooks/browser/useListenForPostMessage.js
```diff
@@ -2,21 +2,32 @@ import { useEffect, useState } from 'react'
 import useConfig from '../utils/useConfig'
 import { getRouteFromWindow } from '../../utils/routes'
 
-export default function useListenForPostmessage(window) {
+export default function useListenForPostmessage(
```
julien51: Do you mind adding JS doc comments to tell us a bit more about that this hook does?

paywall/src/hooks/browser/useListenForPostMessage.js
cnasc: D'oh, can't believe I skimmed right over that. Makes sense!

paywall/src/__tests__/hooks/browser/useListenForPostMessage.test.js
cellog: Just seems weird to define prop types for a test helper, but I can do this if you'd like

paywall/src/__tests__/hooks/browser/useListenForPostMessage.test.js
cellog: yeah isn't this cool? could definitely slim up some of the tests like CreatorLockForm validation

paywall/src/hooks/browser/useListenForPostMessage.js
cellog: done

cellog:
OK, this is ready for a 2nd look. I discovered a bug in the passing of origin to the iframe `postMessage` call, in that it was passing the full paywall URL, which is `https://unlock-protocol.com/paywall` currently. Postmessage expects an origin, which would be `https://unlock-protocol.com`, so I added in a call to `new URL(paywallUrl).origin` in order to retrieve the origin.

While working to test this change, I discovered that there was a gap in coverage for `scrollLoop` inside `paywall-builder/build.js`, and so I pulled it out of the `buildPaywall` function to make this task easier, and added 6 new tests to cover the use cases. I considered splitting this out, but it is a pretty minor change, and would require doing a pre-PR with a buggy, insecure buildPaywall, and then immediately this one would change almost as many lines to fix it.

In any case, PTAL on Monday (or earlier if you're revved up)

paywall/src/hooks/browser/useListenForPostMessage.js
julien51: wow, great job!

julien51: Nit: do you mind using camelCase for this? (I see the previous code was indeed using the wrong case, but let's fix it here! Thanks!

paywall/src/hooks/browser/useListenForPostMessage.js
```diff
@@ -2,21 +2,72 @@ import { useEffect, useState } from 'react'
...
   const { isInIframe, isServer } = useConfig()
   const parent = window && window.parent
-  const [data, setData] = useState()
+  const [data, setData] = useState(defaultValue)
+
   const saveData = event => {
+    // origin is passed in by the paywall, see paywall-builder/build.js
     const { origin } = getRouteFromWindow(window)
+    // **SECURITY CHECKS**
+
```
julien51: nit: remove line?

PR #1845, opened 2019-03-03
Lock names can be updated

### LINKED ISSUE #1771: Updating a lock name
We currently set the lock's name only when it is created. We should add the ability to update a lock's name after it's been updated.

julien51:
This is the final PR to allow the lock owner to update their lock's name.
This fixes #1771

### DIFF
--- a/unlock-app/src/components/creator/CreatorLock.js
+++ b/unlock-app/src/components/creator/CreatorLock.js
@@ -23,7 +23,7 @@ import {
   DoubleHeightCell,
   BalanceContainer,
 } from './LockStyles'
-import { updateKeyPrice } from '../../actions/lock'
+import { updateKeyPrice, updateLockName, updateLock } from '../../actions/lock'
 
 import { INFINITY } from '../../constants'
 
@@ -56,11 +56,6 @@ export class CreatorLock extends React.Component {
     this.toggleKeys = this.toggleKeys.bind(this)
   }
 
-  updateLock(lock) {
-    const { updateKeyPrice } = this.props
-    updateKeyPrice(lock.address, lock.keyPrice)
-  }
-
   toggleEmbedCode() {
     this.setState(previousState => ({
       showEmbedCode: !previousState.showEmbedCode,
@@ -76,7 +71,7 @@ export class CreatorLock extends React.Component {
   render() {
     // TODO add all-time balance to lock
 
-    const { lock } = this.props
+    const { lock, updateLock } = this.props
     const { showEmbedCode, showKeys, editing } = this.state
 
     if (editing) {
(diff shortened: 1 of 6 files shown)


julien51:
image

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -142,11 +137,28 @@ export class CreatorLock extends React.Component {
...
+        dispatch(updateKeyPrice(lock.address, newLock.keyPrice))
+      }
+
+      // If the name has changed
+      if (lock.name !== newLock.name) {
+        dispatch(updateLockName(lock.address, newLock.name))
+      }
+
+      // Reflect all changes
+      dispatch(updateLock(lock.address, newLock))
```
cellog: This is the only part of the PR I have questions about. Obviously, this duplicates the prior actions, in that nothing should have changed except price and/or name. Thus, perhaps we can be less redundant?

What if the first 2 actions were removed? We can perform these checks in the middleware by retrieving the current lock from the store, checking the name and price, and then calling the walletService methods for either change if necessary. The storage service can perform the same intercept and diff of name and use that as a trigger for signing the name change.

This has the benefit of reducing the number of actions dispatched, and simplifying the UI logic. If we choose to ever move away from redux this would also make that move slightly simpler. It also removes some unnecessary processor cycles, but that is pretty minimal. The drawback is it transfers some complexity out of the UI and into the middleware.

julien51: I thought about that a lot but I thing in the end it is a name issue and does not duplicate the prior actions.

The action `updateLock` has the prupose of indicating that the lock object has been modified, without requiring (necessarily) that the updates are propagated on the backend. For example, when loading past transactions, some of them may show a price change. In that case, we will want to to reflect that the lock object needs to be updated "locally", without (obviously!) triggering a transaction to update the lock on chain again.

I thought about performing the checks in the middleware, but in that case we would lose the context of why the object as changed : is that something triggered by the user (for which we would trigger and transaction) or just to reflect a change in chain which was triggered in some other place... Keeping distinct actions allows us to keep them "close" to where the action has been triggered and then have more context about it.

julien51:
Adding a few more reviewers since @cellog is out mañana!

unlock-app/src/components/creator/CreatorLock.js
cellog: ok, thanks, this makes more sense. I think ultimately we will need a different architecture to make this work. I'm going to approve based on your comment.

One thing I did in another app was to have 2 equivalent actions, one for CREATE_ITEM and anotehr for LOAD_ITEM which were handled the same way, but the CREATE_ITEM one triggered async stuff, and the LOAD_ITEM was for loading. This might be a way forward here too, but that's for a future refactor

PR #1846, opened 2019-03-03
using a hardcoded address for the Unlock smart contract

julien51:
# Description

Extracting the lock address from the artifact is creating a lot of friction when in practice the Unlock smart contract address is consistently the same.

This draft is an effort to see if we can hard code the URL for all environments.

# Checklist:

### DIFF
--- a/unlock-app/src/config.js
+++ b/unlock-app/src/config.js
@@ -71,7 +71,8 @@ export default function configure(
   let requiredNetwork = 'Dev'
   let requiredNetworkId = 1984
   let requiredConfirmations = 12
-  let unlockAddress = ''
+  // Unlock address by default (smart contract deployments yield the same address on a "clean" node as long as long as the migration script runs in the same order)
+  let unlockAddress = '0x8DE3f95E2efd3B9704ccb0d0925EC951bC78cb8B'
   let services = {}
   let supportedProviders = []
   let paywallUrl = runtimeConfig.paywallUrl || 'http://localhost:3000/paywall'
@@ -141,7 +142,7 @@ export default function configure(
       'https://staging.unlock-protocol.com/static/paywall.min.js'
 
     // Address for the Unlock smart contract
-    unlockAddress = '0xd8c88be5e8eb88e38e6ff5ce186d764676012b0b'
+    unlockAddress = '0xD8C88BE5e8EB88E38E6ff5cE186d764676012B0b'
 
     // rinkeby block time is roughly same as main net
     blockTime = 8000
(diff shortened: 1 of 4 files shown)


unlock-app/src/services/web3Service.js
```diff
@@ -110,18 +108,9 @@ export default class Web3Service extends EventEmitter {
...
-      )
-    } else if (UnlockContract.networks[requiredNetworkId]) {
-      // If we do not have an address from config let's use the artifact files
-      this.unlockContractAddress = Web3Utils.toChecksumAddress(
-        UnlockContract.networks[requiredNetworkId].address
-      )
-    } else {
-      return this.emit('error', new Error(NON_DEPLOYED_CONTRACT))
-    }
+    this.unlockContractAddress = Web3Utils.toChecksumAddress(
```
cellog: Maybe we can move this to config.js since it is now static?

unlock-app/src/config.js
```diff
@@ -71,7 +71,7 @@ export default function configure(
   let requiredNetwork = 'Dev'
   let requiredNetworkId = 1984
   let requiredConfirmations = 12
-  let unlockAddress = ''
+  let unlockAddress = '0x8DE3f95E2efd3B9704ccb0d0925EC951bC78cb8B' // Unlock address by default (smart contract deployments yield the same address on a "clean" node as long as long as the migration script runs in the same order)
```
cellog: Put the comment on the previous line for line breaks?

unlock-app/src/services/web3Service.js
julien51: That's an option. I think we can even put the checksumed address inside the config so that we do not even have to checksum it.

unlock-app/src/config.js
julien51: sure (note that this is a draft PR, just to get feedback ;P)

unlock-app/src/config.js
```diff
@@ -71,7 +71,8 @@ export default function configure(
   let requiredNetwork = 'Dev'
   let requiredNetworkId = 1984
   let requiredConfirmations = 12
-  let unlockAddress = ''
+  // Unlock address by default (smart contract deployments yield the same address on a "clean" node as long as long as the migration script runs in the same order)
```
cnasc: I think this is a great step overall, just curious whether there's any possibility of the migration scripts running in a different order and messing things up

julien51: In dev, yes there is always a risk and I actually think it *will* happen when/if we change the way we deploy the locks but that should be fairly rare.

julien51: In prod and staging the contract have been deployed and its address will never change.

julien51:
Ok, I think feedback is good and since this is quite simple, I'll actually send a full PR for this next!  Thanks @cellog @cnasc !

PR #1851, opened 2019-03-04
Refactor hooks to use window passed down through context.

### LINKED ISSUE #1828: Refactor dependency injection of window object in paywall hooks
All of the browser paywall hooks use `window`, and accept it as a parameter. This can lead to extremely wonky-looking code.

There are a couple of options for solving this. The most obvious is to mock the window object, something like:

```js
beforeEach(() => {
  global.window = { ... }
})
```

I have tried this in other projects, and it occasionally led to race conditions where unrelated tests were broken by the mocks. This loser posted a question about it on Twitter: https://twitter.com/GregCello/status/1101306198538752000

and the commenter seemed very confident in this approach.

Another alternative is to pass the window global down using context, something like:

```js
import React, { createContext, useContext } from 'react'

export const WindowContext = createContext()

export default function useWindow() {
  return useContext(WindowContext)
}
```

and then in the top-level file:

```js
import { WindowContext } from '../path/to/hooks/useWindow.js'

// ...

return (
  <WindowContext.Provider value={window}>...</WindowContext.Provider>
)
```

Testing is less magical (no mocking needed), at the expense of slightly more boilerplate.  Every hook that uses it would need to add 1 line:

```js
const window = useWindow()
```

But then we have more isolation of the code. Would love to get feedback to make a decision on this one, @julien51 @akeem and @cnasc

cellog:
# Description

This PR refactors the dependency injection in the paywall's hooks.

Currently, the DI is done through function parameters, so every hook is required to explicitly load window as in `useLocalStorage(window, '__unlock__account__')`

This is clunky, and messy, making the code harder to reason through.

The solution implemented here is the one mentioned at the end of the comment thread in #1828, and uses a top-level context to pass the `window` object down through the app. As such, it is actually passed down in `pages/_app.js`. The context is then consumed in hooks by the dead simple `useWindow` hook (a 1-liner that pulls in window using the React `useContext` hook).

This way, the hooks do no need to request the window object from the parent, and testing is simpler.

This PR also takes the opportunity to move the hook tests to the recommended format, which is to have a mock component use the hook and set its value to be inspected directly. This is the format already in use by the `useListenForPostMessage` hook.

Questions to resolve:

1. are the comments clear enough? Do more need to be added in either the tests or the hooks?
2. I can't think of a 2nd question, but I'm sure you can :)

# Issues

Fixes #1828 

# Checklist:


cnasc:
I think this is clear enough. It's not terribly different from our other uses of contexts, right? Just in this case we use it with hooks, and this also allows us to easily mock `window` for tests.

Issue #1856, opened 2019-03-04
set up CHROMATIC_APP_CODE for CI


cellog:
to run chromatic in CI, we need to set the `CHROMATIC_APP_CODE` env variable to our chromaticqa app code, found at https://www.chromaticqa.com/manage?appId=5c7d369cd530c00020ea8e0e&branch=master

julien51:
To do that I think I need a bit more context about when that code is running.
Basically our tests run in docker *inside* CI, and env variables from the "host" are not passed inside the docker processes. Which means we have to manually do that.

Is that code secret? What are the processes which need it?

julien51:
If it is not secret then maybe it is easier to have it inside version control?

cellog:
it's secret. With it, anyone can use our plan to run their chromaticqa tests :)

cellog:
the command that needs it is found in #1857 `npm run chromatic`

julien51:
Just to clarify, our `unlock-app` needs it to run its tests then, right?

cellog:
Yes, in order to get comparisons between PRs and master of the visual changes in storybook stories for `unlock-app` by using chromaticqa, we will need this. Let me know if there is any other assumption I'm not voicing! :)

julien51:
Once merged, I believe the env variable `CHROMATIC_APP_CODE` will be available do use from inside the script file for `unlock-app`!

cellog:
ok, I'll modify the draft PR to see if it runs a build :)

julien51:
Added both to Circle and Travis! Closing now.

Issue #870, opened 2018-12-31
Visual diffing for storybook changes


cellog:
**Is your feature request related to a problem? Please describe.**
Recently, we had a PR that broke storybook, and we didn't catch it. If we were doing visual diffing of storybook run screenshots, this would be caught. In addition, when changes are made to visual display of a storybook component, this would provide a visual diff as well as the snapshot diff.

**Describe the solution you'd like**
The most promising solution I found is called loki:

https://loki.js.org/

The setup works best with chrome in docker, and requires graphicsmagick. Fortunately:

https://hub.docker.com/r/jcsousa/nodejs-alpine-graphicsmagick/

I think we can make a docker image based on this image that will merge with chrome docker and run our storybooks, then we can make this a commit hook like test failures.

**Describe alternatives you've considered**
I considered using some kind of PR web hook, but there is no existing service and so we would have to make one and ... no.

**Additional context**
This is early stage. comments from @julien51 @benwerd @akeem @smombartz ?

cellog:
That makes sense. I think this tool is great, and we should consider it. For this issue, which is just verifying that storybook is actually working, maybe I am thinking too large.

Perhaps we can add 1 new unit test that runs our storybook in a sandbox, gets the html of the first page and creates a snapshot we can compare? This would verify it works

julien51:
I think you're right. I am still not fully sure I understand how the snapshot could be compared if storybook did not run?

cellog:
Could we leverage the integration test infrastructure to run storybook, and grab the html from puppeteer?

cellog:
OK, Here is the setup I think we should explore: https://medium.com/bleeding-edge/screenshot-testing-with-react-and-storybook-19ab7e49ec92

storybook-chrome-screenshot and reg-suit together will give us a way to create screenshots and see the differences with a pixel diff

julien51:
See https://github.com/unlock-protocol/unlock/issues/821 for options of tools.

cellog:
implemented with chromatic

PR #1857, opened 2019-03-04
add chromatic to config

cellog:
# Description

This PR shows how we would enable chromatic. It does not set it up in CI yet, as we need to handle #1856 first.

To use, run with:

```
CHROMATIC_APP_CODE=<our app code> npm run chromatic
```

Replace `<our app code>` with the app code from https://chromaticqa.com (log in with github)

It only enables chromaticqa in unlock-app for now

# Issues

Refs #870 
Refs #1856 

# Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,19 @@ jobs:
           name: Creator Dashboard tests
           command: scripts/tests.sh unlock-app
 
+  unlock-app-chromatic:
+    machine: true
+    environment:
+      DOCKER_REPOSITORY: unlockprotocol
+    steps:
+      - checkout
+      - run:
+          name: Pull Unlock Image
+          command: scripts/pull-image.sh unlock "$CIRCLE_SHA1"
+      - run:
+          name: Creator Dashboard stories
+          command: scripts/storybook.sh unlock-app
+
   smart-contracts-tests:
     machine: true
     environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,19 @@ jobs:
           name: Creator Dashboard stories
           command: scripts/storybook.sh unlock-app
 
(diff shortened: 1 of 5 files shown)


unlock-app/.storybook/config.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react'
 import { configure, addDecorator } from '@storybook/react'
+import 'storybook-chromatic'
 import StoryRouter from 'storybook-react-router'
+
```
julien51: nit: whitespace?

cellog: I have gotten in the habit of putting a space between node_modules and internal imports. I'll remove this

cellog:
> Split the PR to add the dep ;)

I will, this is a draft, just want to make sure it works and is what we are looking for first. Would appreciate another look, when you have the time

.circleci/config.yml
```diff
@@ -40,6 +40,32 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Pull Unlock Image
+          command: scripts/pull-image.sh unlock "$CIRCLE_SHA1"
+      - run:
+          name: Creator Dashboard stories
+          command: scripts/storybook.sh unlock-app
+
+  unlock-app-chromatic-master:
```
julien51: I think rather than have 2 jobs you should push the logic of checking if we are on master down to `scripts/storybook.sh`

julien51: You can pass the branch name using `$CIRCLE_BRANCH` and inside your script, if the branch is master, then you would append `--auto-accept-changes`.

This way we are as decouple as possible from the CI tool we use.

cellog: thanks! this is the piece I needed.

cellog:
@julien51 PTAL

PR #1858, opened 2019-03-04
fix incorrect context usage

cellog:
# Description

There was a bug in the usage of context in _app.js, this fixes the bug.

There are 2 distinct bugs.

1) on the server, window is not defined, and errors out
2) Context must always be referenced as the Provider (the default of the context is the consumer, which is undocumented)

# Issues

Fixes #
Refs #1828 

# Checklist:

### DIFF
--- a/paywall/pages/_app.js
+++ b/paywall/pages/_app.js
@@ -101,14 +101,14 @@ The Unlock team
       <Container>
         <GlobalStyle />
         <Provider store={store}>
-          <WindowContext value={window}>
+          <WindowContext.Provider value={global.window}>
             <WalletCheckOverlay />
             <ConnectedRouter history={history}>
               <ConfigProvider value={config}>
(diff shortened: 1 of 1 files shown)


cnasc:
I will ditto Akeem's comment. And just to add to the discussion, is this something TypeScript could have caught?

cellog:
yes TS would have caught bug (2)

cellog:
regarding regression tests, this one is very tricky, because none of the config is mockable since we need to actually test the behavior of the actual page.

I was getting NON_DEPLOYED_CONTRACT_ERROR. I will try again now that this is ancient history!

cellog:
I'm going to get a regression test in here before I merge

cellog:
gave up, 2 hours of banging my head against the wall of `_app.js` and all I got was despair and gnashing of teeth in the outer darkness. `_app.js` is untestable in its current form.

PR #1863, opened 2019-03-04
Refusing to build the application when some env variables are missing

### LINKED ISSUE #1859: Missing Locksmith config on CircleCI
Naming locks does not currently work because we're missing the required configuration environment variable when starting the application.

Not only should we set the variables but, more importantly, we should fail to "deploy" the application if required environment variables are missing. This will prevent us from pushing non-fnctionning applications.

julien51:
This is a way to ensure that the app is only starting if all the required configuration are available.

Fixes #1859

### DIFF
--- a/unlock-app/next.config.js
+++ b/unlock-app/next.config.js
@@ -1,3 +1,5 @@
+/* eslint no-console: 0 */
+
 const fs = require('fs')
 const { join } = require('path')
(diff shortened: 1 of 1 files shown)


julien51:
Note: the build should fail until we add the required env variables to the CI setup!

Issue #1805, opened 2019-03-01
Add button to navigate to the creator dashboard log


julien51:
Our creator log dashboard is currently not linked from the creator dashboard which makes it impossible to view for users ;)

Issue #1688, opened 2019-02-25
Make transaction log reachable from navigation


cnasc:
The log page is in the app, but nothing points to it yet!

This feature should implement the new header icons found in this Figma: https://www.figma.com/file/gmhVq6Pqeu2gH7e3FC2lr0ax/Unlock-Dashboard?node-id=0%3A67

I think there was some question about the icons, @smombartz I assume the 2 new icons map as follows:

- Home icon -> Dashboard
- List icon -> Log page

Is that correct?

cnasc:
This is blocked until #1694 is fixed

smombartz:
yes correct! unless @julien51 still opposes the home icon thats good to go from my side.

cellog:
You can use https://nextjs.org/docs/#using-a-higher-order-component to access the route information you need without waiting for #1694

cnasc:
Nice, looks like that's a lot easier than keeping Redux in sync with the router, which is dubiously necessary in the first place

cnasc:
Another question: when do these new navigation buttons show up? I think we do not want them on the homepage, since they would allow someone to bypass the terms of service. Same for the non-app pages? In other words, they should only show up on the dashboard or the transaction log?

cellog:
That depends, and perhaps @benwerd can chime in here, but my understanding is that we don't need people to agree to terms unless they are creating locks. Viewing info is free, legally and financially.

If that's true it opens up the possibility of linking from the homepage

smombartz:
i would say only dashboard. thats once you enter into the app. the landing page should be informative.

benwerd:
Sorry for chiming in late - we don't need people to agree to terms for landing pages, but all application pages need to be after terms have been agreed to. It's just not about the activity; it's about any usage, including reading.

PR #1865, opened 2019-03-04
Router is now connected to Redux

### LINKED ISSUE #1694: Router in redux does not update when you change pages
**Describe the bug**
The router in redux never updates after it is first set. So if you just load the root url, `pathname` will be `/`. Visiting any page after that, it stays on `/`.

**To Reproduce**
1. Load the app
2. Take note of the router state
3. Navigate elsewhere without actually reloading the page (that is, click a link, don't modify the url and hit enter)
4. See that router state has not changed

**Expected behavior**
Router information in the state will update to reflect page navigation.

**Screenshots**
My Redux state on the homepage
image

My Redux state after having moved from the homepage to the dashboard via the green button on the homepage
image

cnasc:
# Description

This PR connects Redux with the Next router. Previously, there was a router object in the state that did not track movement between pages--this is because we used `connected-react-router` despite our usage of Next router. I changed the dependency to `connected-next-router`, which handles updating the router state automagically for anything wrapped in a `next/Link` component.

Additionally, changes were made to allow the initial state of the app to include the page you directly queried.

# Issues

Fixes #1694 
Refs #1688 #1805 

# Checklist:

### DIFF
--- a/unlock-app/package-lock.json
+++ b/unlock-app/package-lock.json
@@ -5530,6 +5530,15 @@
       "resolved": "https://registry.npmjs.org/connected-next-router/-/connected-next-router-1.0.1.tgz",
       "integrity": "sha512-R4GOqz6XvRfh1X5PS2XY+lsvxiMlv87v5P8h8JBxrPlBXe0tmHoGXPEND16Mv45ta7ciuG1klyH5eJHmtKb0Mg=="
     },
+    "connected-react-router": {
+      "version": "6.1.0",
+      "resolved": "https://registry.npmjs.org/connected-react-router/-/connected-react-router-6.1.0.tgz",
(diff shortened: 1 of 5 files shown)


cnasc:
So it's:

1. First PR adds `connected-next-router` and hooks it up
2. Second PR removes `connected-react-router`?

cellog:
> So it's:
> 
> 1. First PR adds `connected-next-router` and hooks it up
> 2. Second PR removes `connected-react-router`?

Yep, ping Julien if you want the rationale from the source, it has to do with separating config changes from code changes, in case we need to revert, it's easier (in case future commits depend on that config change)

cnasc:
Yup, makes total sense just wanted to make sure I split it correctly the first time and not have to redo it!

cnasc:
This is ready for another look--I'll follow up shortly with a PR to remove `connected-react-router`

PR #1870, opened 2019-03-04
KEY_PURCHASE transactions now have the lock address

### LINKED ISSUE #1834: Missing lock address on /log for some transactions
The `/log` is very useful! Unfortunately it seems to have a small bug when displaying some transactions.

As you can see some lock addresses are empty, but the link is actually there with `undefined` when pointing to etherscan: https://rinkeby.etherscan.io/address/undefined

cnasc:
# Description

This PR is a second, hopefully cleaner, pass at what #1707 tried to do. When we pull in transactions, any key purchase should have a field that identifies which lock it is associated with.

There is some additional work needed on the log page, however:
#1707 also tried to fix the issue of accidentally mutating the redux state. This PR does not, so that fix will have to come in a later PR.

This is now ready for review, and is much smaller and simpler than the draft PR. Still could use some advice on testing; it actually looks like the handlers aren't directly tested currently.

image

# Issues

Fixes #1834 #1704 
Refs #1707

# Checklist:

### DIFF
--- a/unlock-app/src/services/web3Service.js
+++ b/unlock-app/src/services/web3Service.js
@@ -111,7 +111,7 @@ export default class Web3Service extends EventEmitter {
           balance: '0', // Must be expressed in Eth!
         })
       },
-      purchaseFor: async (transactionHash, contractAddress, params) => {
+      purchaseFor: async (transactionHash, contractAddress) => {
         this.emit('transaction.updated', transactionHash, {
           lock: contractAddress,
         })
--- a/unlock-app/src/services/web3Service.js
+++ b/unlock-app/src/services/web3Service.js
@@ -54,6 +54,7 @@ export default class Web3Service extends EventEmitter {
         const owner = args._to
         this.emit('transaction.updated', transactionHash, {
           key: keyId(contractAddress, owner),
+          lock: contractAddress,
         })
         return this.emit('key.saved', keyId(contractAddress, owner), {
           lock: contractAddress,
--- a/unlock-app/src/services/web3Service.js
+++ b/unlock-app/src/services/web3Service.js
@@ -110,6 +110,11 @@ export default class Web3Service extends EventEmitter {
           balance: '0', // Must be expressed in Eth!
         })
       },
+      purchaseFor: async (transactionHash, contractAddress, params) => {
+        this.emit('transaction.updated', transactionHash, {
+          lock: contractAddress,
+        })
+      },
     }
   }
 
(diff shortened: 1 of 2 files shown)


unlock-app/src/services/web3Service.js
```diff
@@ -319,12 +319,18 @@ export default class Web3Service extends EventEmitter {
   parseTransactionFromInput(transactionHash, contract, input, contractAddress) {
     const transactionType = this.getTransactionType(contract, input)
 
-    this.emit('transaction.updated', transactionHash, {
+    let update = {
```
cellog: const is fine here

unlock-app/src/services/web3Service.js
```diff
@@ -462,14 +468,19 @@ export default class Web3Service extends EventEmitter {
       if (blockNumber - blockTransaction.blockNumber < requiredConfirmations) {
         this._watchTransaction(transactionHash)
       }
-
       // The transaction was mined, so we should have a receipt for it
-      this.emit('transaction.updated', transactionHash, {
+      let update = {
```
cellog: and here too

unlock-app/src/services/web3Service.js
```diff
@@ -319,12 +319,18 @@ export default class Web3Service extends EventEmitter {
...
+    let update = {
       status: 'pending',
       type: transactionType,
       confirmations: 0,
       blockNumber: Number.MAX_SAFE_INTEGER, // Asign the largest block number for sorting purposes
-    })
+    }
+    // The lock a key is associated with is the same as the contract the key
+    // purchase transaction was sent to
+    if (transactionType === TRANSACTION_TYPES.KEY_PURCHASE) {
```
julien51: I would put that logic in the `handler` called at line 352. The code here is not specific to any type of transaction while the handler actually exists for this specific purpose ;)

unlock-app/src/services/web3Service.js
```diff
@@ -462,14 +468,19 @@ export default class Web3Service extends EventEmitter {
...
-      this.emit('transaction.updated', transactionHash, {
+      let update = {
         status: 'mined',
         type: transactionType,
         confirmations: blockNumber - blockTransaction.blockNumber,
         blockNumber: blockTransaction.blockNumber,
-      })
+      }
+      // The lock a key is associated with is the same as the contract the key
+      // purchase transaction was sent to
```
julien51: Same comment as above ;)

unlock-app/src/services/web3Service.js
cnasc: Sure thing! Are these the relevant methods on the contract?
- `purchaseFor`
- `purchaseForFrom`

Issue #754, opened 2018-12-18
Decrease the polling frequency of currency conversion checking within the platform application


akeem:
**Is your feature request related to a problem? Please describe.**
Currently we are polling a service provider to ensure that the platform application has up to date prices to support the pricing of keys (eth/dollars) etc.

**Describe the solution you'd like**
1. Reduce the frequency of fetching of data from the third party API
2. Move the initial fetch to the backend
  a. Start to proxy platform request to the backend
  b. Have the client poll the third party API if there is a need to fail over.

julien51:
I think we can eventually cache the API call inside locksmith?

PR #1871, opened 2019-03-04
Extracts Currency Lookup Service from Middleware

akeem:
# Description

As part of abstracting price detail from the portions of
the creator dashboard the middleware has been split from the
mechanics of how to retrieve price data.

This should ideally give us confidence to receive additional price pairs
from multiple sources without risk of breaking the dashboard.

# Issues

Fixes #
Refs #754

# Checklist:

### DIFF
--- a/smart-contracts/zos.json
+++ b/smart-contracts/zos.json
@@ -1,5 +1,5 @@
 {
-  "zosversion": "2",
+  "zosversion": "2.2",
   "name": "unlock-protocol",
   "version": "0.1.0",
   "contracts": {
--- a/smart-contracts/zos.json
+++ b/smart-contracts/zos.json
@@ -1,5 +1,5 @@
 {
-  "zosversion": "2.2",
+  "zosversion": "2",
   "name": "unlock-protocol",
   "version": "0.1.0",
   "contracts": {
--- a/unlock-app/src/__tests__/middlewares/currencyConversionMiddleware.test.js
+++ b/unlock-app/src/__tests__/middlewares/currencyConversionMiddleware.test.js
@@ -1,19 +1,16 @@
-import mockAxios from 'jest-mock-axios'
 import { setConversionRate } from '../../actions/currencyConvert'
 
-afterEach(() => mockAxios.reset())
+jest.mock('../../services/currencyLookupService', () => () => ({
+  lookupPrice: jest
+    .fn()
+    .mockResolvedValueOnce({ currency: 'USD', amount: '195.99' })
+    .mockResolvedValueOnce({ currency: 'USD', amount: '198.20' })
+    .mockResolvedValueOnce({ currency: 'USD', amount: '195.99' }),
+}))
 
 describe('Currency conversion service retrieval middleware', () => {
-  const response1 = {
-    data: { base: 'ETH', currency: 'USD', amount: '195.99' },
-  }
-  const response2 = {
-    data: { base: 'ETH', currency: 'USD', amount: '198.20' },
-  }
-  const APIaddress = 'https://api.coinbase.com/v2/prices/ETH-USD/buy'
-
-  it('service called, action dispatched to set currency conversion rate', () => {
-    expect.assertions(4)
+  it('service called, action dispatched to set currency conversion rate', async () => {
+    expect.assertions(2)
     jest.useFakeTimers()
     const middleware = require('../../middlewares/currencyConversionMiddleware')
       .default
@@ -24,29 +21,22 @@ describe('Currency conversion service retrieval middleware', () => {
       },
     }
 
-    middleware(store)
+    await middleware(store)
 
-    expect(mockAxios.get).toHaveBeenCalledWith(APIaddress)
-    mockAxios.mockResponse({ data: response1 })
     expect(store.dispatch).toHaveBeenCalledWith(
       setConversionRate('USD', '195.99')
     )
 
     store.dispatch = jest.fn() // reset
-    mockAxios.reset()
-
-    jest.advanceTimersByTime(10000) // test the setInterval
-
-    expect(mockAxios.get).toHaveBeenCalledWith(APIaddress)
-    mockAxios.mockResponse({ data: response2 })
+    await jest.advanceTimersByTime(10000) // test the setInterval
 
     expect(store.dispatch).toHaveBeenCalledWith(
       setConversionRate('USD', '198.20')
     )
   })
 
-  it("service called, values are the same, so don't dispatch", () => {
-    expect.assertions(2)
+  it("service called, values are the same, so don't dispatch", async () => {
+    expect.assertions(1)
     jest.useFakeTimers()
     const middleware = require('../../middlewares/currencyConversionMiddleware')
       .default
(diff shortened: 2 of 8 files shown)


unlock-app/src/__tests__/services/currencyLookupService.test.js
```diff
@@ -0,0 +1,53 @@
...
+        expect.assertions(1)
+        currencyLookupService._handleCoinbaseFetch = jest.fn()
+        currencyLookupService.lookupPrice('ETH', 'GBP')
+        expect(currencyLookupService._handleCoinbaseFetch).toBeCalledWith(
+          'ETH',
+          'GBP'
+        )
+      })
+    })
+    describe('when configured for anything that isnt coinbase', () => {
```
julien51: nit: space above ;)

unlock-app/src/__tests__/services/currencyLookupService.test.js
```diff
@@ -0,0 +1,53 @@
...
+      it('return a rejected promise', () => {
+        expect.assertions(1)
+        currencyLookupService = new CurrencyLookupService(
+          'https://somelocksmithinstallation'
+        )
+
```
julien51: nit: so space above ;p

unlock-app/src/services/currencyLookupService.js
```diff
@@ -0,0 +1,30 @@
...
+  _constructCoinbaseLookupURI(base, currency) {
+    return `https://api.coinbase.com/v2/prices/${base.toUpperCase()}-${currency.toUpperCase()}/buy`
+  }
+
+  async _handleCoinbaseFetch(base, currency) {
+    let result = await axios.get(
+      this._constructCoinbaseLookupURI(base, currency)
+    )
+
+    return Promise.resolve({
```
cellog: I think you can just return the value and it uses it as the return in .then?

unlock-app/src/middlewares/currencyConversionMiddleware.js
```diff
@@ -1,33 +1,30 @@
 // TODO convert to fetch instead of using axios when it is supported well enough
```
julien51: Remove the TODO please

unlock-app/src/middlewares/currencyConversionMiddleware.js
```diff
@@ -1,33 +1,30 @@
...
   setInterval(() => {
-    axios.get(currencyPriceLookupURI).then(info => {
-      const {
-        data: {
-          data: { currency, amount },
-        },
-      } = info
-      const current = store.getState().currency[currency]
-      if (current === +amount) return
+    currencyLookupService.lookupPrice('ETH', 'USD').then(info => {
```
julien51: It looks like this call is duplicate with the above? Any way to refactor to avoid that?

unlock-app/src/middlewares/currencyConversionMiddleware.js
```diff
@@ -1,33 +1,30 @@
 // TODO convert to fetch instead of using axios when it is supported well enough
 /* eslint promise/prefer-await-to-then: 0 */
 
-import axios from 'axios'
 import { setConversionRate } from '../actions/currencyConvert'
 import configure from '../config'
+import CurrencyLookupService from '../services/currencyLookupService'
 
 const { services } = configure(global)
 const currencyPriceLookupURI = services.currencyPriceLookup
```
cellog: I understand why you didn't change this, but perhaps it is better to put this data inside the new service? Perhaps as an exported constant, so that you don't need to use a regexp in the service. This should give us more certainty.

unlock-app/src/middlewares/currencyConversionMiddleware.js
```diff
@@ -1,33 +1,30 @@
...
+const retryInterval = 10000
```
julien51: Put that as a constant in constants.js?

cellog: can you add this to `constants.js`?

unlock-app/src/services/currencyLookupService.js
```diff
@@ -0,0 +1,30 @@
+import axios from 'axios'
+
+export default class CurrencyLookupService {
```
julien51: Do you mind adding comments/JSDoc in the functions

unlock-app/src/middlewares/currencyConversionMiddleware.js
akeem: Yeah, its a temporary shim to ensure that nothing breaks while I get the other side of things up and running. Although we only have provider currently, the regex in the code signals to readers of the code that this could change and please don't build around it as a given currently.

Should all be sorted by wednesday

unlock-app/src/services/currencyLookupService.js
akeem: I think I can have it implicitly return `result.data.data`, the major reason for doing it was to shape the data that the middleware should expect from the service basically a object with only these two properties: currency & amount

cellog: No I mean you don't need the Promise.resolve() around it, I like that you're explicit, sorry I was unclear!

unlock-app/src/middlewares/currencyConversionMiddleware.js
akeem: I'd like to hold off on this change at the moment, to see if a larger case for reuse emerges

julien51: works for me

smart-contracts/zos.json
```diff
@@ -1,5 +1,5 @@
 {
```
julien51: HA! Please remove this file from the PR?

unlock-app/src/__tests__/middlewares/currencyConversionMiddleware.test.js
```diff
@@ -1,19 +1,16 @@
-import mockAxios from 'jest-mock-axios'
```
julien51: Please note to get rid of that dep if we're not using it!

unlock-app/src/services/currencyLookupService.js
```diff
@@ -0,0 +1,59 @@
+import axios from 'axios'
+
+const URL = require('url')
+
+/**
+ *  Provides currency lookup abstraction for down stream consumers.
+ */
+export default class CurrencyLookupService {
+  constructor(uri) {
+    /* eslint-disable node/no-deprecated-api */
```
julien51: I am worried by this. Please add a comment to explain what's going on?

unlock-app/src/services/currencyLookupService.js
julien51: I agree we should not return `result.data.data` which is specific to Axios.
Now indeed like @cellog the Promise.resolve is not required?

unlock-app/src/__tests__/middlewares/currencyConversionMiddleware.test.js
akeem: noted, but I do believe this is being used elsewhere in the code base

unlock-app/src/services/currencyLookupService.js
akeem: Our lint rules currently flag this as an deprecated method for node 11. As we aren't using this version it is added noise and is blocking the build. I'll add a note with the understand that this is should be resolved in the future and this removed

PR #1876, opened 2019-03-05
WIP: Update ZOS migrations

hardlydifficult:
# Description

Goals:
 - Remove force `--reset` on deploy
 - Use ZOS for all tests

Questions:
 - Does this work consistently for `unlock-app`?  Previous error appeared only on the second run of the app.

other questions in github comments

Notes:
 - This fails in Windows with an `out of gas` error but appears to work on Linux

# Issues

Fixes #
Refs #

# Checklist:


package.json
```diff
@@ -38,7 +38,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "start-ganache": "ganache-cli -i 1984 --mnemonic 'hello unlock save the web'",
     "postinstall": "./scripts/postinstall.sh",
-    "deploy-lock": "cd smart-contracts && npm run deploy -- --network local",
+    "deploy-lock": "cd smart-contracts && npm run deploy",
```
hardlydifficult: This should be unnecessary / have 0 impact

smart-contracts/index.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
```
hardlydifficult: This is from the ZOS example.  Should test without this file or add a comment regarding it's purpose and source.

smart-contracts/migrations/1_initial_migration.js
```diff
@@ -1,5 +1,5 @@
 const Migrations = artifacts.require('./Migrations.sol')
 
-module.exports = function initialMigration (deployer) {
+module.exports = function (deployer) {
```
hardlydifficult: This change is just reverting to the default truffle example.

smart-contracts/package.json
```diff
@@ -34,12 +34,12 @@
   },
   "scripts": {
     "remix": "remixd -s contracts --remix-ide http://localhost:8080 --read-only",
-    "test": "cross-env NODE_ENV=test truffle test --network development",
+    "test": "cross-env NODE_ENV=test truffle migrate && truffle test",
     "lint": "solhint \"contracts/**/*.sol\" && solium -d ./contracts/ && eslint .",
     "dev": "npm run lint && npm run build && npm run test",
     "ci": "npm run lint && npm run test && npm run reports",
     "build": "truffle compile --all",
-    "deploy": "truffle deploy --reset",
```
hardlydifficult: This is the workaround we put in to work around an `out of gas` error on the second run of `unlock-app`.  It should not be necessary.

smart-contracts/zos.json
```diff
@@ -1,5 +1,5 @@
 {
-  "zosversion": "2",
+  "zosversion": "2.2",
```
hardlydifficult: We may need the same update in the other zos.*.json files

smart-contracts/package.json
```diff
@@ -34,12 +34,12 @@
   },
   "scripts": {
     "remix": "remixd -s contracts --remix-ide http://localhost:8080 --read-only",
-    "test": "cross-env NODE_ENV=test truffle test --network development",
+    "test": "cross-env NODE_ENV=test truffle migrate && truffle test",
```
hardlydifficult: For some reason all tests fail if I don't migrate first.  ...why?

smart-contracts/index.js
hardlydifficult: When I added this it appeared to fix an issue but I believe that's a red herring and this file is not necessary.

package.json
nfurfaro: Why do we not need to specify the network here?

smart-contracts/migrations/2_deploy_unlock.js
```diff
@@ -1,21 +1,23 @@
...
-        ).code !== 0
-      ) {
-        throw new Error('Migration failed')
-      }
-    }
+    // Push implementation contracts to the network
+    await push(options)
+
+    // Create an instance of MyContract, setting initial value to 42
+    await create(Object.assign({
```
nfurfaro: I wonder about this part. I know it follows the zos example, but I believe there's a good reason to keep the deployment of unlock and the creation of the proxy in separate scripts. 
If this script fails at the proxy deployment (`create`) step, the Unlock contract has already been deployed, but the number of ther last successful migration in `Migrations.sol` won't be updated. So , next time we run `truffle deploy`, the Unlock contract will be deployed again by the `push` step. 
I feel like these migrations scripts should be atomic and do one thing each to avoid duplicating steps unintentionally.

smart-contracts/package.json
nfurfaro: Are you referring to the tests in `smart-contracts` or `unlock-app`? I know with the `unlock-app` tests there were a bunch that would fail unless `npm run dev` was run first...I believe that was to do with contracts needing compilation. But maybe that's outdated now and no longer the case.

smart-contracts/package.json
nfurfaro: Hmm. So this forces All migrations scripts to run each time, which shouldn't matter for dev, but definitely not the behaviour we want on Ropsten or mainnet.

smart-contracts/index.js
nfurfaro: Yes, let's test without first? I believe this is only needed if using truffle exec.

package.json
hardlydifficult: I'm trying to simplify the truffle-config.js down the line.  What value does network `local` offer now - it's effectively being merged with development which means we should be able  to just use default values again.

smart-contracts/package.json
hardlydifficult: Right.  This PR aims to fix that.

smart-contracts/package.json
hardlydifficult: unlock-app, on the second+ run only. (it works only with the first run)

smart-contracts/migrations/2_deploy_unlock.js
hardlydifficult: If any line in the migration step fails, the entire file will run again on next migration - i.e. it should fix itself.  Are you seeing different behavior?

smart-contracts/zos.json
nfurfaro: yes. I can do that in the PR I have going right now.

smart-contracts/migrations/2_deploy_unlock.js
nfurfaro: Well, if the `push` step is successful, Unlock.sol is deployed. But if the script fails after that, the entire file will run again as you say, and will deploy another instance of Unlock.sol. Not a big deal in development, but if we want to use migrations consistently (including for upgrades on rinkeby/mainnet) it's probably better for these scripts to be atomic steps, ie: the script either deploys unlock or it doesn't, then the next one either creates the proxy or doesn't, etc...
Now if there's no intention of using migrations for upgrades, then this isn't as important. But, as all of our actual deployments and upgrades have been done manually so far, there's definitely room to improve on that by scripting the process.

package.json
nfurfaro: I see. "Local" was only added when "development" was needed for solidity coverage. So if we remove local, and don't specify a network, "development" is used. Does this sound right?

smart-contracts/migrations/2_deploy_unlock.js
hardlydifficult: I see... k makes sense, will change

package.json
hardlydifficult: Yea.  I'm going to revert this for now to simplify the PR

smart-contracts/migrations/2_deploy_unlock.js
hardlydifficult: Done

smart-contracts/package.json
nfurfaro: This is weird. I need to think about what's happening here.

package.json
nfurfaro: Sounds good, but I like the direction this is going, towards simplification!

nfurfaro:
Re: 
>"Goals 
>- Use ZOS for all tests"
I'm not sure about this part. Can you tell me more about what htis would look like? How would this affect `upgrades.js`? While I agree that in our tests we should be using zos consistently (all interactions with unlock should be through the proxy), I'm not sure about using migrations for this. That's why I had a specific case for dealing with `network == "test"`, because we handle deployment and proxy creation on a test by test basis which gives us a lot of flexibility.

hardlydifficult:
I just mean remove the if condition in migrations so it always uses the ZOS path (as done in this PR).  But no problem putting that back in and we can defer to a future PR.

nfurfaro:
PR #1935 removes the use of `shell.js` from our migrations scripts. Next, I'll look at removing the `--reset` flag from the deploy script.

PR #1881, opened 2019-03-05
Fix unlocked paywall flag story to display it as it actually looks

### LINKED ISSUE #1404: Refactor paywall story to mock container for unlocked flag
**Describe the bug**
#1402 relies heavily on the iframe to handle the positioning and hiding of the `UnlockFlag` component. Since the storybook does not use the iframe for the paywall, we'll need to massage that story so that it will display in a manner consistent with the actual app.

**To Reproduce**
1. Go to storybook
2. Click on `Paywall > the paywall overlay, unlocked`
3. Notice the vertical positioning and lack of hide behavior on the `UnlockFlag`

cellog:
# Description

screen shot 2019-03-05 at 9 23 37 am
screen shot 2019-03-05 at 9 23 34 am
screen shot 2019-03-05 at 9 23 24 am

# Issues

Fixes #1404

# Checklist:

### DIFF
--- a/paywall/stories/Paywall.stories.js
+++ b/paywall/stories/Paywall.stories.js
@@ -1,5 +1,6 @@
 import { Provider } from 'react-redux'
-import React from 'react'
+import React, { useRef } from 'react'
+import PropTypes from 'prop-types'
 import { storiesOf } from '@storybook/react'
 import Paywall from '../src/components/Paywall'
 import createUnlockStore from '../src/createUnlockStore'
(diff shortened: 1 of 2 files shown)


paywall/stories/Paywall.stories.js
```diff
@@ -79,6 +80,32 @@ const config = {
   },
 }
 
+function FakeItTillYouMakeIt({ children }) {
+  const divit = useRef()
+  return (
+    <div
+      ref={divit}
+      style={{
+        position: 'absolute',
```
cnasc: Do we have a plan for keeping the storybook in sync with the paywall flag? (thinking of what happens if we change some of the surrounding stuff that isn't part of the flag itself, which we mock here)

LG otherwise!

cellog: I am unsure how we could best do this. We could make a parameter to the Paywall which allows us to use that for setting styles instead of the `window.document.body` but we will always have to fake the stuff that happens in `paywall.min.js`

cellog: (ideas appreciated!)

PR #1895, opened 2019-03-05
You can now navigate between dashboard and log pages

### LINKED ISSUE #1688: Make transaction log reachable from navigation

cnasc:
# Description

This PR adds 2 buttons into the header. They are only shown when the user is on an app page--right now that's just the dashboard and the log page. As a result, it is now possible for end-users to use the log page without being clairvoyant and guessing the URL.

Homepage: same as before
image

Dashboard: addition of buttons in the header (active button highlighted)
image

Creator log: new buttons in header, active page highlighted
image

# Issues

Fixes #1688 #1805 

# Checklist:

- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread (TODO)


unlock-app/src/components/interface/Header.js
```diff
@@ -54,6 +62,21 @@ export default class Header extends React.PureComponent {
             {title}
           </Title>
         )}
+        <AppButtons>
+          {onAppPage &&
+            appButtons.map(({ Button, page }) => {
+              const isActivePage = page === pathname
+              return (
+                <Button
```
cellog: can we abstract this behavior into a PageNavButton or something?

cnasc: I don't see why not! Keep it local to `Header` or have it in a separate file?

cellog: I'd put it with the other buttons, in its own file, since we will probably add other pages later, this sets up a pattern to follow

cnasc: Alright, I'll do that as a separate PR then rebase this one just to keep it from swelling too hugely

cnasc: The buttons were kind of mixed in with the work I had going here... Just crammed it all in here at least for the moment, let me know if I should split.

cnasc:
Now that #1917 and #1971 are fixed I think this is ready for another review

unlock-app/src/__tests__/components/interface/GlobalErrorConsumer.test.js
```diff
@@ -1,28 +1,29 @@
...
+import { Provider } from 'react-redux'
 import GlobalErrorConsumer, {
   displayError,
 } from '../../../components/interface/GlobalErrorConsumer'
 import { GlobalErrorContext } from '../../../utils/GlobalErrorProvider'
 import { FATAL_NO_USER_ACCOUNT, FATAL_MISSING_PROVIDER } from '../../../errors'
+import createUnlockStore from '../../../createUnlockStore'
 
-const Provider = GlobalErrorContext.Provider
+const ErrorProvider = GlobalErrorContext.Provider
```
julien51: I am a bit confused as to why these changes need to be in this PR?

cnasc: No problem at all, I will get started on splitting this!

PR #1896, opened 2019-03-05
Updating zos version in config files

nfurfaro:
# Description

This PR simply updates the zos version in 3 files to bring them in line with the version we're currently using. We're using v2.2, and in `zos.json`, `zos.mainnet.json` and `zos.rinkeby.json` we were still using v2.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/zos.json
+++ b/smart-contracts/zos.json
@@ -1,5 +1,5 @@
 {
-  "zosversion": "2",
+  "zosversion": "2.2",
   "name": "unlock-protocol",
   "version": "0.1.0",
   "contracts": {


hardlydifficult:
Do we need to upgrade the proxy before changing zos.mainnet or zos.rink ?  I'm not sure the significance of this version

julien51:
I was about to ask the same question?
I think we should probably update the files only when we actually upgrade the contracts?

nfurfaro:
For this PR I first only changed the version in `zos.json` and tried to run dev. I got an error, something to the effect of:
> An old version of zos has been detected. do you want to migrate to the new version?

I had an old copy of `zos.dev-1984.json` in my repo( with the zos version set to `2`. When I deleted the file and tried again it worked fine. My thought is that when we try to upgrade on rinkeby or mainnet, we'll see the same error message prompting the upgrade, unless we change it now. Maybe it's better to just let zos deal with it when (if?) that is the case? I could remove these two network files from this PR.

Issue #1842, opened 2019-03-02
Deploy contracts from their ABI?


julien51:
Is it possible to deploy smart contract from their ABI (the artifact generated when compiling) rather than from the source code?

This would really make https://github.com/unlock-protocol/unlock/issues/1840 great!

hardlydifficult:
We need bytecode to deploy a contract and ABI to interact with it.  For the bytecode, all we need is bytecode from Unlock (since that includes PublicLock inline).

Both may be included in the json, but it's configurable (e.g. npm public-lock-abi-0 is the abi only).

Today we capture versioned bytecode in `\versions\*.json`.  We could periodically publish an Unlock_V1.json (or maybe Unlock_latest.json)

julien51:
I just checked the module `unlock-abi-0` which I published earlier today and it does have the bytecode!

hardlydifficult:
ohh there's two packages, I didn't realize.  Perfect.

How frequently do we publish v1 and confirm integration before release?

julien51:
I think we should only publish packages with "final" versions. Also when we update the contract, we actually publish a *new* package, not an updated package.

hardlydifficult:
How does the integration test pass look though?  And what if the front-end depends on a new feature?  We could maybe publish v1 as version 0.1 and then upgrade to 1.0 when it's finalized

julien51:
So the integration tests should run with the packaged module.

When we need a new feature on the frontend we publish a new packaged abi which adds the feature needed by the front end first, and then we inplement the feature in the front end application.

So basically here is the process:

1. Create a packaged version of the abi.
2. All applications use the packaged version.
3. Smart contract dev continue to happen.
4. Once we reach a new version on the smart contract dev, we publish a new package.
5. We test the front end app works with the new package (without implementing new features in the front end)
6. If all works as expected, we deploy the new packaged abi to rinkey/olympic.
7. We implement the new features on the front end.
[repeat 3-7].

julien51:
@hardlydifficult @nfurfaro any of you know how to deploy the smart contract from the artifact files in the npm module?

hardlydifficult:
Yes. You could follow the example from `upgrades.js` or use standard Web3 calls.

hardlydifficult:
Believe this is complete, re-open if I missed something.

PR #1906, opened 2019-03-05
Now deploying the Unlock smart contract when starting in dev mode.

julien51:
We are not compiling the smart contract anymore when starting unlock app, but rather using the packaged version.

Refs #1842


julien51:
Dammit absolutely not! Thanks for catching this!

hardlydifficult:
Ideally call `initialize` as well, however that is not important for a test environment.

julien51:
@hardlydifficult Please clarify? I thought this was only useful for zos?

Also @akeem I actually forgot a pretty important part (and it's causing the tests to fail!): we cannot use the artifacts directly anymore! I'll push an update.

hardlydifficult:
We only have an initialize method because of zos, but now that it's there - it's an alternative to using the constructor. 

ATM it only sets the owner address (and resets grossNetworkProduct/totalDiscountGranted values but I'll remove these).  So for testing it's not important, but maybe a best practice in case something important is initialized in the initialize function.

https://github.com/unlock-protocol/unlock/blob/master/smart-contracts/contracts/Unlock.sol#L71

PR #1909, opened 2019-03-06
Remove remix-ide and use hosted IDE instead.

hardlydifficult:
# Description

Removing the `remix-ide` and changing the command to connect to the hosted solution instead.  To use:

 - `npm run remix`
 - Open browser to remix.ethereum.org
 - Click `Connect to localhost`

And the directory appears on the left.  You can edit files and they are synced with your local file system.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/package-lock.json
+++ b/smart-contracts/package-lock.json
@@ -265,7 +265,7 @@
       "dependencies": {
         "readable-stream": {
           "version": "2.3.6",
-          "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
+          "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
           "integrity": "sha512-tQtKA9WIAhBF3+VLAseyMqZeBjW0AHJoxOtYqSUZNJxauErmLbVm2FW1y+J/YA9dUrAC39ITejlZWhVIwawkKw==",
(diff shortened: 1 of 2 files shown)


nfurfaro:
I would be ok with this, but just wondering what benefits it provides over the current setup?

nfurfaro:
I'm ok with this, but would like to know how it's better than our current setup?

hardlydifficult:
One less package to install and update I guess.  I just thought it would be easier to use the hosted version, and might as well turn off the read-only restriction.

That said, neither approach actually works ATM.  The node references do not compile.  So this is not terribly important either way ;)

nfurfaro:
re:
>The node references do not compile.
Is there a specific error you're getting for this?

hardlydifficult:
Reference not found. Does it work for you?

nfurfaro:
Will give it a try now

nfurfaro:
Usually the error about reference to AST nodes is fixed by deleting artifacts and re-compiling contracts

hardlydifficult:
Ah, well there's a good reason for this PR - at least you'll get the IDE to open without issue.

But once you get past that the import will fail.  Remix has a special syntax for importing external packages, which sucks.  But having this integration available can still be handy, especially when doing a prototype.

nfurfaro:
I'm cool with this as long as @julien51 is comfortable with it.

PR #1921, opened 2019-03-06
Take space on the log page

### LINKED ISSUE #1918: Take space on the log page!
The log page is currently not looking great. I understand we do not currently have all the required data to implement @smombartz 's [design](https://www.figma.com/file/gmhVq6Pqeu2gH7e3FC2lr0ax/Unlock-Dashboard?node-id=0%3A67) but we should at least stretch so that the UI does not feel "broken".

I believe the font is too small and I also think we should add space between the rows.

image

cnasc:
# Description

This PR adjusts the font sizes and the row spacing on the creator log page to make it a little less dense and more readable. Additionally, the rows expand now (the blockNumber field takes up a static width for the moment, the others each take half the remainder). I think we can consider this a draft PR until we decide on the exact sizing we like.

The log page as it currently stands:
old creator log

And the log page with my changes:
image

Thoughts @julien51?

# Issues

Fixes #1918 

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
+++ b/unlock-app/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
@@ -54118,12 +54118,12 @@ Object {
 .c26 {
   display: grid;
   grid-template-columns: min-content min-content min-content;
-  grid-auto-rows: 20px;
+  grid-auto-rows: 28px;
   grid-column-gap: 20px;
 }
 
 .c27 {
-  font-size: 8px;
+  font-size: 11px;
   font-weight: normal;
   text-transform: uppercase;
   color: var(--grey);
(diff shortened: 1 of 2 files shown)


smombartz:
@cnasc updated figma with font and spacing suggestion: https://www.figma.com/file/gmhVq6Pqeu2gH7e3FC2lr0ax/Unlock-Dashboard?node-id=0%3A67

i would keep row text size consistent (i.e. make block number the same size). maybe its good to keep the create lock button there (that will take the user back to the dashboard)

cnasc:
I think it makes sense to keep the create lock button there; we'll need to rework the account component to handle that and that can come in a separate change I think. But that will also require some work to the dashboard. I think I will create a few issues to track those things and then try and schedule them. I think the log can get merged before we do those.

cellog:
what if the log page was a tabbed interface on the dashboard? we could use React.lazy() to lazy-load the log component

julien51:
Let's keep things simple for now and not change the stuff we do not need to change in this PR ;)

cnasc:
Updated for consistency with @smombartz figma design!

julien51:
LNTM (Looks Nice to me ;))

Issue #1926, opened 2019-03-06
eth_accounts chattiness caused by unexpected double polling of account on server and client


cellog:
The walletMiddleware sets up 2 polls, one on the server and one on the client. This only affects dev, where we have a provider on the server.

If we simply bail out of checking for account updates on the server, and reduce the polling interval to 2 seconds, this should still work for catching account changes but not overload the console with `eth_accounts` calls

PR #1928, opened 2019-03-06
Remove eth_accounts chattiness in dev environment

cellog:
# Description

***THIS IS A DRAFT AND SHOULD BE REVIEWED ONLY FOR APPROACH AND NOT NIT-PICKED YET***

This PR fixes the `eth_accounts` chattiness in dev environment through a couple of small changes:

1) polling for accounts was happening on both the client and the server, this is changed to only poll on the server
2) polling is changed to every 2 seconds instead of every half second

In order to implement #1, it required a minor change to the `walletMiddleware` that results in many lines of code changed because of indentation. Basically, to detect whether we are running on the server, we need the config, which is not currently available in `walletMiddleware`. We have the config object in `_app.js`, which is where middlewares are instantiated, and so we can pass it to `walletMiddleware` there.

In order to implement this in the paywall, the changes plus a couple of other files relating to signatures that were changed must also be copied over.

This PR will be split into 3 PRs:

1) modify `walletMiddleware` to accept config as an argument
2) change polling interval and disable the account polling on server
3) merge changes to the paywall

# Issues

Refs #1926 

# Checklist:


julien51:
I am not sure we currently need the signature stuff in the paywall app at all?

cellog:
> I am not sure we currently need the signature stuff in the paywall app at all?

no we don't, but it makes synchronizing `walletMiddleware/Service` and `web3Middleware/Service` prohibitively complex if we make any changes. This problem almost goes away when we have a separate location for the shared services. Ultimately, it's a question of diffing across directories. Happy to discuss strategy offline too

Issue #831, opened 2018-12-21
Key Data


julien51:
Each key has an optional data field. This is important because it lets the customer (owner of the key) communicate something to the creator (lock owner).
Use cases for this are for example: encrypted email addresses, encrypted names, images... etc

Our current implementation stores this on chain (the `data` field is actually `bytes` which is not limited. According to @nfurfaro 's research, a challenge introduced by this unbounded field means that it is very hard to quickly paginate keys. We would need a fixed size field to quickly be able to retrieve X keys at once.

Another (more obvious) challenge is obviously that this can be very costly (more on this [old article](https://medium.com/@didil/off-chain-data-storage-ethereum-ipfs-570e030432cf)).

Now this problem is in no way unique to Unlock. Many crypto projects are actually facing similar challenges. The general approach seems to be to store the data "off chain" and only store a pointer to it on chain.

There are multiple choice for the platform to pick, including IPFS, Dat and others.

nfurfaro:
I now think this is considered the "standard " approach. It has the benefit of being storage provider agnostic . https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md

julien51:
But wait this would be lock-level metadata, right?

nfurfaro:
This should be key (NFT)-level metadata, that we're currently using the Key.data field for.

julien51:
Ha! I see you, mean the `tokenURI` stuff?

julien51:
However i guess we could use the `name` stuff (cc @akeem)

nfurfaro:
@julien51 Regarding the `tokenURI `: It is allowed to be mutable, and so could be updated if the storage location changes. Are your concerns more about ensuring the integrity of the data at the storage location?

nfurfaro:
"Metadata Choices (metadata extension)

We have required name and symbol functions in the metadata extension. Every token EIP and draft we reviewed (ERC-20, ERC-223, ERC-677, ERC-777, ERC-827) included these functions.

We remind implementation authors that the empty string is a valid response to name and symbol if you protest to the usage of this mechanism. We also remind everyone that any smart contract can use the same name and symbol as your contract. How a client may determine which ERC-721 smart contracts are well-known (canonical) is outside the scope of this standard.

[...]

PR #1938, opened 2019-03-06
Removing key specific data

hardlydifficult:
# Description

Per recent discussions we are simplifying the contract a bit by removing key specific data.  More info in the issue referenced below.

# Issues

Refs #831

# Checklist:

### DIFF
--- a/smart-contracts/contracts/interfaces/ILockCore.sol
+++ b/smart-contracts/contracts/interfaces/ILockCore.sol
@@ -10,11 +10,9 @@ interface ILockCore {
   /**
   * @dev Purchase function, public version, with no referrer.
   * @param _recipient address of the recipient of the purchased key
-  * @param _data optional marker for the key
   */
   function purchaseFor(
-    address _recipient,
-    bytes calldata _data
+    address _recipient
   )
     external
     payable;
@@ -23,12 +21,10 @@ interface ILockCore {
   * @dev Purchase function, public version, with referrer.
   * @param _recipient address of the recipient of the purchased key
   * @param _referrer address of the user making the referral
-  * @param _data optional marker for the key
   */
   function purchaseForFrom(
     address _recipient,
-    address _referrer,
-    bytes calldata _data
+    address _referrer
   )
     external
     payable;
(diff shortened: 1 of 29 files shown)


cellog:
Can you clarify how it breaks the tests on the paywall (and why, if you know)?

hardlydifficult:
See https://circleci.com/gh/unlock-protocol/unlock/4834?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

e.g. the `keyDataFor` method does not exist anymore and `purchaseFor` expects 1 argument now and not 2.

hardlydifficult:
Regarding the larger process point, anytime an external API changes in the contracts we need a process for updating dependencies such as the paywall.  For this specific change the `Key` struct changed, function `keyDataFor` was removed, and both `purchaseFor` and `purchaseForFrom` requires one less param.

hardlydifficult:
BTW this is the first breaking change in contracts - if not this specific change, I would love to see us see some breaking change implemented end-to-end sooner than later.  It's a great proof of our upgradability.

cellog:
So my specific reaction is this affects all apps, because the functionality is in walletService, which is about to become a shared library. So we will want a way to detect which contract version it is, and to abstract to the correct implementation of the functions you mentioned.

Is there a version explicit that we can grab from the contract?

hardlydifficult:
Yes - in this case we are talking about PublicLock specifically and not the Unlock contract.  

`publicLockVersion` was not implemented in v0, which means call it and you should get the default value / i.e. 0 as expected. 
V0 returns 1.

cellog:
Ok, we definitely need to key @julien51 into this conversation. This decision needs to be made very deliberately so we have a clear path both of how to handle this but also how to make the code intelligible.

julien51:
Thanks for the work on this @hardlydifficult !
My thinking is that we *definetely* need to use the packaged versions in the app so that when we do introduce breaking changes in the smart contract, we do no unintentionally break the apps that are using it.

The way forward, IMO, is to make sure that all apps use the packaged npm, and not the current smart contracts. 
Then we package the *newer* version of the smart contract and add support inside `web3Service` and `walletService` for it.
When we have confidence that all is clean, we can upgrade our on chain contracts.

hardlydifficult:
Sounds good.  Opened #1950 to track this.

hardlydifficult:
Looks like the build is passing.  Promoting to a normal PR for review.

nfurfaro:
@hardlydifficult Can you summarise this in a one-liner so I know what I'm looking at going in? IE: Is key data removed completely, or just to a data structure external to the Key struct?

hardlydifficult:
All references to the key specific data field have been removed.

PR #1951, opened 2019-03-07
Add grantKeys (owner only) functions to PublicLock

### LINKED ISSUE #1450: Enable a migration path for Lock owners
This is a sub-task (1/3) of "Lock Deletion" #1348
If Lock owners want to upgrade and benefit from new features in a new version of the Lock contract, there needs to be a mechanism to "migrate" existing valid keys to the new lock.
 
Either implement a `grantKeys` function as per https://github.com/unlock-protocol/unlock/issues/1348#issuecomment-459974142, 
or build the ability for future Locks to acknowledge keys from a prior version, as per https://github.com/unlock-protocol/unlock/issues/1348#issuecomment-459997864

**Additional context**
- #1348 
> "We need to keep in mind the cancelAndRefund feature here as well. If the Lock owner simply re-creates Keys on the new Lock without first invalidating them on the old lock (or first deprecating the old lock which could also block cancel attempts) then a user could try to cancel both the new and the old and potentially profit."

hardlydifficult:
# Description

Adds the ability for a Lock owner to grantKeys (i.e. gift keys) to users.  There are 3 functions wrapping this capability to provide a little flexibility for the Lock owner, and to save on gas costs for some scenarios (like bulk gifting).

In addition to standard gifting scenarios, this feature can be used to allow a Lock owner to migrate Key owners from an old Lock version to a newer Lock for the same content.

Risk of note: if a Lock owner grants some users keys and sells keys to other users, the key owners which received a free key could 1) call cancelAndRefund and get some ETH, this money comes out of revenue the Lock owner received from paying customers or 2) sell the key on a secondary market potentially taking away from future revenue for the Lock owner.  If desired, either one of those (or both of them) could be blocked in the contract.

# Issues

Fixes #1450

# Checklist:

### DIFF
--- a/smart-contracts/contracts/PublicLock.sol
+++ b/smart-contracts/contracts/PublicLock.sol
@@ -7,13 +7,14 @@ import 'openzeppelin-solidity/contracts/introspection/ERC165.sol';
 import 'openzeppelin-solidity/contracts/token/ERC721/IERC721Receiver.sol';
 import './mixins/MixinApproval.sol';
 import './mixins/MixinDisableAndDestroy.sol';
+import './mixins/MixinGrantKeys.sol';
 import './mixins/MixinKeys.sol';
 import './mixins/MixinLockCore.sol';
-import './mixins/MixinRefunds.sol';
 import './mixins/MixinLockMetadata.sol';
 import './mixins/MixinNoFallback.sol';
-import './mixins/MixinTransfer.sol';
 import './mixins/MixinPurchase.sol';
+import './mixins/MixinRefunds.sol';
+import './mixins/MixinTransfer.sol';
 
 
 /**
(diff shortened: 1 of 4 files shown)


julien51:
So, this is interesting! I talked about this with @cnasc and at the time I assumed granting could still be done at the app level... but I think you have a point regarding "migrating" keys.

I agree we need to careful about the things you highlight, so not 100% sure this is a thing we will expose in the UI at the beginning. The refund part is the most worrisome.

nfurfaro:
We could possibly prevent "gifted" keys from receiving refunds if cancelled to overcome this risk. Allowing a migration path is still needed, not sure that both migration and gifting/granting need to be achieved with the same function(s) but they do seem to fit together well if the risks highlighted can be mitigated.

PR #1966, opened 2019-03-07
pass middleware the top-level config

cellog:
# Description

In preparation for fixing #1928, the walletMiddleware needs to get the configuration. This is a pattern shared by three other middlewares.

In order to easily test changing configuration values (for instance, from running on the server to running on the client), there are 2 ways to do this.

1) mock the configuration by making it a parameter
2) mock the configuration by using jest mocks to mock the entire file

This PR is the first approach. These are the tradeoffs involved.

**mocking the configuration by making it a parameter**
Advantages:
1. The dependency is cleaner in the test, we just need to change the local config like:
```js
it('does not work on server', () => {
  config.isServer = false
  const middleware = create()
//...
```
2. no reliance on third-party solutions in order to mock, it will just work.
3. configuration is re-created every time we call `configure()`, both creating unnecessary redundant runtime calls and introducing a point of failure should something in the env it uses to create the config changes.

Disadvantages:
1. a parameter must be passed to create the middleware. This makes the source slightly more complex:
```diff
+const createWalletMiddleware = config => ({ getState, dispatch }) => {
-export default function walletMiddleware({ getState, dispatch }) {
```
2. The dependency must be passed in the top-level `_app.js`, introducing a potential point of failure should that be forgotten.

**mocking the container through a jest mock**
Advantages:
1. The dependency is cleaner in the source, it's a simple import and function call to configure()
2. top-level _app.js does not need to pass the config to work.
3. the source of the dependency is more explicit (the import is in the same file)

Disadvantages:
1. We have to rely upon the testing framework to properly mock the config, and jest mocks are pretty complex to get right.
2. configuration is re-created every time we call `configure()`, both creating unnecessary redundant runtime calls and introducing a point of failure should something in the env it uses to create the config changes.

**How I made this decision**
This is not an easy one to make, the advantages and disadvantages are almost the same. What tips the balance in favor of this PR's solution is the complexity of jest mocking. In my short time at Unlock, I've already seen many hard-to-track-down bugs caused by weird mock behavior, as well as a large learning curve setting up a mock, and continued confusion even after setting up many mocks previously.

The other factor is that the configuration is not just a static import, it requires calling a function and in the global scope of the file. This introduces the equivalent of a global variable, and also an unnecessary performance hit.

I'm open to the other solution, and I am going to make a competing PR with that implementation in a moment. Note that because the current approach is closer to the jest mocking way, there are fewer lines of code that need to be touched.

# Issues

Refs #1928 
Refs #1967

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/middlewares/currencyConversionMiddleware.test.js
+++ b/unlock-app/src/__tests__/middlewares/currencyConversionMiddleware.test.js
@@ -1,4 +1,5 @@
 import { setConversionRate } from '../../actions/currencyConvert'
+import configure from '../../config'
 
 jest.mock('../../services/currencyLookupService', () => () => ({
   lookupPrice: jest
@@ -9,6 +10,11 @@ jest.mock('../../services/currencyLookupService', () => () => ({
 }))
 
 describe('Currency conversion service retrieval middleware', () => {
+  let config
+  beforeEach(() => {
+    // TODO: write tests that allow testing different services
+    config = configure()
+  })
   it('service called, action dispatched to set currency conversion rate', async () => {
     expect.assertions(2)
     jest.useFakeTimers()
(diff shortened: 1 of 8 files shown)


cellog:
Closing in favor of #1967 after discussing further with Julien

PR #1970, opened 2019-03-08
Running all tests through zos proxy via migrations

### LINKED ISSUE #1841: Switch `truffle migrate` to follow the ZOS example
We are currently using shell commands to interact with ZOS, however they just released a better way.

Follow the example here: https://github.com/zeppelinos/zos/tree/master/examples/truffle-migrate

**Additional context**

nfurfaro:
# Description

This is a WIP PR, and is a proof-of-concept for:
- removing all network conditionals from our migrations scripts
- creating a new "proxy" helper 
- modifying 1 test file (`Lock.js`) to use the new helper

The goal is to run all of our tests against a proxy vs directly against Unlock. This is achieved by letting the migrations scripts handle the Unlock deployment and proxy creation as is done on dev, and pointing our tests at the newly deployed proxy address.

Currently, each time we run `npm run test` a new proxy address is added to `zos.dev-1984.json` so the helper (`proxy.js`)selects the last one in the array. Alternately, we could delete this file each time the test command is run and select the first [0] index in the array of proxies.

Based on feedback, I could break this into 3 PR's.
- remove `IF` statements from migrations
- add `proxy.js` helper
- modify tests to use `proxy.js` consistently

### Still to do:
[ ] modify all the tests that deploy unlock to use the new helper (`/test/helpers/proxy.js`)
# Issues

Fixes #1841
Refs #

# Checklist:


nfurfaro:
Actually, all of these changes must happen in together as you can see from the failed build. The first test `Lock / Lock` was failing after removing the `If`'s until I added and used the new `proxy.js` helper.

nfurfaro:
Update: This has been modified to use the "officially recommended" zos way of getting the proxy address from the network file.
I've also modified `setUnlockProxy` to be more generic. It should accepts an Unlock contract (either from artifacts or from an npm-package) and a network:
```
module.exports = function setUnlockProxy (_Unlock, _network)
```
While this isn't really neccessary for our tests, it may allow this same helper to be used by the `deploy-unlock.js` script.

smart-contracts/test/helpers/proxy.js
```diff
@@ -0,0 +1,26 @@
...
+const packageFile = new ZosPackageFile()
+
+let networkFile
+let proxiesArray
+let proxies
+let mostRecentProxy
+let ProxyAddress
+let proxiedUnlock
+
+module.exports = function setUnlockProxy (_Unlock, _network) {
```
hardlydifficult: minor, but would it be more clear to call this `getUnlockProxy`?  It doesn't actually deploy or change anything.

nfurfaro: I think you're right.

nfurfaro:
So to make this pass I'm going to have to modify (slightly) a LOT of tests to use this new helper. But moving forward we'll be able to make changes in 1 place for all these tests.

nfurfaro:
So just 2 tests to fix in `createLockWithInfiniteKeys`. Not sure what's wrong yet...
```
1) Contract: Unlock / behaviors / createLockWithInfiniteKeys
       Create a Lock with infinite keys
         "before all" hook:
     TypeError: unlock.methods.createLock is not a function
      at Context.<anonymous> (test/Unlock/behaviors/createLockWithInfiniteKeys.js:22:10)

  2) Contract: Unlock / behaviors / createLockWithInfiniteKeys
       Create a Lock with 0 keys
         "before all" hook:
     TypeError: unlock.methods.createLock is not a function
      at Context.<anonymous> (test/Unlock/behaviors/createLockWithInfiniteKeys.js:52:10)
```
Note that I don't think this test actually belongs in the `/behaviours` directory. I'm the one who put it here originally.

hardlydifficult:
That test is going throw downgrade pains ;)  I recently upgraded it to use the web3 beta 1 API, and now this is switching it back.

BTW I would like to remove /behaviours completely in the future. I don't believe we use that pattern any longer and can revert to the standard test style.

nfurfaro:
@hardlydifficult What's the best approach here? Revert this 1 test (`createLockWithInfiniteKeys.js`) to make it pass...or try to modify my `proxy.js` helper to use the web3 beta 1 API, and then upgrade ALL the test which use the helper?
Maybe I'll revert the 1 test, for now, to push this through, and then the rest can be a future PR of its own?

hardlydifficult:
Yea just fix the createLock test.  This PR has enough going on :)

nfurfaro:
Sounds good. That's the approach I took. Just trying to sort out a linter problem to make this thing pass!

nfurfaro:
@hardlydifficult I know you approved this, but I've modified another 30 files or so since then;)

hardlydifficult:
Oh sorry I missed this last time I took a look.  Could we not hardcode a network ID?  Looks like web3 has a method we could use instead:
https://github.com/ethereum/wiki/wiki/JavaScript-API#web3versionnetwork or https://web3js.readthedocs.io/en/1.0/include_package-net.html#getid

nfurfaro:
Sure, I'll check that out.

hardlydifficult:
It's getting better for sure.  I think this is good to merge.  We can always revisit if needed when (if) we attempt to run the test suite against Ropsten.

nfurfaro:
I have to push one last time after fixing linter errors. I can add a case for rinkeby now if you like.

PR #1971, opened 2019-03-08
Set correct lock address on lock creation transactions

### LINKED ISSUE #1917: Log page does not show the right lock address?
When going to the log page on the creator dashboard I see several "lock creation" transactions shown with the same lock address... This can't be right:

image

I suspect this is actually showing the address for the Unlock smart contract?

cnasc:
# Description

image

Before this PR, the `lock` property on lock creation transactions was set to the address of the smart contract, and not the address of the lock that was created. This PR addresses that by setting the lock address to the correct value once we have the transaction receipt.

**This PR is a draft** until I handle what happens with pending lock creations.

It seems that we get the lock address right as we submit, before any new block is mined:
image

That seems sufficient to me, any thoughts?

# Issues

Fixes #1917 

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/services/web3Service.test.js
+++ b/unlock-app/src/__tests__/services/web3Service.test.js
@@ -875,6 +875,27 @@ describe('Web3Service', () => {
       expect(web3Service.getLock).toHaveBeenCalledWith(params.newLockAddress)
     })
 
+    it('should handle NewLock and emit transaction.updated with the new lock address', () => {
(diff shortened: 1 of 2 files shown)


cnasc:
Investigating where the transactions in the state are used right now. From what I can tell, they are used in:

- Wallet middleware
- The creator lock
- The transaction log
- The dashboard

I will take a look at each of these and find out whether or not they rely on the existing behavior (though it would be very peculiar if they did)

Issue #1907, opened 2019-03-05
Update integration test configuration to include running instance of Locksmith


akeem:
As part of adding more coverage of the application with integration test, we should add an instance of Locksmith into the configuration.

Ideally this will allow us to ensure names are being saved, transactions can easily be looked up etc.

akeem:
Closing as this has been resolved

PR #1973, opened 2019-03-08
Adds initial test coverage around the persistence of Lock names

akeem:
# Description

Expanding on the work to provide better integration test coverage for the application, this PR adds  coverage around a name being saved.

# Issues

Fixes #
Refs #1907

# Checklist:


julien51:
Dang. any idea why the tests fail? Could it be that the samrt contract is still not deployed where we expected it to be?

akeem:
@julien51 this is what I was working on this morning. It was/is a configuration issue the contract is there but this is around locksmith

akeem:
I've updated out CI environments include values for the following environment variables:

* CI
* HTTP_PROVIDER
* DB_USERNAME
* DB_PASSWORD
* DB_NAME
* DB_HOSTNAME
* LOCKSMITH_URI

I do believe that our integration test script may not be functioning exactly as we believe it may. https://forums.docker.com/t/environment-variables-in-args/27522/2

cellog:
I wonder... docker created swarms to support what docker-compose does, and has been pushing folks to use that instead of docker-compose. Perhaps we should look at whether that would make things more reliable since it's built into docker?

https://vsupalov.com/difference-docker-compose-and-docker-stack/

julien51:
Let's keep things simple and avoid yet another rabbit hole. Docker compose should work fine for now ;)

docker/docker-compose.ci.yml
```diff
@@ -53,17 +56,20 @@ services:
     depends_on:
       - ganache-integration
       - unlock
+      - locksmith
```
julien51: I think we should not have it both here and in `unlock` above. I'm in favor of just having it here but that's not a huge deal either way.

docker/docker-compose.ci.yml
```diff
@@ -53,17 +56,20 @@ services:
...
     working_dir: /home/unlock/tests
   locksmith:
     environment:
       - DB_USERNAME
       - DB_PASSWORD
       - DB_NAME
       - DB_HOSTNAME
+      - NODE_ENV=production
```
julien51: Hum. Is that required?

docker/docker-compose.ci.yml
```diff
@@ -53,17 +56,20 @@ services:
...
     image: unlock
     user: node
     working_dir: /home/unlock/locksmith
-    ports:
+    command: sh -c './scripts/wait-for.sh db:5432 -- npm run db:migrate && npm run start'
```
julien51: Yikes.

unlock-app/src/config.js
```diff
@@ -97,7 +97,7 @@ export default function configure(
     )
     blockTime = 10 // in mseconds.
     supportedProviders = ['HTTP']
-    services['storage'] = { host: 'http://127.0.0.1:8080' }
+    services['storage'] = { host: runtimeConfig.locksmithHost }
```
julien51: Is there a default value for this set in `config.next.js`? I just want to make sure things run fine in dev...

scripts/integration-tests.sh
```diff
@@ -1,9 +1,20 @@
 #!/usr/bin/env bash
 
+
```
julien51: nit: newline?

scripts/integration-tests.sh
```diff
@@ -1,9 +1,20 @@
...
+export DB_USERNAME='locksmith_test'
+export DB_PASSWORD='password'
+export DB_NAME='locksmith_test'
+export DB_HOSTNAME='db'
+export CI=true
+export HTTP_PROVIDER='ganache-integration'
+export LOCKSMITH_URI='http://locksmith:8080'
+
+
```
julien51: nit: same same ;)

docker/docker-compose.ci.yml
akeem: We could probably come up with another environment for this, but the non-production environment will pull in the dev packages and will bring sqllite along with it.

PR #1978, opened 2019-03-08
walletService triggers an error when the Unlock smart contract is not available at the supplied address

julien51:
Adding this function will let the application detect when it is noti pointing to the actual Unlock smart contract.
This will also make it easier to debug integration tests.

### DIFF
--- a/unlock-app/src/components/helpers/ErrorMessage.js
+++ b/unlock-app/src/components/helpers/ErrorMessage.js
@@ -14,7 +14,7 @@ const genericError = message => (
 const defaultMessages = {
   MISSING_PROVIDER: <p>You need a web3 provider to use Unlock.</p>,
   NOT_ENABLED_IN_PROVIDER: <p>Please enable Unlock in your web3 Provider</p>,
-  NON_DEPLOYED_CONTRACT: (
+  FATAL_NON_DEPLOYED_CONTRACT: (
     <p>Unlock has not been deployed on the current network.</p>
   ),
 
--- a/unlock-app/src/__tests__/services/walletService.test.js
+++ b/unlock-app/src/__tests__/services/walletService.test.js
@@ -14,6 +14,7 @@ import {
   FAILED_TO_PURCHASE_KEY,
   FAILED_TO_UPDATE_KEY_PRICE,
   FAILED_TO_WITHDRAW_FROM_LOCK,
+  FATAL_NON_DEPLOYED_CONTRACT,
 } from '../../errors'
 
 jest.mock('../../utils/promises')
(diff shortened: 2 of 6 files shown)


julien51:
Hum. Have I uncovered issues in the integration tests?

akeem:
Yeah, this is a good find!

julien51:
I am slowly going down the rabbit hole here.
What we find is that the smart contracts are not deployed... We're moving to using the npm module by changing how `start.sh` works.
Additionally, since `deploy-unlock-contract` actually does not set the `UNLOCK_ADDRESS` value we change how it is being called.

I also found out that for some reason docker is still starting the `ganache` image (and the `ganache-integration` image) when we should only be using `ganache-integration`.

Finally our `deploy-unlock.js` script was only connecting to `localhost` when it should connect to `ganache-integration`.

julien51:
Another interesting realization is that the deploy-unlock script is "too fast" and tries to run even though the ganache node is not "ready".

julien51:
At this point, I am running the integration tests locally and they fail, but I think everything works _as expected_, so I start to think that maybe the tests are wrong.

julien51:
🎉

julien51:
Ok, splitting that PR ;)

julien51:
Split the irrelevant changes into https://github.com/unlock-protocol/unlock/pull/1980

unlock-app/src/components/helpers/ErrorMessage.js
```diff
@@ -14,7 +14,7 @@ const genericError = message => (
 const defaultMessages = {
   MISSING_PROVIDER: <p>You need a web3 provider to use Unlock.</p>,
   NOT_ENABLED_IN_PROVIDER: <p>Please enable Unlock in your web3 Provider</p>,
-  NON_DEPLOYED_CONTRACT: (
+  FATAL_NON_DEPLOYED_CONTRACT: (
```
cellog: This introduces an inconsistency. `FATAL_` errors are intended to only be handled by the `GlobalErrorConsumer` and generated by the `GlobalErrorProvider` at startup, not by redux. Redux-based errors are handled by components and converted from error codes to static markup here.

This is also true of a recent change to the `FATAL_NO_USER_ACCOUNT` constant made in a separate PR.

julien51:
Once https://github.com/unlock-protocol/unlock/pull/2004 has been merged, I can make sure that this error actually shows up ;)

unlock-app/src/components/helpers/ErrorMessage.js
julien51: We are changing the way the errors are handled by getting rid of the provider and just using redux to dispatch the errors.
The fatal one will be rendered and block the app.

julien51:
image

julien51:
Rebased, screenshot added, ready for another look ;)

PR #1980, opened 2019-03-09
Fixing integration tests

julien51:
So, this PR does a bunch of things to fix the integration tests.
I was not able to split and keep passing tests.

* we're using the packaged contract artifacts to deploy contracts
* changed the Unlock contract address to reflect the change above
* changed the deployed lock contract to reflect the change above (it changes since it's deployed by a different contract)
* updated the script to deploy Unlock to add a "waiting" period


docker/docker-compose.ci.yml
```diff
@@ -24,8 +24,6 @@ services:
     image: unlock
     user: node
     working_dir: /home/unlock
-    depends_on:
```
akeem: Even though we are now polling via the configuration, I think we might probably need to ensure that at least one of the instances of ganache is ready/available here

julien51: Oh, so technically I think the `unlock` image itself does not need `ganache-integration` integration to run (it could theoretically be running in other contexts), but the `integration-test` does (and has it!)

PR #1989, opened 2019-03-10
Lock Form Button Design Fixes

smombartz:
# Description
* removed blue focus boder on active state
* increased spacing of button on right side on expanded card
* shortened preview lock button label text to label


julien51:
@smombartz you can click on the "Details" link and then make sure this looks like what you expect ;)

Issue #1449, opened 2019-02-06
Decide whether and how to use TypeScript


cnasc:
We need to determine if we are going to use TypeScript when we spin off the paywall app, and how we will use it if we do.

cellog:
@cnasc do you want to take the reins on deciding a limited way we can test this? Or lead an experiment? Part of this will be consulting with @julien51 to determine the priority and when this should be considered

cnasc:
I am 100% willing to take the reins on an experiment for this.

Here's what I propose:

1. We get the app building with TypeScript, but don't annotate any types just yet (our JS is legal TypeScript, it just isn't type checked)
  a. The end result of this should be an app that works just the same as before, but with a whole boatload of errors for things that haven't been typed
2. We prioritize modules of the code to be annotated. My personal preference on this is to start at the boundaries. The idea is that we have a sort of "ring" around the codebase. If the things on the ring boundary can be trusted, then the things inside the ring which are not checked still gain some robustness. As time goes on, we shrink the ring piece by piece until the whole app is typed.
3. Once the whole app is typed, we can start setting lint configs to fail rather than warn on code that isn't typed or which has an `any` type. In fact, we can probably do this module-by-module, so that as each module gets typed, we also clamp down the rules so that no new dynamically typed code sneaks in.

julien51:
@cnasc works for me. Let's focus on the `unlock-app` for now on this one since it probably has less moving part (despite being bigger!).
Let's do 1. and reconvene before you attack 2.

cnasc:
Just to confirm:

- The TS experiment will be scoped to the `unlock-app` project
- For now, the task is just to make `unlock-app` build in its current state with `tsc` and associated tooling
- After that we will meet to discuss before going any further

cnasc:
I think at this point we have decided what the path forward is for TypeScript, and given that today I should merge TS support for both `unlock-app` and `locksmith`, I think we can close this issue.

PR #1996, opened 2019-03-11
Experiment: build unlock-app with TypeScript

cnasc:
# Description

This PR is an experiment to start building `unlock-app` with TypeScript. Because this is an experiment there are some commingled purposes here:

- Various packages installed and configuration changes made
- 2 files (one `.ts` and one `.tsx`) converted to TypeScript just to get an idea of the development workflow.

Right now the build will not pass, because the TypeScript options are set very strictly. These can be eased up, but my preference would be to keep them strict--loose TypeScript is barely better than JS, so we should make sure that anything we convert gets the full gamut of typing.

Here's some truncated output from a test run:
```
    src/components/creator/CreatorLock.tsx:30:28 - error TS7031: Binding element 'lock' implicitly has an 'any' type.

    30 const LockKeysNumbers = ({ lock }) => (
                                  ~~~~

    src/components/creator/CreatorLock.tsx:139:13 - error TS2339: Property 'propTypes' does not exist on type 'typeof CreatorLock'.

    139 CreatorLock.propTypes = {
                    ~~~~~~~~~
```

Both of these errors occur from just moving a React component into a `.tsx` file (incidentally, it looks like the `.tsx` extension is mandatory for any TypeScript code using JSX. Build fails if you just use `.ts`) without adding any types at all.

The first error could be resolved by simply annotating the `any` type. That should be avoided, because it confers no benefit over standard JS. The preferred solution is to annotate the actual type, luckily for many cases we should be able to use the PropTypes as a template if the expected type is unclear.

The second error demonstrates TS' strictness a little further, although in this case it's ok because TS supersedes PropTypes (or at least, it will in the future; interop with JS still needs to be evaluated). Solution to this particular case of the error is to just not assign the PropTypes once we have added the TS types.

There wasn't much in terms of objective criteria for this experiment, since the idea was simply to see how difficult (or not) it is to migrate to TS. My take on that: it's pretty easy. Once the config is set up, it just works.

My recommendation for moving forward:
- Continue building with TS
- Migrate file-by-file as needed.
  - Perhaps we should have a cultural rule that when we start work on an area of the codebase that hasn't had types added yet, we first do a PR to add those types?
  - Some low-hanging fruit on that may be just to install any type definitions we can find for any dependencies we use. (I've already added types for Jest, React, and Redux since they were necessary for the build)
- New subrepos should use TypeScript from the beginning

I'm going to invite everyone to review this so we can hopefully start up some discussion around this. Feel free to ask any questions about things I may have glossed over! I'll add some comments inline in the code on places where I ran up against some difficulties.
# Issues

Refs #1449 

# Checklist:

- [x] 1 PR, many purposes: my Pull Request modifies tons of different things


unlock-app/next.config.js
```diff
@@ -3,7 +3,7 @@
 const fs = require('fs')
 const { join } = require('path')
 const { promisify } = require('util')
-const withSourceMaps = require('@zeit/next-source-maps')
```
cellog: can the two extensions co-exist? Or does withTypescript auto-export source maps? If the latter, a comment above would be useful

unlock-app/src/constants.ts
```diff
@@ -18,7 +18,7 @@ export const ETHEREUM_NETWORKS_NAMES = {
  * @param title
  * @returns {string}
  */
-export const pageTitle = title => {
+export function pageTitle(title: string): string {
```
cellog: more a question than anything else. Is this valid TS too?

```ts
export const pageTitle = (title: string): string => {
```

package.json
```diff
@@ -26,7 +28,8 @@
     "link-parent-bin": "^1.0.0",
     "netlify-cli": "^2.6.0",
     "prettier": "^1.15.3",
-    "prettier-eslint-cli": "^4.7.1"
+    "prettier-eslint-cli": "^4.7.1",
+    "typescript": "^3.3.3333"
```
cnasc: All of these deps are added at the root level because eslign plugins need to be installed in the same location as eslint. It would be the same if eslint were global; the plugins would need to be global.

Typescript is added here because otherwise the plugin would use an older version of Typescript which is not actually supported by the plugin anymore.

unlock-app/.babelrc
```diff
@@ -38,7 +40,8 @@
             "preset-env": {
               "modules": "commonjs"
             }
-          }
+          },
+          "@zeit/next-typescript/babel"
```
cnasc: I'm not 100% sure why, but the order of the plugins mattered on this environment. Putting `next-typescript` in as the second element was an error.

All of these are added so that we can build typescript files. This was a pretty easy change--after this config all the building scripts worked exactly as they did before.

unlock-app/.eslintrc.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
     browser: true,
     jest: true,
   },
-  plugins: ['jest', 'mocha', 'promise'],
+  plugins: ['jest', 'mocha', 'promise', 'import'],
```
cnasc: The changes here allow eslint to correctly resolve imports of TS files. Without this, you'd get a "module not found" error in any file that imported some TS.

unlock-app/jest.config.js
```diff
@@ -8,8 +8,9 @@ module.exports = {
   ],
   transform: {
     '^.+\\.(js|jsx)$': 'babel-jest',
+    '^.+\\.(ts|tsx)$': 'ts-jest',
```
cnasc: Like eslint, Jest doesn't support TS by default. A package had to be added and some minor config changes. 
Nothing esoteric.

unlock-app/next.config.js
```diff
@@ -32,7 +32,7 @@ Object.keys(requiredConfigVariables).forEach(configVariableName => {
   }
 })
 
-module.exports = withSourceMaps({
+module.exports = withTypescript({
```
cnasc: You'll note that in TSConfig, we set `sourceMap: true`, so we don't need `withSourceMap` anymore because it's handled by the TypeScript compiler. We do need `withTypescript` though, that's what hooks all the Next stuff up so we can just call `next build` and so on.

unlock-app/package.json
```diff
@@ -10,7 +10,11 @@
     "@storybook/addon-viewport": "^4.1.13",
     "@storybook/addons": "^4.1.13",
     "@storybook/react": "4.1.13",
+    "@types/jest": "^24.0.11",
+    "@types/react": "^16.8.7",
+    "@types/react-redux": "^7.0.3",
```
cnasc: Types! These are additional packages that define types for these libraries. We'll probably do a lot of this, luckily just about every major JS library has a types package (whether official or community-run)

unlock-app/src/constants.ts
cnasc: An example of migrating a function to have type annotations. Before this change, the tests complained about the implicit `any` type.

It is not an error to manually specify the `any` type, but you really shouldn't.

cnasc: Going to comment here, since GitHub doesn't let you comment on files that only had their extensions changed.

For the `CreatorLock` component, migrating it was as simple as changing the file extension to `.tsx`. Then you just have to suppress any errors--ideally by annotating the correct types for each declaration. Most of our files are not enormous, so doing this incrementally should not be a great hardship.

unlock-app/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
```
cnasc: This line allows us to have a hybrid JS/TS project. If we get to 100% TS in the future, we should set this to `false` to avoid the introduction of any additional JS.

unlock-app/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "moduleResolution": "node",
+    "noEmit": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "preserveConstEnums": true,
+    "removeComments": false,
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "strict": true,
+    "target": "esnext"
```
cnasc: This is a question for now, we may wish to choose a different compilation target.

unlock-app/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noEmit": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "preserveConstEnums": true,
+    "removeComments": false,
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "strict": true,
```
cnasc:
```
Enable all strict type checking options.
Enabling --strict enables
  --noImplicitAny,
  --noImplicitThis,
  --alwaysStrict,
  --strictBindCallApply,
  --strictNullChecks,
  --strictFunctionTypes and
  --strictPropertyInitialization
```

See: https://www.typescriptlang.org/docs/handbook/compiler-options.html

unlock-app/src/constants.ts
cnasc: It is, and I really should have kept a consistent type of declaration between them.

unlock-app/next.config.js
cnasc: Added a comment just after your question--the TypeScript compiler actually emits sourcemaps, so the next extension shouldn't be necessary.

unlock-app/src/constants.ts
julien51: > It is not an error to manually specify the any type, but you really shouldn't.

Is there an eslint (not sure how that would work because eslint run on transpiled files?) to enforce that we do not use `any`

julien51:
> Continue building with TS
Yes!

> Perhaps we should have a cultural rule that when we start work on an area of the codebase that hasn't had types added yet, we first do a PR to add those types?
I want to make sure we keep track of the fact that product should come before... so let's remain practical on this. That said TS-porting can/should be a good task to work on when blocked on another thing.

> Some low-hanging fruit on that may be just to install any type definitions we can find for any dependencies we use. (I've already added types for Jest, React, and Redux since they were necessary for the build)
Yes!

> New subrepos should use TypeScript from the beginning
Works!

One more thing: could you just see if some of the locksmith is as easily converting?

unlock-app/src/constants.ts
cnasc: TSLint has a `no-any` rule. I haven't brought TSLint in yet, and I think we'll have to have a more involved discussion around the rules we'd want to enable, but this particular case is definitely something we can handle.

cnasc:
I'll take a look at locksmith now, I think if anything it should be a little easier to transition, but we'll see.

cnasc:
I spent a few hours yesterday looking at moving locksmith into TS. It was not difficult to simply make `tsc` a part of the build process since locksmith runs on node and doesn't require any bundling or transpilation. So `tsc` just output the unmodified JS files into a build folder, and I updated the run script to watch that folder.

Because that's not a very interesting result, I then decided to start migrating a few files into TS. That revealed some weirdness with the build process. Essentially, some files were importing things the wrong way, but `tsc` wouldn't report that on a full build. You'd see an error in the IDE, and you'd see an error if you compiled the file individually, but not when you do a full build. I'm certain that can't be correct, so I want to do a little more research before pushing any locksmith changes.

Additionally, from a tooling perspective in order to maintain an automatic-reload workflow for locksmith some other packages are necessary (Next handles all that for us in unlock-app, but for locksmith we have to orchestrate that ourselves).

julien51:
So, what's your final take on that?

cnasc:
The plan is to merge a minimal TS configuration with master by the end of the week for `unlock-app` and `locksmith`, unless there are any objections.

PR #2004, opened 2019-03-11
Deprecating the global error consumer by first using errors passed from the redux store to the consumer

julien51:
We are building an approach where rather than *detecting* errors in a context producer, we use redux to
dispatch them.
When they're fatal the global error consumer will just show them and block the app.

### DIFF
--- a/unlock-app/src/components/interface/GlobalErrorConsumer.js
+++ b/unlock-app/src/components/interface/GlobalErrorConsumer.js
@@ -1,12 +1,14 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-
+import { connect } from 'react-redux'
 import { GlobalErrorContext } from '../../utils/GlobalErrorProvider'
 import Layout from './Layout'
 import { mapErrorToComponent } from '../creator/FatalError'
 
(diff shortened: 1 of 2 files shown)


julien51:
Hum. I apparently added 1 untested line. Trying to uncover which one ;)

unlock-app/src/components/interface/GlobalErrorConsumer.js
```diff
@@ -1,12 +1,14 @@
...
-
+import { connect } from 'react-redux'
 import { GlobalErrorContext } from '../../utils/GlobalErrorProvider'
 import Layout from './Layout'
 import { mapErrorToComponent } from '../creator/FatalError'
 
+// DEPRECATED we should not use the Context to detect errors
 const Consumer = GlobalErrorContext.Consumer
 
+// TODO: move that logic back to <GlobalErrorConsumer> once the Context logic is gone
```
cellog: I would recommend we keep this behavior, as it allows changing the way errors are displayed. Although the paywall is splitting, we will add new apps that may desire to keep this feature. Your call, though, since it will soon be obsolete for the dashboard.

Issue #1890, opened 2019-03-05
Integration testing


cellog:
The unlock app needs integration tests that load the full application to ensure that the frontend works with smart contracts and the paywall iframe properly communicates with the underlying paywall application

julien51:
This previous issue has a list of integration tests which we probably need: https://github.com/unlock-protocol/unlock/issues/1399

julien51:
I think we can close this issue/epic because it is a bit too vague for now and we have indeed implemented a bunch of integration tests!

PR #2009, opened 2019-03-12
developer-friendly script for running integration tests locally

cellog:
# Description

This PR introduces a new script, which is designed to make it easy to run the integration tests locally in a docker context repeatedly. It starts by using `docker down` to remove any running instances of the unlock integration tests, mainly `unlock`, `ganache-integration`. This also removes networks (a critical step). Next, it builds (or re-builds) the docker images locally using local cache, and finally it runs the integration tests

# Issues

Refs #1890 

# Checklist:


julien51:
Please add nore info inside the file itself about its purpose and how to use it ;)

cellog:
> Please add nore info inside the file itself about its purpose and how to use it ;)

done, PTAL

scripts/local-docker-integration-tests.sh
```diff
@@ -0,0 +1,35 @@
...
+export DB_HOSTNAME='db'
+export CI=true
+export HTTP_PROVIDER='ganache-integration'
+export LOCKSMITH_URI='http://locksmith:8080'
+export PAYWALL_URL=http://unlock:3000/paywall
+export PAYWALL_SCRIPT_URL=http://unlock:3000/static/paywall.min.js
+
+# if the integration test images are running, this ensures we remove any state
+# prior to attempting to run the tests. This line is critical!
+docker-compose -f $DOCKER_COMPOSE_FILE down --volumes
```
akeem: Is `volumes` needed here?

cellog: I guess not, we don't define any do we? I'll pull that out

cellog:
> I worry about the duplicate information between this file and `integration-tests.sh`.
> Could we only have the strictly necessary stuff in here and then invoke integration-tests.sh as the last step?

sure, I'll look into that, good idea!

cellog:
ok @julien51 PTAL.

We ended up needing the environment variables in both scripts, so I extracted them out. The odd syntax:

```bash
. $REPO_ROOT/scripts/integration-envs.sh
```

is needed to source the env variables. It's not at all intuitive, so I added a comment above to help clarify.

cellog:
update: this approach does not work. I don't think it's worth the effort to remove duplication, but I will have it invoke the integration_tests.sh script

PR #2018, opened 2019-03-12
Changing the shape of Errors

julien51:
Our errors are just a string which means that it is hard to pass data relevant to the error which
could be helpful for the user. The GlobalErrorProvider/Consumer uses a metadata object, but since
we're moving away from that approach to a redux based error handling we need a similar capaibility.

This PR has more changes that I'd like to see but I could not find a better way to split it.

### DIFF
--- a/unlock-app/src/reducers/errorsReducer.js
+++ b/unlock-app/src/reducers/errorsReducer.js
@@ -25,16 +25,10 @@ const errorsReducer = (state = initialState, action) => {
     if (!action.error) return initialState
 
     // If the error to be reset is not in the list, nothing changes
-    if (!state.map(error => error.name).includes(action.error)) return state
+    if (!state.find(error => action.error === error.name)) return state
 
     // Otherwise, only push to new state the all the other errors
-    const newState = []
-    state.forEach(error => {
-      if (error.name !== action.error) {
-        newState.push(error)
-      }
-    })
-    return newState
+    return state.filter(error => error.name !== action.error)
   }
 
   return state
--- a/unlock-app/src/reducers/errorsReducer.js
+++ b/unlock-app/src/reducers/errorsReducer.js
@@ -10,16 +10,30 @@ const errorsReducer = (state = initialState, action) => {
   }
 
   if (action.type === SET_ERROR) {
-    if (state.includes(action.error)) return state
-    return [...state, action.error]
+    if (state.map(error => error.name).includes(action.error)) return state
+    return [
+      ...state,
+      {
+        name: action.error,
+        data: action.data,
+      },
+    ]
   }
 
   if (action.type === RESET_ERROR) {
+    // The no error is passed, reset all errors
     if (!action.error) return initialState
-    if (!state.includes(action.error)) return state
-    const newState = [...state]
-    const newIndex = newState.indexOf(action.error)
-    newState.splice(newIndex, 1)
+
+    // If the error to be reset is not in the list, nothing changes
+    if (!state.map(error => error.name).includes(action.error)) return state
+
+    // Otherwise, only push to new state the all the other errors
+    const newState = []
+    state.forEach(error => {
+      if (error.name !== action.error) {
+        newState.push(error)
+      }
+    })
     return newState
   }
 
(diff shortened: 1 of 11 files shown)


unlock-app/src/reducers/errorsReducer.js
```diff
@@ -10,16 +10,30 @@ const errorsReducer = (state = initialState, action) => {
...
   if (action.type === RESET_ERROR) {
+    // The no error is passed, reset all errors
     if (!action.error) return initialState
-    if (!state.includes(action.error)) return state
-    const newState = [...state]
-    const newIndex = newState.indexOf(action.error)
-    newState.splice(newIndex, 1)
+
+    // If the error to be reset is not in the list, nothing changes
+    if (!state.map(error => error.name).includes(action.error)) return state
```
cellog:
```suggestion
if (!state.find(error => action.error === error.name)) return state
```

unlock-app/src/reducers/errorsReducer.js
```diff
@@ -10,16 +10,30 @@ const errorsReducer = (state = initialState, action) => {
   }
 
   if (action.type === SET_ERROR) {
-    if (state.includes(action.error)) return state
-    return [...state, action.error]
+    if (state.map(error => error.name).includes(action.error)) return state
```
cellog:
```js
if (state.find(error => action.error === error.name && action.data !== error.data)) return state
```

also need to test for the equality of data

unlock-app/src/reducers/errorsReducer.js
```diff
@@ -10,16 +10,30 @@ const errorsReducer = (state = initialState, action) => {
...
-    if (!state.includes(action.error)) return state
-    const newState = [...state]
-    const newIndex = newState.indexOf(action.error)
-    newState.splice(newIndex, 1)
+
+    // If the error to be reset is not in the list, nothing changes
+    if (!state.map(error => error.name).includes(action.error)) return state
+
+    // Otherwise, only push to new state the all the other errors
+    const newState = []
```
cellog:
```js
return state.filter(error => error.name !== action.error)
```

can replace all that stuff

unlock-app/src/reducers/errorsReducer.js
julien51: Nope. I actually think this is fine. The data is not defining what the error is.

julien51:
So, I think we should actually *not* use error inside of error. 
Name is probably fine because this is actually not what is displayed to users, thanks to the mapping that we have for fatal errors for example (and we should have the same eventually for other errors but I did not want to do that in that already big PR.).

julien51:
Changes fixed! PTAL

unlock-app/src/reducers/errorsReducer.js
cellog: ok. this is a nit, but still more efficient to use:

```js
if (state.find(error => action.error === error.name)) return state
```

However, there is an example of a situation where the data defines what the error is:

```
user navigates to paywall, but is on rinkeby
"wrong network" error is dispatched, with data being rinkeby
user quickly tries to change to mainnet, but accidentally clicks Ropsten
"wrong network" error is dispatched, with data being ropsten
error message does not update
```

Technically, it is still the same kind of error, but if we don't update the state, the change in the data is lost in the UI's display of the error.

One possible solution is to make separate errors for each data possibility, but I think in this case it may be difficult, as a user could use their own custom ethereum network, and we would not have an error for it.

julien51: Well you can;t be on 2 networks at once, can you? So in practice the metadata helps us make a better tailored message to the user but does not define the error.

julien51:
Think of the data as stuff for the UI...

unlock-app/src/reducers/errorsReducer.js
julien51: When swicting the network the app  will reload... so state is reset no matter what.

cellog: This is just an example. If we don't save the error, the change in data is lost to the UI. I'll approve this, but with a strong recommendation of changing this since it will be a difficult thing to debug if it ever happens in production.

julien51:
@greg: you deleted the comment from CLA so I cannot click on it to recheck that I signed it :( 😭

cellog:
Ack, I clicked recheck and it said it had passed :/

PR #2019, opened 2019-03-12
console.log helper for debugging integration tests

cellog:
# Description

This helper script is for debugging integration test problems locally. Usage:

```js
const debugPage = require('../helpers/debugging')

describe('The Unlock Dashboard', () => {
  beforeAll(async () => {
    debugPage(page, true) // pass false to disable
    await page.goto(url('/dashboard'))
  })
// ...
```

It will output all usages of console (log, error, info, etc.) inside the integration test context. In addition to displaying the output of console.log, it shows the URL the console references (if any), and it properly displays simple types that were logged.

Initially, the helper used the `text()` method of the `ConsoleMessage` but it turns out that for simple things like logging a `lock`, this would display `JSHandle@object` instead of the actual console output. Using `jsonValue()` displays the full object, which is much more useful for debugging.

With this helper, you can put console.log() calls in either the tests to run in evaluate:

```js
 await page.evaluate(() => console.log(document.querySelector('#whatever'))
```

or in the unlock app source code, and both will be displayed.

Here is a sample:

```js
  it('should load the creator dashboard', async () => {
    page.evaluate(() => console.log({ this: 'is', a: 'complex', object: true }))
    await expect(page).toMatch('Creator Dashboard')
  })
```

and the output:

Screen Shot 2019-03-12 at 10 56 51 AM

# Issues

Refs #1890 

# Checklist:

### DIFF
--- a/tests/.eslintrc.js
+++ b/tests/.eslintrc.js
@@ -45,5 +45,6 @@ module.exports = {
         "eol-last": [
             "error"
         ],
+        "no-console": [ "error" ]
     }
 };
--- a/tests/helpers/debugging.js
+++ b/tests/helpers/debugging.js
@@ -1,0 +1,12 @@
+module.exports = function debugPage(page, debug = false) {
+  if (debug) {
+    page.on('console', async (msg) => {
+      const locationInfo = msg.location()
+      const fileInfo = `${locationInfo.url ? locationInfo.url : ''}${
+        locationInfo.lineNumber !== undefined ? ` Line ${locationInfo.lineNumber}` : ''
+      }`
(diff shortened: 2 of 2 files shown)


cellog:
that brings up an interesting point: I don't think linting is enabled for tests/ yet!

cellog:
> I'm game for this, if we also add a lint rule to ensure that its usage doesn't make its way into master

OK, so I pushed a change to make console a linting error in the tests. There is no way to explicitly disable the use of debugPage, unless we write a custom linter rule (https://eslint.org/docs/developer-guide/working-with-rules). I'll go ahead and say I don't think it's worth it at this point, but if it becomes a problem, we can certainly take those steps. PTAL @akeem

Issue #1653, opened 2019-02-22
connect to walletService provider just-in-time


cellog:
Currently, the walletService attempts to connect to the provider right away in order to retrieve the user account, and also the network the wallet is using. This results in some odd UX on two separate occasions.

1. When viewing unlock-protocol.com, metamask immediately asks to log in, even though there is absolutely nothing ethereum-related until one clicks on the dashboard
2. When viewing a locked paywall. Until the user clicks on the lock, we don't need the account. However, we *do* need the account if the user has purchased a key.

One consideration is whether it might make sense to always store the user's account in localStorage, but also always check to see if we can enable the wallet when the user takes action.

With this system, a user's unlocked content will display if they have ever visited the content with their wallet, regardless of whether they are logged in at that moment.

For the dashboard, the wallet would be requested immediately.

cellog:
love to get your thoughts on this, @julien51 @benwerd @akeem @cnasc

julien51:
> Currently, the walletService attempts to connect to the provider right away in order to retrieve the user account, and also the network the wallet is using. 

I actually think this is good practice because it means we can retrieve information right away and avoid latency... It also "fails" silently if the user does not have a wallet, and will only show an error message if the user tries to load the dashboard and/or click on a "purchase" button on the paywall.

> When viewing unlock-protocol.com, metamask immediately asks to log in, even though there is absolutely nothing ethereum-related until one clicks on the dashboard
Indeed. I have been thinking about this one. But I am honestly not sure this is a big problem? Again, if the user has no wallet, no error is shown and if they have one, they are asked to authorize the app, which they can still decline and use the website (until the dashboard).

> When viewing a locked paywall. Until the user clicks on the lock, we don't need the account. However, we _do_ need the account if the user has purchased a key.

Actually we do need the account when the user clicks on the lock because they do that to purchase the key?

cellog:
> > When viewing a locked paywall. Until the user clicks on the lock, we don't need the account. However, we _do_ need the account if the user has purchased a key.
>
> Actually we do need the account when the user clicks on the lock because they do that to purchase the key?

Yes. We only need it to *purchase* a lock, not to display it. So it is required to purchase, but not to display the lock.

It's also currently required to retrieve keys, but as demonstrated in the coinbase wallet stuff, that is not necessary. We can store the account in localStorage and use it without authorizing on a new URL to retrieve keys.

[...]

PR #2041, opened 2019-03-12
fixing the connect function for walletMiddleware

julien51:
We should only not try to connect when the provider is set previously.

### DIFF
--- a/unlock-app/src/middlewares/walletMiddleware.js
+++ b/unlock-app/src/middlewares/walletMiddleware.js
@@ -151,7 +151,10 @@ export default function walletMiddleware({ getState, dispatch }) {
 
   return function(next) {
     // Connect to the current provider
-    walletService.connect(getState().provider)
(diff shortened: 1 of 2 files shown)


unlock-app/src/middlewares/walletMiddleware.js
```diff
@@ -151,7 +151,10 @@ export default function walletMiddleware({ getState, dispatch }) {
 
   return function(next) {
     // Connect to the current provider
-    walletService.connect(getState().provider)
+    // We connect once the middleware has been initialized, using setTimout (warning: fragile?)
```
cellog: This solution should be fine for now, and I don't believe it is fragile even though it feels shitty :)

When we implement #1653 it will move the connect call out of this place in the middleware anyways, as we will want to call it when it is needed instead. In other words, it will be triggered by the UI instead of by redux directly.

PR #2050, opened 2019-03-13
Experiment: build Locksmith with TypeScript

cnasc:
# Description

Following on PR #1996, where we looked at a simple migration of `unlock-app` to TypeScript, this PR does the same for `locksmith`.

The major difference here is that `unlock-app` has a lot more "magic" in its build process because Next takes care of so much for us. This effort has more done from scratch, but I think you'll see that overall it's not too bad.

Most of my commentary from the last PR also applies here, but I will step through this PR with additional comments on parts I found tricky or interesting.

# Issues

Refs #1996 #1449 

# Checklist:


locksmith/__tests__/app.test.ts
```diff
@@ -1,7 +1,7 @@
-const request = require('supertest')
-const app = require('../src/app')
-const Lock = require('../src/lock')
-const Transaction = require('../src/transaction')
+import request = require('supertest')
```
cnasc: TypeScript was fairly keen on having me change from `const x = require...` to `import x = require...`. 

It seems that is the current best practice for importing non-ES6 modules: https://stackoverflow.com/questions/29596714/new-es6-syntax-for-importing-commonjs-amd-modules-i-e-import-foo-require

locksmith/__tests__/app.test.ts
```diff
@@ -1,7 +1,7 @@
...
+import app from '../src/app'
```
cnasc: For modules we create internally, there is no reason not to make them be ES6, so I converted a few of them to see how it would go. More on that later.

locksmith/__tests__/app.test.ts
```diff
@@ -76,16 +76,15 @@ describe('Requesting lock details', () => {
     test('should return an OK status code', async () => {
       expect.assertions(1)
       let response = await request(app).get(`/lock/${testLockDetails.address}`)
-      expect(response.statusCode).toBe(200)
+      expect(response.status).toBe(200)
```
cnasc: This was interesting. Once I converted this file to TS and installed the typings for `supertest`, it complained that `statusCode` was not a valid property. Changing it to  `status` satisfied the compiler, and the tests still pass so 🤷‍♂️

locksmith/__tests__/app.test.ts
```diff
@@ -97,7 +96,7 @@ describe('Setting lock details', () => {
 
     await request(app)
       .post('/lock')
-      .set('Accept', /json/)
+      .set('Accept', 'json')
```
cnasc: No idea why these were RegExps in the first place, but TS caught that as a `TypeError` and I corrected it.

locksmith/__tests__/signatureValidationMiddleware.test.ts
```diff
@@ -1,7 +1,8 @@
-var httpMocks = require('node-mocks-http')
-var signatureValidationMiddleware = require('../src/signatureValidationMiddleware')
+import httpMocks = require('node-mocks-http')
+import express = require('express')
+import signatureValidationMiddleware = require('../src/signatureValidationMiddleware')
 
-let request, response
+let request: httpMocks.MockRequest<express.Request>, response: httpMocks.MockResponse<express.Response>
```
cnasc: Generics! Was slightly difficult to figure out what the correct type was here; TS didn't infer it so it was  `any`

locksmith/config/config.js
```diff
@@ -7,10 +7,10 @@ module.exports = {
...
   },
   test: {
-    username: process.env.DB_USERNAME,
-    password: process.env.DB_PASSWORD,
-    database: process.env.DB_NAME,
-    host: process.env.DB_HOSTNAME,
+    username: 'locksmith_test',
+    password: 'password',
+    database: 'locksmith_test',
+    host: 'localhost',
```
cnasc: You can safely disregard this, these are just the creds used for local testing

locksmith/jest.config.js
```diff
@@ -9,4 +9,5 @@ module.exports = {
       statements: 91.63,
     },
   },
+  preset: 'ts-jest/presets/js-with-ts',
```
cnasc: Allows `jest` to process both JS and TS files

locksmith/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "server.js",
   "scripts": {
     "predev": "NODE_ENV=development sequelize db:migrate",
-    "dev": "nodemon server.js",
+    "dev": "tsc-watch --onSuccess 'node build/server.js' --onFailure 'echo Beep! Compilation Failed'",
```
cnasc: This command will run `tsc` in response to any change of watched files. It does incremental builds, so this is quick.

When the build finishes, it will run the compiled server. Any time it rebuilds, it kills the last server process and starts a new one.

When it fails it prints a cute message.

locksmith/package.json
```diff
@@ -14,7 +14,8 @@
     "db:migrate": "sequelize db:migrate",
     "fail-pending-changes": "../scripts/pending-changes.sh",
     "ci": "npm run lint && npm test && npm run reformat && npm run fail-pending-changes",
-    "build": "echo 'Nothing to build'"
+    "build": "echo 'Nothing to build'",
+    "tsc": "tsc"
```
cnasc: We can probably turn `build` into an alias for `tsc`?

locksmith/package.json
```diff
@@ -28,6 +29,12 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@types/cors": "^2.8.4",
+    "@types/es6-promise": "^3.3.0",
+    "@types/express": "^4.16.1",
+    "@types/jest": "^24.0.11",
+    "@types/nock": "^9.3.1",
+    "@types/supertest": "^2.0.7",
```
cnasc: Added some types definitions for third-party libs

locksmith/package.json
```diff
@@ -37,22 +44,25 @@
     "ethers": "^4.0.26",
     "express": "^4.16.4",
     "express-jwt": "^5.3.1",
-    "jest": "^23.6.0",
+    "jest": "^24.4.0",
```
cnasc: Jest had to be upgraded to work with `ts-jest`

locksmith/src/app.ts
```diff
@@ -19,4 +19,4 @@ app.use('/', transactionRouter)
 app.use('/', lockRouter)
 app.use('/block', blockRouter)
 
-module.exports = app
+export default app
```
cnasc: Pretty easy to convert this to an ES6 module!

locksmith/package.json
akeem: Are these used after the code has been transpiled? Should we move these to the development dependencies?

locksmith/src/lock.js
```diff
@@ -2,3 +2,4 @@ const models = require('../models')
 
 models.Lock.removeAttribute('id')
 module.exports = models.Lock
+export default models.Lock
```
cnasc: For the moment I kept both kinds of exports on these, the line addition is just so I could import as an ES6 module from TS

locksmith/tsconfig.json
```diff
@@ -0,0 +1,66 @@
...
+    // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
+
+    /* Experimental Options */
+    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+  },
+  "exclude": [
+    "node_modules",
+    "build",
+    "__tests__"
```
cnasc: Don't build the `build` folder, that will cause a weird circularity. Also don't build `__tests__` because then Jest will detect both of the `__tests__` folders and run each test twice.

locksmith/package.json
cnasc: There are a handful of cases where it's beneficial to have the type definitions in the regular dependencies: https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies

I think we should be safe keeping them as dev dependencies.

locksmith/__tests__/app.test.ts
julien51: what happens if you use the regular `import request from 'supertest'`?

locksmith/package.json
julien51: yup.

locksmith/package.json
julien51: so, dev dependencies will not be installed by CI... which means that if they are required to run tests, then the tests will fail.

locksmith/src/lock.js
julien51: I think this created conflicts :( sorry.

locksmith/package.json
cnasc: That's what we did for #2076, that way locksmith's TypeScript will get built before deploy

Issue #1205, opened 2019-01-24
Splitting up the creator dashboard and the paywall app


julien51:
Both the creator dashboard an the paywall are currently inside a single app... I believe we should strongly consider splitting them in 2 different apps.

The main rationale for that is that the creator dashboard does a lot more things and loads a lot more data than the paywall... which makes the paywall app particularly bloated and slow while it needs to be especially speedy and efficient.

I am sure there would be shared components and libraries between the 2, but I think we should start considering splitting them because the longer we wait the more painful it will be.

cnasc:
This is definitely a smart idea

PR #2052, opened 2019-03-13
Fix serving, enhance logging in paywall

cellog:
# Description

This PR has 3 small but significant changes:

1. the source served was incorrectly looking in `paywall` instead of `paywall/src` which resulted in a `404` for every page requested
2. logging now prepends with `PAYWALL ` in order to differentiate from logging of the unlock-app. This is for `npm run dev` purposes, so I could modify it to check `UNLOCK_ENV` if desired
3. the paywall should only match paths that start with a lock address, and since we no longer serve from `https://<paywall server>/paywall/` but instead from `https://<paywall server>/` the check is adjusted to look for a lock address prefix.

# Issues

Refs #1205 

# Checklist:

### DIFF
--- a/paywall/src/_server.js
+++ b/paywall/src/_server.js
@@ -12,7 +12,7 @@ function _server(port, dev) {
 
     app.prepare().then(() => {
       let server = createServer((req, res) => {
-        console.info(`PAYWALL ${req.method} ${req.url} > ${res.statusCode} `)
+        console.info(`${req.method} ${req.url} > ${res.statusCode} `)
         try {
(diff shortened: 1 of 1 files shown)


cellog:
> > logging now prepends with PAYWALL in order to differentiate from logging of the unlock-app. This is for npm run dev purposes, so I could modify it to check UNLOCK_ENV if desired
> 
> Let's make things consistent by indeed using `UNLOCK_ENV` here please.

Removed the prefix, as we discussed, PTAL @julien51

PR #2054, opened 2019-03-13
removed handling of FATAL_NO_USER_ACCOUNT inside global error provider

julien51:
In our quest to remove the Global Error Provider, I am removing the last bit that was used by it:
creating a FATAL_NO_USER_ACCOUNT. Note that this error is disaptched when the user tries to perform
and action.

### DIFF
--- a/unlock-app/src/__tests__/pages/pages.test.js
+++ b/unlock-app/src/__tests__/pages/pages.test.js
@@ -8,6 +8,10 @@ import { About } from '../../pages/about'
 import { Dashboard } from '../../pages/dashboard'
 import { pageTitle, ETHEREUM_NETWORKS_NAMES } from '../../constants'
 import createUnlockStore from '../../createUnlockStore'
+import { ConfigContext } from '../../utils/withConfig'
+import configure from '../../config'
+
+const config = configure()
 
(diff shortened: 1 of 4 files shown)


unlock-app/src/__tests__/pages/pages.test.js
```diff
@@ -59,15 +63,18 @@ describe('Pages', () => {
         privateKey: 'deadbeef',
         balance: '200',
       }
+      const ConfigProvider = ConfigContext.Provider
       rtl.render(
-        <Provider store={store}>
```
julien51: Previously the global error provider would actually show a "missing account" error here so it was never reaching the point in the react tree where a component needs the config... since the global error provider does not do that anymore, I've had to add the `ConfigProvider` in this test.

julien51:
After that one is merged, the final step will be to completely get rid of the global error provider now that it does not do anything anymore!

cnasc:
> In our quest to remove the Global Error Provider, I am removing the last bit that was used by it:
> creating a FATAL_NO_USER_ACCOUNT. Note that this error is disaptched when the user tries to perform
> and action.

I think there's a typo in there, which actions is the error meant to be displayed for?

PR #2058, opened 2019-03-13
Don't poll for scroll position unless the paywall is locked

cellog:
# Description

The `paywall.min.js` script unnecessarily polls for changes in scroll position and attempts to post the scroll position even when the paywall is unlocked. This can be a major performance problem on slow phones. This PR enables polling if the paywall is marked as locked, and disables it if it is marked as unlocked.

# Issues

Refs #1205 

# Checklist:

### DIFF
--- a/paywall/src/paywall-builder/build.js
+++ b/paywall/src/paywall-builder/build.js
@@ -77,19 +77,19 @@ export default function buildPaywall(window, document, lockAddress, blocker) {
   const origin = new window.URL(paywallUrl).origin
 
   let locked = false
-  let disablePolling
+  let disableScrollPolling = () => {}
   window.addEventListener(
     'message',
     event => {
       if (event.data === POST_MESSAGE_LOCKED && !locked) {
         locked = true
         show(iframe, document)
-        disablePolling = scrollLoop(window, document, iframe, origin)
+        disableScrollPolling = scrollLoop(window, document, iframe, origin)
         blocker.remove()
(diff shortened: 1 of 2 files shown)


paywall/src/paywall-builder/build.js
```diff
@@ -17,19 +17,24 @@ const baseBannerHeight = window => {
   // can use it as the basis to scroll from
   return minHeightPct > 30 ? minHeightPct : 30
 }
+let scrollPolling = true
 // iOS allows the page to scroll even when the paywall is up. Our
```
cnasc: This actually reminds me that right now, the scrollPosition stuff is only relevant on iOS devices. I believe @julien51 was interested in making it the default behavior across the board. I guess we should decide whether that's still a goal and whether it justifies making an issue.

cellog: I think that this should be default. Any device that allows scrolling should be fubared by the paywall

cnasc: Oh no, I agree with that. I think the discussion was around whether to always allow scrolling so that behavior is consistent across as many devices as possible instead of having iOS able to scroll (and the paywall grows to obscure) and having every other device stay frozen.

cellog: ok. That's something to open an issue for, yes. This only fixes a potential problem where if the iframe is small, we don't want to post messages

PR #2077, opened 2019-03-14
Cap refund on cancel Key

### LINKED ISSUE #2000: Cap the max refund a Key owner can get on `cancelAndRefund`
**Describe the bug**
When a user cancels their Key, they get a prorated refund based on the time remaining.  If the time remaining is greater than the default `expirationDuration` than the user could potentially get a refund which is larger than `keyPrice`.

This may or may not be desirable.

**To Reproduce**
This could happen if the user:
 - Purchases more time by buying Keys for the same Lock (either directly from the contract or 2nd hand).
 - The Key owner was gifted a key by the Lock owner using `grantKey`.  The Lock owner may have intended to gift a Key which never expires and have that backfire when the recipient drains the Lock of all its funds instead.

**Additional context**
The formula today is:
`keyPrice * (timeRemaining / expirationDuration) - penalty`

We could change it to:
`min(keyPrice, keyPrice * timeRemaining / expirationDuration) - penalty`

hardlydifficult:
# Description

If a key owner has a long time remaining relative to the expirationDuration, `cancelAndRefund` would have previously refunded more than `keyPrice`.  This may be desirable but I suspect it is not, and it's a risk if a Lock owner uses `grantKey(address, -1)` without realizing the implications - this avoids that concern.

A bit more detail in the issue.  We can reject this if I'm wrong here.

# Issues

Fixes #2000 
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinRefunds.sol
+++ b/smart-contracts/contracts/mixins/MixinRefunds.sol
@@ -102,8 +102,12 @@ contract MixinRefunds is
     Key storage key = _getKeyFor(_owner);
     // Math: safeSub is not required since `hasValidKey` confirms timeRemaining is positive
     uint timeRemaining = key.expirationTimestamp - block.timestamp;
-    // Math: using safeMul in case keyPrice or timeRemaining is very large
-    refund = keyPrice.mul(timeRemaining) / expirationDuration;
+    if(timeRemaining >= expirationDuration) {
+      refund = keyPrice;
+    } else {
+      // Math: using safeMul in case keyPrice or timeRemaining is very large
+      refund = keyPrice.mul(timeRemaining) / expirationDuration;
+    }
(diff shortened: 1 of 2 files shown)


smart-contracts/contracts/mixins/MixinRefunds.sol
```diff
@@ -102,8 +102,12 @@ contract MixinRefunds is
     Key storage key = _getKeyFor(_owner);
     // Math: safeSub is not required since `hasValidKey` confirms timeRemaining is positive
     uint timeRemaining = key.expirationTimestamp - block.timestamp;
-    // Math: using safeMul in case keyPrice or timeRemaining is very large
-    refund = keyPrice.mul(timeRemaining) / expirationDuration;
+    if(timeRemaining >= expirationDuration) {
```
julien51: 👍 This is indeed a good fix!

Issue #1997, opened 2019-03-11
Charge a fee on transfer


hardlydifficult:
**User stories**
 - As a Lock owner, I would like some compensation when Keys are traded 2nd hand.
 - As a user interested in content, I would like the content creator to be compensated even when I buy a discounted Key 2nd hand.
 - As a Lock owner, I would like control over this fee including the ability to turn it off if I prefer.
 - As a Key owner, I want any pre-paid fee refunded if the key does not sell or if the Lock owner upgrades to a new Lock version for the same content (or have the pre-paid fee be recognized on the new Lock version).

**Requirements**
For ETH locks only:
 - Allow the key owner to optionally pay the fee on approve
 - Require the fee either be pre-paid or included with the tx to transfer
 - Allow the key owner to refund any unused pre-paid fees
 - Allow the lock owner to refund any unused pre-paid fees

For ERC20 locks only:
 - Charge fee from the Key seller at time of transfer (the user must have previously called `approve` on the token's contract)

For both:
 - The fee must be compatible with trading on a 0x marketplace
 - Allow the lock owner to change the percent fee (possibly to 0%)
 - Provide function to estimate fee
 - Provide function for isFeeCovered (or similar, this returns true if the user pre-paid enough or has called approved and has enough tokens in their wallet to cover the fee)

**Questions/concerns**
 - Supporting this feature for ERC20 based locks is much more simple than ETH based locks.  Do we consider only implementing this for ERC20?  Or maybe for ETH based locks, this fee is charged in WETH (while normal key purchases could continue in ETH)?

**Assumptions**
 - The fee is defined as a percent of keyPrice, prorated (i.e. considering % of time remaining on the key).
 - The fee is paid by the seller, not the buyer.
 - The fee is charged in the same token as normal key sales (e.g. an ERC20 lock cannot accept this fee paid in ETH).

**Risks**
 - If a key owner pre-pays the transfer fee and then the Lock owner withdraws funds, refunding will not be possible until the Lock acquires more funds.

**Implementation**
This seems very straightforward for the ERC20 based locks - add a function to estimate the fee and use that to collect the funds on `transfer`... and if that fails then revert.

For ETH based locks we would need a new data field added to the contract to track pre-paid fees.  Then it's similar to above, but with an additional option to refund anything remaining.  I added a requirement above for allowing the Lock owner to initial the refund as well, as this seems important for enabling them to comfortably perform an upgrade to a new Lock version.

PR #2078, opened 2019-03-14
Define transferFee and allow it to be changed

hardlydifficult:
# Description

Defining a transfer fee amount, and allowing the Lock owner to change it from its default value.

The transfer fee itself is not yet implemented.  This is a first step towards completing that task.

# Issues

Fixes #
Refs #1997 

# Checklist:

### DIFF
--- a/smart-contracts/contracts/UnlockErrors.sol
+++ b/smart-contracts/contracts/UnlockErrors.sol
@@ -19,6 +19,10 @@ contract UnlockErrors {
   // The specified address is not valid (i.e. must not be 0).
   string public constant INVALID_ADDRESS = 'INVALID_ADDRESS';
 
+  // The specified rate is not valid, the denominator must be greater than 0.
+  // To set the rate to 0, set the numerator to 0 and the denominator to 1.
+  string public constant INVALID_RATE = 'INVALID_RATE';
+
   // The user does not have a valid key, they may or may not have purchased one previously.
   string public constant KEY_NOT_VALID = 'KEY_NOT_VALID';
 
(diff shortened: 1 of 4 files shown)


julien51:
First: I think ideally it needs to be a %. Does that make this hard-er to implement?
The default should be a rather small %, like maybe 5%.

hardlydifficult:
There's no decimal support directly, using a denominator only is a gas optimization.  But maybe that's overkill and we should use percent instead.  To that end, the refund penalty is currently implemented the same way.

We could simply do a `value.mul(percent).div(100)` to work with percents as people expect - but they must be in whole numbers (i.e. 0.5% is not possible).  ....unless we have them enter basis points instead of percent.

The most flexible approach is to accept both a numerator and a denominator (as two uints).  This would allow any value to be expressed.

hardlydifficult:
Oh btw if we go with numerator / denominator for the frontend a library like https://www.npmjs.com/package/fractional does the conversion to make it easy for users to simply enter x%

julien51:
> The most flexible approach is to accept both a numerator and a denominator (as two uints). This would allow any value to be expressed.

👍 let's do that!

hardlydifficult:
done.  Ready for review.

I'll go ahead and change the refund rate as well to be consistent.

hardlydifficult:
It'll default to 5%, and then the Lock owner can make a tx to change it.  We should support changing it to 0 as well.

Looking at this again, maybe we should require the denominator is > 0 on update

hardlydifficult:
Added new require statement and a test for it.

PR #2095, opened 2019-03-15
Rename outstandingKeys to totalSupply

### LINKED ISSUE #2082: Add ERC721 enumerable extension
We have functionality already implemented which is similar to APIs suggested in the ERC721 spec.

Consider using: https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC721/IERC721Enumerable.sol

Maybe https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC721/ERC721Enumerable.sol but that includes a lot.  We should take a close look at this class before picking up this task to decide if we can use this or just the interface.

**Additional context**
See also https://eips.ethereum.org/EIPS/eip-721 and search for "enumeration extension"

hardlydifficult:
# Description

I started this task with the intention of implementing the ERC721 enumerable standard, however as I started on this I realized that 2 of 3 functions in that standard do not seem to offer any value for our use case.  However 1 of the functions is `totalSupply` which functionally matches what we had previously called `outstandingKeys` given the current implementation.  

I suggest we match the standard where we can, therefore renaming this function.  However we will not be registering as supporting the enumerable extension with ERC165 since we do not have implementations of the other functions.

The excluded methods:
```
    function tokenByIndex(uint256 _index) external view returns (uint256);
```
For our use case this would return the `_index` provided.

```
    function tokenOfOwnerByIndex(address _owner, uint256 _index) external view returns (uint256);
```
For our use case this requires `_index == 0`, and then returns what we currently call `getTokenIdFor`

We could reconsider and include these functions if we would prefer to implement the enumerable extension.  `tokenByIndex` is basically a no-op, but I could see an argument in favor of switching `getTokenIdFor` to `tokenOfOwnerByIndex` even though `_index` is not relevant for us..

# Issues

Fixes #2082
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinKeys.sol
@@ -201,7 +201,7 @@ contract MixinKeys is
 
   /**
    * Public function which returns the total number of unique owners (both expired
-   * and valid).  This may be larger than outstandingKeys.
+   * and valid).  This may be larger than totalSupply.
    */
   function numberOfOwners()
     public
--- a/smart-contracts/README.md
+++ b/smart-contracts/README.md
@@ -63,8 +63,8 @@ It has the following fields:
 These are the function which change some parameters for the Lock or check its status
 
 - setKeyPrice (TODO) : updates the price of new keys
-- outstandingKeys : get the number of outstanding keys
-- numberOfOwners : get the total number of unique owners (both expired and valid).  This may be larger than outstandingKeys.
+- totalSupply : get the number of outstanding keys
+- numberOfOwners : get the total number of unique owners (both expired and valid).  This may be larger than totalSupply.
 - balanceOf (ERC721) : get the number of keys for a given owner (0 or 1)
 - ownerOf (ERC721) : get the owner of a key (if applicable)
 - keyDataFor : get the value of the data field for a key
--- a/smart-contracts/contracts/interfaces/ILockCore.sol
+++ b/smart-contracts/contracts/interfaces/ILockCore.sol
@@ -112,7 +112,7 @@ interface ILockCore {
 
   /**
    * Public function which returns the total number of unique owners (both expired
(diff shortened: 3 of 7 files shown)


smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -200,14 +198,13 @@ contract MixinKeys is
   }
 
   /**
-   * Public function which returns the total number of unique owners (both expired
+  * @dev Public function which returns the total number of unique owners (both expired
    * and valid).  This may be larger than outstandingKeys.
```
nfurfaro: I need to look at this closer, but wouldn't `outstandingKeys` be a closer match for `totalSupply` than `numberOf Owners` (which may be greater than `outstandingKeys`)?

nfurfaro: I guess `numberOfKeysSold` is what I meant here. But looking closer,
If one key has been purchased on a lock and the owner transfers it to a new owner:
- there are now 2 owners in the ownersAray
- there are technically 2 keys on the lock, both with the same `tokenId`, but one is expired. I feel like we should be treating this as though there's only one key, and in that case `numberOfOwners` would not be equal to `totalSupply`.

nfurfaro: The fact that we technically have 2 keys in this case is more an artifact of the way we exacute a transfer than a desired outcome. At least that's what my caffeine-deprived brain says at the momment;)

hardlydifficult: Yea you are right, my mistake.  Good catch - will switch to replacing outstandingKeys.

hardlydifficult: K fixed.  Thanks again, good catch.

Issue #1852, opened 2019-03-04
PayWall Reader Landing Page


smombartz:
# Description
This page exists to inform readers and content consumers about what Unlock is. Readers should be able to get to this page from a locked as well as unlocked page when clicking on the Unlock logo in the flag or powered buy Unlock underneath the Payment Box.

The text that on this page is located in this doc (feel free to suggest changes)
https://docs.google.com/document/d/14Kzinz9-T-1xUZOYGuwSllld1x2SuhMhJx7f_m4zbKA

# Design
https://www.figma.com/proto/LkWSSnCGlKEITBB5T5NYXW/Unlock-Paywall?node-id=264%3A1159&viewport=2734%2C732%2C0.354305&scaling=min-zoom

julien51:
@cellog have you been able to work on that?

cellog:
Just as soon as the split is finalized it's in the next batch of stories, so perhaps tomorrow or Wednesday!

julien51:
@cellog I think this can soon be closed?

cellog:
as soon as #2108 is merged, it will close it (I put a "Fixes" in the description)

PR #2106, opened 2019-03-15
Layout component for the paywall landing page

cellog:
# Description

Screen Shot 2019-03-15 at 10 16 31 AM
Screen Shot 2019-03-15 at 10 16 21 AM
Screen Shot 2019-03-15 at 10 16 04 AM

# Issues

Refs #1852 

# Checklist:


paywall/src/stories/interface/Layout.stories.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Provider } from 'react-redux'
```
cnasc: Are there components in `Layout` other than the header that rely on the state? The test only passes in the router state, which isn't used in the header, so we can probably get rid of it here.

cellog: yep, just removed it, good catch, PTAL

PR #2115, opened 2019-03-15
Add unlockUrl to configuration

benwerd:
# Description

This provides a value for `unlockUrl` in `runtimeConfiguration`, starting with an env variable if it exists, and falling back to sensible defaults. We'll later use this to make server-side asynchronous callbacks, eg for use in the blogging platform.

# Checklist:


unlock-app/next.config.js
```diff
@@ -19,6 +20,11 @@ const requiredConfigVariables = {
     process.env.UNLOCK_ADDRESS || '0x885EF47c3439ADE0CB9b33a4D3c534C99964Db93', // default for CI
 }
 
+requiredConfigVariables.unlockUrl = unlockUrl(
+  process.env,
```
julien51: I think you need to put an env variable in there too?

unlock-app/src/utils/unlockUrl.js
```diff
@@ -0,0 +1,20 @@
+const unlockUrl = (env, config) => {
+  if (env && env.UNLOCK_URL) return env.UNLOCK_URL
```
julien51: I now understand the thing above better. Would you rather pass `env.UNLOCK_URL` as a variable named `default` maybe? I think it's better if we try to isolate the use of `env`.

julien51: Also I am a bit confused. It looks like this is duplicating the logic in `config.js` ?

benwerd: I didn't see any comparable logic in `config.js`, but happy to be corrected. I originally had it there but had feedback that I should move it into next.config.js, so that's what I did :)

The reason why I passed the whole logic here is so that I could test it - unfortunately I can't write tests for `next.config.js` itself. But yeah, I'll refactor to pass the env property differently.

julien51: HA! I meant the logic below: setting the URL based on the env (staging, dev or prod...)

benwerd:
@julien51 this is now a much smaller PR ;) PTAL

PR #2107, opened 2019-03-15
Refactor log page (now with 40% more TypeScript)

cnasc:
# Description

This PR refactors the log page in preparation for future enhancements. Given the difficulty of testing pages, the component definition has been separated from the page definition. Additionally, this PR is something of a landmark because it's the first actual use of TypeScript in our codebase!

# Issues

Refs #2096 #1691 #1689 

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [x] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [x] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [x] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #2119, opened 2019-03-15
Add master type file

cnasc:
# Description

As a prelude to #2107, this PR creates a master type file where we can define the interfaces we use throughout `unlock-app`. There was one question: how to handle `TransactionType`, which takes the place of `TRANSACTION_TYPES` in `constants.js`. The 2 options:

1. Change all uses of `TRANSACTION_TYPES` to uses of the enum
  a. This will be a big PR, but ultimately cleaner
2. Change the values of `TRANSACTION_TYPES` to indices of the enum (so `TRANSACTION_TYPES['LOCK_CREATION'] = TransactionType.LOCK_CREATION`)
  b. This will be a small PR, but all those usages will have to be cleaned up at some point. This just kicks that can down the road a bit.

We chose to go with option 1.

# Checklist:


cellog:
I wonder if this shouldn't be above unlock-app? this way we can share them eventually with both Locksmith and paywall and future apps. What do you think @julien51 ?

cnasc:
One comment I want to add just for historical record: initially I used a `.d.ts` file since the file is meant only to include types. Unfortunately, TS special cases that file, so it doesn't actually get compiled. It's useful for JS->TS, but not from TS->JS which is what we need in this case. So sticking with a regular `.ts` file.

PR #2130, opened 2019-03-16
fix key purchase on coinbase wallet

### LINKED ISSUE #2129: Coinbase Wallet does not open new window when tapping on a lock to purchase
Open up any locked page, and tap on the lock in coinbase wallet, and it will not redirect to a new page. It appears that the postMessage to the main window is not happening.

cellog:
# Description

This PR fixes a final piece of the puzzle missed in the refactor of the paywall to use hooks. Redirecting to a new window uses postMessage, but the `Lock` component did not use this. Because the old system did not pass the origin in messages, which was insecure, the `interWindowCommunicationMiddleware` call to `window.parent.postMessage` was being discarded by the paywall.

This PR fixes that issue by using the new (far more secure) `usePostMessage` hook, which pulls the parent window's origin from the URL and uses that to call `postMessage` properly.

This PR also introduces new testing for the Lock using `react-testing-library` which will make it far less fragile when making future changes, and much more likely to catch regressions.

# Issues

Fixes #2129 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/components/lock/Lock.test.js
+++ b/paywall/src/__tests__/components/lock/Lock.test.js
@@ -1,10 +1,17 @@
-import {
+import React from 'react'
+import * as rtl from 'react-testing-library'
+import { Provider } from 'react-redux'
+
+import Lock, {
   mapStateToProps,
   mapDispatchToProps,
 } from '../../../components/lock/Lock'
 import { purchaseKey } from '../../../actions/key'
 import { TRANSACTION_TYPES } from '../../../constants'
-import { openNewWindowModal } from '../../../actions/modal'
+import createUnlockStore from '../../../createUnlockStore'
+import { ConfigContext } from '../../../utils/withConfig'
+import { WindowContext } from '../../../hooks/browser/useWindow'
+import { POST_MESSAGE_REDIRECT } from '../../../paywall-builder/constants'
 
 describe('Lock', () => {
   describe('mapDispatchToProps', () => {
@@ -21,22 +28,6 @@ describe('Lock', () => {
       newProps.purchaseKey(key)
       expect(props.showModal).toHaveBeenCalledWith()
       expect(dispatch).toHaveBeenCalledWith(purchaseKey(key))
-    })
-
-    it('should dispatch openNewWindowModal if the openNewWindow prop is truthy', () => {
-      expect.assertions(2)
-      const dispatch = jest.fn()
-      const props = {
-        showModal: jest.fn(),
-        openInNewWindow: true,
-      }
-      const key = {}
-
-      const newProps = mapDispatchToProps(dispatch, props)
-
-      newProps.purchaseKey(key)
-      expect(props.showModal).not.toHaveBeenCalled()
-      expect(dispatch).toHaveBeenCalledWith(openNewWindowModal())
     })
   })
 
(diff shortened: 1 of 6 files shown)


paywall/src/components/lock/Lock.js
```diff
@@ -22,7 +23,15 @@ export const Lock = ({
   config,
   disabled,
   hideModal,
+  openInNewWindow,
 }) => {
+  const { postMessage } = usePostMessage()
+  const purchase = key => {
```
julien51: Wouldn't it make more sense to put that logic inside `mapStateToProps`? This would at least make it testable much more easily?

cellog: Hooks cannot be outside of components. I could make a new custom hook if you'd like?

cellog:
refactored, PTAL

paywall/src/__tests__/components/lock/Lock.test.js
```diff
@@ -103,4 +94,57 @@ describe('Lock', () => {
       expect(newProps.transaction).toEqual(state.transactions['0x777'])
     })
   })
+  describe('usePurchaseKey is called for purchases', () => {
```
julien51: nit: space above

paywall/src/components/lock/Lock.js
```diff
@@ -68,6 +70,7 @@ Lock.propTypes = {
   purchaseKey: PropTypes.func.isRequired,
   config: UnlockPropTypes.configuration.isRequired,
   hideModal: PropTypes.func.isRequired,
+  showModal: PropTypes.func.isRequired,
```
julien51: weird. Was it missing before? It looks like your changes do not use it?

paywall/src/hooks/usePurchaseKey.js
```diff
@@ -0,0 +1,17 @@
...
+  const purchase = useCallback(
+    key => {
+      if (openInNewWindow) {
+        return postMessage(POST_MESSAGE_REDIRECT)
+      }
+      purchaseKey(key)
+    },
+    [purchaseKey, postMessage, openInNewWindow]
+  )
+  return purchase
```
julien51: nit: move the `return` statement to `const purchase`?

paywall/src/stories/lock/Lock.stories.js
```diff
@@ -5,11 +5,22 @@ import { Lock } from '../../components/lock/Lock'
 import createUnlockStore from '../../createUnlockStore'
 import { ConfigContext } from '../../utils/withConfig'
 import { UNLIMITED_KEYS_COUNT } from '../../constants'
+import { WindowContext } from '../../hooks/browser/useWindow'
```
julien51: I am confused by the choice of variable name. It looks like most (all?) hooks use `useXX`. why here a name which looks like a component?

julien51: But that is probably for another PR: maybe add a `TODO rename` comment so we can keep track of this?

paywall/src/__tests__/hooks/usePurchaseKey.test.js
```diff
@@ -0,0 +1,56 @@
...
+      const component = rtl.render(<MockPurchaseKey />)
+
+      rtl.act(() => {
+        rtl.fireEvent.click(component.getByText('Click'))
+      })
+
+      expect(purchaseKey).toHaveBeenCalledWith('hi')
+      expect(postMessage).not.toHaveBeenCalled()
+    })
+    it('calls postMessage if openInNewWindow is true', () => {
```
julien51: nit: space above?

paywall/src/__tests__/hooks/usePurchaseKey.test.js
```diff
@@ -0,0 +1,56 @@
...
+  })
+  function MockPurchaseKey() {
+    const purchase = usePurchaseKey(purchaseKey, openInNewWindow)
+    return (
+      <button type="button" onClick={() => purchase('hi')}>
+        Click
+      </button>
+    )
+  }
+  describe('purchase', () => {
```
julien51: nit: space above?

paywall/src/stories/lock/Lock.stories.js
cellog: this is the context, defined with `React.createContext` that is used to pass the `window` down. `useWindow.js` exports both this and the hook `useWindow`.

paywall/src/components/lock/Lock.js
cellog: it was in fact missing. This is not part of the limited scope of this PR, want me to move it to a separate PR?

PR #2134, opened 2019-03-16
redirect to valid URLs from invalid URLs on the paywall

### LINKED ISSUE #2103: paywall needs to allow //0x... as well as /0x.... for lock addresses
The paywall should be more forgiving of malformed URLs when attempted to detect a lock in the pathname, and allow multiple leading slashes

cellog:
# Description

This PR fixes invalid URLs by redirecting to a valid URL. Next's router does not process invalid URLs correctly, causing internal errors we don't have control over. This solution converts double `//` (or more than double) to single `/` and compares it to the current URL. If they're different, it redirects permanently to the valid URL. The screencast shows it working in action.

In addition, I discovered that the server hangs on errors in URL parsing, so this causes the reject to throw. I'm not sure this is the best approach, but it does work here in that it kills the server.

out

# Issues

Fixes #2103

# Checklist:

### DIFF
--- a/paywall/src/_server.js
+++ b/paywall/src/_server.js
@@ -19,7 +19,8 @@ function _server(port, dev) {
 
           // assigning `query` into the params means that we still
           // get the query string passed to our application
-          if (pathname.match('/0x')) {
+          const fixedPathname = pathname.replace(/\/+/g, '/')
+          if (fixedPathname.match('/0x')) {
             const params = route('/:lockAddress/:redirect?')(pathname)
             app.render(req, res, '/', Object.assign(params, query))
           } else {
--- a/paywall/src/_server.js
+++ b/paywall/src/_server.js
@@ -12,6 +12,15 @@ function _server(port, dev) {
 
     app.prepare().then(() => {
       let server = createServer((req, res) => {
+        const fixedPathname = req.url.replace(/\/+/g, '/')
+        if (req.url !== fixedPathname) {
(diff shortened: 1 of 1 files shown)


cellog:
well in this case, it is actually worse, no? Because if we mask the problem in dev only, then it is bad. I'm going to close this and try again, with a fix in the code, if possible.

This affects the entire app:

https://unlock-protocol.com//dashboard
https://unlock-protocol.com// (loads, but errors silently)

I think we may want to mark as "won't fix", what do you think? It's clearly an upstream bug

julien51:
Hum. Do we know why this happens?  Where is the double slash coming from?

cellog:
it happens when someone (who shall remain nameless: @cellog) sets `PAYWALL_NETLIFY_PAYWALL_URL` to `https://paywall.unlock-protocol.com/` with a trailing slash :). The paywall script expects no trailing slash. I could fix it there, I suppose, but the problem remains if someone manually types in the wrong address, and I worry that it could be an attack vector, since we don't yet understand why it happens

paywall/src/_server.js
```diff
@@ -27,7 +36,7 @@ function _server(port, dev) {
             return
           }
         } catch (error) {
-          reject(error)
+          throw reject(error)
```
julien51: Is that working? I would expect `throw` to require en `Error`?

cellog:
this solution is incomplete, there is a deeper issue inside the next router, I'm going to close this until we know what is going on

PR #2139, opened 2019-03-16
new screenshot helper for integration testing

cellog:
# Description

This provides a simple mechanism for taking screenshots for debugging integration tests locally. In addition, it is designed to fail on CI if accidentally committed.

It works by mounting the current directory as a directory `/screenshots` in the integration tests image. Then, a helper function `screenshot` can be used in the integration test like so:

```js
const { screenshot } = require('../helpers/debugging')

// ...

  it('should load the creator dashboard', async () => {
    const paywallIframe = page.mainFrame().childFrames()[0]
    screenshot(page)
  })
```

It saves screenshots as files named with this scheme: `test-<num>.jpg` where `<num>` is the screenshot number. Thus, with 4 calls to `screenshot` you will have `test-1.jpg` through `test-4.jpg` in the current directory that can be inspected.

If you pass a name as the second parameter, screenshot will use that as the filename. The example below will create a file named `fail-<num>.jpg` if an exception is thrown, where `<num>` is the numth screenshot:

```js
  beforeAll(async () => {
    try {
      await page.goto(url('/dashboard'))
      await expect(page).toMatch('Creator Dashboard')
      await page.waitFor('#CreateLockButton')
      screenshot(page)
      await page.waitForSelector(`#LockEmbeddCode_${testLockAddress}`)
      screenshot(page)
      await expect(page).toMatch('30 days')
      await Promise.all([
        page.goto(url(`/demo/${testLockAddress}`)),
        page.waitForNavigation(),
      ])
      screenshot(page)
      await page.waitForFunction(() => window.frames.length)
      screenshot(page)
    } catch (e) {
      screenshot(page, 'fail')
      throw e
    }
  })
```

I used the example above to determine whether the screenshot was in the normal flow or caused by an assertion fail when designing the paywall tests.

Because `/screenshots` is only mounted by the local integration test, any calls to `screenshot` will fail on the server with `path not found`, so they will never sneak into master.

# Issues

Refs #1890 

# Checklist:

### DIFF
--- a/scripts/local-docker-integration-tests.sh
+++ b/scripts/local-docker-integration-tests.sh
@@ -31,5 +31,5 @@ docker build -t unlock -f "$REPO_ROOT/docker/unlock.dockerfile" $REPO_ROOT
 docker build -t unlock-integration -f "$REPO_ROOT/docker/unlock-integration.dockerfile" $REPO_ROOT
 
 # Run the tests
-$REPO_ROOT/scripts/integration-tests.sh $EXTRA_ARGS
+$REPO_ROOT/scripts/integration-tests.sh -v $PWD:/screenshots $EXTRA_ARGS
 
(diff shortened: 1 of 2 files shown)


cellog:
sample screenshot:

lock-3

PR #2142, opened 2019-03-17
Paywall: fix npm start to work

cellog:
# Description

While working on moving the integration tests for the paywall to use the new split-off paywall, I discovered that the `npm start` command differed substantially from the one inside `unlock-app` when it should be identical. This was causing some weird errors when running in the integration tests, and I assume also in production that we just had not seen yet.

In addition, it was doing a build step that belonged in `npm run build`. This PR corrects both problems.

# Issues

Refs #1205 

# Checklist:

### DIFF
--- a/paywall/package.json
+++ b/paywall/package.json
@@ -9,7 +9,7 @@
     "build": "npm run before && next build src",
     "deploy": "next export src -o out",
     "deploy-netlify": "./scripts/deploy-netlify.sh",
-    "start": "npm run before && next start",
+    "start": "cross-env NODE_ENV=production node src/server.js",
     "test": "cross-env UNLOCK_ENV=test jest --env=jsdom",
     "lint": "eslint .",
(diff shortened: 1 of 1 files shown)


paywall/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "npm run before && next build src",
     "deploy": "next export src -o out",
     "deploy-netlify": "./scripts/deploy-netlify.sh",
-    "start": "npm run before && next start",
+    "start": "cross-env NODE_ENV=production node src/server.js",
```
julien51: I strongly think we should not set env variables in these commands... any way we can avoid that?

cellog: perhaps. But for consistency, I think we should use this. It's a literal copy/paste from `unlock-app`. Certainly, in a follow-up PR, we could address it in both apps, but that is outside the narrow scope of this particular PR. What would you like me to do?

julien51: Ok, please open an issue about this then...

PR #2143, opened 2019-03-17
fix docker environment for paywall and integration tests

cellog:
# Description

While working on making the paywall integration tests work for the split-off paywall, I discovered a few errors in the `docker-compose.ci.yml` file.

1. environment variables we thought would exist in the integration tests environment did not.
2. same for the paywall
3. the paywall was not actually being run and was not exposing its port in the right location, it needed to copy the way locksmith was set up (hats off to Akeem)

With these changes and the fix to #2142 the paywall service actually runs now in the integration test environment, which is a prerequisite to switching the tests to use the split-off paywall

# Issues

Refs #1205 #1882 

# Checklist:

### DIFF
--- a/docker/docker-compose.ci.yml
+++ b/docker/docker-compose.ci.yml
@@ -26,7 +26,6 @@ services:
     working_dir: /home/unlock
     expose:
       - 3000
-      - 3001
     environment:
       CI: "${CI}"
       HTTP_PROVIDER: "${HTTP_PROVIDER}"
@@ -48,11 +47,20 @@ services:
     image: unlock
     user: node
     working_dir: /home/unlock/paywall
+    expose:
+      - 3001
+    environment:
+      CI: "${CI}"
+      HTTP_PROVIDER: "${HTTP_PROVIDER}"
+    command: ["npm", "run", "start"]
   integration-tests:
     image: unlock-integration
     user: node
     environment:
       UNLOCK_HOST: "unlock"
+      PAYWALL_URL: "${PAYWALL_URL}"
+      PAYWALL_SCRIPT_URL: "${PAYWALL_SCRIPT_URL}"
+      LOCKSMITH_URI: "${LOCKSMITH_URI}"
       CI: "${CI}"
     depends_on:
       - ganache-integration


docker/docker-compose.ci.yml
```diff
@@ -48,11 +47,20 @@ services:
     image: unlock
     user: node
     working_dir: /home/unlock/paywall
+    expose:
+      - 3001
+    environment:
+      CI: "${CI}"
+      HTTP_PROVIDER: "${HTTP_PROVIDER}"
+    command: ["npm", "run", "start"]
```
julien51: Ha! I knew adding a port 3001 to the `unlock` image was probably not solving the problem you were looking to solve ;)

What I am thinking now is that we do not need both `unlock` and `unlock-app`. Maybe worth considering for another PR.

docker/docker-compose.ci.yml
```diff
@@ -48,11 +47,20 @@ services:
...
   integration-tests:
     image: unlock-integration
     user: node
     environment:
       UNLOCK_HOST: "unlock"
+      PAYWALL_URL: "${PAYWALL_URL}"
```
julien51: I am not sure this is a good approach. Since we know that the service will be at `paywall` (that is how docker works?) could we just use these values rather than force ourselves to set them?

julien51: For consistency, you would probably need to use `PAYWALL_HOST: "paywall"` and actually build the URL from the host?

docker/docker-compose.ci.yml
```diff
@@ -48,11 +47,20 @@ services:
...
+      PAYWALL_SCRIPT_URL: "${PAYWALL_SCRIPT_URL}"
```
julien51: Same as above? Generally since we moved to using the paywall app, could we just rebuild `PAYWALL_SCRIPT_URL` as we need it from the `PAYWALL_HOST` rather than reset it independently every time (more error prone)

docker/docker-compose.ci.yml
```diff
@@ -48,11 +47,20 @@ services:
...
+      LOCKSMITH_URI: "${LOCKSMITH_URI}"
```
julien51: Same here, please let's use `LOCKSMITH_HOST` and let the app build the URL as it needs it.

cellog:
Respectfully, I think that this is not the PR to change these things, the env variables are used all over all 3 apps, so can we open an issue for the refactors to follow this one?

PR #2076, opened 2019-03-14
Locksmith now builds with TypeScript

cnasc:
# Description

Following on #2050, the experiment to build Locksmith with TypeScript, this PR introduces the minimum configuration necessary to achieve that state.

Change summary:
1. Install dependent packages
2. Add `tsc` script to `package.json`
2. Generate and customize `tsconfig.json`
3. Modify `dev` script in `package.json` to allow the app to automatically build and restart with TypeScript
4. Update `.gitignore` and `.eslintignore` to avoid build artifacts
5. Upgrade Jest
  a. `ts-jest` requires v24 of `jest`
  b. upgrading to `jest@24.1` revealed that we had a malformed test, which was fixed
  c. upgrading to a new major version of `jest` changed the way code coverage is calculated, so the thresholds had to be adjusted for the build to pass.

As with #2071, no actual TypeScript is included in this PR. This only allows the ability to use TypeScript in the future.

# Issues

Fixes #2050 

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [x] This PR only contains configuration changes (package.json, etc.) **[and a test, but the config changes couldn't pass without it]**
  - [ ] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [x] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [x] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #2145, opened 2019-03-17
adding the lock operations layer

julien51:
I believe this provides better isolation and will allow us to re-use more code.
For example, when doing an update, we need to first check if the object exists, which now re-uses the logic to retrieve a lock by its address rather than re-implement that logic inside the controller.

I am in favor of keeping most controller tests as they behave like integration tests: actually making requests to the database
and actually yielding HTTP responses, while the operation tests only test the logic in isolation (mocking the underlying models).

I propose we follow the same kind of architecture for the future locksmith features.


julien51:
Note : it looks like https://github.com/unlock-protocol/unlock/pull/2076 re-added the `app.test.js` which had been deleted (refactored into controller tests in a previous PR!).

PR #2147, opened 2019-03-18
Page components include as little logic as possible to make them more testable and re-usable. Fixes #1740

### LINKED ISSUE #1740: pages should have as little React in them as possible to make them testable
In doing a refactor of the paywall app, I found that `pages/paywall.js` contained a lot of React code. Pages are more difficult to test because they have a fair amount of next.js magic in them, such as `getInitialProps`. We should extract as many of these pages into components that accept props from the pages as possible. This will allow more comprehensive testing of our most critical top-level items, and make changing less likely to fail silently

julien51:
This looks bigger than what it is, but I basically took each page and made a corresponding PageContent component
which includes the logic from the page.
This makes the components more testable.

Fixes #1740


julien51:
I tried that first but next is *very* opinionated and this would break the convention of naming files like the components.
Basically if we name the page for Dashboard as `DashboardPage`, then we'd need to name the file `dashboardPage.js` and that would break the next.js routing (this page would be accessible at `/dashboardPage` instead of `/dashboard`...

So, even if I agree that using `dashboardContent` is not very elegant, it is at least quite explicit and works with next.js...

cellog:
Really? I did not find this in the PR to enable the paywall landing page. The file is named `home.js` and the component is `PaywallHome`. I wonder if it was something else next-y like perhaps a next.config.js thing?

julien51:
Merging to avoid more conflicts, and we can certainly rename things afterwards...

PR #2152, opened 2019-03-18
remove conflicting rule, add eslint-check to paywall

### LINKED ISSUE #2151: eslint/prettier conflicts
running `npm run eslint-check` in `unlock-app` reveals a conflict between eslint and prettier: the `linebreak-style` rule. This should be removed from `.eslintrc`

cellog:
# Description

This PR removes a conflicting rule between eslint and prettier, and may fix the `npm run reformat` issues blocking some PRs

# Issues

Refs #2151

# Checklist:

### DIFF
--- a/paywall/.eslintrc.js
+++ b/paywall/.eslintrc.js
@@ -23,7 +23,6 @@ module.exports = {
   },
   rules: {
     'react/prefer-stateless-function': [2],
-    'linebreak-style': ['error', 'unix'],
     quotes: [
       'error',
       'single',
--- a/paywall/.prettierrc
+++ b/paywall/.prettierrc
@@ -2,4 +2,5 @@
   "trailingComma": "es5",
   "singleQuote": true,
   "semi": false,
+  "endOfLine": "lf"
 }
--- a/paywall/package.json
+++ b/paywall/package.json
@@ -14,6 +14,7 @@
     "lint": "eslint .",
     "reformat": "prettier-eslint \"src/**/*.js\" --write",
     "fail-pending-changes": "../scripts/pending-changes.sh",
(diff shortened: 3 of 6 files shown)


cellog:
P.S. this doesn't fix the issue, just checked. Someone (probably me :) will need to do a deep dive into prettier-eslint logs to find the rule that is screwing things up

julien51:
I think we should actually rather change the prettier config, rather than the eslint config if you do no mind?

cellog:
https://github.com/prettier/eslint-config-prettier

cellog:
```
arrow-body-style and prefer-arrow-callback
These rules might cause problems if using eslint-plugin-prettier and --fix.

If you use any of these rules together with the prettier/prettier rule from eslint-plugin-prettier, you can in some cases end up with invalid code due to a bug in ESLint’s autofix.

```

I think this is what is causing the problem

julien51:
I think this is noop? It looks like the problem was in fact due to prettier not using the same parser locally and on CI?

Please confirm and close ;)

julien51:
Could you clarify if just adding the prettier rule actually solves the problem?

cellog:
the fix for the blocker is already in master. This fixes a potential conflict between eslint and prettier. Prettier automatically ensures consistency of line endings in every file, but by default it uses the most common line ending (`"auto"`). To force it to use only `LF` as the line endings, we need the setting explicitly set in `.prettierrc`. Once this is done, the eslint rule for line endings becomes superfluous and can be removed. That is what this PR does.

julien51:
> Once this is done, the eslint rule for line endings becomes superfluous and can be removed. 

I'd much rather keep it for now because I feel like it is better to have a "stricter" Eslint should we eventually change our usage of prettier. Could you please update that? Thanks!

PR #2193, opened 2019-03-19
adding transactionOperations

julien51:
Using the same model as the block Operations this PR introduces transaction operations!

### DIFF
--- a/locksmith/__tests__/operations/transactionOperations.test.js
+++ b/locksmith/__tests__/operations/transactionOperations.test.js
@@ -1,0 +1,59 @@
+import {
+  findOrCreateTransaction,
+  getTransactionsBySender,
+} from '../../src/operations/transactionOperations'
+
+const Sequelize = require('sequelize')
+
+const models = require('../../src/models')
+
+let Transaction = models.Transaction
+const Op = Sequelize.Op
+
+beforeEach(() => {
+  Transaction = models.Transaction // resetting Lock for before each test
+})
+
+describe('lockOperations', () => {
+  describe('findOrCreateTransaction', () => {
+    it('should call findOrCreate on the transaction object', async () => {
+      expect.assertions(5)
+      const transactionHash =
+        '0x7d6289db59c3434b6a034b4f211be52ca34f05e2aa856fc3e69b8af101355842'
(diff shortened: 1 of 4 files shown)


locksmith/src/controllers/transactionController.js
```diff
@@ -13,33 +12,20 @@ const transactionCreate = async (req, res) => {
...
-const transactionGet = async (req, res) => {
-  const sender = ethJsUtil.toChecksumAddress(req.query.sender)
-
-  let transactions = await Transaction.findAll({
-    where: {
-      sender: { [Op.eq]: sender },
-    },
-  })
-
+const transactionsGet = async (req, res) => {
```
benwerd: Is this the best name for this? Are there likely to be other transaction retrieval methods in the future? It's not obvious to me that this gets transactions owned by a sender.

julien51: So, not a huge fan of the name either but I think this is consistent with the names picked by @akeem previously and with the actual REST URL `/transactions`. I just put a plural because it indicates many transactions ;p

locksmith/__tests__/operations/transactionOperations.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+describe('lockOperations', () => {
+  describe('findOrCreateTransaction', () => {
+    it('should call findOrCreate on the transaction object', async () => {
+      expect.assertions(5)
+      const transactionHash =
+        '0x7d6289db59c3434b6a034b4f211be52ca34f05e2aa856fc3e69b8af101355842'
+      const sender = '0x0x77cc4f1fe4555f9b9e0d1e918cac211915b079e5'
+      const recipient = '0xca750f9232c1c38e34d27e77534e1631526ec99e'
+      Transaction.findOrCreate = jest.fn(query => {
```
akeem: I'm a bit concerned about us mocking out these methods that we don't own. I kind of glazed over this in LockOperation but do think we might need to consider some other options. 

The test know deeply about the implementation of these methods, if we change the implementation we'll have to update the tests as well

julien51: This is right. I think eventually we should have one more level between the 'orm-models' and the actual app models.

julien51:
Something is going on here.

PR #2202, opened 2019-03-19
Routing emails based on their templates

julien51:
Emails can now be sent! Well, since there is no template, none will actually be sent, but everything
is in place for them to be sent!

### DIFF
--- a/wedlocks/src/config.js
+++ b/wedlocks/src/config.js
@@ -1,0 +1,11 @@
+export default {
+  host: process.env.SMTP_HOST,
+  port: process.env.SMTP_PORT || 587,
+  secure: false,
+  auth: {
+    user: process.env.SMTP_USERNAME,
+    pass: process.env.SMTP_PASSWORD
+  },
+  sender:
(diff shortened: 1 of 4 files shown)


wedlocks/src/config.js
```diff
@@ -0,0 +1,11 @@
+export default {
+  host: process.env.SMTP_HOST,
+  port: process.env.SMTP_PORT || 587,
+  secure: false,
+  auth: {
+    user: process.env.SMTP_USERNAME,
+    pass: process.env.SMTP_PASSWORD
+  },
+  sender:
+    process.env.SMTP_FROM_ADDRESS || 'Unlock <no-reply@unlock-protocol.com>' // TODO: can we do better eventually?
```
benwerd: Idea: every email comes from support@unlock-protocol.com and people can hit reply if they run into a problem.

Possible contradiction: we should certainly _never_ do that for key recovery emails.

OTOH: maybe we could be smart enough to autorespond to key recovery emails and make sure we never see them on receipt.

julien51: I think we can eventually have a `no-reply` address for the recovery key emails and have a reply address for the other ones...

PR #2205, opened 2019-03-19
Initial pass at  api to support user accounts

akeem:
# Description

This change exposes the ability to store user accounts and the ability to retrieve an encrypted private key based on a provided email address. 

Notes:

* We are JSON column to store the password encrypted private key. This is the least common denominator between postgres and sqllite. It might be worth a discussion around changing to the Postgres specific JSONB column type for performance resasons.

* Indices will be added in following change

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Wow, I did not expect you to have the whole thing a single PR ;) I was thinking you'd only work on the low level model stuff!

locksmith/src/operations/userOperations.ts
```diff
@@ -0,0 +1,41 @@
+const ethJsUtil = require('ethereumjs-util')
+const models = require('../models')
+const { User, UserReference } = models
+import * as types from '../types'
+
+class UserOperations {
```
julien51: I actually intentionally did not create classes for the other operations. Could you expand on why you think they're required here?

locksmith/src/operations/userOperations.ts
```diff
@@ -0,0 +1,41 @@
...
+      recoverySeed: input.recoverySeed,
+      encryptedPrivateKey: input.encryptedPrivateKey,
+    })
+
+    let userReference = await UserReference.create({
+      emailAddress: input.emailAddress.toLowerCase(),
+    })
+
+    if (userReference) {
+      userReference.setUser(user)
```
julien51: Could we do that above directly when creating the object? This would save at least on SQL request?

locksmith/src/operations/userOperations.ts
```diff
@@ -0,0 +1,41 @@
...
+      userReference.save()
+      return true
+    } else {
+      return false
+    }
+  }
+
+  static async privateKeyLookup(emailAddress: String): Promise<String | null> {
```
julien51: `getUserPrivateKeyByEmailAddress` would probably be more consistent with how we named other functions?

locksmith/src/operations/userOperations.ts
```diff
@@ -0,0 +1,41 @@
...
+    let user = await UserReference.findOne({
+      where: { emailAddress: emailAddress.toLowerCase() },
+      include: [{ model: User, attributes: ['encryptedPrivateKey'] }],
+    })
+
+    if (user) {
+      return user.User.encryptedPrivateKey
```
julien51: Is there no way to do that in a single query?

locksmith/src/models/userReference.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
```
julien51: I do not think you need these ;)

locksmith/src/models/userReference.js
```diff
@@ -0,0 +1,18 @@
...
+    },
+    {}
+  )
+  UserReference.associate = function(models) {
+    UserReference.belongsTo(models.User, {
+      foreignKey: 'publicKey',
+      targetKey: 'publicKey',
+    })
+  }
+  sequelize.sync()
```
julien51: What is this for?

julien51: Would you want to add a `UserReference.removeAttribute('id')`?

locksmith/src/controllers/userController.ts
```diff
@@ -0,0 +1,40 @@
+import UserOperations = require('../operations/userOperations')
+import { Request, Response } from 'express-serve-static-core'
+
+class UserController {
```
julien51: Same here. It looks like we are not using classes for the other controllers. Any reason for the change?

locksmith/migrations/20190318152733-create-user.js
```diff
@@ -0,0 +1,32 @@
...
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('Users', {
+      publicKey: {
+        type: Sequelize.STRING,
+        primaryKey: true,
+        allowNull: false,
+        unqiue: true
+      },
+      recoverySeed: {
```
julien51: I am not sure `seed` is the right word here. `recoveryPhrase` is probably more accurate? Arguably, a seed is something which "generates" something else. This is used to 'encrypt' the private key.

locksmith/migrations/20190318152733-create-user.js
```diff
@@ -0,0 +1,32 @@
...
+        type: Sequelize.STRING,
+        allowNull: false
+      },
+      encryptedPrivateKey: {
```
julien51: Let's be more explicit and call it `passwordEncryptedPrivateKey`.

julien51: Also, I believe we will use `web3.eth.accounts.encrypt(privateKey, password);` which yields a pretty complex object: https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#encrypt

Is a string enough to store that?

locksmith/src/operations/userOperations.ts
akeem: this was/is me reacting to some typescript things or at least the way that I understand them.  Our js approach written in typescript puts the methods in the global scope, there may be a better way to approach this

locksmith/src/operations/userOperations.ts
akeem: Agreed, i'll put this and the other call in a transaction. I've started down this direction but paused due to some of the ripples caused

locksmith/src/models/userReference.js
akeem: this is all part of the sequelize black magic, https://sequelize.readthedocs.io/en/latest/docs/associations/ . From what I understand it glues the models together and provides some of the dynamic methods for convenience

locksmith/migrations/20190318152733-create-user.js
akeem: Interesting I was thinking that we would use something like HMAC256. Let's discuss the details so that we can get the storage to match our approach

locksmith/src/operations/userOperations.ts
cnasc: Maybe you're looking for namespaces? https://www.typescriptlang.org/docs/handbook/namespaces.html

Alternatively, just declaring modules.

cnasc:
This looks really good overall. So I can make sure I understand completely, right now this draft API supports 2 operations:
1. Create a user account by POSTing some JSON to the endpoint
2. Retrieve the password-encrypted private key with a GET

I actually have some questions about the second operation, but I think that will be a better discussion in person that we can summarize in text format for the record afterward.

julien51:
@akeem would you be able to start working on splitting this so @cnasc @benwerd and I could work on other related parts?

akeem:
@julien51 how do you envision this being split? At the moment, I am still making adjustments to this but will move this to a non-draft PR once I believe it is in a solid state

julien51:
Split in horizontal layers:
1. the model/migration
2. Operations
3. Controllers

locksmith/src/models/userReference.js
akeem: this file was generated by Sequelize

locksmith/__tests__/operations/userOperations.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    it('should normalize the public key/address', async () => {
+      UserReference.create = jest.fn(() => {})
+      expect.assertions(1)
+      await UserOperations.createUser(userCreationDetails)
+      expect(UserReference.create).toHaveBeenCalledWith(
+        {
+          User: {
+            passwordEncryptedPrivateKey: '{"data" : "encryptedPassword"}',
+            publicKey: '0x21cC9C438D9751A3225496F6FD1F1215C7bd5D83',
```
julien51: Could you actually denormalize (lowercase) the `publicKey` in the `userCreationDetails` so that we know we're actually normalzing it?

locksmith/__tests__/operations/userOperations.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    })
+  })
+
+  describe('when the email address does not exist in persistence', () => {
+    it('returns null', async () => {
+      expect.assertions(1)
+      let result = await UserOperations.getUserPrivateKeyByEmailAddress(
+        'non-existant@example.com'
+      )
+      expect(result).toEqual(null)
```
julien51: I really think it is important that we actually do not return null but a new encrypted private key so that we are not leaking user data! We should encrypt the email address with a consistent string (some kind of hash of the email address?)

locksmith/migrations/20190318152733-create-user.js
```diff
@@ -0,0 +1,32 @@
...
+        primaryKey: true,
+        allowNull: false,
+        unique: true
+      },
+      recoveryPhrase: {
+        type: Sequelize.STRING,
+        allowNull: false
+      },
+      passwordEncryptedPrivateKey: {
+        type: Sequelize.JSON,
```
julien51: that's cool!

locksmith/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "an unlock storage service",
   "main": "server.js",
   "scripts": {
-    "predev": "NODE_ENV=development sequelize db:migrate",
+    "predev": "NODE_ENV=development npm run db:migrate",
```
julien51: Could you explain what this changes? Put that into another PR?

locksmith/package.json
```diff
@@ -29,6 +29,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@types/jest": "^24.0.11",
```
julien51: Please add dependencies separately!

locksmith/package.json
```diff
@@ -48,6 +49,8 @@
     "sequelize-cli": "^5.4.0",
     "supertest": "^3.3.0",
     "ts-jest": "^24.0.0",
+    "@types/express": "^4.16.1",
```
julien51: ditto

locksmith/package.json
```diff
@@ -48,6 +49,8 @@
...
+    "@types/node": "^11.11.3",
```
julien51: ditto

locksmith/src/controllers/userController.ts
```diff
@@ -0,0 +1,40 @@
...
+          publicKey: user.publicKey,
+          passwordEncryptedPrivateKey: user.passwordEncryptedPrivateKey,
+          recoveryPhrase: user.recoveryPhrase}
+        )
+
+        let status = creationStatus ? 200 : 400
+        return res.sendStatus(status)
+      }
+    } catch (e) {
+      return res.sendStatus(400)
```
julien51: 500 would be more appropriate here? Also, I think we should log this?

locksmith/src/models/userReference.js
julien51: Please remove!

locksmith/src/operations/userOperations.ts
```diff
@@ -0,0 +1,47 @@
...
+const { User, UserReference } = models
+import * as types from '../types'
+
+namespace UserOperations {
+  export async function createUser(
+    input: types.UserCreationInput
+  ): Promise<Boolean> {
+    let userReference = await UserReference.create(
+      {
+        emailAddress: input.emailAddress.toLowerCase(),
```
julien51: I don't remember seeing a test where we check with an UPPERCASE email address and we still return the private key corresponding to the lowercase one? Could you add this please?

locksmith/src/operations/userOperations.ts
```diff
@@ -0,0 +1,47 @@
...
+          passwordEncryptedPrivateKey: input.passwordEncryptedPrivateKey,
+        },
+      },
+      {
+        include: User,
+      }
+    )
+
+    if (userReference) {
+      return true
```
julien51: I think we can just `return userReference`?

locksmith/src/operations/userOperations.ts
```diff
@@ -0,0 +1,47 @@
...
+  ): Promise<String | null> {
+    let user = await UserReference.findOne({
+      where: { emailAddress: emailAddress.toLowerCase() },
+      include: [{ model: User, attributes: ['passwordEncryptedPrivateKey'] }],
+    })
+
+    if (user) {
+      return user.User.passwordEncryptedPrivateKey
+    } else {
+      return null
```
julien51: That's where we should generate a private key and encrypt it with a salted hash of the email address

locksmith/src/routes/user.js
```diff
@@ -0,0 +1,10 @@
+let express = require('express')
+
+let router = express.Router()
+/* eslint-disable import/no-unresolved*/
```
julien51: Could you explain what this is?

locksmith/src/routes/user.js
```diff
@@ -0,0 +1,10 @@
...
+let userController = require('../controllers/userController')
+
+router.post('/', userController.createUser)
+router.get('/:emailAddress', userController.retrieveEncryptedPrivatekey)
```
julien51: Could we do `/user/:emailAddress/privatekey` because I think this route is currently too generic and we may create conflicts?

akeem: It was designed with the idea that /user/<email address> returns the relevant information that we are willing to show about the address passed. In this case the encrypted private key, the idea here being that the data returned could be expanded upon. This can be constricted to only support this specific use case

locksmith/src/routes/user.js
akeem: this is referencing the result of a typescript file that will exist, when build. As there actually isn't a javascript file for this controller prior to being built the linter raises an error

locksmith/src/operations/userOperations.ts
akeem: This branch of the conditional represents the encrypted private key not existing in storage. From our last conversation, I believe that we were/are not generating private keys at all on the back end and this falls under the concern of the front end

locksmith/src/operations/userOperations.ts
akeem: At the moment method's return type is `Promise<Boolean>` we could change it to be an any type but I am not sure that it would provide a lot of value to downs stream consumers in this fashion

docker/docker-compose.ci.yml
```diff
@@ -84,7 +84,7 @@ services:
     image: unlock
     user: node
     working_dir: /home/unlock/locksmith
-    command: sh -c './scripts/wait-for.sh db:5432 -- npm run db:migrate && npm run start'
+    command: sh -c './scripts/wait-for.sh db:5432 -- npm run db:migrate && npm run build && node ./build/server.js'
```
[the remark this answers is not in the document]
akeem: yep, this is pending a conversation with @cnasc

locksmith/src/controllers/userController.ts
akeem: Can add the logging. Not sure about this being a 500. https://twitter.com/stevelosh/status/372740571749572610?lang=en 

lol

locksmith/__tests__/operations/userOperations.test.ts
akeem: We should have a quick chat about security in general, since we hadn't discussed it in the context of these initial stories. I assumed that we would do it prior to this code went live but probably not in the introduction of this into the code base

locksmith/src/operations/userOperations.ts
julien51: good point

locksmith/src/routes/user.js
julien51: awww, this is weird. @cnasc is there a way to deal with that in a better way?

locksmith/src/routes/user.js
julien51: I guess than `/user/:emailAddress` is fine, but let's not use `/:emailAddress`.

julien51: Sorry I misunderstood how the controller prefixes the routes.

akeem:
Closing this as pull requests of the component parts of this have made their way to production

PR #2214, opened 2019-03-20
Define new interfaces for upcoming work

cnasc:
# Description

This PR is another slice of #2107. It updates our Unlock type definitions with some new interfaces and fields.

There are two pieces of this that I'm not in love with: `TransactionMetadata` and `TransactionMetadataMap`. These types will be used in the Log page, essentially to pass down additional information that (probably?) doesn't belong in the state. Right now that's the link to the chain explorer for transaction address, and the readable name of the transaction type.

One solution here is to pass down less information--the link to the chain explorer really doesn't belong in the state, and passing down an object of `string:string`  pairs from hash to href will look a little cleaner. The names could possibly just be changed to readable names in the enum? I'd be interested to know if there are any downsides to that.

# Checklist:

### DIFF
--- a/unlock-app/src/unlock.ts
+++ b/unlock-app/src/unlock.ts
@@ -28,16 +28,17 @@ export interface Transactions {
 
 export interface Account {
   address: string
-  balance: string
+  balance: string // TODO: Stricter balance type (enforce currency, precision)
 }
 
 export interface Network {
-  name: number
+  name: number // TODO: This is very misleading, change property name to id
 }
 
 // TransactionMetadata and TransactionMetadataMap are used in the log page to
 // pass additional information about transactions that do not belong in the
-// state.
+// state. TODO: There is certainly a better way to handle this, find out what
+// that is and do it.
 export interface TransactionMetadata {
   href: string
   readableName: string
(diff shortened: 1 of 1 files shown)


unlock-app/src/unlock.ts
```diff
@@ -9,11 +9,44 @@ export enum TransactionType {
...
+}
+
+// Analogous to the transactions in Redux
+export interface Transactions {
+  [hash: string]: Transaction
+}
+
+export interface Account {
+  address: string
+  balance: string
```
julien51: Later: I think we'll eventually need a `Balance` type so that we can be tighter on the currency as well as precision.

unlock-app/src/unlock.ts
```diff
@@ -9,11 +9,44 @@ export enum TransactionType {
...
+}
+
+export interface Network {
+  name: number
```
julien51: lol: this naming things will keep haunting me! Please add a TODO to rename that property to `id`.

unlock-app/src/unlock.ts
```diff
@@ -9,11 +9,44 @@ export enum TransactionType {
...
+}
+
+// TransactionMetadata and TransactionMetadataMap are used in the log page to
+// pass additional information about transactions that do not belong in the
+// state.
+export interface TransactionMetadata {
```
julien51: Indeed, this is weird. Could you add a TODO to look into doing something better here at some point?

Issue #1887, opened 2019-03-05
integration test: Paywall redirects back to content if key has already been purchased in main window


cellog:

cellog:
dupe of #1886

Issue #1886, opened 2019-03-05
integration test: Paywall redirects back to content after key purchase in main window


cellog:

cellog:
dupe of #1884

PR #2217, opened 2019-03-20
Add a helper for retrieving paywall URL in integration tests

cellog:
# Description

This PR modifies the `url` helper for integration tests, to allow it to pull in both the main `unlock-app` url and the `paywall` url.

This will be used to write integration tests for the paywall in the main window

# Issues

Refs #1884 #1886 #1887 

# Checklist:

### DIFF
--- a/tests/helpers/url.js
+++ b/tests/helpers/url.js
@@ -1,6 +1,6 @@
 const host = process.env.UNLOCK_HOST || '127.0.0.1'
 const port = process.env.UNLOCK_PORT || 3000
-const paywall = process.env.PAYWALL_URL || '127.0.0.1:3001'
+const paywall = process.env.PAYWALL_URL || 'http://127.0.0.1:3001'
 
 module.exports = {
   main: (path = '/') => `http://${host}:${port}${path}`,
--- a/tests/helpers/url.js
+++ b/tests/helpers/url.js
@@ -4,5 +4,5 @@ const paywall = process.env.PAYWALL_URL || '127.0.0.1:3001'
 
 module.exports = {
   main: (path = '/') => `http://${host}:${port}${path}`,
-  paywall: (path = '/') => `${paywall}/${path}`,
+  paywall: (path = '/') => `${paywall}${path}`,
 }
--- a/tests/helpers/url.js
+++ b/tests/helpers/url.js
@@ -1,4 +1,8 @@
 const host = process.env.UNLOCK_HOST || '127.0.0.1'
 const port = process.env.UNLOCK_PORT || 3000
+const paywall = process.env.PAYWALL_URL || '127.0.0.1:3001'
 
-module.exports = (path = '/') => `http://${host}:${port}${path}`
+module.exports = {
+  main: (path = '/') => `http://${host}:${port}${path}`,
+  paywall: (path = '/') => `${paywall}/${path}`,
+}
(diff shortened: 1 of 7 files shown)


tests/helpers/url.js
```diff
@@ -1,4 +1,8 @@
 const host = process.env.UNLOCK_HOST || '127.0.0.1'
 const port = process.env.UNLOCK_PORT || 3000
+const paywall = process.env.PAYWALL_URL || '127.0.0.1:3001'
 
-module.exports = (path = '/') => `http://${host}:${port}${path}`
+module.exports = {
+  main: (path = '/') => `http://${host}:${port}${path}`,
+  paywall: (path = '/') => `${paywall}/${path}`,
```
julien51: There is something fishy here. If the `PAYWALL_URL` is not set, then that will build `127.0.0.1:3001/` which I am not sure is correct?

tests/test/paywall.test.js
```diff
@@ -1,4 +1,4 @@
-const url = require('../helpers/url')
+const { main: url } = require('../helpers/url')
```
julien51: any reason you use a different syntax for this one vs. home.test.js or dashboard.test.js?

tests/helpers/url.js
cellog: this is correct. If we are running the integration tests locally outside of docker, then the PAYWALL_URL is http://127.0.0.1:3001

tests/test/paywall.test.js
cellog: no, I hadn't noticed that.

cellog:
@julien51 fixed the consistency issue. The other one had a bug in that by default it would actually return `<paywallurl>//`. The correct default address for local paywall is either `127.0.0.1:3001` or `localhost:3001`, so that part is correct. PTAL

tests/helpers/url.js
julien51: Well clearly that line above `const paywall = process.env.PAYWALL_URL || '127.0.0.1:3001'`
if `PAYWALL_URL` happens to be falsy (unset), then we end up with `127.0.0.1:3001/` which will not work...

tests/test/paywall.test.js
julien51: self reviews ;)

julien51:
I still think there is a problem. if we set the env variable to `http://127.0.0.1:3001` then yes we build the URL like `http://127.0.0.1:3001/blabla`. If however that env variable is not set, we end up with `127.0.0.1:3001/blabla` which will probably not work down in the app.

Please write tests for the url.js file since we do not seem to agree on the logic.

cellog:
Oh I see what you were saying, not that 127.0.0.1 is the wrong host, just that we need the `http://`. I have fixed that. Pretty sure we won't need tests :). I will be using this helper in the as-yet-to-be-written tests for the paywall in the main window, and they simply won't work if it is wrong. testing by indirection

tests/test/url-with-defaults.test.js
```diff
@@ -0,0 +1,15 @@
+delete process.env.UNLOCK_HOST
```
julien51: Is that going to have an effect only on this file?

tests/test/url.test.js
```diff
@@ -0,0 +1,15 @@
+process.env.UNLOCK_HOST = 'host'
```
julien51: Same question ;)

tests/test/url-with-defaults.test.js
cellog: I guess we will find out, but since each test runs in its own process, I assumed yes.

cellog: the answer is it only affects this file

PR #2224, opened 2019-03-20
Log Page refactored to use TypeScript

cnasc:
# Description

This PR updates the Log page to use TypeScript throughout in anticipation of future enhancements. This is a replacement for #2107, which got out of sync with the main repo.

# Checklist:


unlock-app/src/__tests__/config.test.js
```diff
@@ -146,7 +146,7 @@ describe('config', () => {
     it('should contain the right URLs for chain explorers', () => {
       expect.assertions(2)
       expect(Object.keys(config.chainExplorerUrlBuilders)).toHaveLength(1)
-      expect(config.chainExplorerUrlBuilders.etherScan('0x0')).toEqual(false)
+      expect(config.chainExplorerUrlBuilders.etherScan('0x0')).toEqual('')
```
cellog: let's use `toBe`

unlock-app/src/components/content/LogContent.tsx
```diff
@@ -0,0 +1,75 @@
...
+import * as UnlockTypes from '../../unlock'
+
+interface Props {
+  account: UnlockTypes.Account
+  network: UnlockTypes.Network
+  transactionFeed: UnlockTypes.Transaction[]
+  transactionMetadataMap: UnlockTypes.TransactionMetadataMap
+}
+
+export const LogContent = (props: Props) => {
```
cellog: can we destructure in the function declaration? (don't need to change this, just wondering about the complexity level of that)

unlock-app/src/__tests__/components/creator/CreatorLog.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+afterEach(rtl.cleanup)
+describe('CreatorLog', () => {
+  beforeEach(() => {
+    wrapper = rtl.render(
+      <CreatorLog
+        transactionFeed={transactionFeed}
+        transactionMetadata={transactionMetadata} />
+    )
+  })
+  it('should show the blockNumber of the transaction', () => {
```
julien51: nit: space above

unlock-app/src/components/content/LogContent.tsx
julien51: it becomes a mess with TS ;)

unlock-app/src/components/content/LogContent.tsx
```diff
@@ -0,0 +1,75 @@
...
+              transactionMetadata={transactionMetadataMap}
+            />
+          </BrowserOnly>
+        )}
+      </Layout>
+    </GlobalErrorConsumer>
+  )
+}
+
+export const readable = (type: UnlockTypes.TransactionType) => {
```
julien51: Could you please add a TODO to eventually use a better mechanism (string mapping!) because I am pretty sure in the long term this will come and bite us!

unlock-app/src/components/content/LogContent.tsx
```diff
@@ -0,0 +1,75 @@
...
+  let parts = type.split('_')
+  return parts.map(part => part[0] + part.slice(1).toLowerCase()).join(' ')
+}
+
+export const mapStateToProps = (
```
julien51: Hum, this ts is gross/hard to read? can we use the same trick as we do for components? (define the interface/s above and then destructure inside the function)?

unlock-app/src/components/content/LogContent.tsx
```diff
@@ -0,0 +1,75 @@
...
+  { config: { chainExplorerUrlBuilders } } : { config: { chainExplorerUrlBuilders: UnlockTypes.ChainExplorerURLBuilders } }
+) => {
+  const transactionFeed = Object.values(transactions).sort(
+    (a, b) => b.blockNumber - a.blockNumber
+  )
+
+  const transactionMetadataMap: UnlockTypes.TransactionMetadataMap = {}
+
+  transactionFeed.forEach(tx => {
+    transactionMetadataMap[tx.hash] = {
```
julien51: now that I read this I wonder if instead of passing `transactionMetadataMap` we should not use a react component like `<TransactionMetadata>` which would show what we need to show? That would probably be cleaner?

Let's open an issue for this and do that in a separate PR (you can get rid of the strange `TransactionMetadataMap` type too!

unlock-app/src/components/creator/CreatorLog.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const { transactionFeed, transactionMetadata } = props
+  return (
+    <Grid>
+      <HeaderItem>Block Number</HeaderItem>
+      <HeaderItem>Lock Name/Address</HeaderItem>
+      <HeaderItem>Type</HeaderItem>
+      {transactionFeed.map(tx => {
+        const { href, readableName } = transactionMetadata[tx.hash]
+        return (
+          <React.Fragment key={tx.hash}>
```
julien51: All of this could be a `<Transaction>` component which would display the readable name without having to pass it at state.

unlock-app/src/__tests__/components/content/LogContent.test.ts
```diff
@@ -0,0 +1,81 @@
+import {
+  mapStateToProps,
+  readable,
+} from '../../../components/content/LogContent'
+import * as UnlockTypes from '../../../unlock'
+
+const { LOCK_CREATION, WITHDRAWAL, UPDATE_KEY_PRICE } = UnlockTypes.TransactionType
+const transactions: UnlockTypes.Transactions = {
+  '0x12345678': {
```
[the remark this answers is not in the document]
cnasc: TypeScript doesn't do dependent types, that's going to have to be a runtime check

unlock-app/src/components/content/LogContent.tsx
cnasc: Haha yup, you can destructure, but the type declaration gets very ugly.

unlock-app/src/components/content/LogContent.tsx
cnasc: Yes, we can definitely do that. I'll make that change.

unlock-app/src/components/content/LogContent.tsx
cnasc: Oooh, did you mean can we do
```
({ account, network, transactionFeed, transactionMetadataMap}: Props) => ...
```

🤔 I didn't try, but I'm going to now.

unlock-app/src/components/content/LogContent.tsx
cnasc: My preferred solution for this would be to just change the enum values to the readable forms. Insofar as I'm aware, we never actually use those bare string values for anything but this.

unlock-app/src/components/content/LogContent.tsx
cnasc: I would just as soon get rid of the concept of transaction metadata altogether. I think the right solution for the readable type name is to simply change the enum (as mentioned above). Assuming that is done, we're just trying to pass one additional piece of information, which I think we can do in a very lightweight manner.

unlock-app/src/components/creator/CreatorLog.tsx
cnasc: Can you elaborate a bit on this?

cnasc:
Closing in favor of #2279

Issue #2209, opened 2019-03-20
document integration test helpers


cellog:
Integration test helpers need documentation in-file to help people use them!

Perhaps some commented out examples in the tests as well.

Issue #2191, opened 2019-03-19
consider setting up some kind of "screenshot on fail" system for the integration tests


cellog:
Currently, when integration tests fail, we don't have a good way to see what happened. It may be worth setting up some kind of cheap cloud storage to which we can send a screenshot when a test fails. We could easily define a test helper that wraps `it` to catch exceptions, take a screenshot and save it, then throw the exception.  In this way, we can see what the page state is and possibly debug why a test fails more accurately.

There are some considerations, such as making sure we purge storage periodically, depending on a third party for storage, and so on. This is just to get the conversation started.

cellog:
CircleCI can store artifacts after a test build, and they specifically mention screenshots. This is promising. (https://circleci.com/docs/2.0/artifacts/)

cellog:
implemented

PR #2235, opened 2019-03-20
add screenshotOnFail and use .png for screenshots

cellog:
# Description

This helper creates a new `screenshotOnFail` helper which is designed to automatically take screenshots when integration tests fail, and save them to `/screenshots`. This is designed to work inside a docker context. `/screenshots` will be mapped to `/tmp/screenshots` for both local and CI work.

## Usage

At the top of the file
```js
const { screenshotOnFail } = require('../helpers/debugging')

const it = screenshotOnFail(page)
```

This replaces the jest `it` seamlessly. Passing tests run normally, failing ones take a screenshot with the test title replaced.

Note that documentation for all helpers is part of #2209 and separate from this PR.

This PR also standardizes screenshots to use png, which is better than jpg for our use cases (fewer visual artifacts)

# Issues

Refs #2191 

# Checklist:

### DIFF
--- a/tests/helpers/debugging.js
+++ b/tests/helpers/debugging.js
@@ -74,7 +74,7 @@ module.exports = {
     const runTest = async () => {
       try {
         // await will do nothing for non-async code, and make async tests work
-        module.exports.debugPage(page, true, true)
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await
         await code()
         // if the code does not throw, we did not fail
       } catch (e) {
--- a/tests/helpers/debugging.js
+++ b/tests/helpers/debugging.js
@@ -1,30 +1,51 @@
 /* eslint no-console: 0 */
 
 /**
- * debugPage
- ***********
+ * debugPage [local dev only]
  *
- * debugPage is used to turn on the display of console.log calls within the browser context.
- * This includes calls within page.evaluate and page.waitForFunction.
- *
- * Usage:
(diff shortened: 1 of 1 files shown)


julien51:
> Note that documentation for all helpers is part of #2209 and separate from this PR.

I'd rather have the doc added in the same PR so that at no point our code base has code that is undocumented. Should we need to remove this it's better to have everything in a single commit which can reverted.

Generally I have to say that I am a little worries about wrapping jest stuff inside some "custom" wrapper. Isn't there a jest way to deal with that? (in the form of some kind of callback maybe?)

cellog:
The way jest plugins work is to wrap `it` in similar fashion (for an example, see https://github.com/negativetwelve/jest-plugins/blob/master/packages/jest-plugin-its/src/its.js)

Nothing requires us to override it, we could just as easily use `itWithScreenshotOnFail`. The solution I propose doesn't have a huge learning curve for new devs, with some added risk of failure if the source is incorrect.

julien51:
Can we look into that at a later point? I think we need to avoid making this kind of change without a proper research and both you or I have a lot of other things to focus on at this point.

cellog:
closing for now

cellog:
re-opening. Starting to get many integration test failures, and this will help a lot.

cnasc:
@cellog is this still in progress? Just realized that we have a very old PR lingering here

cellog:
yes, this is what Julien was referring to yesterday in the team meeting. good eye for the ancient roman ruins of this PR ;)

PR #2240, opened 2019-03-21
show an error if the paywall crashes

### LINKED ISSUE #2117: Upon error, paywall needs to display something
There have been a number of instances where the paywall experiences an error, and it just hangs indefinitely with the "Checking access rights.." blocker.

This is ultimately a UX issue, we need to let them know that something went wrong internally, and possibly provide a way for them to notify us of the error. I am imagining something like a link that if they click will auto-open a github issue with the info we need to debug it.

At the very least, it needs to switch to a "we're having technical difficulties" page

cellog:
# Description

This PR adds a small UI clue that something has gone wrong when the paywall crashes outright.

Screen Shot 2019-03-20 at 11 50 37 PM

# Issues

Fixes #2117 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/paywall-builder/blocker.test.js
+++ b/paywall/src/__tests__/paywall-builder/blocker.test.js
@@ -74,5 +74,27 @@ describe('paywall builder', () => {
 
       expect(blocker.remove).toHaveBeenCalled()
     })
+
+    it('errorBlocker', () => {
+      expect.assertions(1)
+      const document = {
+        createElement() {
+          return { style: {} }
+        },
+      }
+
+      const blocker = {
+        appendChild: jest.fn(),
+      }
+
+      blockerManager.errorBlocker(document, blocker)
+
+      expect(blocker.appendChild).toHaveBeenCalledWith({
+        innerText: 'An error occurred, please refresh the page',
+        style: {
+          color: 'red',
+        },
+      })
+    })
   })
 })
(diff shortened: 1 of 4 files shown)


julien51:
Could you clarify about the need for this?
I worry that the "refresh the page" thing may not be quite enough? Arguably, if there is an error, refreshing is very likely to trigger the same error?

cnasc:
This might be the foot in the door for a broader discussion about error states: fail open vs. fail closed and so on.

cellog:
The other alternative is to display "something went wrong" and not suggest an action. The current state is the loading state displays infinitely. I don't care what it says, so please decide and I will make it do that.

julien51:
I agree with @cnasc about having a broader discussion about that. Maybe the best approach at this point (least disruptive) is to just remove the paywall app in case of failure.

cellog:
ok, so just show the content? if I understand, that is no problem

cnasc:
Yeah, that's what I meant by fail closed (currently, with the blocker staying indefinitely) vs fail open (remove the paywall if we can't recover).

cellog:
here's what I propose. In env and test environments, we fail closed, and fail open in staging and production. This way when the paywall breaks in integration tests, we will see an indication of the error

cellog:
closing, I have a new implementation that is fail open in all envs.

PR #2238, opened 2019-03-21
isolating the web3Service to pull it off into its own js module

julien51:
The web3Service will be pulled away from the various apps and in its own JS module so it can be re-used more easily accross apps.

Refs #1873

- [X] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [X] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [X] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [X] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #2250, opened 2019-03-21
Make TransactionType values more human-friendly

### LINKED ISSUE #2247: Replace values of TransactionType enum with "readable" names
In the course of #2224, we realized that some of our data representation led to inconvenient/ugly code patterns, such as passing a metadata object along with each transaction to access some information that isn't actually part of the transaction type:

- readableName (covered in this issue)
- href (mentioned here for completeness, but not covered by this issue)

Additionally, the procedure for creating a `readableName` is brittle. An easier solution is to simply replace the values of the enum with their readable equivalents, `'LOCK_CREATION' => 'Lock Creation'`. This is feasible because we never actually use these bare string values anywhere except when trying to render the type to the log page (the lone exception is in some mock data used in the Paywall story in `unlock-app`, that would be a single line change).

cnasc:
# Description

This PR changes the values of the `TransactionType` enum to be nicer looking strings. For example,
```
'LOCK_CREATION' => 'Lock Creation'
```
Because of this change, the `humanize` function in `LogContent` is superfluous, so it has been removed along with its tests. Additionally, an update was made so that the correct color for the transaction type can still be selected given this change. This PR smooths the road for #2224.

# Issues

Fixes #2247 
Refs #2238 **This one may need to be updated, since it duplicates the TransactionType (hopefully temporarily)**

# Checklist:

### DIFF
--- a/paywall/src/constants.js
+++ b/paywall/src/constants.js
@@ -28,10 +28,10 @@ export const pageTitle = title => {
  * Transaction types
  */
 export const TRANSACTION_TYPES = {
-  LOCK_CREATION: 'LOCK_CREATION',
-  KEY_PURCHASE: 'KEY_PURCHASE',
-  WITHDRAWAL: 'WITHDRAWAL',
-  UPDATE_KEY_PRICE: 'UPDATE_KEY_PRICE',
+  LOCK_CREATION: 'Lock Creation',
+  KEY_PURCHASE: 'Key Purchase',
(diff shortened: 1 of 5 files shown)


julien51:
Pleaase make sure you change the same things in paywall app tho!

cnasc:
A bit of difficulty there--the enum requires TypeScript and I don't believe the paywall currently uses TS.

PR #2264, opened 2019-03-21
Remove the paywall and blocker if there is an unexpected error thrown in paywall.min.js

### LINKED ISSUE #2117: Upon error, paywall needs to display something

cellog:
# Description

This PR ensures that users never see a hanging paywall if there is some kind of unexpected error thrown in paywall.min.js

The screenshots show an artificial setup. The `kill` function simply throws an error. There are 2 scenarios in which an error could occur. The first is synchronous when we are setting up the iframe, and the second is asynchronous, when listening for `postMessage` calls from the paywall inside the iframe. The screenshots highlight the calls to kill and the message showing it worked, as well as showing that the paywall and blocker are not present.

Screen Shot 2019-03-21 at 3 09 47 PM
Screen Shot 2019-03-21 at 3 09 08 PM

# Issues

Fixes #2117 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/paywall-builder/build.test.js
+++ b/paywall/src/__tests__/paywall-builder/build.test.js
@@ -204,7 +204,6 @@ describe('buildPaywall', () => {
         }
         blocker = {
           remove: jest.fn(),
-          appendChild: jest.fn(),
         }
         mockShow = jest.spyOn(iframeManager, 'show')
         mockShow.mockImplementation(() => {})
--- a/paywall/src/__tests__/paywall-builder/build.test.js
+++ b/paywall/src/__tests__/paywall-builder/build.test.js
@@ -75,6 +75,7 @@ describe('buildPaywall', () => {
           postMessage,
         },
         style: {},
+        remove: jest.fn(),
       }
 
       mockAdd = jest.spyOn(iframeManager, 'add')
@@ -99,7 +100,7 @@ describe('buildPaywall', () => {
         },
       }
       blocker = {
-        remove() {},
+        remove: jest.fn(),
       }
     })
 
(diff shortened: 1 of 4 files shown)


paywall/src/__tests__/paywall-builder/build.test.js
```diff
@@ -188,28 +209,67 @@ describe('buildPaywall', () => {
...
+        }).toThrow()
+        expect(iframe.remove).toHaveBeenCalled()
+        expect(blocker.remove).toHaveBeenCalled()
+      })
+
       it('triggers show on locked event', () => {
         expect.assertions(2)
         callbacks.message({ data: POST_MESSAGE_LOCKED })
 
-        expect(mockShow).toHaveBeenCalledWith(mockIframeImpl, document)
```
cellog: this test should never have been passing, but the changes made unearthed its incorrect variable naming

paywall/src/__tests__/paywall-builder/build.test.js
```diff
@@ -218,19 +278,33 @@ describe('buildPaywall', () => {
...
+        expect(() => {
+          callbacks.message({ data: POST_MESSAGE_REDIRECT })
+        }).toThrow()
+        expect(iframe.remove).toHaveBeenCalled()
+        expect(blocker.remove).toHaveBeenCalled()
+      })
+
       it('calls redirect on redirect event', () => {
         expect.assertions(1)
-        callbacks.message({ data: 'redirect' })
```
cellog: this constant should have been used, but wasn't

benwerd:
To clarify, does this mean that in production if the paywall errors, readers will be able to get full access to the content?

cellog:
Yes this is what it means. That's why we need full integration and unit tests on the paywall and the paywall.min.js script. The alternative is displaying some kind of error and blocking access even to those who have paid.

PR #2272, opened 2019-03-21
Introduces user operations into the codebase for Locksmith

akeem:
# Description

Introduces user operations and associated files into the code base.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/__tests__/operations/userOperations.test.ts
+++ b/locksmith/__tests__/operations/userOperations.test.ts
@@ -1,0 +1,86 @@
+import UserOperations from '../../src/operations/userOperations'
+
+const models = require('../../src/models')
+let User: any = models.User
+let UserReference: any = models.UserReference
+
+describe('User creation', () => {
+  let userCreationDetails = {
+    emailAddress: 'USER@EXAMPLE.COM',
+    publicKey: '0x21cc9c438d9751a3225496f6fd1f1215c7bd5d83',
+    passwordEncryptedPrivateKey: '{"data" : "encryptedPassword"}',
+    recoveryPhrase: 'recoveryPhrase',
+  }
+
+  describe("when a user for the public key doesn't exist", () => {
+    it('should create a user and associated data', async () => {
+      UserReference.create = jest.fn(() => {})
+
+      expect.assertions(1)
+      await UserOperations.createUser(userCreationDetails)
+      expect(UserReference.create).toHaveBeenCalledWith(
+        {
+          User: {
+            passwordEncryptedPrivateKey: '{"data" : "encryptedPassword"}',
+            publicKey: '0x21cC9C438D9751A3225496F6FD1F1215C7bd5D83',
+            recoveryPhrase: 'recoveryPhrase',
+          },
+          emailAddress: 'user@example.com',
+        },
+        { include: User }
+      )
+    })
+
+    it('should normalize the public key/address & email address', async () => {
+      UserReference.create = jest.fn(() => {})
+      expect.assertions(1)
+      await UserOperations.createUser(userCreationDetails)
+      expect(UserReference.create).toHaveBeenCalledWith(
+        {
+          User: {
+            passwordEncryptedPrivateKey: '{"data" : "encryptedPassword"}',
+            publicKey: '0x21cC9C438D9751A3225496F6FD1F1215C7bd5D83',
+            recoveryPhrase: 'recoveryPhrase',
+          },
+          emailAddress: 'user@example.com',
+        },
+        { include: User }
+      )
+    })
+  })
+})
+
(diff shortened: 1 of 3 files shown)


locksmith/__tests__/operations/userOperations.test.ts
```diff
@@ -0,0 +1,86 @@
+import UserOperations from '../../src/operations/userOperations'
+
+const models = require('../../src/models')
+let User: any = models.User
```
julien51: I think we want to avoid `any`, right @cnasc ?

locksmith/__tests__/operations/userOperations.test.ts
```diff
@@ -0,0 +1,86 @@
...
+let UserReference: any = models.UserReference
+
+describe('User creation', () => {
+  let userCreationDetails = {
+    emailAddress: 'USER@EXAMPLE.COM',
```
julien51: want to put a quick // saying we use upercase to ensure that data is stored in lower case?

locksmith/__tests__/operations/userOperations.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    publicKey: '0x21cc9c438d9751a3225496f6fd1f1215c7bd5d83',
+    passwordEncryptedPrivateKey: '{"data" : "encryptedPassword"}',
+    recoveryPhrase: 'recoveryPhrase',
+  }
+
+  describe("when a user for the public key doesn't exist", () => {
+    it('should create a user and associated data', async () => {
+      UserReference.create = jest.fn(() => {})
+
+      expect.assertions(1)
```
julien51: I think eslint should complain about this not being the first line in the block?

locksmith/__tests__/operations/userOperations.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      await UserOperations.createUser(userCreationDetails)
+      expect(UserReference.create).toHaveBeenCalledWith(
+        {
```
julien51: The logic of the params is actually tested in the test below?
I would rather test that it returns `true`?

julien51: So that the tests are not identical and we do actually check the output?

locksmith/__tests__/operations/userOperations.test.ts
```diff
@@ -0,0 +1,86 @@
...
+            publicKey: '0x21cC9C438D9751A3225496F6FD1F1215C7bd5D83',
+            recoveryPhrase: 'recoveryPhrase',
+          },
+          emailAddress: 'user@example.com',
+        },
+        { include: User }
+      )
+    })
+
+    it('should normalize the public key/address & email address', async () => {
```
julien51: Cpuld you add a test where it returns `false` if the we try to create the same user twice?

locksmith/src/operations/userOperations.ts
```diff
@@ -0,0 +1,47 @@
+const ethJsUtil = require('ethereumjs-util')
+const models = require('../models')
+const { User, UserReference } = models
+import * as types from '../types'
```
julien51: Since you only use `UserCreationInput` maube use 
```suggestion
import { UserCreationInput } as types from '../types'
```

locksmith/src/operations/userOperations.ts
```diff
@@ -0,0 +1,47 @@
...
+
+namespace UserOperations {
+  export async function createUser(
+    input: types.UserCreationInput
```
julien51:
```suggestion
    input: UserCreationInput
```

locksmith/__tests__/operations/userOperations.test.ts
akeem: At the moment, there are not type definitions defined for these generated objects. This can be done, but I think that it can be done after this lands

cnasc: I agree with @akeem on this one. `any` is undesirable, and in many cases it's trivially easy to get rid of it. But if in this case there's no provided typedef, it's probably more important to take it down as a TODO so we can keep moving forward.

locksmith/__tests__/operations/userOperations.test.ts
akeem: i'm not sure if these test will provide a lot of value if we are mocking out the implementation of its internals

locksmith/__tests__/operations/userOperations.test.ts
akeem: A similar sentiment to the other test above

akeem: This logic is tested in the controller's tests

PR #2277, opened 2019-03-22
waiting for postgres to be up before running locksmith

julien51:
When looking at the logs on integration tests (like https://circleci.com/gh/unlock-protocol/unlock/13929) it appears that locksmith is making requests to the database server even though this one is not up yet.

This PR forces locksmith to wait for postgres to be up!


locksmith/server.js
```diff
@@ -1,4 +1,48 @@
...
+const config = require('./config/config')[env]
+
+const port = process.env.PORT || 8080
+
+/**
+ * This is a helper function to ensure that we start the test suite only when the server is up
+ * We will retry for
+ */
+const serverIsUp = (host, port, delay, maxAttempts, callback) => {
+  let attempts = 1
```
benwerd: It feels more "correct" to me to have this initialized at 0, but this definitely works.

locksmith/server.js
```diff
@@ -1,4 +1,48 @@
...
+  }
+  tryConnecting()
+}
+
+if (!config.host) {
+  return app.listen(port)
+}
+
+// We wait for the db server to be up before starting the app
+serverIsUp(config.host, config.port, 100, 120, error => {
```
akeem: Can we isolate all of the changes so that they only occur when in the test environment? I am concerned that this may introduce variability into that app startup in production where currently do not have an issue (at least where this is concerned)

julien51: Good point!

julien51: PTAL

julien51:
@akeem I added a check on the if statement. PTAL ;)

PR #2279, opened 2019-03-22
Log Page Refactored (for real this time)

cnasc:
# Description

This is a replacement for #2224, which wasn't as clean as it needed to be.

The key difference between this one and that one is that there is no longer an ugly metadata object threaded through. The transaction type strings are readable by default, and a simple map from `hash: url` is generated for the chain explorer URLs and passed down.

# Checklist:


unlock-app/src/components/content/LogContent.tsx
```diff
@@ -0,0 +1,70 @@
...
+import * as UnlockTypes from '../../unlock'
+
+interface Props {
+  account: UnlockTypes.Account
+  network: UnlockTypes.Network
+  transactionFeed: UnlockTypes.Transaction[]
+  explorerLinks: { [key: string]: string }
+}
+
+export const LogContent = ({ account, network, transactionFeed, explorerLinks }: Props) => {
```
cnasc: @cellog

cellog: woot!

unlock-app/src/__tests__/config.test.js
```diff
@@ -146,7 +146,7 @@ describe('config', () => {
     it('should contain the right URLs for chain explorers', () => {
       expect.assertions(2)
       expect(Object.keys(config.chainExplorerUrlBuilders)).toHaveLength(1)
-      expect(config.chainExplorerUrlBuilders.etherScan('0x0')).toEqual(false)
+      expect(config.chainExplorerUrlBuilders.etherScan('0x0')).toEqual('')
```
cellog: let's use `toBe`

unlock-app/src/components/content/LogContent.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <Account network={network} account={account} />
+            <CreatorLog transactionFeed={transactionFeed} explorerLinks={explorerLinks} />
+          </BrowserOnly>
+        )}
+      </Layout>
+    </GlobalErrorConsumer>
+  )
+}
+
+interface State {
```
cellog: I wonder if we can find a more specific name for this and `Config`? Let's open up a follow-up issue and discuss there?

unlock-app/src/components/content/LogContent.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+interface Config {
+  chainExplorerUrlBuilders: UnlockTypes.ChainExplorerURLBuilders
+}
+
+export const mapStateToProps = (
+  { account, network, transactions }: State,
+  { config: { chainExplorerUrlBuilders } }: { config: Config }
+) => {
```
cellog: can we specify a return type as well here? Or does it mess with redux typings?

unlock-app/src/__tests__/config.test.js
cnasc: I think if that's an important change to be made, it should be done to that file as a whole in a separate PR instead of introducing it to just one of the many test cases using `toEqual` in this PR.

unlock-app/src/components/content/LogContent.tsx
cnasc: I'm definitely open to suggestions, note that these types never leave this file though. They're really just there to keep `no-any` off our backs.

unlock-app/src/components/content/LogContent.tsx
cnasc: Adding a return type here works, I've pushed a change with that in there (it's just `: Props` before the `=>`)

unlock-app/src/__tests__/config.test.js
cellog: that makes sense, I did not see the wider context of the needed change. Open an issue?

Issue #1945, opened 2019-03-07
Account header on Dashboard and Log should be updated to match Figma


cnasc:
![image.png](https://images.zenhubusercontent.com/5c7808ab9546cc1a88f4f8bd/2d0b15fd-c8d9-42cd-9f05-1360233212c1)

The account header has been redesigned, and some of our assumptions moving forward rely on this layout.

Some requirements:

[ ] The "Create Lock" button has 2-pronged behavior
  [ ] On the dashboard, it opens the Create Lock form
  [ ] On the log page, it moves you to the Dashboard (and also opens the form?)

[ ] The search will have to be implemented
  [ ] Initial version can be fairly naive
  [ ] Later versions will likely involve some interaction with Locksmith.

Complexity estimates:
 - Header Layout: simple
 - Button Behavior: medium difficult
 - Search: complex

julien51:
I think we deprecated search a long time ago as there is no real use case for it (even in the mid term future...) ;) @smombartz

smombartz:
will remove search for ever ever!

PR #2301, opened 2019-03-22
Add lock creation form visibility to state

cnasc:
# Description

In order to make #1945 work, we need a way to communicate whether the lock creation form should open across components. This PR adds the actions and reducer necessary for the "Create Lock" button to do its thing  across pages.

# Checklist:

### DIFF
--- a/unlock-app/src/reducers/lockFormVisibilityReducer.ts
+++ b/unlock-app/src/reducers/lockFormVisibilityReducer.ts
@@ -5,7 +5,7 @@ import { SET_PROVIDER } from '../actions/provider'
 import { SET_NETWORK } from '../actions/network'
 
 export const initialState = {
-  showLockCreationForm: false,
+  visible: false,
 }
 
 const lockFormVisibilityReducer = (state = initialState,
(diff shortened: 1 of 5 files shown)


unlock-app/src/reducers/lockFormVisibilityReducer.ts
```diff
@@ -0,0 +1,30 @@
...
+import { SET_PROVIDER } from '../actions/provider'
+import { SET_NETWORK } from '../actions/network'
+
+export const initialState = {
+  visible: false,
+}
+
+const lockFormVisibilityReducer = (state = initialState,
+                                   action: { type: string }) => {
+  if ([SET_PROVIDER, SET_NETWORK].indexOf(action.type) > -1) {
```
cellog: I think we also need `SET_ACCOUNT` here?

cnasc: Oh, I think you're right. I used the accounts reducer as my template which handled that separately, but if I'd picked any other one I'd have gotten it right. I'll add that one in.

PR #2305, opened 2019-03-23
Loading the transaction rather than loading all past transactions.

julien51:
This a small improvment: it loads the lock creation transaction rather than loading all past lock
transactions. This allows for a slightly faster rendering of the UI, as well as saves 1 transaction per lock...


julien51:
It still happens, but lower in the chain.
Basically when we retreive this specific transaction we find that it's a lock, and then retrieve its pas transactions. (we were doing that twice...)

Issue #2307, opened 2019-03-23
merge back in changes from #2250 into the paywall


cellog:
The TransactionType refactoring from #2250 needs to be merged back into the paywall

Issue #2284, opened 2019-03-22
Paywall should show "pending" state prior to "confirming" state (the blue design) after key purchase clicked


cellog:
Currently, when clicking on the "purchase key" button, nothing is shown for up to 5 seconds. It would be good to show a purchase pending state, using Sascha's blue design.

cnasc:
#1383?

cellog:
nope this is the blue lock, which should show in between the green lock and yellow lock

cnasc:
Right, "between when a user clicks on the purchase button and when the lock shows that it has started confirming"

cellog:
So this is actually caused by a deeper issue, one which I am unsure how to address best. Basically, the problem is at the intersection of the transaction creation and the parsing of receipt.

Here is the progression for key purchase:

1) pending transaction emitted
2) hash received, new transaction emitted
3) <long delay>
4) first transaction receipt is emitted

The first 2 events are emitted by walletService, the last by web3Service. Only web3Service has the logic that sets the metadata on a transaction, which in the case of key purchase is the key and lock that is used by the UI to figure out what to display. Thus, in order to properly address this issue, it looks like some of the logic that is in web3Service will need to be available inside the walletService as well.

This is where my question comes in. We could extract the common logic from web3Service, which would mean taking `eventsHandlers` and making that a shared export. We could also duplicate the logic, since this is only used for pre-populating pending transactions.

Another less invasive alternative would be to have `transaction.new` basically assume that the transaction is a key purchase on the paywall, and fake the lock id that is looked for by the `Lock` component. This is a brittle solution, but would work for now.

So the question is do we want to do something more generic inside walletService, or resort to a bit of a hack inside the paywall only in the middleware?

I'm pretty sure the same issue is responsible for the weird behavior of the `LockIconBar` on the creator dashboard as well, otherwise I would just do the paywall solution and forget about it.

Opinions, @julien51 ?

cnasc:
Just to add one more hacky idea: this intermediate state is really just to paper over the hang between when a transaction starts and when it first gets mined. So the solution could be entirely local to the button: once you click submit, it renders the pending state until something else (error, mining success) happens to change it.

That doesn't tackle the `LockIconBar` problem, of course.

cellog:
that is probably the simplest solution, thanks Chris

cellog:
this is fixed

PR #2308, opened 2019-03-23
pass transaction type to emission of transaction.pending

### LINKED ISSUE #2300: Add transaction type to walletService._sendTransaction()
In order to implement an immediate state change for transactions, it would be great if `walletService._sendTransaction` accepted a 3rd parameter, which would simply be the type of transaction initiated. This can then be passed to the `this.emit('transaction.pending')` as a 2nd parameter, which would set the transaction type, and allow immediate updating of the UI to show that the transaction has started.

cellog:
# Description

In preparation for #2284 this PR passes the transaction type to `_sendTransaction`, and then emits it with the `'transaction.pending'` event so that the UI can use this to display a pending state as soon as the user initiates the transaction by clicking/tapping on the button that sends it.

(Note that #2307 addresses the inconsistency across the apps and will follow this one)

# Issues

Fixes #2300 

# Checklist:

### DIFF
--- a/paywall/src/services/walletService.js
+++ b/paywall/src/services/walletService.js
@@ -12,6 +12,7 @@ import {
   FAILED_TO_UPDATE_KEY_PRICE,
   FAILED_TO_WITHDRAW_FROM_LOCK,
 } from '../errors'
+import { TRANSACTION_TYPES } from '../constants'
 
 export const keyId = (lock, owner) => [lock, owner].join('-')
 
@@ -139,7 +140,7 @@ export default class WalletService extends EventEmitter {
    * receipt... etc)
    * @private
    */
-  _sendTransaction({ to, from, data, value, gas }, callback) {
+  _sendTransaction({ to, from, data, value, gas }, callback, transactionType) {
     const web3TransactionPromise = this.web3.eth.sendTransaction({
       to,
       from,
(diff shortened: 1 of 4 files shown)


paywall/src/services/walletService.js
```diff
@@ -139,7 +140,7 @@ export default class WalletService extends EventEmitter {
    * receipt... etc)
    * @private
    */
-  _sendTransaction({ to, from, data, value, gas }, callback) {
+  _sendTransaction({ to, from, data, value, gas }, callback, transactionType) {
```
julien51: Please always use the callback as the _last_ argument on a function...

unlock-app/src/services/walletService.js
```diff
@@ -139,7 +140,7 @@ export default class WalletService extends EventEmitter {
    * receipt... etc)
    * @private
    */
-  _sendTransaction({ to, from, data, value, gas }, callback) {
+  _sendTransaction({ to, from, data, value, gas }, callback, transactionType) {
```
julien51: Same comment as above ;)

paywall/src/services/walletService.js
cellog: Ok. Another thought: should I put the translation type inside the object we pass instead of passing a new parameter?

julien51: I'd say no because I think this type is really specific to Unlock while the other fields are generic in Ethereum.

cellog: Sounds good.

cellog:
all done, PTAL

cellog:
all changes requested are up-to-date, PTAL @julien51

PR #2314, opened 2019-03-23
using prettier inside eslint

julien51:
Greg recently found that the recommended way to use prettier and eslint is explained on this page: https://prettier.io/docs/en/eslint.html

This PR introduces the required plugin.
The next Pr on this front will remove the `reformat` script and its associated check for changed files since eslint will fail when files are not respecting prettier's rules.

PS: some files may be failing formatting... I'll add them to this PR.


julien51:
Finally, I'll experiment using a single eslint config at the root of the repo rather than have each repo implement its own rules!

julien51:
We are also using both `prettier` and `plugin:prettier/recommended` which is probably too much.
I strongly believe we should not _only_ rely on prettier for code styles and still enforce some rules that eslint provides.

cellog:
> We are also using both `prettier` and `plugin:prettier/recommended` which is probably too much.
> I strongly believe we should not _only_ rely on prettier for code styles and still enforce some rules that eslint provides.

The reason for disabling rules is because we will get the same error twice, once from prettier, and once from eslint. Here's a quote:

"Whether you run Prettier via ESLint or run both tools separately, you probably only want to hear about each formatting issue once, and you especially don't want ESLint to complain about formatting "issues" which are simply a different preference than what Prettier does."

https://prettier.io/docs/en/eslint.html

Additionally, using them both together triggers reformatting that generates invalid javascript in some cases (https://github.com/prettier/eslint-config-prettier/blob/2c842675e55b91aecaef6f997d234ebf2d220ffb/README.md#arrow-body-style-and-prefer-arrow-callback and https://gitlab.com/gitlab-org/gitlab-ce/issues/52233)

It is for these 2 reasons that they have made the recommendation of turning off certain eslint rules.

eslint provides other benefits, such as detecting unused variables, missing prop types, and so on, none of which prettier can help us with.

julien51:
> The reason for disabling rules is because we will get the same error twice, once from prettier, and once from eslint. 

I think this is a minor inconvenience vs. not knowing that a rule which we used to enforce is not enforce anymore (if we disable eslint's formatting...), so, we will keep eslint's rules. 

I think there is a confusion between `plugin` and the `extends` lines.
My understanding is that there is no need to use `"extends": ["prettier"]` when using `"extends": ["plugin:prettier/recommended"]`, but eslint still needs to know about the prettier rules, which it does thanks to `plugins: ['prettier']`, so I do not think we should remove that line.

cellog:
https://prettier.io/docs/en/eslint.html#use-both

`"extends": ["plugin:prettier/recommended"]` is a combination of:

`"extends": ["prettier"]`

and

`"plugins": ["prettier"],`

There is no need for `"extends": ["prettier"]`

julien51:
Thanks! Changes done. PTAL.

PR #2331, opened 2019-03-25
enable configuration for paywall demo

cellog:
# Description

In preparation for the move of the demo page to the paywall app, this adds 2 variables to the paywall. The first is the path to the paywall, which is defined by `PAYWALL_URL` in the environment. The second (and this is a departure from unlock-app) is `PAYWALL_SCRIPT_PATH` which is `'/static/paywall.min.js'` by default. The demo page uses the paywall URL plus the path to locate the paywall script. Doing things this way means we don't need a separate value for dev, staging and master, and removes an unnecessary point of failure.

# Issues

Refs #2324 

# Checklist:

### DIFF
--- a/paywall/next.config.js
+++ b/paywall/next.config.js
@@ -10,6 +10,8 @@ module.exports = withSourceMaps({
     unlockEnv: process.env.UNLOCK_ENV || 'dev',
     httpProvider: process.env.HTTP_PROVIDER || '127.0.0.1',
     readOnlyProvider: process.env.READ_ONLY_PROVIDER,
+    paywallUrl: process.env.PAYWALL_URL,
+    paywallScriptUrl: process.env.PAYWALL_SCRIPT_PATH,
     locksmithHost: process.env.LOCKSMITH_URI || 'http://127.0.0.1:8080',
     unlockAddress:
       process.env.UNLOCK_ADDRESS ||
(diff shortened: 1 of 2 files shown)


paywall/src/config.js
```diff
@@ -68,6 +68,9 @@ export default function configure(
 
   const env = runtimeConfig.unlockEnv
 
+  let paywallUrl = runtimeConfig.paywallUrl || 'http://127.0.0.1:3001'
```
cellog: note: the choice of `127.0.0.1:3001` is very intentional. This will trigger cross-domain protections when accessed from `localhost:3001`, which is where the demo will live.

julien51: Could you please add that as a comment above that line? I worry that we will forget the reason for this in the future..

cellog: good idea!

paywall/next.config.js
```diff
@@ -10,6 +10,8 @@ module.exports = withSourceMaps({
     unlockEnv: process.env.UNLOCK_ENV || 'dev',
     httpProvider: process.env.HTTP_PROVIDER || '127.0.0.1',
     readOnlyProvider: process.env.READ_ONLY_PROVIDER,
+    paywallUrl: process.env.PAYWALL_URL,
+    paywallScriptUrl: process.env.PAYWALL_SCRIPT_PATH,
```
julien51: `paywallScriptPath`?

julien51: At least this would be better in terms of name since this is a path... and it looks like this is what you use below ;)

cellog: Weird, I thought I did that :). Definitely intended to, maybe I didn't push the commit. In any case, will do!

PR #2335, opened 2019-03-25
Svm to typescript

akeem:
# Description

To be rebased against master after #2334 is merged, this convert the library to typescript with some very basic type support. 

This is in effort to support signed user creation requests.

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/middlewares/signatureValidationMiddleware.ts
```diff
@@ -0,0 +1,70 @@
...
+  function validatePayloadContent(payload: any): Boolean{
+    if (payload.lock) {
+      if (payload.lock.name && payload.lock.owner && payload.lock.address) {
+        return true
+      }
+    }
+    return false
+  }
+
+  export function process(req: any, res: Response, next: any) {
```
cellog: can we use `next: Function` or whatever the typescripty way to do this is?

locksmith/src/middlewares/signatureValidationMiddleware.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (payload.lock) {
+      return (
+        ethJsUtil.toChecksumAddress(payload.lock.owner) ===
+        ethJsUtil.toChecksumAddress(signee)
+      )
+    } else {
+      return false
+    }
+  }
+  function validatePayloadContent(payload: any): Boolean{
```
cellog: I think we need a type that includes lock as an optional type for payload

locksmith/src/middlewares/signatureValidationMiddleware.ts
```diff
@@ -0,0 +1,70 @@
...
+      req.headers.authorization &&
+      req.headers.authorization.split(' ')[0] === 'Bearer'
+    ) {
+      return req.headers.authorization.split(' ')[1]
+    } else {
+      return null
+    }
+  }
+
+  function validatePayload(payload: any, signee: string): Boolean{
```
cellog: I think we need a type with lock as an optional sub-thing of payload

cnasc: This would be pretty easy to do.

```typescript
interface Payload {
  lock: Lock // NB. assumes that there is a type for Lock.
}
```

Since locks will probably be used and passed around a lot, it would be helpful to type them. And I think it should be pretty straightforward to do, no rabbit holes to fall into.

locksmith/src/middlewares/signatureValidationMiddleware.ts
cnasc: I think `next` functions in Express never return anything, so you could do `next: () => void`

akeem: yeah the reason I held off on setting this is in the case that somewhere in the chain of middleware there is something that returns disaster ensues

cnasc: It shouldn't cause any concerns at runtime, so as long as it builds you'll be fine

locksmith/src/middlewares/signatureValidationMiddleware.ts
akeem: So there is a longer story here, but the short version is that we have a representation of this in the creator dashboard https://github.com/unlock-protocol/unlock/blob/master/unlock-app/src/structured_data/unlockLock.js

While it is trivial to add it here, it will be pulled out in the subsequent code to follow this PR to unify what this code represents. If possible i'd like to do this only once

cnasc: That sounds reasonable and prudent to me

PR #2330, opened 2019-03-25
Update account header to match Figma

cnasc:
# Description

This PR is a huge sweeping change to bring our UI back into sync with the Figma. It lifts the create lock button out of the list of locks and into the omnipresent account header within `unlock-app`.

I will keep this as a draft for the moment.

image

# Issues

Fixes #1945 

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [x] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [x] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [x] I have performed a self-review of my own code
- [x] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #2338, opened 2019-03-26
Add types to theme/media

cnasc:
# Description

Part of splitting #2330. This turns the `theme/media` file into TypeScript, a change required for TS files to be able to use these declarations in `styled-components`. Right now these types are vague, with lots of `any`. A brief attempt at making them more concrete was quickly stymied, and in the future I'd recommend just leaving this file alone since it's unlikely to cause any problems.

# Checklist:

### DIFF
--- a/unlock-app/src/theme/media.ts
+++ b/unlock-app/src/theme/media.ts
@@ -1,7 +1,7 @@
 import styled, { css } from 'styled-components'
 import { MAX_DEVICE_WIDTHS } from '../constants'
 
-let sizes = {
+let sizes: { [key: string]: any } = {
   desktop: {
     min: MAX_DEVICE_WIDTHS.TABLET,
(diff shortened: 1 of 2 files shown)


unlock-app/src/theme/media.ts
```diff
@@ -1,7 +1,7 @@
 import styled, { css } from 'styled-components'
 import { MAX_DEVICE_WIDTHS } from '../constants'
 
-let sizes = {
+let sizes: { [key: string]: any } = {
```
cellog: this is because they can be either string or number, right? Is there any chance of using a union type like `sizes: { [key: string]: number | string }`?

cnasc: Actually, they can be number or boolean. However, with that `number | boolean` type in there, the code that generates the Media object starts to error and I decided that it wasn't worth the time to go further.

PR #2350, opened 2019-03-26
Update Account to match Figma.

cnasc:
# Description

Part of splitting up #2330. This rearranges the `Account` component so that its visual style matches Figma. Additionally, the file was moved to TypeScript.

One of the deps we rely on for this component has no type definition package available. As a result, a mock definition was added so that `tsc` wouldn't fail with a `no implicit any` error. I don't think we have time at the moment to define more robust types for third-party deps, so for now it's just a stub module declaration.

image

# Checklist:


cnasc:
Thanks Greg!

PR #2383, opened 2019-03-27
don't connect until the middleware chain is complete

### LINKED ISSUE #2381: walletMiddleware can under certain circumstances dispatch before middlewares are ready
If the walletService constructor emits an `error`, this will cause the walletMiddleware to attempt to dispatch 2 actions before the middleware chain is constructed, which will throw an error. This affects ~~both `unlock-app` and~~ the `paywall`

cellog:
# Description

This fixes a potential crash if the provider is missing on the paywall, by merging in a change made to walletMiddleware on `unlock-app`. The change guarantees that connection errors will not trigger until the middleware chain is complete. Using `setTimeout` simply guarantees the connection attempt will not happen until the synchronous middleware registration is complete.

# Issues

Fixes #2381

# Checklist:

### DIFF
--- a/paywall/src/middlewares/walletMiddleware.js
+++ b/paywall/src/middlewares/walletMiddleware.js
@@ -114,7 +114,10 @@ export default function walletMiddleware({ getState, dispatch }) {
 
   return function(next) {
     // Connect to the current provider
-    // We connect once the middleware has been initialized, using setTimout (warning: fragile?)
+    // We connect once the middleware has been initialized, using setTimout
+    // The redux event loop is synchronous, so using setTimeout guarantees
+    // that connect will only be called after all of the other middleware have
+    // been initialized
     setTimeout(() => {
       walletService.connect(getState().provider)
     })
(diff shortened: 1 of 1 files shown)


paywall/src/middlewares/walletMiddleware.js
```diff
@@ -114,7 +114,10 @@ export default function walletMiddleware({ getState, dispatch }) {
 
   return function(next) {
     // Connect to the current provider
-    walletService.connect(getState().provider)
+    // We connect once the middleware has been initialized, using setTimout (warning: fragile?)
+    setTimeout(() => {
```
benwerd: This does seem really fragile - is it guaranteed that the middleware will have initialized when `setTimeout` is called?

cellog: Yes it is, because the entirety of redux is synchronous. Middleware are allowed to do other async stuff, but when it hooks back into redux via dispatch, the entire set of middlewares/reducers is run before the process releases control back. In this case, the `setTimeout` will simply postpone the connect call until after that synchronous loop has finished, which basically looks like this:

```
createStore is called
middleware 1 is initialized
middleware 2 is initialized
store is returned
(whatever else happens in the script, setting up React tree, etc.
...
process concludes and releases control to the event loop
--> setTimeout callback is called
```

Redux has had some pressure to change this from the React devs, but because redux supports anything, and is used in all kinds of frameworks, this will never happen, and we can rely upon it being that way until the end of time. amen.

PR #2390, opened 2019-03-28
Fix locksmith eslint

cnasc:
# Description

This PR brings `locksmith` up to speed with the changes to the eslint config. Additonally, it updates the various files that snuck through our old linting process.

# Checklist:


locksmith/src/controllers/userController.ts
```diff
@@ -1,9 +1,13 @@
-import UserOperations = require('../operations/userOperations')
-import { Request, Response } from 'express-serve-static-core'
+import { Request, Response } from 'express-serve-static-core' // eslint-disable-line no-unused-vars, import/no-unresolved
 import { DecoyUser } from '../utils/decoyUser'
 
+import UserOperations = require('../operations/userOperations')
+
 namespace UserController {
-  export async function createUser(req: Request, res: Response): Promise<any> {
+  export const createUser = async (
```
akeem: Do you have a link to any documentation around with function definitions in this fashion are preferred?

locksmith/src/middlewares/signatureValidationMiddleware.ts
```diff
@@ -1,10 +1,10 @@
 import sigUtil from 'eth-sig-util'
 import ethJsUtil from 'ethereumjs-util'
+import { Request, Response } from 'express-serve-static-core' // eslint-disable-line no-unused-vars, import/no-unresolved
```
akeem: Is this around the use of Response?

locksmith/src/types.ts
```diff
@@ -4,3 +4,10 @@ export interface UserCreationInput {
   passwordEncryptedPrivateKey: String
   recoveryPhrase: String
 }
+
```
akeem: Can we disable this eslint rule? If possible, I dont think we should add code that we arent going to use to make it happy

locksmith/src/controllers/userController.ts
cnasc: This is one example of eslint doing the wrong thing. This is totally valid TS, but eslint doesn't like having a function declaration at anything other than the program root. Creating the functions as a value is allowed everywhere, so this should have the same exact behavior while causing eslint to be quiet.

On a side note, doing things inside namespaces may not be necessary with ES6 modules, but we can concern ourselves with that at a later date.

locksmith/src/middlewares/signatureValidationMiddleware.ts
cnasc: Eslint thinks that when you import types, they always go unused. So we have to disable that rule whenever we import types.

locksmith/src/types.ts
cnasc: I tried disabling it, unfortunately it didn't work. Since you have a PR in the pipeline that will add a type here, we can probably coordinate on making this only a very temporary change.

cnasc: And/or add that new type as part of this PR

cnasc:
Looks like this is all resolved, @akeem PTAL

PR #2459, opened 2019-04-01
extracted shared config for eslint to be shared among all folders

julien51:
The next step is making sure this shared config is also used for other projects.


julien51:
I am honestly flabbergasted that linting fails in 2 other folders which i did not change.

smart-contracts/reports/gas.js
```diff
@@ -19,6 +20,7 @@ contract('Reports', accounts => {
   })
 
   it('gas usage report', async () => {
+    expect.assertions(1)
```
hardlydifficult: this is causing the build fail.  not sure where this line came from, but not valid syntax for these tests

hardlydifficult:
The merge conflict is from #2382
Just adding a couple new globals `after` and `afterEach`

smart-contracts/reports/gas.js
julien51: Yeah, this is `jest` and a pretty useful thing that we enforce on all other tests. I would be on favor of moving over to jest instead of mocha too for smart contract but not sure what is involved.

julien51:
Ok, tests are passing now... PTAL!

PR #2460, opened 2019-04-01
Provides signature validation to User endpoints

akeem:
# Description

We are cutting over from unvalidated endpoints to those
utilizing the middleware that covers Lock related endpoints

This should be relatively straight foward, ideally a renaming of some
of the existing components will probably be a worthwhile fast follow.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/__tests__/operations/userOperations.test.ts
+++ b/locksmith/__tests__/operations/userOperations.test.ts
@@ -1,4 +1,7 @@
 import UserOperations from '../../src/operations/userOperations'
+import RecoveryPhrase from '../../src/utils/recoveryPhrase'
+
+jest.mock('../../src/utils/recoveryPhrase', () => ({}))
 
 const models = require('../../src/models')
 
@@ -18,6 +21,7 @@ describe('User creation', () => {
     it('should normalize the public key/address & email address', async () => {
       expect.assertions(1)
       UserReference.create = jest.fn(() => {})
+      RecoveryPhrase.generate = jest.fn(() => 'generated phrase')
 
       await UserOperations.createUser(userCreationDetails)
       expect(UserReference.create).toHaveBeenCalledWith(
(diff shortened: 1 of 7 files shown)


locksmith/src/controllers/userController.ts
```diff
@@ -1,23 +1,21 @@
...
-    let user = req.body.user
+    let user = req.body.message.user
 
     try {
       if (user) {
         let creationStatus: Boolean = await UserOperations.createUser({
           emailAddress: user.emailAddress,
           publicKey: user.publicKey,
           passwordEncryptedPrivateKey: user.passwordEncryptedPrivateKey,
-          recoveryPhrase: RecoveryPhrase.generate(),
```
julien51: I see this is the place where you moved it one level down, right?

locksmith/src/operations/userOperations.ts
```diff
@@ -1,5 +1,7 @@
 import { UserCreationInput } from '../types' // eslint-disable-line no-unused-vars
-import * as Normalizer from '../utils/normalizer' // eslint-disable-line no-unused-vars
+import * as Normalizer from '../utils/normalizer'
+// eslint-disable-line no-unused-vars
```
julien51: is that eslint being sily?

locksmith/__tests__/operations/userOperations.test.ts
```diff
@@ -20,14 +21,14 @@ describe('User creation', () => {
 
       await UserOperations.createUser(userCreationDetails)
```
julien51: Could you add a mock on `RecoveryPhrase.generate()` so that we know that the string below is actually coming from it?

locksmith/src/operations/userOperations.ts
akeem: yeah, I am not sure what that is about. I'll see if I can save it without it being re-adjusted

locksmith/src/controllers/userController.ts
akeem: yes , that is correct!

locksmith/__tests__/operations/userOperations.test.ts
akeem: yes, i'll adjust this

akeem:
@julien51 would you mind giving this another review when you have a moment?

locksmith/src/operations/userOperations.ts
cnasc: Is this just barely over the line-length limit? That's annoying. You can use disable-next-line (or something like that) if it persists.

Issue #2393, opened 2019-03-28
add optimism to the paywall UI


cellog:
With optimism of either `1` for "yes, expect that this will successfully unlock content" and `0` for "no, show the paywall until everything is confirmed", the UI should either display the new optimistic flag or the full paywall.

cellog:
This is fully implemented.

PR #2461, opened 2019-04-01
make optimistic locking fully operational (frontend-side)

cellog:
# Description

This enables the `useOptimism` hook in the paywall, so that it actively retrieves the current optimistic success from locksmith, and the past value from cache, in order to update the UI accordingly

# Issues

Refs #2393 

# Checklist:

### DIFF
--- a/paywall/src/components/Paywall.js
+++ b/paywall/src/components/Paywall.js
@@ -17,11 +17,13 @@ import {
 } from '../paywall-builder/constants'
 import { isPositiveNumber } from '../utils/validators'
 import useWindow from '../hooks/browser/useWindow'
+import useOptimism from '../hooks/useOptimism'
+import { TRANSACTION_TYPES } from '../constants'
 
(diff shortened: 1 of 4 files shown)


paywall/src/components/Paywall.js
```diff
@@ -110,12 +121,29 @@ export const mapStateToProps = ({ locks, keys, modals, router, account }) => {
     }
   })
 
+  const lock = locksFromUri.length ? locksFromUri[0] : { address: 0 }
```
julien51: `{ address: 0 }` is that supposed to a no-op lock?
Not feeling very good about this. can you explain what is going on?

cellog: yes. Basically, as I wrote out an explanation, I figured out a simpler way to do this. I'll send it up and request another look.

cellog:
ok @julien51 take a look. Instead of using `{ address: 0 }` it is using `null` and adding a check for whether the lock exists on line 129. This should be much less wtf-y :)

PR #2463, opened 2019-04-02
Removing static content from app and pointing to static site

benwerd:
# Description

This PR removes static pages from `unlock-app`, re-links buttons to the static site, and updates stories and tests accordingly.

# Checklist:


unlock-app/next.config.js
```diff
@@ -23,6 +23,9 @@ let requiredConfigVariables = {
 if (process.env.UNLOCK_URL)
   requiredConfigVariables.unlockUrl = process.env.UNLOCK_URL
 
+if (process.env.UNLOCK_STATIC_URL)
+  requiredConfigVariables.unlockStaticUrl = process.env.UNLOCK_STATIC_URL
```
julien51: I am not sure this is the right approach because this bypasses the mechanism below which would fail deploys if this is unset. I think `UNLOCK_STATIC_URL` and `UNLOCK_URL` should be set above inside the `requiredConfigVariables` and left to `null`/`undefined` if unset. This means that they will be overriden by config in dev but if none is set we cannot deploy to production (which is good because that would break links!)

unlock-app/src/__tests__/config.test.js
```diff
@@ -3,6 +3,9 @@ import {
   CANONICAL_BASE_DEV_URL,
   CANONICAL_BASE_STAGING_URL,
   CANONICAL_BASE_URL,
+  CANONICAL_STATIC_URL,
+  CANONICAL_STATIC_DEV_URL,
```
julien51: So I am not sure this approach is right here. We should not put have different variables but a single one set based on the actual env. This would let us re-deploy with just a env var change, without a code change.

julien51: [I know this is what happens wfor `CANONICAL_BASE_URL` but I think this is wrong too.

unlock-app/src/config.js
```diff
@@ -4,6 +4,9 @@ import {
   CANONICAL_BASE_DEV_URL,
   CANONICAL_BASE_STAGING_URL,
   CANONICAL_BASE_URL,
+  CANONICAL_STATIC_DEV_URL,
+  CANONICAL_STATIC_STAGING_URL,
+  CANONICAL_STATIC_URL,
```
julien51: See comment above but I think this is not the right approach. We need a single `CANONICAL_STATIC_URL` set as an env variable

benwerd:
PTAL. Thanks!

benwerd:
PTAL. Please note that I decreased the function coverage threshold to 88.7% because removing code reduced coverage in a way that's irrelevant to the content of this PR. @julien51

unlock-app/src/components/interface/buttons/homepage/HomepageButton.js
```diff
@@ -41,11 +44,11 @@ export class HomepageButton extends React.Component {
           <TermsBox>
             <div>
               To access the dashboard you need to agree to our&nbsp;
-              <Link href="/terms">
+              <Link href={config.unlockStaticUrl + '/terms'}>
```
julien51: We need a URL Builder at some point ;p

benwerd: Yes, but that's something for another PR ;)

unlock-app/src/__tests__/components/content/LogContent.test.tsx
```diff
@@ -60,6 +61,8 @@ const config = {
     etherScan: (address: string) =>
       `https://blockchain.party/address/${address}/`,
   },
+  unlockUrl: 'https://foo',
```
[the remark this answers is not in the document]
benwerd: Good catch! I think I added that during a previous iteration. Removed.

PR #2464, opened 2019-04-02
re-enabled the integration tests for the paywall

julien51:
This re-enables the integration tests for the paywall.
The PR does a lot of things and is probably not fully clean, but I wanted to get it out the door.

I'll leave comments in there to explain some of the logic.


paywall/src/components/lock/ConfirmingKeyLock.js
```diff
@@ -14,7 +14,7 @@ import BalanceProvider from '../helpers/BalanceProvider'
 import withConfig from '../../utils/withConfig'
 
 export const ConfirmingKeyLock = ({ lock, transaction, config }) => (
-  <LockWrapper id={`Confirming_${lock.address}`}>
+  <LockWrapper lock={lock}>
```
julien51: I think that rather than using `id` which are adding too much coupling we can use a `data-address` attribute to identify locks based on their addresses. The `LockWrapper` component does that.

paywall/src/components/lock/UnlockFlag.js
```diff
@@ -24,14 +24,16 @@ export const UnlockedFlag = () => {
     return () => clearTimeout(timeout)
   }, [setHidden])
   return (
-    <Flag id="UnlockFlag" hidden={hidden}>
+    <Flag hidden={hidden}>
```
julien51: Generally I think it is better to use the styled components to set attributes which need to be set for all instances of a given component, rather than expect/hope that the implementer will mot forget to add the `id` or `className`.

tests/test/dashboard.test.js
```diff
@@ -55,7 +47,7 @@ describe('The Unlock Dashboard', () => {
         .substring(7)}`
       const expirationDuration = Math.floor(1000 * Math.random()).toString(10)
       const maxNumberOfKeys = Math.floor(100000 * Math.random()).toString(10)
-      const keyPrice = (Math.floor(10000 * Math.random()) / 100.0).toString(10)
+      const keyPrice = (Math.floor(990 * Math.random()) / 100.0).toString(10) // we want key prices less than 10 Eth
```
julien51: The prices above 100 meant that the testing user could not purchase keys :(

julien51:
The irony of having these integration tests fail is not lost on me!

julien51:
I'll try to debug that and see exactly what is going on on this one because it passes locally consistently!

cellog:
Don't worry, we empathize strongly - every time the tests seemed solid in my past work, they freaked out on CI :). We'll get there!

PR #2479, opened 2019-04-02
Provider Middleware

cnasc:
# Description

This draft PR illustrates the introduction of a middleware to handle providers. Right now, initialization of a provider happens in the wallet middleware in response to the `SET_PROVIDER` action. Conceptually, that isn't very clean. We also need to do some special work for the providers anyway so that we can effectively use the (upcoming) Unlock provider, so now seems like the time to split out management of providers into its own middleware.

The idea here is that the provider will be a resident of the config file, which will be a singleton (ref #2470). The provider middleware will trigger the initialization of the provider (in particular, the `enable` call). The provider will then be passed down through the config to those components that require it.

https://github.com/unlock-protocol/unlock/blob/8c63ea46d388a5e933ef9b4f6d5bac6166825aee/unlock-app/src/services/walletService.js#L54-L95

The `connect` method will still exist on the wallet service, but it will just be the part from line 87 onward, where it creates an instance of Web3 based on the given provider.

This PR needs a little more time to bake, but I wanted to put it up to solicit comments about this approach.

# Checklist:

### DIFF
--- a/unlock-app/src/middlewares/providerMiddleware.ts
+++ b/unlock-app/src/middlewares/providerMiddleware.ts
@@ -1,0 +1,51 @@
+import { SET_PROVIDER } from '../actions/provider'
+import { setError } from '../actions/error'
+import { FATAL_MISSING_PROVIDER } from '../errors'
+
+interface Action {
+  type: string
+  [key: string]: any
+}
+
+async function initializeProvider(
+  providerName: string,
+  providers: { [key: string]: any },
+  dispatch: any
+) {
+  const provider = providers[providerName]
+  if (!provider) {
+    // Once Unlock Provider is in place, it should be highly unusual for this error to occur.
+    dispatch(setError(FATAL_MISSING_PROVIDER))
+    return
+  }
+
+  try {
+    if (provider.enable) {
+      // this exists for metamask and other modern dapp wallets and must be
+      // called, see:
+      // https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8
+      // TODO: unlock provider specific code to open communication channels for prompts
+      await provider.enable()
+    }
+  } catch (err) {
+    // do a thing with err
+  }
+}
+
+const providerMiddleware = (config: any) => {
+  return ({ getState, dispatch }: { [key: string]: any }) => {
+    return function(next: any) {
+      return function(action: Action) {
+        if (action.type === SET_PROVIDER) {
+          // Only initialize the provider if we haven't already done so.
+          if (action.provider !== getState().provider) {
+            initializeProvider(action.provider, config.providers, dispatch)
+          }
+          next(action)
+        }
+      }
+    }
+  }
+}
+
+export default providerMiddleware
--- a/unlock-app/src/middlewares/providerMiddleware.ts
+++ b/unlock-app/src/middlewares/providerMiddleware.ts
@@ -7,18 +7,7 @@ interface Action {
   [key: string]: any
 }
 
-async function initializeProvider(
-  providerName: string,
-  providers: { [key: string]: any },
-  dispatch: any
-) {
-  const provider = providers[providerName]
-  if (!provider) {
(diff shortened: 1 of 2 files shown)


unlock-app/src/middlewares/providerMiddleware.ts
```diff
@@ -0,0 +1,51 @@
...
+import { setError } from '../actions/error'
+import { FATAL_MISSING_PROVIDER } from '../errors'
+
+interface Action {
+  type: string
+  [key: string]: any
+}
+
+async function initializeProvider(
+  providerName: string,
```
akeem: Is there a back story around us passing a string for a provider, then looking up in a dictionary of providers? As the code reads now it looks like we should be able to pass the provide to the initialization method directly

cnasc: We could do that, I think that would be a readability win.

cellog: the back story is that the provider in redux state must be serializable

cnasc: Right, but we can get the provider one level up and pass it directly into `initializeProvider` instead of passing a bunch of arguments to retrieve within the function

cnasc: Such as in the change I just pushed

cellog:
I like this a lot. One question is whether the config file is the best place for providers to live? Configuration should be pretty static. I definitely get the convenience of having it there, as a way to pass it to the middlewares. Definitely take into consideration the fact that this code will need to be shared with the paywall, and the configuration is not something we want to share.

Another possibility is that providers live in their own middleware, and dispatch themselves when they are constructed. The other middlewares then can simply listen for an action. No reducer would be necessary in this case, as the action would strictly be for communicating the provider to the middlewares. A side benefit is that in the future, if a hooks-based component wants to do something specific with the provider, it can also listen for the action and use that to retrieve the provider. I'm thinking of the scenario I will need to execute soon, where I need a provider that has no account to test that scenario in integration tests.

julien51:
LGTM ;)

cnasc:
> Another possibility is that providers live in their own middleware, and dispatch themselves when they are constructed. The other middlewares then can simply listen for an action.

I think that can be a goal for the future. This PR keeps us in line with the current handling of providers for the most part, and doesn't preclude us from eventually moving to another architecture, but it lets us get moving quickly with fairly minimal code change (and the changes to the services, for example, should be fully compatible with an alternative approach in the future).

cnasc:
Based on a test branch that integrated this into the app (not included here, changes will be made later), this approach seems to work nicely. One change had to be made (the `next(action)` call was inside the `if`, so it silently ditched all other actions), but this is now ready for a full review!

Required changes to activate this involve:

- mostly stripping the `connect` method of `walletService`
- removing most of the tests for the `connect` method (since it doesn't do those things anymore)
- importing and using the `providerMiddleware` in `_app.js`

Those will be included in a future PR.

unlock-app/src/middlewares/providerMiddleware.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+const providerMiddleware = (config: any) => {
+  return ({ getState, dispatch }: { [key: string]: any }) => {
+    return function(next: any) {
+      return function(action: Action) {
+        if (action.type === SET_PROVIDER) {
+          // Only initialize the provider if we haven't already done so.
+          if (action.provider !== getState().provider) {
```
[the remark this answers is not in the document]
cnasc: This is a good question, and I'm not 100% sure of all the cases when this situation might come up. Since the initial state for the provider is `null`, we know that it will do this the first time. I don't believe that the provider should ever change beyond that, barring use of the developer overlay.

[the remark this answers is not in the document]
julien51: Yeah I think this is just for the developer overlay

Issue #2351, opened 2019-03-26
Paywall cannot do optimistic unlocking when opened in a new window


cellog:
When the paywall opens in a new window, such as in a coinbase wallet purchase, we cannot properly handle optimistic unlocking. The reason is that if we do the optimistic unlock in the new window, there is no content to view. If we redirect back to the old window, then we don't have access to the transaction any longer, and the content is simply displayed as purchased.

This will require larger changes to the app, specifically enabling the same kind of ability to retrieve the transaction for a key purchase that is used on the creator dashboard to retrieve the transaction for lock creation.

cellog:
The solution is to pass the transaction hash instead of user account when redirecting. The user account can be retrieved from the transaction

PR #2490, opened 2019-04-03
VASTLY simplify the iframe css by using actual css classes

cellog:
# Description

This replaces all the fancy javascript for managing the iframe's position and size and animation with simple CSS. Also, because we are using fancy minimizing of CSS, the resulting `paywall.min.js` file is fully compressed and contains the CSS inline. In addition, because the .css file is external, it can be imported into the paywall app storybook stories to more accurately simulate how the iframe causes our components to look. win-win-win.

*Note: although this has the fixes in place for mobile view, the rest of the app is not fixed yet, so the screencast for that view will be in the subsequent PR.*

out

# Issues

Refs #2351

# Checklist:

### DIFF
--- a/paywall/src/__tests__/paywall-builder/iframe.test.js
+++ b/paywall/src/__tests__/paywall-builder/iframe.test.js
@@ -1,10 +1,4 @@
-import {
-  getIframe,
-  iframeStyles,
-  add,
-  show,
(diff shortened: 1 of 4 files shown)


cellog:
will re-open once this is rebased off of #2492

PR #2496, opened 2019-04-03
mapping the types from unlock-js

julien51:
Unlock.js has different transction types than the application.
This PR introduces a mapping for it in creator dashboard.

### DIFF
--- a/unlock-app/src/__tests__/utils/types.test.js
+++ b/unlock-app/src/__tests__/utils/types.test.js
@@ -1,0 +1,20 @@
+import { transactionTypeMapping } from '../../utils/types'
+import { TransactionType } from '../../unlockTypes'
+
+describe('transactionTypeMapping', () => {
+  it('should map the types from unlock-js', () => {
+    expect.assertions(4)
+    expect(transactionTypeMapping('LOCK_CREATION')).toEqual(
+      TransactionType.LOCK_CREATION
+    )
+    expect(transactionTypeMapping('KEY_PURCHASE')).toEqual(
+      TransactionType.KEY_PURCHASE
+    )
+    expect(transactionTypeMapping('WITHDRAWAL')).toEqual(
+      TransactionType.WITHDRAWAL
+    )
+    expect(transactionTypeMapping('UPDATE_KEY_PRICE')).toEqual(
+      TransactionType.UPDATE_KEY_PRICE
+    )
+  })
+})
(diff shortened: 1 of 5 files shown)


unlock-app/src/utils/types.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (type === 'LOCK_CREATION') {
+    return TransactionType.LOCK_CREATION
+  } else if (type === 'KEY_PURCHASE') {
+    return TransactionType.KEY_PURCHASE
+  } else if (type === 'WITHDRAWAL') {
+    return TransactionType.WITHDRAWAL
+  } else if (type === 'UPDATE_KEY_PRICE') {
+    return TransactionType.UPDATE_KEY_PRICE
+  } else {
+    // Not sure but that should not happen
```
cellog: let's throw here?

unlock-app/src/__tests__/utils/types.test.js
```diff
@@ -0,0 +1,20 @@
+import { transactionTypeMapping } from '../../utils/types'
+import { TransactionType } from '../../unlockTypes'
+
+describe('transactionTypeMapping', () => {
+  it('should map the types from unlock-js', () => {
+    expect.assertions(4)
+    expect(transactionTypeMapping('LOCK_CREATION')).toEqual(
```
cellog: mapping from raw strings makes me feel nervous, but as long as this is the only place they happen, and we always use this file, it should be fabulous :)

unlock-app/src/utils/types.ts
julien51: Welll, that's what I had initially but then remembered this is called inside the middleware... where we do not want to throw, right?

unlock-app/src/__tests__/utils/types.test.js
julien51: Yeah, I should have added a comment that we should actually load the values from the expored types in unlock-js and then test on all of them.

unlock-app/src/utils/types.ts
cellog: Maybe we can wrap in try/catch and dispatch setError, but we definitely want to catch this edge case somehow?

julien51: Since this should never happen I am actually in favor of keeping things simple here for now.

cellog: How about we console.log() here? If it does happen, we would have a devil of a time tracking the bug down otherwise :)

cellog: or better yet, can we use an enum for the parameter? maybe static typing will help us catch it?

PR #2509, opened 2019-04-04
pass withConfig in the correct location

### LINKED ISSUE #2498: withConfig should wrap the component actually using it
in the paywall, some uses of `withConfig` are not wrapping the immediate component that uses the configuration. For example, `ConfirmingFlag`, `Lock`, `Overlay`. This can result in the config not being propagated when the connected component is instead wrapped.

cellog:
# Description

This passes withConfig in the correct location to ensure it is directly available to the component using it, and fixes a crash when transaction is available prior to its key.

# Issues

Fixes #2498 
Refs #2351 

# Checklist:

### DIFF
--- a/paywall/src/components/lock/ConfirmingFlag.js
+++ b/paywall/src/components/lock/ConfirmingFlag.js
@@ -83,4 +83,4 @@ export const mapStateToProps = ({ account, keys, transactions }, { lock }) => {
   }
 }
 
-export default withConfig(connect(mapStateToProps)(ConfirmingFlag))
+export default connect(mapStateToProps)(withConfig(ConfirmingFlag))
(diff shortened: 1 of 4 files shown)


paywall/src/components/lock/ConfirmingFlag.js
```diff
@@ -83,4 +83,4 @@ export const mapStateToProps = ({ account, keys, transactions }, { lock }) => {
   }
 }
 
-export default withConfig(connect(mapStateToProps)(ConfirmingFlag))
+export default connect(mapStateToProps)(withConfig(ConfirmingFlag))
```
julien51: Wow, does the order actually matter?

cellog: I did not dig into why, but yes it mattered with the ConfirmingFlag component for some reason.

PR #2514, opened 2019-04-04
Lock name signature

### LINKED ISSUE #2322: Asking to sign a lock name when rejecting lock creation
When a user creates a Lock, they are being asked to confirm the transaction using their web3 provider (metamask..). If they click on "reject", they are still asked to sign a message to name the lock.

We should not ask users to sign messages when the lock is not going to be created.

julien51:
only save the lock name when it has been actually created. This fixes #2322

Fixes #2322


julien51:
> I think in general this looks good, but we do need to ensure that locks that havent been created do not end up in the redux store

Well they actually _are_ in the store. There is currently a bug I believe where we should actually remove them when they failed to be created,.

julien51:
Thanks!

PR #2516, opened 2019-04-04
refresh transaction from URL if needed, fix crash on key save if account is null

### LINKED ISSUE #2505: web3Middleware crashes on key.saved event if account is not set
In a race condition, if the `'key.saved'` event is emitted by the read-only web3Service before account has been set, the app crashes, because `'key.saved'` handler assumes the account is set.

### LINKED ISSUE #2506: web3Middleware does not properly retrieve transaction from URL
Currently, the `web3Middleware` only retrieves the transaction on route change, but since both `SET_PROVIDER` and `SET_NETWORK` actions will nuke transaction state, we need to retrieve it then as well. In addition, calling `web3Service.getTransaction` will trigger a race condition where transaction updates are sent prior to the transaction existing in redux, so instead, `ADD_TRANSACTION` should be dispatched and the `getTransaction` call will then happen in the correct order.

cellog:
# Description

If account changes, network changes, provider changes, or the actual URL changes, we need to refresh the transaction state.

In addition, if a `key.save` is emitted, there is a chance that the account is not yet set (race condition), so this removes a crash.

# Issues

Fixes #2505 
Fixes #2506 
Refs #2351

# Checklist:

### DIFF
--- a/paywall/src/middlewares/web3Middleware.js
+++ b/paywall/src/middlewares/web3Middleware.js
@@ -72,8 +72,9 @@ export default function web3Middleware({ getState, dispatch }) {
    * it might have been updated (balance, outstanding keys...)
    */
   web3Service.on('key.saved', (keyId, key) => {
+    const { account } = getState()
     web3Service.getLock(key.lock)
-    if (getState().account.address === key.owner) {
+    if (account && account.address === key.owner) {
       web3Service.refreshAccountBalance(getState().account)
     }
     web3Service.getKeyByLockForOwner(key.lock, key.owner)
@@ -116,10 +117,31 @@ export default function web3Middleware({ getState, dispatch }) {
 
       next(action)
 
+      const {
+        account,
+        router: {
+          location: { pathname, hash },
+        },
+        locks,
+      } = getState()
+      const { lockAddress, transaction } = lockRoute(pathname + hash)
+
+      if (
+        [SET_PROVIDER, SET_NETWORK, SET_ACCOUNT, LOCATION_CHANGE].includes(
+          action.type
+        )
+      ) {
+        if (transaction) {
+          dispatch(
+            addTransaction({
+              hash: transaction,
+            })
+          )
+        }
+      }
+
       if (action.type === SET_PROVIDER || action.type === SET_NETWORK) {
         // for both of these actions, the lock state is invalid, and must be refreshed.
-        // Location was changed, get the matching lock
-        const { lockAddress } = lockRoute(getState().router.location.pathname)
         if (lockAddress) {
           web3Service.getLock(lockAddress)
         }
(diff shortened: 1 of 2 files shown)


paywall/src/middlewares/web3Middleware.js
```diff
@@ -116,10 +117,31 @@ export default function web3Middleware({ getState, dispatch }) {
...
       next(action)
 
+      const {
+        account,
+        router: {
+          location: { pathname, hash },
+        },
+        locks,
+      } = getState()
+      const { lockAddress, transaction } = lockRoute(pathname + hash)
```
julien51: Is that true on all routes? What happens on `/`?

julien51: I guess it returns null for both of these?

paywall/src/middlewares/web3Middleware.js
```diff
@@ -134,41 +156,25 @@ export default function web3Middleware({ getState, dispatch }) {
...
 
-        const {
-          router: {
-            location: { pathname },
-          },
-        } = getState()
-
-        const { lockAddress } = lockRoute(pathname)
         if (lockAddress) {
           web3Service.getKeyByLockForOwner(lockAddress, action.account.address)
```
julien51: Combine this with line 146 above?

paywall/src/middlewares/web3Middleware.js
```diff
@@ -134,41 +156,25 @@ export default function web3Middleware({ getState, dispatch }) {
...
 
       if (action.type === ADD_LOCK || action.type == UPDATE_LOCK) {
-        const lock = getState().locks[action.address]
-        if (getState().account) {
-          web3Service.getKeyByLockForOwner(
-            lock.address,
-            getState().account.address
-          )
+        const lock = locks[action.address]
+        if (account) {
```
julien51: is account assigned here?

paywall/src/middlewares/web3Middleware.js
```diff
@@ -134,41 +156,25 @@ export default function web3Middleware({ getState, dispatch }) {
...
+          web3Service.getKeyByLockForOwner(lock.address, account.address)
         }
       } else if (
         action.type === LOCATION_CHANGE &&
         action.payload.location &&
         action.payload.location.pathname
       ) {
         // Location was changed, get the matching lock, if we are on a paywall page
```
julien51: Please update the comment ;)

paywall/src/middlewares/web3Middleware.js
cellog: yes

paywall/src/middlewares/web3Middleware.js
cellog: I tried, but these are actually quite different. 146 uses the URL lock, but this line uses the lock from the action, which may be completely different. I'll add a comment

paywall/src/middlewares/web3Middleware.js
cellog: yes

PR #2518, opened 2019-04-04
fully fix optimistic unlocking in coinbase wallet-like browsers

### LINKED ISSUE #2351: Paywall cannot do optimistic unlocking when opened in a new window

cellog:
# Description

This is the final piece of logic in the quest to enable optimistic unlocking in coinbase wallet-like browsers. It redirects to the main window when the transaction becomes pending, and keeps the paywall locked until the transaction is confirmed.

This also means that the paywall will work even on refresh!

# Issues

Fixes #2351 

# Checklist:

### DIFF
--- a/paywall/src/components/Paywall.js
+++ b/paywall/src/components/Paywall.js
@@ -19,10 +19,10 @@ import { isPositiveNumber } from '../utils/validators'
 import useWindow from '../hooks/browser/useWindow'
 import useOptimism from '../hooks/useOptimism'
 import { TRANSACTION_TYPES } from '../constants'
+import withConfig from '../utils/withConfig'
 
 // TODO: mobile formatting for unlocked and optimistic unlocking views
 export function Paywall({ locks, locked, redirect, account, transaction }) {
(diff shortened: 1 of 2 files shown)


paywall/src/components/Paywall.js
```diff
@@ -142,7 +144,22 @@ export const mapStateToProps = ({
...
+  }
+  if (!locked && !validKeys.length) {
+    locked = true
+  }
+  if (!locked && transaction) {
+    if (
+      !transaction.confirmations ||
+      transaction.confirmations < requiredConfirmations
+    ) {
+      locked = true
```
julien51: Good one!

PR #2519, opened 2019-04-04
enable mobile flags, both unlocked and optimistic

### LINKED ISSUE #2356: Optimistic Unlocking
Optimistic or Trusted Unlocking: At the time of transaction we quickly assess whether a transaction is likely to be confirmed in a reasonable amount of time. If it is likely enough the paywall will unlock the page and confirm the transactions in the background and show the status in the Unlock flag.

It should also be noted that the likeliness of the transaction going through should quickly decline with time. The reasoning being if the transaction is not confirmed quickly the "purchaser" has more time to "double spend" the funds which would make the transaction fail altogether.

More details in this [gdoc](https://docs.google.com/document/d/1oNna06wdbcfjyNW17ghDWHTQ8vop_jYaIhw4QUFlHRo/).

## Creator Flow

- Creators should have an option of how optimistic their unlocking should be. This will factor into the algorithm.
- OFF Not Optimist: don’t unlock until confirmed
- ON let the algorithm determine the best unlocking mechanism
- I’ve also called it Trusted Unlocking in the UI
- [Mockup](https://www.figma.com/file/gmhVq6Pqeu2gH7e3FC2lr0ax/Unlock-Dashboard?node-id=0%3A1457)
image

## User Flow
- User comes to a locked page
- User initiates payment, metamask opens
- User sends transaction, metamask closes
- Algorithm determines if its trusted-user, unsure-user or suspicious-user 
- trusted-user: page is immediately unlocks, while showing confirmation progress in  the unlock flag
- new-user: page unlocks after a certain time/count of confirmations or page unlocks only after transaction is confirmed
- [Mockup](https://www.figma.com/proto/LkWSSnCGlKEITBB5T5NYXW/Unlock-Paywall?node-id=0%3A817&viewport=1521%2C-1188%2C0.253176&scaling=min-zoom): Click Purchase then in MetaMask Trusted Purchase
image

### Relocking the Page
- If a trusted-users transaction is not confirmed (due to suspicious activity or confirmation taking too long) the page will lock again and the user will see informative messaging to what happened
- [Mockup](https://www.figma.com/proto/LkWSSnCGlKEITBB5T5NYXW/Unlock-Paywall?node-id=289%3A0&viewport=1521%2C-1188%2C0.253176&scaling=min-zoom): Select "Failed Transaction"
image

### LINKED ISSUE #2393: add optimism to the paywall UI

cellog:
# Description

This enables mobile formatting for the unlocked flag, locked flag and both optimistic and pessimistic unlocking

Screen Shot 2019-04-04 at 2 17 33 PM
Screen Shot 2019-04-04 at 2 17 30 PM
Screen Shot 2019-04-04 at 2 17 28 PM
Screen Shot 2019-04-04 at 2 17 24 PM
Screen Shot 2019-04-04 at 2 17 22 PM
Screen Shot 2019-04-04 at 2 17 17 PM
Screen Shot 2019-04-04 at 2 17 11 PM
Screen Shot 2019-04-04 at 2 17 07 PM
Screen Shot 2019-04-04 at 2 17 01 PM
Screen Shot 2019-04-04 at 2 16 56 PM
Screen Shot 2019-04-04 at 2 16 52 PM
Screen Shot 2019-04-04 at 2 16 48 PM
Screen Shot 2019-04-04 at 2 16 45 PM
Screen Shot 2019-04-04 at 2 16 41 PM
Screen Shot 2019-04-04 at 2 16 37 PM
Screen Shot 2019-04-04 at 2 16 27 PM
Screen Shot 2019-04-04 at 2 16 24 PM
Screen Shot 2019-04-04 at 2 16 20 PM
Screen Shot 2019-04-04 at 2 16 13 PM
Screen Shot 2019-04-04 at 2 16 07 PM

# Issues

Fixes #2393 
Fixes #2356 

# Checklist:

### DIFF
--- a/paywall/src/components/lock/UnlockFlag.js
+++ b/paywall/src/components/lock/UnlockFlag.js
@@ -10,7 +10,9 @@ export function LockedFlag() {
   return (
     <Colophon>
       <RoundedLogo size="28px" />
-      <p>Powered by Unlock</p>
+      <p>
+        Powered by&nbsp;<a href="https://paywall.unlock-protocol.com">Unlock</a>
+      </p>
     </Colophon>
   )
 }
@@ -26,7 +28,10 @@ export const UnlockedFlag = () => {
   return (
     <Flag id="UnlockFlag" hidden={hidden}>
       <RoundedLogo size="28px" />
-      <p>Subscribed with Unlock</p>
+      <p>
+        Subscribed with&nbsp;
+        <a href="https://paywall.unlock-protocol.com">Unlock</a>
+      </p>
     </Flag>
   )
 }
@@ -43,8 +48,34 @@ const Flag = styled(Colophon)`
     opacity: 1;
     transition: opacity 0.4s ease-in;
   }
+  & a {
+    color: var(--red);
+  }
 
   ${Media.phone`
     display: flex;
-`}
+    background-color: var(--white);
+    box-shadow: 14px 0px 40px rgba(0, 0, 0, 0.08);
+    grid-row: 2;
+    grid-column: 1;
+    width: 120px;
+    align-self: center;
+    justify-self: center;
+    height: 50px;
+    margin-right: -33px;
+    flex-direction: row;
+    float: none;
+    margin-right: auto;
+    margin-left: auto;
+  
+    & > div {
+      display: block;
+    }
+    & > p {
+      width: 63px;
+      display: block;
+      grid-row: 2;
+    }
+    opacity: 1;
+  `}
 `
--- a/paywall/src/components/lock/UnlockFlag.js
+++ b/paywall/src/components/lock/UnlockFlag.js
@@ -10,7 +10,7 @@ export function LockedFlag() {
   return (
     <Colophon>
       <p>Powered by</p>
-      <RoundedLogo size="28px" />
+      <RoundedLogo />
       <a href="/">Unlock</a>
     </Colophon>
   )
@@ -26,11 +26,9 @@ export const UnlockedFlag = () => {
   }, [setHidden])
   return (
     <Flag id="UnlockFlag" hidden={hidden}>
-      <RoundedLogo size="28px" />
-      <p>
-        Subscribed with&nbsp;
-        <a href="/">Unlock</a>
-      </p>
+      <RoundedLogo />
+      <p>Powered by</p>
+      <a href="/">Unlock</a>
     </Flag>
   )
 }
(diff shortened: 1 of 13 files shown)


julien51:
One thing that I note from the screenshot is that the paywall is not "blurry" at the top on mobile?

cellog:
> One thing that I note from the screenshot is that the paywall is not "blurry" at the top on mobile?

The blur happens in the static demo and the real app only, it is still there. This PR did not touch the code that generates the blur, which is in `paywall.min.js`. The image attached shows the real demo with blurring

Screen Shot 2019-04-04 at 6 11 56 PM

paywall/src/components/lock/UnlockFlag.js
```diff
@@ -10,7 +10,9 @@ export function LockedFlag() {
   return (
     <Colophon>
       <RoundedLogo size="28px" />
-      <p>Powered by Unlock</p>
+      <p>
+        Powered by&nbsp;<a href="https://paywall.unlock-protocol.com">Unlock</a>
```
julien51: I think we should use `/` here rather?

paywall/src/components/lock/UnlockFlag.js
```diff
@@ -26,7 +28,10 @@ export const UnlockedFlag = () => {
   return (
     <Flag id="UnlockFlag" hidden={hidden}>
       <RoundedLogo size="28px" />
-      <p>Subscribed with Unlock</p>
+      <p>
+        Subscribed with&nbsp;
+        <a href="https://paywall.unlock-protocol.com">Unlock</a>
```
julien51: Same as above? I worry this will break on local dev and/or staging and be hard to change when/if we change domains?

julien51:
I am surprised by the UI for mobile. Is that really what @smombartz  designed? I assumed the icon would be at the top of the box and that the box would be both wider (all screen wide maybe) and less tall?

paywall/src/components/lock/UnlockFlag.js
cellog: Definitely open to using config.paywallUrl here, does that make sense to you too?

paywall/src/components/lock/UnlockFlag.js
cellog: Oh you mean linking to `/` instead of a full url, that's not a bad idea.

cellog:
As for design, I had probably too much liberty, the figma I based it on has 5 very different designs, and I just chose a couple to use ;). 

How about this: if we can get this or something this-like in as a stopgap since mobile literally doesn't work atm (the flag is partially off the screen for optimistic unlocking and the unlocked flag), and then I can refine based on the new design? Hopefully before tomorrow at 4:30 to catch the Tuesday cutoff for productions deploy.

paywall/src/components/lock/UnlockFlag.js
julien51: Yes

paywall/src/components/lock/UnlockFlag.js
julien51: exactly ;)

cellog:
ok, because of the volume of screenshots, I'm going to open a new PR

PR #2543, opened 2019-04-08
Initial Key Chain authentication screen

cnasc:
# Description

As part of the user accounts epic, we need a way to get input from a user so that we can authenticate their account. This PR creates a form component to handle that process and adds it to the key chain page.

- The form is only visible when there is no account
- The key chain page is currently not accessible
- Submitting this form does nothing at the moment because we don't have the infrastructure necessary to actually log someone in.

I'll keep this as a draft for the moment, unless we think it's OK to merge it given the caveats above.

image

# Checklist:


julien51:
Could you write a couople stories for this please?

PR #2545, opened 2019-04-08
Adding signature verification in front of purchase endpoint

akeem:
# Description

Adding signature verification in front of purchase endpoint

# Issues

Fixes #
Refs #

# Checklist:


locksmith/__tests__/controllers/purchaseController.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+    primaryType: 'PurchaseRequest',
+    message: message,
+  }
+}
+
+describe('Purchase Controller', () => {
+  describe('purchase initiation', () => {
+    describe('when the purchase request is appropriately signed', () => {
+      it('returns a 401 status code', async () => {
```
cnasc: It reads like this returns `401 Unauthorized` when the key is signed the right way, and `202 Accepted` when the key is signed the wrong way. I feel like this must be a case of swapped descriptions?

akeem: good catch, i've updated this to reflect the feedback

Issue #2174, opened 2019-03-18
Run contract tests w/ and w/o ERC20 tokens


NickCuso:
**Is your feature request related to a problem? Please describe.**
The current ERC20 implementation selectively tests functions believed to be impacted by using ERC20 instead of the usual ETH pricing.  This approach assumes that we have selected the correct subset of tests.

**Describe the solution you'd like**
Switch the tests to use a behavior pattern (or similar) and have them run twice - once when priced in ETH and once when priced in ERC20.

**Describe alternatives you've considered**

**Additional context**

I think it may be best to have all tests run for both scenarios.  I think this will be easier on us when we add other scenarios in the future.  When a test needs ERC20 specific changes, there's an if condition in the test.

hardlydifficult:
Maybe using this approach: https://mochajs.org/#dynamically-generating-tests

hardlydifficult:
This is important for the transfer fee implementation as well.  #1997

julien51:
I actually want to do something like that at the `unlock-js` level: we deploy multiple lock "flavors" (version, currency... etc) and we make sure the "core" function:
- create lock
- withdraw
- change price
- purchase key
...
 behave as they should!

julien51:
I have implemented this inside of `unlock-js` in the integration tests for walletService. I'll close but feel free to re-open if applicable!

NickCuso:
I do still want to implement this.  Although maybe it's a low priority.

If we consider the test design it shouldn't be a large maintenance expense.  After the initial migration, we just need to get in the habit of using a slightly different pattern.

If we do encounter an issue it'd be best to catch it when introduced into the contract instead of after starting an app upgrade.

NickCuso:
We do have a place for this now, but still need to port a few more tests over to use it.

PR #2551, opened 2019-04-08
Require a small transfer fee

### LINKED ISSUE #1997: Charge a fee on transfer

hardlydifficult:
# Description

Requires the transfer fee is paid anytime a key is moved from one account to another.  

This implementation should work for both ETH and ERC20 tokens, however at the moment only ETH is tested.  We will follow up with testing ERC20 via #2174

# Issues

Fixes #1997
Refs #2174

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -3,6 +3,7 @@ pragma solidity 0.5.7;
 import './MixinDisableAndDestroy.sol';
 import './MixinApproval.sol';
 import './MixinKeys.sol';
+import './MixinFunds.sol';
 import './MixinLockCore.sol';
 import 'openzeppelin-solidity/contracts/utils/Address.sol';
 import 'openzeppelin-solidity/contracts/token/ERC721/IERC721Receiver.sol';
@@ -17,6 +18,7 @@ import 'openzeppelin-solidity/contracts/math/SafeMath.sol';
  */
 
 contract MixinTransfer is
+  MixinFunds,
   MixinLockCore,
   MixinKeys,
   MixinApproval
@@ -57,6 +59,7 @@ contract MixinTransfer is
     onlyKeyOwnerOrApproved(_tokenId)
   {
     require(_recipient != address(0), 'INVALID_ADDRESS');
+    _chargeAtLeast(getTransferFee(_from));
 
     Key storage fromKey = _getKeyFor(_from);
     Key storage toKey = _getKeyFor(_recipient);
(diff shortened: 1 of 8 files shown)


hardlydifficult:
This broke the gas report.  I opened #2556 to avoid this in the future.

hardlydifficult:
Pushed a fix for the gas report.

PR #2552, opened 2019-04-08
Pin all dependencies in smart-contracts

hardlydifficult:
# Description

Changing all dependencies in the smart-contracts directory to be pinned to a specific version.

Rational: we use greenkeeper and `npm ci`.

This combination causes us to miss updates.  With the previous configuration Greenkeeper will test every new update, and if it passes it then deletes the branch without any notification to us.  We are only alerted if there is a major version change or if the latest breaks something.  They do this because if you were to install with `npm i` then you would have the updated version without changing package.json.  However we use `npm ci` which uses the exact versions from the package-lock.json, meaning we don't use the latest and greatest.

By switching to this new approach, Greenkeeper will PR every individual update.  The potential downside of course is that this may be noisy. 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/package-lock.json
+++ b/smart-contracts/package-lock.json
@@ -5,17 +5,17 @@
   "requires": true,
   "dependencies": {
     "@babel/runtime": {
-      "version": "7.3.4",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.3.4.tgz",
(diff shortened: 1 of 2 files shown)


julien51:
Interesting. I am not sure I understand what you mean when writting " Greenkeeper will test every new update, and if it passes it then deletes the branch without any notification to us. " They don't update the package, do they? AFAIK they do not have commit rights on the master branch anyway?

hardlydifficult:
Greenkeeper will create a branch with the update, and then delete the branch.  So the master branch is never touched.

This change would cause greenkeeper to create a PR instead of ignoring the update.

See also "in-range" here https://greenkeeper.io/docs.html#greenkeeper-step-by-step

PR #2554, opened 2019-04-08
Remove ILockCore

### LINKED ISSUE #2180: Remove the Lock interface until there's a use-case for it
I have not been consistently updating the ILockCore.sol interface file.  At the moment, we only use this in Unlock.sol which does not depend on most of the features it includes.

Either:
 - Remove the file completely use the PublicLock type instead
 - Reduce the scope of this interface file (and maybe rename it) to limit it for use-cases Unlock.sol cares about

**Additional context**

hardlydifficult:
# Description

Remove the lock interface for now.  We can restore this once there is a use case.  See #2180 for more info.

# Issues

Fixes #2180
Refs #

# Checklist:


smart-contracts/contracts/interfaces/IUnlock.sol
```diff
@@ -32,7 +32,7 @@ interface IUnlock {
     uint _maxNumberOfKeys
   )
     external
-    returns (ILockCore lock);
```
nfurfaro: We may need to change this again in the future...the idea here was that `createLock` would always return an `IlockCore`, whether it's a PublicLock, PrivateLock, etc... But. this can be addressed when we have more lock types I guess.

Issue #2571, opened 2019-04-09
retrieve transaction hash from locksmith on the paywall to preserve information across page navigation/refreshes


cellog:
Currently, the paywall does not have a way of knowing the transaction if refreshed in most situations. We need a clear way of handling this. Locksmith can retrieve transactions, and should be used to retrieve active key purchase transactions for the current user and lock.

PR #2589, opened 2019-04-10
typescript configuration fixes for paywall/

cellog:
# Description

This PR sets up `paywall/` so that it will work with typescript files. It also pulls in the default `unlockTypes.ts` file from `unlock-app`. This is a prerequisite to the fix for #2571 which is written in typescript.

# Issues

Refs #2571

# Checklist:


paywall/src/unlockTypes.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  lock?: string
+  name?: string
+  key?: string // TODO: tighten up our types, hopefully we won't have too many
+  // optional properties.
+}
+
+// Analogous to the transactions in Redux
+export interface Transactions {
+  [hash: string]: Transaction
```
akeem: How strongly do we believe that this type declaration appropriately represents a collection of Transactions as they exist today?

cnasc: For the collection part, I think 100%. They're kept as a mapping from hashes to transactions. For the actual individual transactions, somewhat less confident. This definition basically started with what was always in Redux, and grew some optional properties because sometimes transactions get extra ones. There may be others that were missed.

PR #2621, opened 2019-04-12
DRY up integration test scripts, fix local development running of integration tests

cellog:
# Description

This PR DRYs up our integration test scripts. Environment variables needed for testing are extracted into a separate script. Using a bash trick, the env script simply echos the values needed, and then the scripts eval the text to set the environment variables in their local context. ([reference](https://stackoverflow.com/questions/16618071/can-i-export-a-variable-to-the-environment-from-a-bash-script-without-sourcing-i))

In addition, the local docker integration test script now exactly mirrors CI. It takes advantage of the shell source command to pass the environment variables to the docker build script.

(NOTE: the "source" command is unintuitively named `.`, only `bash` supports the more verbose `source` command, so the PR uses `.` in case we ever need to move away from bash. If reviewers prefer, changing it to `source` would be a trivial change).

# Issues

# Checklist:


scripts/docker-compose-build.sh
```diff
@@ -11,13 +11,13 @@ ARGS="--cache-from $DOCKER_REPOSITORY/unlock-core:master"
...
 
 # First build all of the images
-$REPO_ROOT/scripts/build-image.sh unlock-app true &
-$REPO_ROOT/scripts/build-image.sh wedlocks true &
-$REPO_ROOT/scripts/build-image.sh smart-contracts true &
-$REPO_ROOT/scripts/build-image.sh paywall true &
-$REPO_ROOT/scripts/build-image.sh locksmith true &
-$REPO_ROOT/scripts/build-image.sh unlock-protocol-com true &
-$REPO_ROOT/scripts/build-image.sh integration-tests true &
+. $REPO_ROOT/scripts/build-image.sh unlock-app true &
```
cellog: note: this script was not passing environment variables to the child scripts, which caused them to rebuild every single time they run rather than pulling from cache locally. I don't think this change will affect remote CI, but it's possible that it might speed things up, we'll see.

scripts/local-docker-integration-tests.sh
```diff
@@ -11,25 +11,29 @@ REPO_ROOT=`dirname "$0"`/..
...
+eval "$($REPO_ROOT/scripts/set-integration-tests-env-variables.sh)"
 
 # if the integration test images are running, this ensures we remove any state
 # prior to attempting to run the tests. This line is critical!
 docker-compose -f $DOCKER_COMPOSE_FILE down
 
 # re-build the images. This will use local docker cache
-docker build -t unlock -f "$REPO_ROOT/docker/unlock.dockerfile" $REPO_ROOT
-docker build -t unlock-integration -f "$REPO_ROOT/docker/integration-tests.dockerfile" $REPO_ROOT
+# and because we source the script, it will inherit our environment variables
```
cellog: bad news: this is the only method by which the 2nd build uses the cached builds. To make the docker-compose-build script work for local runs will require extensive re-writing. With the below script, the 2nd (and consequent) runs of integration tests takes only 45 seconds to rebuild

PR #2630, opened 2019-04-12
Adds endpoints supporting the creation and fetching of Events

akeem:
# Description

We are beginning to add backend support for the initial release
of the events application.

adds `POST - /events`
adds `GET - /events/lockAddress`

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/app.js
+++ b/locksmith/src/app.js
@@ -11,11 +11,18 @@ var transactionRouter = require('./routes/transaction')
 var lockRouter = require('./routes/lock')
 var blockRouter = require('./routes/block')
 var userRouter = require('./routes/user')
+var eventRouter = require('./routes/event')
 var purchaseRouter = require('./routes/purchase')
 
 let lockSignatureConfiguration = {
   name: 'lock',
   required: ['name', 'owner', 'address'],
+  signee: 'owner',
+}
+
+let eventConfiguration = {
+  name: 'event',
+  required: ['lockAddress', 'name', 'location', 'date', 'owner'],
   signee: 'owner',
 }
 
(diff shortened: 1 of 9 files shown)


locksmith/src/controllers/eventController.ts
```diff
@@ -0,0 +1,43 @@
...
+
+      return res.sendStatus(200)
+    } catch (e) {
+      return res.sendStatus(409)
+    }
+  }
+
+  export const find = async (req: Request, res: Response): Promise<any> => {
+    let lockAddress = req.params.lockAddress
+    let event = await Event.findOne({
```
julien51: Would you be ok to add an event Operations? I think this would be more consistent with the rest of locksmith?

locksmith/src/app.js
```diff
@@ -29,6 +36,12 @@ app.post(
   /^\/lock$/i,
   signatureValidationMiddleware.generateProcessor(lockSignatureConfiguration)
 )
+
+app.post(
+  /^\/events$/i,
```
julien51: Just want to make sure that we add a test to ensure that this middleware gets triggered when a event is created?

locksmith/src/controllers/eventController.ts
akeem: yeah, we can give it a go. I was a bit reluctant to do it this early as both of these methods are calling Sequelize methods without a ton of value add at this point

locksmith/src/app.js
akeem: I'll add a negative test case to indicate that it is being triggered, similar to what we currently have with other paths that require signatures. 

In another PR i'll address the regexes in the middleware to ensure that callers don't fall through with missing/invalid signatures

PR #2631, opened 2019-04-12
adding a date util which we will use for the date picker in tickets

julien51:
This will be useful to let users pick a date in the future for their event!

### DIFF
--- a/tickets/src/utils/dates.js
+++ b/tickets/src/utils/dates.js
@@ -1,0 +1,53 @@
+/**
+ * This function returns the future days, months and up to 3 years based on a current
+ * date.
+ * @param {*} now
+ * @param {*} year
+ * @param {*} month
+ */
+export const getDaysMonthsAndYears = (now, year, month) => {
+  if (!year) {
+    year = now.getFullYear()
+  }
+  if (!month) {
+    month = now.getMonth() + 1
+  }
+
+  // Years are the next 3 years
+  const years = [0, 1, 2].map(n => now.getFullYear() + n)
+
+  let months = Array.from(new Array(12), (val, index) => index + 1)
+
+  if (year === now.getFullYear()) {
+    months = months.slice(now.getMonth(), months.length)
+  }
+
+  let days = Array.from(new Array(31), (val, index) => index + 1)
+  if ([4, 6, 9, 11].indexOf(month) > -1) {
+    // only 30 days!
+    days.pop()
+  }
+  if (month === 2) {
+    // bissextile years?
+    if (year % 4 === 0) {
+      // Are we leap?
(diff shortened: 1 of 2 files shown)


julien51:
I looked at what moment.js does and I could not see a feature which would let us build a dropdown of dates based on various months/days or years :(

cnasc:
Yeah, moment is great for date normalization and operations on a single date (like "5 days from now"), but it doesn't quite do this.

tickets/src/utils/dates.js
```diff
@@ -0,0 +1,59 @@
...
+    'Jul',
+    'Aug',
+    'Sep',
+    'Oct',
+    'Nov',
+    'Dec',
+  ]
+
+  // Years are the next 3 years
+  const years = [
```
cnasc: If you wanted to compress this a bit, the following should be equivalent. Just an option
```javascript
const years = [0, 1, 2].map(n => (now.getFullYear() + n).toString
// => [ '2019', '2020', '2021' ]
```

tickets/src/utils/dates.js
```diff
@@ -0,0 +1,59 @@
...
+    now.getFullYear(),
+    now.getFullYear() + 1,
+    now.getFullYear() + 2,
+  ].map(d => d.toString())
+
+  let months = [...monthNames]
+
+  if (year === `${now.getFullYear()}`) {
```
cnasc: This might be more explicit that we are considering the first year of our range?
```suggestion
  if (year === years[0]) {
```

tickets/src/utils/dates.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  let days = Array.from(new Array(31), (val, index) => (index + 1).toString())
+  if (['Apr', 'Jun', 'Sep', 'Nov'].indexOf(month) > -1) {
+    // only 30 days!
+    days.pop()
+  }
+  if (month === 'Feb') {
+    // bissextile years?
+    if (year % 4 === 0) {
```
cnasc: This isn't quite correct, leap years aren't only the ones divisible by 4.

cellog: In addition, every 100 years is not a leap year (although we will not live to see the next one probably)

julien51: Ha! Yeah I assumed this would be good enough but I guess I'll appluy this:

```if (year is not divisible by 4) then (it is a common year)
else if (year is not divisible by 100) then (it is a leap year)
else if (year is not divisible by 400) then (it is a common year)
else (it is a leap year) ```

julien51:
Ok, all fixed, PTAL!

julien51:
cc @cnasc

PR #2639, opened 2019-04-15
introducing a unlockService parent class which extracts shared functionnality

julien51:
All in the title!
Idnetifing the version is shared between both services...


cnasc:
Looks good to me! I was going to suggest doing something to prevent the superclass from being initialized, but I don't think there's any good way to do that in JS.

julien51:
Yeah, that would be possible in TS though so maybe I'll swicth to typescript at some point...

Issue #2111, opened 2019-03-15
User Accounts


julien51:
We want to let any user unlock the content they care about, whether they are crypto users (and have a crypto wallet) or whether they do not know or care about cryptocurrencies.

This means that we have to implement a way for these users to "host" their ethereum accounts with Unlock.

PR #2669, opened 2019-04-16
Add component for finishing signup process

cnasc:
# Description

Another part of #2111, this PR introduces a component that completes the sign up process (after the user has already confirmed their email). It takes a password and a confirmation, then dispatches the create account action.

Note: there is at present no form validation here. My plan is to get the minimum viable user account stuff in place so that we can run it through its paces before enhancing with additional features, but I'm fine if we want to put that in now before merging this.

image

# Checklist:


julien51:
> My plan is to get the minimum viable user account stuff in place so that we can run it through its paces before enhancing with additional features, but I'm fine if we want to put that in now before merging this.

Ok to do this later... BUT please let;s keep track of it!

PR #2670, opened 2019-04-16
add functionality to unlock-js to remove all web3 from the other apps

cellog:
# Description

This PR adds a few new exports to `unlock-js` that can replace all uses of `web3` inside the `unlock-app`, `unlock-protocol-com` app, `paywall` app, and `tickets` app.

Most of the features are within `config.js`, which instantiates an HTTP provider. For that, we now export `getWeb3Provider` and `getCurrentProvider` (the latter is unused, but could be potentially?).

For the usage in `storageMiddleware` to create a new account, we export `createAccountAndPasswordEncryptKey`, which accepts a password and returns the new address and encrypted private key.

For the `deploy-unlock.js` script, we provide a new async function `deploy` which accepts a host and port for the running ethereum node, and a contract abi package name, defaulting to `unlock-abi-0`. This also accepts a callback which is called with the new contract (used to console.error the new address out in the deploy script), and then returns the result of the deploy initialization transaction.

This reduces the deploy script to:

```js
const net = require('net')
const { deploy } = require('@unlock-protocol/unlock-js')


/*
 * This script is meant to be used in dev environment to deploy a version of the Unlock smart
 * contract from the packaged version to the local ganache server.
 */

const host = process.env.HTTP_PROVIDER || '127.0.0.1'
const port = 8545
const serverIsUp = (delay, maxAttempts) =>
  new Promise((resolve, reject) => {
    let attempts = 1
    const tryConnecting = () => {
      const socket = net.connect(port, host, () => {
        resolve()
        return socket.end() // clean-up
      })

      socket.on('error', error => {
        if (error.code === 'ECONNREFUSED') {
          if (attempts < maxAttempts) {
            attempts += 1
            return setTimeout(tryConnecting, delay)
          }
          return reject(error)
        }
        return reject(error)
      })
    }
    tryConnecting()
  })
serverIsUp(1000 /* every second */, 120 /* up to 2 minutes */)
  .then(() => {
    return deploy(host, port, 'unlock-abi-0', newContractInstance => console.log(newContractInstance.options.address))
  })
  .catch(error => {
    console.error(error)
    process.exit(1)
  })
  
```

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-js/index.js
+++ b/unlock-js/index.js
@@ -1,8 +1,8 @@
 const Web3Service = require('./lib/web3Service')
 const WalletService = require('./lib/walletService')
-const { createAccountAndPasswordEncryptKey } = require('./lib/accounts')
+const createAccountAndPasswordEncryptKey = require('./lib/accounts').default
 const { getCurrentProvider, getWeb3Provider } = require('./lib/providers')
-const { deploy } = require('./lib/deploy')
+const deploy = require('./lib/deploy').default
 
 module.exports = {
   Web3Service: Web3Service.default,
--- a/unlock-js/index.js
+++ b/unlock-js/index.js
@@ -1,7 +1,14 @@
 const Web3Service = require('./lib/web3Service')
 const WalletService = require('./lib/walletService')
+const { createAccountAndPasswordEncryptKey } = require('./lib/accounts')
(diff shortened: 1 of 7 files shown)


julien51:
Great work on this @cellog

PR #2679, opened 2019-04-16
Production Automated Deploy production-20190416-203034

unlock-deployer:


julien51:
Hum, how can we tell @CLAassistant to go away on production builds?

julien51:
I have disabled the required check to approve this but I think we really need a way to not require `license/cla` here.

PR #2680, opened 2019-04-16
Making our unlock-js module compatible with our newest version of the smart contracts

julien51:
This was less work/issues than what I expected...
But I think I need to refactor how versions are detected.
1. For Unlock's smart contract, the current version uses opCode which would only yield the address of the proxy... and not the address of the actual implementation...
2. For Locks, we need something custom, and not assume their version based on the version of Unlock being deployed.

These are for the next 2 PRs


julien51:
Reviewing by looking at the individual commits will likely be much easier because the bulk of changed files is just renaming of variables...

unlock-js/src/v01/getLock.js
```diff
@@ -48,6 +49,9 @@ export default function(address) {
 
   // Once all lock attributes have been fetched
   return Promise.all(constantPromises).then(() => {
+    // numberOfOwners was previously called outstandingKeys. In order to keep compatibility
+    // we also assign it. This behavior will eventually be deprecated
+    update.outstandingKeys = update.numberOfOwners
```
hardlydifficult: We renamed outstanding keys to `totalSupply` to follow the ERC721 spec.  `numberOfOwners` may be slightly different

julien51: Ho, that is a good point. I'll update that! Thanks.

julien51: Done ;)

PR #2684, opened 2019-04-16
Content component for RSVP pages

benwerd:
# Description

Introduces the content component for event RSVP pages.

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [x] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [x] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] If my code adds or changes components, I have written corresponding stories with Storybook
- [x] I have performed a self-review of my own code
- [x] If my code involves visual changes, I am adding applicable screenshots to this thread

Screen Shot 2019-04-16 at 4 01 40 PM
Screen Shot 2019-04-16 at 4 01 06 PM

PR #2690, opened 2019-04-17
Revert "Content component for RSVP pages"

benwerd:
Reverts unlock-protocol/unlock#2684

Tests were passing - but I'll figure it out.


julien51:
This is not the problem... we can probably close this one?

Issue #2725, opened 2019-04-18
Paywall does not allow purchasing a new key when the old one expires


cellog:
This is a critical problem, highest priority. fix is forthcoming

PR #2727, opened 2019-04-18
attach KEY_PURCHASE transactions to the keys they represent in redux

cellog:
# Description

This adds the missing information needed to determine the status of a key directly to the key object in the redux store. This will make it possible to distinguish between expired keys and a pending key purchase transaction, which is currently very difficult to do accurately.

# Issues

Refs #2725 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/middlewares/web3Middleware.test.js
+++ b/paywall/src/__tests__/middlewares/web3Middleware.test.js
@@ -2,7 +2,7 @@ import EventEmitter from 'events'
 import { LOCATION_CHANGE } from 'connected-react-router'
 import web3Middleware from '../../middlewares/web3Middleware'
 import { ADD_LOCK, UPDATE_LOCK } from '../../actions/lock'
-import { UPDATE_KEY } from '../../actions/key'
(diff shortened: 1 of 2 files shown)


paywall/src/__tests__/middlewares/web3Middleware.test.js
```diff
@@ -287,14 +301,26 @@ describe('Web3 middleware', () => {
...
+      foo: 'bar',
+    }
     mockWeb3Service.emit('transaction.updated', transaction.hash, update)
     expect(store.dispatch).toHaveBeenCalledWith(
       expect.objectContaining({
         type: UPDATE_TRANSACTION,
         hash: transaction.hash,
-        update,
+        update: {
+          ...transaction,
```
julien51: Shouldn't the logic of combining the previous transaction with the new update actually be in the reducer? This would be a lot cleaner and less error prone?

cellog: I'll open an issue to refactor this. We will need some new actions to do this cleanly

cellog: #2737

PR #2743, opened 2019-04-18
Introducing typescript version of Locksmith's models

akeem:
# Description

Recently we have been running into issues where part of the Locksmith code base
has become a pain to work with in Typescript files. Specifically when using
libraries involved in the persistence of data.

The adjustment here will ideally eliminate this without introducing many side effects.

Please raise any questions that you may have about this, id like to ensure that we are on the same page before this lands into production.

# Issues

Fixes #
Refs #

# Checklist:


locksmith/tsconfig.json
```diff
@@ -56,8 +56,8 @@
     // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
 
     /* Experimental Options */
-    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
-    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+    "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
```
cnasc: Just some context around this, decorators are currently a Stage 2 draft: https://tc39.github.io/proposal-decorators/

That is, they're not quite standard but other than minor changes, they almost certainly will be. `tsc` compiles them into normal JS, so compatibility should not be a concern.

locksmith/src/models/authorizedLock.ts
```diff
@@ -0,0 +1,11 @@
+import { Table, Column, Model } from 'sequelize-typescript'
+
+@Table({ tableName: 'AuthorizedLocks', timestamps: true })
+// eslint-disable-next-line import/prefer-default-export
+export class AuthorizedLock extends Model<AuthorizedLock> {
+  @Column
+  address!: string
```
cnasc: TS idiom for the uninitiated: the `!` is an assertion that a variable will not be null. The compiler can't always statically determine whether something will be defined or not, so this is us saying that we are 100% confident it will not be null.

https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator

PR #2755, opened 2019-04-19
use gas price for contract deploy from constants.js

### LINKED ISSUE #2754: contract deploy uses more gas than we allow in unlock-js/deploy
as seen in this dump from `npm run dev` inside `unlock-app`, we use too much gas (more than 300000):

```

  Transaction: 0x280e65631d5bd777f3260945f4469acc82cf4012032388690a92c4f24ebdc11e
  Gas usage: 3053213
  Block Number: 21
  Block Time: Thu Apr 18 2019 23:13:15 GMT-0400 (EDT)

```

To fix, I propose allowing passing in the gas used for contract creation, so we don't have to release a new version of unlock-js just to increase the gas cost

cellog:
# Description

This draws the gas price for deploy from the `constants.js` file, and adds a new `deployContract` field in `GAS_AMOUNTS` which must be kept up-to-date

# Issues

Fixes #2754 

# Checklist:

### DIFF
--- a/unlock-js/src/constants.js
+++ b/unlock-js/src/constants.js
@@ -4,7 +4,7 @@ export const GAS_AMOUNTS = {
   purchaseKey: 300000,
   withdrawFromLock: 1000000,
   partialWithdrawFromLock: 1000000,
-  deployContract: 4500000,
+  deployContract: 6000000,
 }
 
 export const MAX_UINT =
--- a/unlock-js/src/constants.js
+++ b/unlock-js/src/constants.js
@@ -4,6 +4,7 @@ export const GAS_AMOUNTS = {
   purchaseKey: 300000,
   withdrawFromLock: 1000000,
   partialWithdrawFromLock: 1000000,
+  deployContract: 4500000,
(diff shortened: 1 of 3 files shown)


cellog:
CC @hardlydifficult to verify the idea of adding a new field to `GAS_AMOUNTS` is kosher

unlock-js/src/constants.js
```diff
@@ -4,6 +4,7 @@ export const GAS_AMOUNTS = {
   purchaseKey: 300000,
   withdrawFromLock: 1000000,
   partialWithdrawFromLock: 1000000,
+  deployContract: 4500000,
```
hardlydifficult: Ah so it was deploy you were talking about.  I think this limit is aggressive and likely to break again in the future.  I'd recommend maybe 6mil instead, which is close to the network max of approx 6.4mil.  This is not a call we wish to optimize in any way, so no need to set an aggressive max.

cellog: great thank you

PR #2771, opened 2019-04-19
refactor paywall Overlay: remove redundant mapStateToProps

cellog:
# Description

This PR refactors `Overlay` to remove its `mapStateToProps`. As part of the refactor, I caught a small section that still used `transaction` instead of `keyStatus` and fixed that. Also, the `'submitted'` key status was not being handled, so that is fixed.

`requiredConfirmations` is pulled in by `withConfig` to Overlay, so redundant passing of the value to `Overlay` from `Paywall` is removed.

There are no changes to any storybook rendered stories.

Note that `Paywall` is already passing down `account` and `expired` and `transaction` and `lockKey`, which is why behavior is seamlessly the same.

# Issues

Refs #2725 

# Checklist:

### DIFF
--- a/paywall/src/stories/lock/Overlay-Optimistic.stories.js
+++ b/paywall/src/stories/lock/Overlay-Optimistic.stories.js
@@ -67,6 +67,15 @@ const defaultState = {
       id: '0x1234567890123456789012345678901234567890-0x456',
       lock: '0x1234567890123456789012345678901234567890',
       owner: '0x456',
+      transactions: {
+        transaction: {
+          id: 'transaction',
+          status: 'pending',
+          confirmations: 0,
+          type: TRANSACTION_TYPES.KEY_PURCHASE,
+          key: '0x1234567890123456789012345678901234567890-0x456',
+        },
+      },
     },
   },
 }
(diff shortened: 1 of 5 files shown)


paywall/src/stories/lock/Overlay-Optimistic.stories.js
```diff
@@ -154,7 +173,7 @@ const render = (
 storiesOf('Overlay/Optimistic Unlocking', module)
   .add('beginning purchase', () => {
     makeStore()
-    return render(locks, 'pending', undefined, undefined, {
+    return render(locks, 'pending', transaction, undefined, undefined, {
```
julien51: nit, when you do this `undefined`, could you name the param so we know what this expects?
```suggestion
    return render(locks, 'pending', transaction, undefined /* errors */, undefined /* thisConfig*/, {
```

For example here, I was expecting to see `optimism` passed in too but that would be 6 args and we seem to expect 5?

cellog: 👍it does expect 6 args, the 6th is optimism. I implemented your suggestion throughout the file

Issue #2782, opened 2019-04-20
make unlock-js serviceWorker-compatible


cellog:
This will either involve upgrading web3 to beta38 or higher, or using ethers or web3x

PR #2791, opened 2019-04-21
remove all mocking of web3.eth from walletService test

cellog:
# Description

This test removes meddling with the internal details of `web3.eth` using `jest.fn()` from `walletService.test.js`. However, it leaves in place a mock provider, which will need to be updated with all upgrades.

# Issues

Refs #2782

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/walletService.test.js
+++ b/unlock-js/src/__tests__/walletService.test.js
@@ -1,6 +1,5 @@
 /* eslint no-console: 0 */
 import Web3 from 'web3'
-import EventEmitter from 'events'
 import NockHelper from './helpers/nockHelper'
 
 import WalletService from '../walletService'
@@ -78,11 +77,10 @@ describe('WalletService', () => {
       it('should yield an error if we could not retrieve the opCode', done => {
         expect.assertions(2)
         const err = new Error('getCode failed')
-        walletService.web3.eth.getCode = jest.fn(() => {
-          throw err
-        })
+
+        nock.ethGetCodeAndYield(unlockAddress, '0x', err)
         walletService.isUnlockContractDeployed((error, isDeployed) => {
-          expect(error).toBe(err)
+          expect(error).toBeInstanceOf(Error)
           expect(isDeployed).toBe(undefined)
           done()
         })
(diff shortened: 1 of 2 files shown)


unlock-js/src/__tests__/walletService.test.js
```diff
@@ -119,13 +117,7 @@ describe('WalletService', () => {
...
           const newAccount = '0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1'
 
-          walletService.web3.eth.accounts.create = jest.fn(() => {
-            return Promise.resolve({
-              address: newAccount,
-            })
-          })
-
-          nock.accountsAndYield([])
+          nock.accountsAndYield([newAccount])
```
julien51: This is not right. The test actually verifies that when no account is available on the node, one is created by web3... So we cannot have a mock actually return an account.

unlock-js/src/__tests__/walletService.test.js
```diff
@@ -152,110 +144,127 @@ describe('WalletService', () => {
...
+            gasUsed: '0x16e360',
+            cumulativeGasUsed: '0x16e360',
+            logs: [],
+          })
+        })
 
-      it('should handle cases where the transaction is sent via a provider', () => {
-        expect.assertions(1)
+        it('should handle cases where the transaction is sent via a provider', async () => {
+          expect.assertions(0)
```
julien51: hu, wat? This is probabluy not a good test if there are no assertions ;p

unlock-js/src/__tests__/walletService.test.js
cellog: ah that makes sense, I'll add comments for the next victim so we don't forget :)

unlock-js/src/__tests__/walletService.test.js
cellog: I'll read up on nock, see if we can get something to assert against from it

cellog:
ok @julien51 PTAL. The solution I came up with was to make a test where the nock calls are explicit, and that we then have a helper that verifies we have used all nocks, and not had any mis-matched nocks in the event that a test does not fail but the nocks mismatch. I also fixed that test so that it calls the real account creation method, and verifies that we got an account hash back.

unlock-js/src/__tests__/walletService.test.js
```diff
@@ -152,110 +146,155 @@ describe('WalletService', () => {
...
-            expect(input).toEqual(data)
-            expect(type).toEqual('type')
-            expect(status).toEqual('submitted')
-            done()
-          }
-        )
+        // in the place of assertions, we explicitly verify that all of the JSON-RPC
+        // calls were completed in the exact order specified above, and that all
+        // of them were called, and nothing else
+        nock.ensureAllNocksUsed()
```
julien51: Works for me.

PR #2792, opened 2019-04-21
refactor deploy test to remove web3-specific code

cellog:
# Description

This removes usage of `web3` inside the deploy test, and instead relies upon the correct JSON-RPC calls being made to the node, which will allow us to verify deployment after swapping implementations without changing the test.

# Issues

Refs #2782

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/deploy.test.js
+++ b/unlock-js/src/__tests__/deploy.test.js
@@ -1,89 +1,15 @@
-import nock from 'nock'
-import Web3 from 'web3'
 import { Unlock } from 'unlock-abi-0'
 
 import deploy from '../deploy'
 import { GAS_AMOUNTS } from '../constants'
+import NockHelper from './helpers/nockHelper'
 
 const host = '127.0.0.1'
 const port = 8545
-const readOnlyProvider = `http://${host}:${port}`
 
-const nockScope = nock(readOnlyProvider, { encodedQueryParams: true })
-
-let rpcRequestId = 0
-
-let debug = false // set to true to see more logging statements
-
-function logNock(...args) {
-  if (debug) {
-    /* eslint-disable-next-line */
-    console.log(...args)
-  }
-}
-
-// Generic call
-const jsonRpcRequest = (method, params, result, error) => {
-  rpcRequestId += 1
-  nockScope
-    .post('/', { jsonrpc: '2.0', id: rpcRequestId, method, params })
-    .reply(200, { id: rpcRequestId, jsonrpc: '2.0', result, error })
-    .log(logNock)
-}
-
-// eth_accounts
-const accountsAndYield = accounts => {
-  return jsonRpcRequest('eth_accounts', [], accounts)
-}
-
-// eth_gasPrice
-const ethGasPriceAndYield = () => {
-  return jsonRpcRequest('eth_gasPrice', [], 8000000)
-}
-
-// eth_sendTransaction
-const sendTransactionAndYield = (
-  from,
-  result,
-  data = Unlock.bytecode,
-  to = false,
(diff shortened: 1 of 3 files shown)


unlock-js/src/__tests__/deploy.test.js
```diff
@@ -114,29 +40,46 @@ describe('contract deployer', () => {
...
+      Unlock.bytecode
+    )
+    nock.ethGasPriceAndYield(gasPrice)
+    // initialize call
+    nock.ethSendTransactionAndYield(
+      {
+        to: '0xbbbdeed4c0b861cb36f4ce006a9c90ba2e43fdc2', // contract address
+        from: unlockAccountsOnNode[0],
+        data:
+          '0xc4d66de8000000000000000000000000aaadeed4c0b861cb36f4ce006a9c90ba2e43fdc2', // data (no idea what this is, but there it is)
```
julien51: I think we need to clarify what this is otherwise we would have no idea why this may fail?

julien51: I understand this is the behavior previously but if we're refactoring, we might as well clean things up and remove some tech debt.

unlock-js/src/__tests__/deploy.test.js
```diff
@@ -114,29 +40,46 @@ describe('contract deployer', () => {
...
+      {
+        from: unlockAccountsOnNode[0],
+        data: Unlock.bytecode,
+        gas: '0x' + GAS_AMOUNTS.deployContract.toString(16),
+      },
+      gasPrice,
+      transaction.hash
+    )
+    nock.ethGetTransactionReceipt(transaction.hash, transactionReceipt)
+    // get the contract bytecode
```
julien51: One test we need to do is see how the web3Service behaves when this yields `0x` (it should throw) This is a key test because if the contract is not deployed, nothing will succeed after that so we would want to know about failures.

unlock-js/src/__tests__/deploy.test.js
```diff
@@ -114,29 +40,46 @@ describe('contract deployer', () => {
...
   })
-  it('retrieves accounts', async () => {
-    expect.assertions(1)
-    const web3 = new Web3(`http://${host}:${port}`)
-    const getAccounts = web3.eth.getAccounts
-    web3.eth.getAccounts = jest.fn(() => getAccounts())
 
-    await deploy(host, port, Unlock, () => {}, web3)
-    expect(web3.eth.getAccounts).toHaveBeenCalled()
+  it('does not throw exception (json-rpc calls are accurate)', async () => {
```
julien51: I am not sure it is great that we only have a single test left which is not really a test because everything happens in the `before` block.

We need to decompose what is actually happening...

unlock-js/src/__tests__/deploy.test.js
julien51: So I think we'd first need to understand which call in the deploy script makes this specific request. Then we can probably have a better way of understanding what it is and "rebuild" it from there?

cellog:
needs to be rebased on #2808, but PTAL @julien51 at the test. Also found a naming "bug" and corrected that, it will show up clearer in the rebase, but the name was `ethGetGasPriceAndYield` instead of just `ethGasPriceAndYield`

cellog:
@julien51 added a failure test as well, this addresses all feedback, PTAL

cellog:
ack, hold off a sec, sorry

cellog:
OK! @julien51 all set, PTAL

unlock-js/src/__tests__/deploy.test.js
cellog: 👍added new test

PR #2793, opened 2019-04-22
Introducing the Event Page

cnasc:
# Description

Since Ben is asleep, I figured I may as well put this up as a draft so we can do some review/comments on this. Of course, I don't know what he would have written here, but he can always come and backfill.

**In Progress: Splitting this into more manageable chunks**
# Issues

Fixes #
Refs #

# Checklist:

- [ ] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [ ] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [ ] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #2796, opened 2019-04-22
Update tickets constants from paywall

cnasc:
# Description

Part of the #2793 split, this PR migrates some constants over from the paywall. Some are already in use, some will be used to further update other utils in the app in a future PR.

# Checklist:


cnasc:
> On its face this looks fine, but I am not clear on exactly where this will be utilized. I think there is high value in using one of the libraries available to verify ethereum addresses. There may be functionality available to do the same for transaction hashes as well. https://github.com/ethereumjs/ethereumjs-util/blob/master/docs/README.md#isvalidaddress

I believe in this case it's convenient to have it as a regex so that we can pull these values out of the URL. So in this case, we're not really validating an ethereum address, we're just looking for something that looks like an ethereum address somewhere in the URL.

Issue #2721, opened 2019-04-18
Be able to log in to an account


cnasc:
Currently it is possible (in development) to create an account. However, there is no way to log in to an account, which makes the account dubiously useful at best.

- [x] Page to log in from
- [x] Link to that page/state from the sign up page, which is the default if there is no account
- [x] Use storage service to authenticate and put the account into state

PR #2804, opened 2019-04-22
Add getAddressFromPrivateKey helper

cnasc:
# Description

For dealing with user accounts, it will help us to have a consistent method for decrypting private keys. This method does that and just returns the address as a string.

# Issues
Refs #2721 

# Checklist:

### DIFF
--- a/unlock-js/src/accounts.js
+++ b/unlock-js/src/accounts.js
@@ -1,6 +1,6 @@
 const Web3 = require('web3')
 
-export default function createAccountAndPasswordEncryptKey(password) {
+export function createAccountAndPasswordEncryptKey(password) {
   const web3 = new Web3()
   const { address, privateKey } = web3.eth.accounts.create()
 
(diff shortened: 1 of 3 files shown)


unlock-js/src/accounts.js
```diff
@@ -13,3 +13,18 @@ export default function createAccountAndPasswordEncryptKey(password) {
...
+ * Given an encrypted private key and a password, decrypts the private key and
+ * gets the public address from it.
+ * @param {*} encryptedPrivateKey
+ * @param {string} password
+ * @throws Throws an error if password does not decrypt private key.
+ */
+export function getAddressFromPrivateKey(encryptedPrivateKey, password) {
+  const web3 = new Web3()
+
+  const { address } = web3.eth.accounts.decrypt(encryptedPrivateKey, password)
```
benwerd: Q: what happens if address isn't present for whatever reason? Should we throw something, or are we okay with leaving it undefined?

cnasc: I was curious about that too, and I verified that web3 will throw an error along the lines of `Key derivation failed - possibly wrong password`. I _believe_ that should cover it, barring bugs in web3. In the end, if account is `undefined` that should halt everything since if no account gets into the state they won't be able to do anything.

Would definitely like to hear the perspective of others on this as well.

PR #2817, opened 2019-04-23
Ability to update payment details

akeem:
# Description

A controller method allowing a user to update their payment details and associated bits. While adding this functionality into the application, we are not mounting this into this main router in this pull request

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/config/config.js
+++ b/locksmith/config/config.js
@@ -5,7 +5,7 @@ module.exports = {
     dialect: 'sqlite',
     storage: 'development.sqlite3',
     operatorsAliases: false,
-    stripe_secret: process.env.STRIPE_SECRET,
+    stripeSecret: process.env.STRIPE_SECRET,
   },
   test: {
     username: process.env.DB_USERNAME,
@@ -16,7 +16,7 @@ module.exports = {
     logging: false,
     dialect: 'postgres',
     operatorsAliases: false,
-    stripe_secret: process.env.STRIPE_SECRET,
+    stripeSecret: process.env.STRIPE_SECRET,
   },
   production: {
     username: process.env.DB_USERNAME,
@@ -24,7 +24,7 @@ module.exports = {
     database: process.env.DB_NAME,
     port: process.env.DB_PORT || 5432,
     host: process.env.DB_HOSTNAME,
-    stripe_secret: process.env.STRIPE_SECRET,
+    stripeSecret: process.env.STRIPE_SECRET,
     dialect: 'postgres',
     operatorsAliases: false,
   },
--- a/locksmith/config/config.js
+++ b/locksmith/config/config.js
@@ -5,6 +5,7 @@ module.exports = {
     dialect: 'sqlite',
     storage: 'development.sqlite3',
     operatorsAliases: false,
+    stripe_secret: process.env.STRIPE_SECRET,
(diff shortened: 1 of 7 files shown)


locksmith/config/config.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
     dialect: 'sqlite',
     storage: 'development.sqlite3',
     operatorsAliases: false,
+    stripe_secret: process.env.STRIPE_SECRET,
```
julien51: Let's stick to the camel case config? `stripeSecret` ;)

locksmith/config/config.js
```diff
@@ -15,13 +16,15 @@ module.exports = {
     logging: false,
     dialect: 'postgres',
     operatorsAliases: false,
+    stripe_secret: process.env.STRIPE_SECRET,
```
julien51: ditto

locksmith/config/config.js
```diff
@@ -15,13 +16,15 @@ module.exports = {
...
   },
   production: {
     username: process.env.DB_USERNAME,
     password: process.env.DB_PASSWORD,
     database: process.env.DB_NAME,
     port: process.env.DB_PORT || 5432,
     host: process.env.DB_HOSTNAME,
+    stripe_secret: process.env.STRIPE_SECRET,
```
julien51: ditto

locksmith/src/controllers/userController.ts
```diff
@@ -83,6 +83,22 @@ namespace UserController {
...
   }
+
+  export const updatePaymentDetails = async (
+    req: Request,
+    res: Response
+  ): Promise<any> => {
+    let emailAddress = req.params.emailAddress
+    let token = req.body.token
+
+    let result = await UserOperations.updatePaymentDetails(token, emailAddress)
```
julien51: IIRC we decided to not use emailAddress but `publicKey` as the universal account identifier?

locksmith/src/routes/user.js
```diff
@@ -13,5 +13,9 @@ router.get(
   userController.retrieveRecoveryPhrase
 )
 router.put('/:emailAddress', userController.updateUser)
+router.put(
+  '/:emailAddress/payment_details',
```
julien51: I think the difference wuth the previous route is that this one only happens when the user is logged in?

akeem: There is more of an operational difference around not modifying the working end point while we worth through rolling this out. The service layer doesn't exactly have a concept of a logged in user, at the time of writing the caller of the API determines when this will be called. 

With a valid stripe token, the users details will be updated.  When comfortable with the code we can merge this into the standard user updated endpoint if that seems like the right path

locksmith/src/controllers/userController.ts
akeem: In the data it absolutely is, i'll swap this endpoint to use that as the lookup key. The general use case for this endpoint a calling application generates a strike token and passes it forward with the  information on hand.

locksmith/src/routes/user.js
julien51: nit do you mind using `paymentdetails` since we use `recoveryphrase` or `privatekey`?

PR #2820, opened 2019-04-23
Updating datepicker to display UTC time

benwerd:
# Description

The setters are using UTC, so the display should too. Fixes an issue where the displayed date didn't quite correspond with the selected date.

# Checklist:

### DIFF
--- a/tickets/src/components/interface/DatePicker.js
+++ b/tickets/src/components/interface/DatePicker.js
@@ -78,8 +78,8 @@ export default class DatePicker extends Component {
       value: date.getMonth(),
       label: MONTH_NAMES[date.getMonth()],
     }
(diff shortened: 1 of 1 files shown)


cnasc:
> The setters are using UTC, so the display should too.

Is that necessarily the case? I think it might be correct to display adjusted to the local time. Or _a_ local time (i.e., if the event is in San Francisco it should display time local to the event, not the user)

julien51:
Agree with @cnasc on this one?

benwerd:
I tend to agree, so I'm revisiting this. (Although I do have to wonder: does it actually need to be a date structure at all? We're not performing date operations in any way.)

benwerd:
OK, I fixed this another way. Closing this PR.

PR #2822, opened 2019-04-23
Keep track of whether event creation form is submitted

cnasc:
# Description

This PR adds one piece of state to the Event creation form in the tickets app. Right now there is no visual impact, but we can use this state to change the page once the form is submitted. This should help us make a nice UX improvement once we have an idea what we'd like to do to indicate that the form was submitted.

# Checklist:

### DIFF
--- a/tickets/src/components/content/CreateContent.js
+++ b/tickets/src/components/content/CreateContent.js
@@ -36,6 +36,7 @@ export class CreateContent extends Component {
       description: '',
       location: '',
       date: now,
+      submitted: false,
     }
 
     this.onChange = this.onChange.bind(this)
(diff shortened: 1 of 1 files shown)


julien51:
test?

cnasc:
> test?

I don't think we have a good way to test this right now since `rtl` discourages actually testing methods. We can sit on this PR until we have some kind of effect dependent on this state, and then test for the presence or absence of that element.

cellog:
because this is local state, I think you should implement the UI in this PR, then it can be tested

cellog:
(i.e. there are no possible side effects, so it is not like redux in that you need to split the PR)

PR #2827, opened 2019-04-23
Add href to link on event creation page

cnasc:
# Description

The link on the event creation page wasn't clickable because it didn't have an href property. This PR adds that and a regression test.

# Checklist:

### DIFF
--- a/tickets/src/components/content/CreateContent.js
+++ b/tickets/src/components/content/CreateContent.js
@@ -163,7 +163,7 @@ export class CreateContent extends Component {
                     <SaveButton type="submit">Save Event</SaveButton>
                     <Text>
                       Your event link: <br />
-                      <Cta>
+                      <Cta href={lockAddress}>
                         <EventUrl address={lockAddress} />
                       </Cta>
                     </Text>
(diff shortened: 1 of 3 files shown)


cnasc:
Oh. Hm, this doesn't quite work. I think maybe `EventUrl` should be refactored as a function rather than a component, then we can use it to construct the href.

tickets/src/components/content/CreateContent.js
```diff
@@ -163,7 +163,7 @@ export class CreateContent extends Component {
                     <SaveButton type="submit">Save Event</SaveButton>
                     <Text>
                       Your event link: <br />
-                      <Cta>
+                      <Cta href={lockAddress}>
```
benwerd: Should this open in a new tab?

cnasc: It probably should — I think an overall better approach will be to make `EventUrl` return an `<a>` tag with that baked in.

julien51:
I think this is a better approach? https://github.com/unlock-protocol/unlock/pull/2850

julien51:
I think this is a better approach? https://github.com/unlock-protocol/unlock/pull/2850

cnasc:
Going to close this since we have other approaches in the works.

PR #2845, opened 2019-04-24
Event page adjusts to key and transaction status

benwerd:
# Description

The event page now reacts properly to key status. It will show as confirmed if you've bought a key, confirming if this is in progress, etc.

**Please only evaluate latest commit.** There will be one more cascading commit tonight, to produce the QR code that sits on the event page itself. Obviously this should be rebased before it's merged in, with the QR code component coming first. [There is a configuration change here, which will disappear when rebased. Please ignore.]

# Checklist:

### DIFF
--- a/tickets/src/components/content/purchase/PayButton.js
+++ b/tickets/src/components/content/purchase/PayButton.js
@@ -5,17 +5,19 @@ import PropTypes from 'prop-types'
 import withConfig from '../../../utils/withConfig'
 import UnlockPropTypes from '../../../propTypes'
 import Media from '../../../theme/media'
+import { KeyStatus } from '../../../selectors/keys'
 
-export const PayButton = ({ transaction, purchaseKey, config }) => {
+export const PayButton = ({ transaction, keyStatus, purchaseKey, config }) => {
   const { requiredConfirmations } = config
-  if (
-    transaction &&
-    ['submitted', 'pending'].indexOf(transaction.status) > -1
-  ) {
+  let transactionStatus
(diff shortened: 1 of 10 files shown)


tickets/.eslintrc.js
```diff
@@ -14,6 +14,14 @@ module.exports = {
   rules: {
     'react/prefer-stateless-function': [2],
     'react/forbid-prop-types': 2,
+    '@typescript-eslint/no-unused-vars': [
```
julien51: Any reason why you add this eslint rule? Maybe better on another PR?

tickets/src/components/content/purchase/PayButton.js
```diff
@@ -24,9 +26,13 @@ export const PayButton = ({ transaction, purchaseKey, config }) => {
         {requiredConfirmations})
       </PayInfo>
     )
-  } else if (transaction && transaction.status === 'mined') {
+  } else if (keyStatus === KeyStatus.CONFIRMING) {
+    // Key is being confirmed but we don't have transaction
```
julien51: Hum. How is that possible? The transaction should come before the key?

tickets/.eslintrc.js
cnasc: This rule should already be covered by our eslint config (and `tsc` also catches it)

tickets/src/components/content/purchase/PayButton.js
cellog: I think you should copy the logic from `paywall/src/components/lock/Lock.js` here

PR #2875, opened 2019-04-25
Disallow locks with 0 duration as well as those longer than 100 years

### LINKED ISSUE #2758: Lock form duration validation incorrect
**Describe the bug**
The lock form validation prevents users from creating locks with a negative duration with the error message displayed below. However, it does not prevent users from creating a lock with a duration of zero.

image

**To Reproduce**
1. Create a lock with a duration of zero

**Expected behavior**
Validation should not allow a lock with zero duration

**Additional context**
The validation is checking for non-negative numbers instead of positive numbers.

### LINKED ISSUE #2864: Max key duration is 100 years, we need to validate that on dashbaord UI
A user may try to create a lock whose keys are set to expire after 100 years and the contract will fail.

We need to inform the user that this is not a valid duration.

cnasc:
# Description

This PR fixes #2758 and fixes #2864 by making the lock form validation reject values of 0 and values greater than 100 years for the duration.

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/components/creator/CreatorLockForm.test.js
+++ b/unlock-app/src/__tests__/components/creator/CreatorLockForm.test.js
@@ -16,7 +16,7 @@ import {
 import {
   INFINITY,
   UNLIMITED_KEYS_COUNT,
-  oneHundredYearsInDays,
+  ONE_HUNDRED_YEARS_IN_DAYS,
 } from '../../../constants'
 
 describe('lockToFormValues', () => {
@@ -180,7 +180,7 @@ describe('CreatorLockForm', () => {
     it('key expiration is greater than 100 years', () => {
       expect.assertions(1)
       const { container } = makeLockForm({
-        expirationDuration: oneHundredYearsInDays + secondsInADay,
+        expirationDuration: ONE_HUNDRED_YEARS_IN_DAYS + 1,
       })
       expect(container.querySelector('.duration > input').dataset.valid).toBe(
         'false'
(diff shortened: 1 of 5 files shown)


unlock-app/src/__tests__/components/creator/CreatorLockForm.test.js
```diff
@@ -13,7 +13,11 @@ import {
   FORM_KEY_PRICE_INVALID,
 } from '../../../errors'
 
-import { INFINITY, UNLIMITED_KEYS_COUNT } from '../../../constants'
+import {
+  INFINITY,
+  UNLIMITED_KEYS_COUNT,
+  oneHundredYearsInDays,
```
julien51: Let's stick to UPERCASE FOR CONSTANTS!

unlock-app/src/utils/validators.js
```diff
@@ -1,10 +1,17 @@
...
 
-// tests whether a number is non-negative and not a decimal number
+// tests whether a number is positive and not a decimal number
 export const isPositiveInteger = val => {
   const parsedInt = parseInt(val)
-  return !isNaN(parsedInt) && val == parsedInt && +val >= 0
+  return !isNaN(parsedInt) && val == parsedInt && +val > 0
+}
+
+export const isLTEOneHundredYearsInDays = val => {
```
julien51: Do we want to make this customizable? `isLTE(x, max)` so we can re-use it?

cnasc:
Updated, @julien51 PTAL!

Issue #1764, opened 2019-02-28
Review wallet expectations for contracts


hardlydifficult:
**Is your feature request related to a problem? Please describe.**
When viewing a Lock or a Key on a blockexplorer or in a wallet, what information might the wallet try reading from the contract in order to make that a good experience for users?

**Describe the solution you'd like**
We should review any plans regarding ERC721 in wallets such as Metamask, Fortmatic, Portis.  Do they look for standard metadata fields? as an example

**Describe alternatives you've considered**
n/a

**Additional context**
Etherscan recently added ERC721 with name, description, and icon - although that info may be gathered off-chain like source code is today.

hardlydifficult:
Also review other popular NFTs to see what metadata they leverage and how

julien51:
This is a great point. Since we now have ERC721 full support, maybe we should plan an upgrade to our deployed smart contract to see if this works better?

hardlydifficult:
Yes, we should probably plan for an upgrade.  But for this, I believe we could test it before then - e.g. with Metamask I could point to the local ganache and see how things look.  I have not tried this yet though.

hardlydifficult:
related to this we should answer if we want to ERC165 register support the the ERC721 metadata extension.

julien51:
I think this will require deploying our latests smart contract to see how "compatible" they are with a list of the most popular crypto wallets, but ideally also market places such as opensea and rarebits.

nfurfaro:
I just wanted to start keeping track of this stuff somewhere. Can be moved to a google doc when I have time.

WALLET             ERC721 Spec             MetaData
======================================
OpenSea           OZ                              TokenURI, erc721 metadata extension
Rare Bits            EIP-721                      Custom?

PR #2894, opened 2019-04-29
Adding UnlockUtils contract with string concat methods

nfurfaro:
# Description

This is to support the erc721 metadata spec. It allows us to construct unique URI's for each token by concatenating the `tokenID` onto the base tokenURI.
Tests to come with a PR which allows Locks to start using these methods.

The code waws borrowed from here:
https://github.com/oraclize/ethereum-api/blob/master/oraclizeAPI_0.5.sol#L943

Note that I've simplified it to accept only 2 strings (instead of 5 as in the original), but it might be best to make this 3 strings, as we then have an easy way to add the lock address to the URI.

Not sure about how to go about accrediting the original author or if I need to be concerned? Here's a link to the License:
https://github.com/oraclize/ethereum-api/blob/master/LICENSE

# Issues

Fixes #
Refs #1764

# Checklist:

### DIFF
--- a/smart-contracts/contracts/UnlockUtils.sol
+++ b/smart-contracts/contracts/UnlockUtils.sol
@@ -1,0 +1,63 @@
+pragma solidity 0.5.7;
+
+// This contract provides some utility methods for use with the unlock protocol smart contracts.
+// Borrowed from:
+// https://github.com/oraclize/ethereum-api/blob/master/oraclizeAPI_0.5.sol#L943
+
+contract UnlockUtils {
+
+  function strConcat(
+    string memory _a,
+    string memory _b
+  ) internal
+    pure
+    returns(string memory)
+ {
+    bytes memory _ba = bytes(_a);
+    bytes memory _bb = bytes(_b);
+    string memory ab = new string(_ba.length + _bb.length);
+    bytes memory bab = bytes(ab);
+    uint k = 0;
+    for(uint i = 0; i < _ba.length; i++) {
+      bab[k++] = _ba[i];
+    }
+    for(uint i = 0; i < _bb.length; i++) {
+      bab[k++] = _bb[i];
+    }
+    return string(bab);
+  }
+
+  function strConcat(
+    string memory _a,
+    string memory _b
+  ) internal
+    pure
+    returns(string memory)
+  {
+    return strConcat(_a, _b, "", "", "");
+  }
+
+  function uint2str(
+    uint i
+  ) internal
+    pure
+    returns (string memory)
+  {
+    if (i == 0) {
+      return "0";
+    }
+    uint j = i;
+    uint len;
+    while(j != 0) {
+      len++;
+      j /= 10;
+    }
+    bytes memory bstr = new bytes(len);
+    uint k = len - 1;
+    while(i != 0) {
+      bstr[k--] = byte(48 + i % 10);
+      i /= 10;
+    }
+    return string(bstr);
+  }
+}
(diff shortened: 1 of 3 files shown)


smart-contracts/contracts/UnlockUtils.sol
```diff
@@ -0,0 +1,63 @@
...
+    for(uint i = 0; i < _ba.length; i++) {
+      bab[k++] = _ba[i];
+    }
+    for(uint i = 0; i < _bb.length; i++) {
+      bab[k++] = _bb[i];
+    }
+    return string(bab);
+  }
+
+  function strConcat(
```
hardlydifficult: These helpers are optional - but in the end I think we want 5 variables, so it would be:
tokenUri == baseUri + lockAddress + somethingInBetween + tokenId + extension
e.g. `https://unlock-protocol.com/metadata/0x1288...8181/101.json`

which we could do with `strConcat(strConcat(strConcat(strConcat(baseUri, lockAddress), somethingInBetween), tokenId), extension)`

smart-contracts/contracts/UnlockUtils.sol
```diff
@@ -0,0 +1,63 @@
...
+    string memory _a,
+    string memory _b
+  ) internal
+    pure
+    returns(string memory)
+  {
+    return strConcat(_a, _b, "", "", "");
+  }
+
+  function uint2str(
```
hardlydifficult: Do we need this one?

smart-contracts/contracts/UnlockUtils.sol
hardlydifficult: at the moment this is breaking CI

smart-contracts/contracts/UnlockUtils.sol
nfurfaro: I think so. This is how it would be used in PublicLock:
```
function tokenURI(
    uint256 _tokenId
  ) external
    view
    isKey(_tokenId)
    returns(string memory)
  {
    string memory URI;
    if(baseTokenURI == '') {
      URI = unlockProtocol.getGlobalBaseURI;
    } else {
      URI = baseTokenURI;
    }
    return UnlockUtils.strConcat(
      URI,
      UnlockUtils.uint2str(_tokenId)
    );
  }
```

nfurfaro: maybe there's an easier way to convert uints to strings though.

smart-contracts/contracts/UnlockUtils.sol
nfurfaro: Yeah, i see that. will fix it up.

smart-contracts/contracts/UnlockUtils.sol
hardlydifficult: ah, makes sense.  lg

hardlydifficult:
I think we may need an address2String function as well, to identify the Lock.

nfurfaro:
>'I think we may need an address2String function as well,'

I was thinking the same thing. Unless we can cast the address as a Uint first, then use `uint2str`. haven't tried yet.

hardlydifficult:
Yea casting to a uint first is a valid option - it'll just create a longer string than a hex address.

nfurfaro:
I updated a few things here. Still working on an `addressToStr` function to complete this.

nfurfaro:
Think this will work for us. The `address2Str` func returns the address as all lowercase characters, but otherwise works as expected.

PR #2905, opened 2019-04-30
send the current ethereum account to the iframe when it is ready

cellog:
# Description

Sending the current account will allow the iframe to look at its account saved in localStorage and change it if it is stale.

# Issues

Refs #2902

# Checklist:

### DIFF
--- a/paywall/src/paywall-builder/config.js
+++ b/paywall/src/paywall-builder/config.js
@@ -17,13 +17,22 @@ export function sendConfig(config, iframe, origin) {
 }
 
 function getAccount(window, iframe, origin) {
-  window.web3.eth.getAccounts((error, result) => {
-    if (error) return
-    iframe.contentWindow.postMessage(
-      { type: POST_MESSAGE_ACCOUNT, payload: result.result[0] },
-      origin
-    )
-  })
+  const id = new Date().getTime()
+  window.ethereum.sendAsync(
+    {
+      method: 'eth_accounts',
+      params: [],
+      jsonrpc: '2.0',
+      id,
+    },
+    (error, result) => {
+      if (error) return
+      iframe.contentWindow.postMessage(
+        { type: POST_MESSAGE_ACCOUNT, payload: result.result[0] },
+        origin
+      )
+    }
+  )
 }
 // this listens for the "ready" message from the iframe
 export function setupReadyListener(window, iframe, origin) {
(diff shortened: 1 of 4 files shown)


paywall/src/paywall-builder/config.js
```diff
@@ -26,6 +30,26 @@ export function setupReadyListener(window, iframe, origin) {
       // <script type="text/javascript">window.unlockConfig = {...}</script>
       // immediately before paywall.min.js is loaded
       sendConfig(window.unlockConfig, iframe, origin)
+      if (window.ethereum) {
+        window.ethereum.enable().then(() => {
```
julien51: I think not all web3 provider actually implement `.enable`. We probably want to check if it exists before.

paywall/src/paywall-builder/config.js
```diff
@@ -26,6 +30,26 @@ export function setupReadyListener(window, iframe, origin) {
...
+          const id = new Date().getTime()
+          window.ethereum.sendAsync(
```
julien51: Yikes. It is really gross that we have to do this rpc call ourselves :/

paywall/src/paywall-builder/config.js
cellog: good catch!

cellog:
all tidied up, PTAL @julien51

cellog:
> Have you been able to test this with our usual suspects?
> 
> * [ ]  FF/Chrome + Metamask
> * [ ]  Coinbase Wallet
> * [ ]  Opera on mobile

This cannot be tested until https://github.com/unlock-protocol/unlock/pull/2916 is merged

cellog:
just rebased, once it builds I'll use the temporary site to test it on coinbase wallet and company. It definitely works in Metamask because that's what I developed it in

cellog:
had to revert - web3.eth does not exist in Opera. sendAsync does not exist either, but send() does (and is standard across providers). Still no way to test in Coinbase unless we actually commit it to master, and I test on staging, so we may have to just do that tomorrow in a controlled environment so we can quickly revert if needed. Let's talk tomorrow about it briefly

cellog:
just added some code that also accepts the account if we have none at all

julien51:
LG with one comment/nit

paywall/src/paywall-builder/config.js
```diff
@@ -26,6 +48,15 @@ export function setupReadyListener(window, iframe, origin) {
       // <script type="text/javascript">window.unlockConfig = {...}</script>
       // immediately before paywall.min.js is loaded
       sendConfig(window.unlockConfig, iframe, origin)
+      if (window.web3 && window.web3.currentProvider) {
```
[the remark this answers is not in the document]
cellog: the logic inside `getAccount` needs to be called twice, so the result of moving it inside getAccount would be to repeat the `send` call twice. I actually extracted `getAccount` because of this need.  However, I *can* make a function that either waits for enable or executes a callback right away, which would be cleaner

PR #2920, opened 2019-04-30
implement ethers-ready versions of unlockService

cellog:
# Description

In order to implement the ethers-based version of unlockService, this PR uses the `ethers_` prefix for any methods that are modified. All changes are at the bottom of the file. The existing methods:

```
unlockContractAbiVersion()
lockContractAbiVersion()
_getPublicLockVersionFromContract()
_getVersionFromContract()
```

are all prefixed to become:

```
ethers_unlockContractAbiVersion()
ethers_lockContractAbiVersion()
_ethers_getPublicLockVersionFromContract()
_ethers_getVersionFromContract()
```

new methods added are:

```
getContract
getWritableContract
getLockContract
getUnlockContract
```

These are ethers-specific. In ethers, contracts are first-class citizens (https://docs.ethers.io/ethers.js/html/api-contract.html) and have providers assigned to them. Ethers additionally makes a strong separation between readable contracts and writable contracts.

Providers or signers must be connected to contracts before they can be used on-chain. There are 2 ways to do this. One is to create a contract, and then call `connect`. This PR chooses the other option, which is to create a new contract by passing the provider, or the signer, as the 3rd parameter.

This logic is handled by `getContract`, which uses the `writable` property (set in the constructor) to determine whether to make a read-only (web3Service) or a write-only (walletService) contract.

After this PR is approved and merged, the next steps are:

1. pass `writable: true` to `unlockService` from `walletService` to ensure we get a writable contract
2. set up `this.provider` in both `web3Service` and `walletService`, and add `_handleMethodCall` to walletService
3. migrate the stuff in `v0/`, `v01/`, and `v02/` over to use ethers. (this we can collaborate on really easily)
4. migrate the remaining `walletService` and `web3Service` methods
5. migrate `deploy.js` and `account.js` and `providers`
6. profit

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-js/src/unlockService.js
+++ b/unlock-js/src/unlockService.js
@@ -161,7 +161,7 @@ export default class UnlockService extends EventEmitter {
     let version = this.ethers_versionForAddress[this.unlockContractAddress]
     if (version === undefined) {
       // This was not memo-ized
-      version = await this.ethers_getVersionFromContract(
+      version = await this._ethers_getVersionFromContract(
         this.unlockContractAddress
       )
       this.ethers_versionForAddress[this.unlockContractAddress] = version
@@ -191,7 +191,7 @@ export default class UnlockService extends EventEmitter {
     let version = this.ethers_versionForAddress[address.toLowerCase()]
     if (version === undefined) {
       // This was not memo-ized
-      version = await this.ethers_getPublicLockVersionFromContract(address)
+      version = await this._ethers_getPublicLockVersionFromContract(address)
       this.ethers_versionForAddress[address.toLowerCase()] = version
     }
 
@@ -205,7 +205,7 @@ export default class UnlockService extends EventEmitter {
     return v0
   }
 
-  async ethers_getPublicLockVersionFromContract(address) {
+  async _ethers_getPublicLockVersionFromContract(address) {
     const contract = new Contract(
       address,
       ['function publicLockVersion() view returns (uint8)'],
@@ -225,7 +225,7 @@ export default class UnlockService extends EventEmitter {
    * Private method, which given an address will query the contract and return the corresponding method
    * @param {*} address
    */
-  async ethers_getVersionFromContract(address) {
+  async _ethers_getVersionFromContract(address) {
     const contract = new Contract(
       address,
       ['function unlockVersion() view returns (uint8)'],
--- a/unlock-js/src/unlockService.js
+++ b/unlock-js/src/unlockService.js
@@ -152,22 +152,17 @@ export default class UnlockService extends EventEmitter {
     return version
   }
 
-  /**
-   * Returns the implementation based on the deployed version
-   * @param {*} address
-   */
-  async ethers_unlockContractAbiVersion() {
+  async contractAbiVersion(address, versionRetrievalMethodName) {
     if (!this.provider) {
       throw new Error(Errors.MISSING_WEB3)
     }
 
-    let version = this.ethers_versionForAddress[this.unlockContractAddress]
+    const contractAddress = address.toLowerCase()
+    let version = this.ethers_versionForAddress[contractAddress]
     if (version === undefined) {
       // This was not memo-ized
-      version = await this._ethers_getVersionFromContract(
-        this.unlockContractAddress
-      )
-      this.ethers_versionForAddress[this.unlockContractAddress] = version
+      version = await this[versionRetrievalMethodName](contractAddress)
+      this.ethers_versionForAddress[contractAddress] = version
     }
 
     if (1 === version) {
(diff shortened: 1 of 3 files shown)


unlock-js/src/unlockService.js
```diff
@@ -136,4 +151,132 @@ export default class UnlockService extends EventEmitter {
...
+      this.provider
+    )
+    return this.lockContracts[lockAddress]
+  }
+
+  async getUnlockContract() {
+    if (this.unlockContract) {
+      return this.unlockContract
+    }
+    const version = await this.ethers_unlockContractAbiVersion()
```
akeem: how would you feel about wrapping this in an else? It is essentially the same thing but should highlight that all branches of this are covered?

unlock-js/src/unlockService.js
```diff
@@ -136,4 +151,132 @@ export default class UnlockService extends EventEmitter {
...
+  async getWritableContract(address, contract) {
+    const signer = this.provider.getSigner()
+    return new Contract(address, contract.abi, signer)
+  }
+
+  async getLockContract(lockAddress) {
+    if (this.lockContracts[lockAddress]) {
+      return this.lockContracts[lockAddress]
+    }
+    const version = await this.ethers_lockContractAbiVersion(lockAddress)
```
akeem: Would you be against the remaining portion of this in an else?

unlock-js/src/unlockService.js
```diff
@@ -56,11 +71,11 @@ export default class UnlockService extends EventEmitter {
       throw new Error(Errors.MISSING_WEB3)
     }
 
-    let version = this.versionForAddress[address]
-    if (!version) {
+    let version = this.versionForAddress[address.toLowerCase()]
```
akeem: Does the address need to be lower case here? What occurs when it isn't?

cellog: there are 2 versions of addresses used in the ethereum world, the all-lower-cased address, and the checksummed address, which has a mixture of upper and lower case letters. If we don't lower-case the address, we will end up constructing 2 copies of the contract and memoizing both, which uses quite a bit of memory unnecessarily (about 3k per contract)

unlock-js/src/unlockService.js
cellog: not against it, I tend to prefer nesting as little as possible, but we aren't in any danger here of bracket hell :)

cellog: done

unlock-js/src/unlockService.js
cellog: done

unlock-js/src/unlockService.js
```diff
@@ -136,4 +151,134 @@ export default class UnlockService extends EventEmitter {
...
+    if (version === undefined) {
+      // This was not memo-ized
+      version = await this._ethers_getVersionFromContract(
+        this.unlockContractAddress
+      )
+      this.ethers_versionForAddress[this.unlockContractAddress] = version
+    }
+
+    if (1 === version) {
+      return v01
```
cnasc: Same question as L54

unlock-js/src/unlockService.js
```diff
@@ -136,4 +151,134 @@ export default class UnlockService extends EventEmitter {
...
+  async ethers_lockContractAbiVersion(address) {
+    if (!this.provider) {
+      throw new Error(Errors.MISSING_WEB3)
+    }
+
+    let version = this.ethers_versionForAddress[address.toLowerCase()]
+    if (version === undefined) {
+      // This was not memo-ized
+      version = await this._ethers_getPublicLockVersionFromContract(address)
+      this.ethers_versionForAddress[address.toLowerCase()] = version
```
cnasc: Same as above questions about the lowercasing. Would it be a good idea to have a setter for `ethers_versionForAddress` method that handles conversion to lowercase?

unlock-js/src/unlockService.js
```diff
@@ -136,4 +151,134 @@ export default class UnlockService extends EventEmitter {
...
+    // The good news (luck) is that no contract was ever deployed as v01
+    if (1 === version) {
+      return v02
+    }
+
+    // Defaults to v0
+    return v0
+  }
+
+  async _ethers_getPublicLockVersionFromContract(address) {
```
cnasc: This and `_ethers_getVersionFromContract` are almost the exact same. Worth abstracting a bit?

unlock-js/src/unlockService.js
cellog: actually, this change is not needed in the web3 side of things, I'll remove it from there, and we can apply your suggestions below

unlock-js/src/unlockService.js
cellog: since it is now only used in 2 places, I think I'll just use a local variable, but if this becomes a larger question, we could use the checksummed address as our default

unlock-js/src/unlockService.js
```diff
@@ -33,12 +44,16 @@ export default class UnlockService extends EventEmitter {
...
-    if (!version) {
-      // This was no memo-ized
+    if (version === undefined) {
+      // This was not memo-ized
       version = await this._getVersionFromContract(this.unlockContractAddress)
       this.versionForAddress[this.unlockContractAddress] = version
     }
 
+    if (1 === version) {
+      return v01
```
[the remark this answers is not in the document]
cellog: fixed the thing - now we validly retrieve v01 to support those few staging locks

unlock-js/src/unlockService.js
cellog: yes!

cellog:
ok @akeem and @cnasc a few more changes uploaded based on your feedback, PTAL when you have a chance

unlock-js/src/unlockService.js
```diff
@@ -181,31 +176,26 @@ export default class UnlockService extends EventEmitter {
     // Defaults to v0
     return v0
   }
+  /**
+   * Returns the implementation based on the deployed version
+   * @param {*} address
```
cnasc: Erroneous JSDoc?

PR #2930, opened 2019-05-01
Pass confirmLink to wedlocks

cnasc:
# Description

Part of #2111 and #2721, this PR adds the confirmation link href to the call to wedlocks, which will allow us to actually use account creation end-to-end.

Example email from signup:

image

# Checklist:

### DIFF
--- a/unlock-app/src/middlewares/wedlocksMiddleware.ts
+++ b/unlock-app/src/middlewares/wedlocksMiddleware.ts
@@ -11,7 +11,7 @@ const wedlocksMiddleware = (config: any) => {
       return (action: Action) => {
         if (action.type === SIGNUP_EMAIL) {
           // TODO: then and catch? I think we really only need to worry about errors.
-          wedlockService.confirmEmail(action.emailAddress)
+          wedlockService.confirmEmail(action.emailAddress, '')
         }
         next(action)
       }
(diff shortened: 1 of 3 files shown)


unlock-app/src/middlewares/wedlocksMiddleware.ts
```diff
@@ -10,8 +10,12 @@ const wedlocksMiddleware = (config: any) => {
     return (next: any) => {
       return (action: Action) => {
         if (action.type === SIGNUP_EMAIL) {
+          const { origin } = window.location
```
cellog: window may not be available on server-side, you should use `global.location`

cnasc: Typescript doesn't like referencing `global.location`. I think that there's no way for the `SIGNUP_EMAIL` action to be produced server-side since it's only emitted as a result of a user interacting with the browser.

If it's important I can look into what kind of type patching I'll need to do to make it work with global, but I think this is a case where use of `window` shouldn't be a problem.

cellog: I think that because next aliases global to window, perhaps they have figured out the typing issues?

cnasc: Looks like the solution is to patch the `global` type. Rather than doing that, which may hide errors in the future (if we change frameworks or something), why don't we add a check that window is actually available?

cellog: sounds like a good solution

PR #2933, opened 2019-05-01
new utility file for ethers

cellog:
# Description

This is the utility file that ethers will use. There are some significant differences in the formats that ethers expects (some values must be BigNumbers instead of strings, for example) that makes a seamless transition with the same Web3Utils API not feasible

# Issues

Refs #2782

# Checklist:

### DIFF
--- a/unlock-js/src/utils.ethers.js
+++ b/unlock-js/src/utils.ethers.js
@@ -1,0 +1,27 @@
+import { utils, constants } from 'ethers'
+
+// this allows us to flexibly upgrade web3 and fix bugs as they surface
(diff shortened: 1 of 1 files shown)


cellog:
> I think it would probably be ideal if we had some tests ensuring that these shim utils return the same results as the non-ethers versions. This is a great approach though

Unfortunately, what I have discovered is that the shim utils don't return the same values, and can't, because ethers expects different values. So the well-intentioned idea may not work. Perhaps the best approach will be to remove this PR, and just do utils updates piecemeal as the actual code comes in.

cnasc:
Ah. That's no fun.

PR #2937, opened 2019-05-01
Correctly index into return value of getUserPrivateKey

### LINKED ISSUE #2721: Be able to log in to an account

cnasc:
# Description

This PR corrects how `storageMiddleware` accessed the account data provided by `locksmith`. Once this is merged, the end-to-end account creation and login milestone will be complete 🎉

# Issues

Fixes #2721 
Refs #2111 

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/middlewares/storageMiddleware.test.js
+++ b/unlock-app/src/__tests__/middlewares/storageMiddleware.test.js
@@ -1,7 +1,4 @@
-import {
-  createAccountAndPasswordEncryptKey,
-  getAccountFromPrivateKey,
-} from '@unlock-protocol/unlock-js'
+import * as UnlockJS from '@unlock-protocol/unlock-js'
 import storageMiddleware from '../../middlewares/storageMiddleware'
 import { UPDATE_LOCK, updateLock, UPDATE_LOCK_NAME } from '../../actions/lock'
 import { STORE_LOCK_NAME } from '../../actions/storage'
@@ -13,8 +10,6 @@ import { startLoading, doneLoading } from '../../actions/loading'
 import configure from '../../config'
 import { SIGNUP_CREDENTIALS } from '../../actions/signUp'
 import { LOGIN_CREDENTIALS } from '../../actions/login'
-
-jest.mock('@unlock-protocol/unlock-js')
 
 /**
  * This is a "fake" middleware caller
@@ -269,10 +264,6 @@ describe('Storage middleware', () => {
   describe('SIGNUP_CREDENTIALS', () => {
     it('should create a user and then set the account', () => {
       expect.assertions(2)
-      createAccountAndPasswordEncryptKey.mockReturnValue({
-        address: '',
-        passwordEncryptedPrivateKey: '',
-      })
       const emailAddress = 'tim@cern.ch'
       const password = 'guest'
       const { next, invoke } = create()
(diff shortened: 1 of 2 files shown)


cnasc:
These tests pass locally 🤔

cellog:
yeah that unlock-js test is a transient failure, no idea why. I'm hoping it dies with the ethers migration...

cnasc:
OK, failures gone. PTAL!

unlock-app/src/__tests__/middlewares/storageMiddleware.test.js
```diff
@@ -10,6 +14,8 @@ import configure from '../../config'
 import { SIGNUP_CREDENTIALS } from '../../actions/signUp'
 import { LOGIN_CREDENTIALS } from '../../actions/login'
 
+jest.mock('@unlock-protocol/unlock-js')
```
cellog: this makes me nervous - the test will pass even when the underlying implementation shifts.

Could you add a separate test in a new file that just verifies assumptions about the underlying API? That way we won't blow up when the next version is published

cellog: alternative: rewrite the test to not mock, and take advantage of `expect.objectContaining` (https://jestjs.io/docs/en/expect#expectobjectcontainingobject) and `expect.any` (https://jestjs.io/docs/en/expect#expectanyconstructor) to ignore the implementation details of what the private key looks like and so on

cnasc: This is like this because when I tried to just spy on `getAccountFromPrivateKey` and it didn't seem to work without doing a whole mock. If you have any suggestions on how to fix that I'm all ears, because I don't like this either

cellog: Ok, I'll take a look, heading in to the office now

cnasc: No worries, I'll continue to play with it as well

cnasc:
Just pushed what I was doing with `jest.spyOn` as an example -- canceled the CI because it fails with the following error: 
```
Storage middleware › LOGIN_CREDENTIALS › Should call getAccountFromPrivateKey with correct data

    expect(jest.fn()).toHaveBeenCalledWith(expected)

    Expected mock function to have been called with:
      ["Private Key reporting for duty", "guest"]
    But it was not called.

      311 |         emailAddress
      312 |       )
    > 313 |       expect(spy).toHaveBeenCalledWith(
          |                   ^
      314 |         'Private Key reporting for duty',
      315 |         'guest'
      316 |       )
```

cnasc:
More robust test in place, PTAL

unlock-app/src/middlewares/storageMiddleware.js
```diff
@@ -140,7 +140,8 @@ const storageMiddleware = config => {
           const { emailAddress, password } = action
           storageService
             .getUserPrivateKey(emailAddress)
-            .then(key => {
+            .then(result => {
+              const key = result.data.passwordEncryptedPrivateKey
```
julien51: `storageService` should actually handle the logic of extracttibg the  `data` from `result` and only yield `passwordEncryptedPrivateKey`. Could you change that?

PR #2948, opened 2019-05-02
migrate deploy script to ethers

cellog:
# Description

This PR migrates the contract deploy script for local dev to ethers. Note that this is the only place that introduces a backwards-compatibility break. The contract address was `contract.options.address` with web3, and is `contract.address` in ethers, so `deploy-unlock.js` will need to be updated when upgrading to the next version of unlock-js

Also, this really shows where ethers shines. The code is concise and easy to reason through, as well as easy to debug (in my experience)

# Issues

Refs #2782 

# Checklist:

### DIFF
--- a/unlock-js/src/deploy.js
+++ b/unlock-js/src/deploy.js
@@ -1,33 +1,31 @@
-const Web3 = require('web3')
+const ethers = require('ethers')
 const gas = require('./constants').GAS_AMOUNTS
 
 export default async function deploy(
   host,
   port,
   Unlock,
-  onNewContractInstance = () => {},
-  web3Object
+  onNewContractInstance = () => {}
 ) {
-  const web3 = web3Object || new Web3(`http://${host}:${port}`)
-  const unlock = new web3.eth.Contract(Unlock.abi)
+  const provider = new ethers.providers.JsonRpcProvider(
+    `http://${host}:${port}`
+  )
+  // Load the wallet to deploy the contract with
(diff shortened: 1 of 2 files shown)


cnasc:
The Ethers API is definitely more fluent than Web3. Looks great, but I'll let someone more familiar with contract deployment give it the thumbs up

unlock-js/src/deploy.js
```diff
@@ -1,33 +1,31 @@
...
+  // Load the wallet to deploy the contract with
+  const wallet = provider.getSigner(0)
+  const factory = new ethers.ContractFactory(
+    Unlock.abi,
+    Unlock.bytecode,
+    wallet
+  )
+  const accounts = await provider.listAccounts()
+  const unlockContract = await factory.deploy({ gasLimit: gas.deployContract })
+  onNewContractInstance(unlockContract)
```
julien51: Shouldn't this be after `await unlockContract.deployed()` to make sure that the callback is only called after the contract has effectively been deplpyed?

julien51: or even after `initialize`?

cellog: That is fine if you'd like. The current (web3-based) implementation calls as soon as we get the contract address, which is what this does. If you'd like to change it to after init that's a trivial change, your call.

cellog:
all fixed up, PTAL @julien51

PR #2968, opened 2019-05-02
WIP: passing token name via createLock()

nfurfaro:
# Description

This PR is to support the erc721 metadata functions.
Currently failing with an "out of gas" error.

As we already have the UI for naming a lock, it should be straightforward to pass that name to the `createLock()` function, or default to something like "Unlock-Protocol Lock Contract" if no name is set. This will need a little work on the front-end to pass the name to the contract.

# Issues

Fixes #
Refs #1764

# Checklist:


nfurfaro:
I was originally going to have all 3 (URI, Symbol & Name) set as defaults in `Unlock.sol` , but Jul;ian pointed out that in the dashboard, we already ask the creator to name it, so passing that name in to the createLock function made sense. As far as a default name, I think we just ensure that the front end always passes something, even if the creator doesn't enter a custom name.
I think for the Locks we'll use the symbol "KEY" as it describes well what each token is. That said, the owner should be able to override this if they choose to.

PR #2971, opened 2019-05-02
Initial validation work on user account creation

cnasc:
# Description

Going to leave this as a draft for now unless we think it's important enough to go in as-is.
This PR introduces some basic password validation. Right now it requires that the password not be empty and that the password and its confirmation match.

In order to enforce this, it uses the validation results to disable the form when the inputs aren't valid.

Future work:

- Add message/indicator showing validity
- Grey out disabled submit button

# Issues

Refs #2111

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/components/interface/FinishSignup.test.tsx
+++ b/unlock-app/src/__tests__/components/interface/FinishSignup.test.tsx
@@ -1,6 +1,8 @@
 import React from 'react'
 import * as rtl from 'react-testing-library'
 import {
+  mapDispatchToProps,
+  validatePassword,
   FinishSignup,
   Credentials, // eslint-disable-line no-unused-vars
 } from '../../../components/interface/FinishSignup'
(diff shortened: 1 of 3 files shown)


unlock-app/src/__tests__/components/interface/FinishSignup.test.tsx
```diff
@@ -43,4 +45,54 @@ describe('FinishSignup', () => {
...
+      expect(validatePassword('guest', 'ghost').isValid).toBeFalsy()
+    })
+  })
+
+  describe('mapDispatchToProps', () => {
+    it('maps the dispatch to the props', () => {
+      expect.assertions(1)
+      expect(mapDispatchToProps((arg: any) => arg)).toEqual(
+        expect.objectContaining({
+          signupCredentials: expect.any(Function),
```
cellog: can also add an assertion that calls the signupCredentials and verifies it spits out the right action too

unlock-app/src/components/interface/FinishSignup.tsx
```diff
@@ -16,6 +16,24 @@ interface Props {
...
+  errors: string[]
+}
+
+export const validatePassword = (
+  password: string,
+  passwordConfirmation: string
+) => {
+  let errors: string[] = []
+
+  if (password.length < 1) {
```
cellog: maybe choose a number a bit longer, and probably put that in constants.js?

cnasc: This is only there to ensure it's not empty. Ultimately there will be an additional check for password complexity, ideally including a meter indicating whether a password is weak, medium, or strong.

PR #2993, opened 2019-05-04
convert ethers's representation of infinite keys to UNLIMITED_KEYS_COUNT

cellog:
# Description

One of the details of Unlock that will cause issues when upgrading to use ethers is that infinite keys is represented internally slightly differently.

This PR provides a thin layer of abstraction that will allow us to clearly communicate to ethers the value of infinity without hard-coding it in the apps.

When installing the ethers version of the app, we will need to update walletMiddleware listening to `CREATE_LOCK` and then convert our app representation of infinity (`-1`) to `unlock-js`'s representation (`INFINITE_KEYS`). Note that web3 used a totally different system to represent the uint256 value we use for infinity, so this breaks creation of infinite keys

# Issues

Refs #2782 

# Checklist:

### DIFF
--- a/unlock-js/index.js
+++ b/unlock-js/index.js
@@ -6,12 +6,10 @@ const {
 } = require('./lib/accounts')
 const { getCurrentProvider, getWeb3Provider } = require('./lib/providers')
 const deploy = require('./lib/deploy').default
-const { ETHERS_MAX_UINT } = require('./lib/constants')
 
 module.exports = {
   Web3Service: Web3Service.default,
   WalletService: WalletService.default,
-  INFINITE_KEYS: ETHERS_MAX_UINT,
   createAccountAndPasswordEncryptKey,
   getAccountFromPrivateKey,
   getCurrentProvider,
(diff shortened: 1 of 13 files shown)


julien51:
It'd much prefer if we did not have to change the logic of the apps that implement this.
Can we rather make sure that the value is converted to what we used to have with web3js?

In the same way that we mapped `totalSupply` to `outstandingKeys`?

cellog:
Yeah I think that's best, I'll refactor this to do that

cellog:
ok @julien51 I refactored to convert from `UNLIMITED_KEYS_COUNT` to `ETHERS_MAX_UINT256` inside `createLock`. I also added a new test to verify that the conversion happens but we still emit the correct value of `UNLIMITED_KEYS_COUNT` in `'lock.updated'`.

I also amended the `getLock` tests to ensure that when we have a lock with unlimited keys, we return `UNLIMITED_KEYS_COUNT` and not `ETHERS_MAX_UINT256`

PR #2994, opened 2019-05-04
Refactored layout and adding calendar link invite

julien51:
All in the title!

### DIFF
--- a/tickets/src/components/content/EventContent.js
+++ b/tickets/src/components/content/EventContent.js
@@ -4,7 +4,7 @@ import { connect } from 'react-redux'
 import Head from 'next/head'
 import PropTypes from 'prop-types'
 import GlobalErrorConsumer from '../interface/GlobalErrorConsumer'
-
+import { googleCalendarLinkBuilder } from '../../utils/links.ts'
 import { Field, Label } from './CreateContent'
 import { MONTH_NAMES, pageTitle, TRANSACTION_TYPES } from '../../constants'
 import UnlockPropTypes from '../../propTypes'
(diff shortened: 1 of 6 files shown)


julien51:
image

tickets/src/components/content/EventContent.js
```diff
@@ -201,10 +251,11 @@ export const mapStateToProps = (
           title: '💬 Telegram Group',
           href: 'https://t.me/joinchat/GzMTuRZfLMHZ1n2EMmF0UQ',
         },
-        {
-          title: 'ℹ️ Meetup Page',
-          href: 'TK',
-        },
+        // {
```
julien51: For now we do not want to point to the eventbrite page...

tickets/src/components/content/EventContent.js
```diff
@@ -58,22 +58,43 @@ export class EventContent extends Component {
 
     if (!lock.address || !event.name) return null // Wait for the lock and event to load
 
-    const { name, description, location, date, links, image } = event
+    const { name, description, location, date, links = [], image } = event
```
cnasc: I assume this is because not all events have an associated links field, so this serves as the default. If that's the case, would it make more sense to insert this default where events are pulled in from storage?

tickets/src/components/content/EventContent.js
```diff
@@ -58,22 +58,43 @@ export class EventContent extends Component {
...
         <li key={href}>
-          <a href={href}>{title}</a>
+          <a target="_blank" rel="noopener noreferrer" href={href}>
+            {title}
+          </a>
         </li>
       )
     })
 
+    const formatDateForCalendar = date => {
```
cnasc: It looks like lines 79 through 97 could be pulled into a helper function just to avoid expanding the render method too greatly. Matter of taste though.

tickets/src/components/content/EventContent.js
julien51: You are right! I will actually add another PR soon which adds the ability to add links.

tickets/src/components/content/EventContent.js
julien51: Good point.

PR #3001, opened 2019-05-05
don't retrieve a web3 provider at all, just return the URL, and let walletService make its own

cellog:
# Description

So this is a short PR that requires a long explanation.

In our web3 version of `unlock-js`, we were instantiating providers in `config.js` and passing them down to set up the `walletService`. In the preparation to move away from `web3`, this code was extracted to `unlock-js` `providers.js`.  Originally, `walletService` simply accepted a provider as its argument to `connect`.

Now, the relevant code in `walletService.connect` is:

```js
    if (typeof provider === 'string') {
      this.provider = new ethersProviders.JsonRpcProvider(provider)
      this.web3Provider = false
    } else {
      this.provider = new ethersProviders.Web3Provider(provider)
      this.web3Provider = provider
    }
```

Thus, we can accept a string (the URL we will find our HTTP provider at), and create our own ethers `JsonRpcProvider`. On the other hand, if the provider passed in is (for instance) metamask, which is a `web3` provider, we will use it with an ethers `Web3Provider`.

The only instance in which we will use an HTTP provider is in local development (ganache), and so having `getWeb3Provider` return the URL it is passed means we can remove the only remaining dependency on `web3` in the `unlock-js` source.

At a future date, we may wish to change this behavior, so I think it's worth keeping this.

For instance, we might decide to create an ethers JsonRpcProvider or even a webservice provider from the URL passed in. So, instead of removing `getWeb3Provider`, this PR has it return its URL argument.

# Issues

Refs #2782

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/providers.test.js
+++ b/unlock-js/src/__tests__/providers.test.js
@@ -1,12 +1,11 @@
 import { getWeb3Provider } from '../providers'
 
 describe('web3 provider creator', () => {
-  it('getWeb3Provider returns a valid web3 provider for a URL', () => {
-    expect.assertions(2)
+  it('getWeb3Provider returns the URL it is given', () => {
+    expect.assertions(1)
 
     // using ws so that the test will still pass when we move to WebSocketProvider
     const provider = getWeb3Provider('ws://1.2.3.4:1234')
-    expect(provider.send).toBeInstanceOf(Function)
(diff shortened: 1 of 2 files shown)


cellog:
I tested it with all of them before I did the PR. When I do the release PR I'll post screencasts for everything, should set us at ease?

PR #3002, opened 2019-05-05
fix signing in walletService

cellog:
# Description

While testing `unlock-js` with the tickets app, several problems with the ethers migration of signatures surfaced that were not being caught by the tests. This PR fixes the problems and updates the tests so that they verify correct interaction with the underlying JSON-RPC library.

The calls to `eth_signTypedData` were totally wrong (passing string in the case of non-metamask providers), and the calls to personal_sign and eth_sign were all mixed up as well.

The tests now verify that the expected JSON-RPC calls are sent instead of mocking out the provider. This is a crucial step in verifying correctness, while still hiding implementation details.

# Issues

Refs #2782 

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/helpers/nockHelper.js
+++ b/unlock-js/src/__tests__/helpers/nockHelper.js
@@ -214,6 +214,10 @@ export class NockHelper {
     return this._jsonRpcRequest('personal_sign', [hash, account], result, error)
   }
 
+  ethSignAndYield(hash, account, result, error) {
+    return this._jsonRpcRequest('eth_sign', [account, hash], result, error)
+  }
+
   getTransactionCount(address, count) {
     return this._jsonRpcRequest(
       'eth_getTransactionCount',
--- a/unlock-js/src/__tests__/helpers/nockHelper.js
+++ b/unlock-js/src/__tests__/helpers/nockHelper.js
@@ -221,7 +221,7 @@ export class NockHelper {
   ethSignTypedDataAndYield(account, data, result, error) {
     return this._jsonRpcRequest(
       'eth_signTypedData',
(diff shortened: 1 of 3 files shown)


julien51:
@cellog @akeem Can you confirm that this is __not__ required to make tickets work like _right now_?

cellog:
There is one change that may be needed, which is the change to signData for metamask. The current code uses `JSON.stringify()` which probably fails, as metamask (as documented https://medium.com/metamask/eip712-is-coming-what-to-expect-and-how-to-use-it-bb92fd1a7a26) requires a struct. The rest should be functionally equivalent to what is there now.

julien51:
I am not 100% sure I understand and I need a yes/no.

cellog:
OK, I found the line we need here: https://github.com/MetaMask/metamask-extension/blob/c4caba131776ff7397d3a4071d7cc84907ac9a43/app/scripts/metamask-controller.js#L997

Current code is correct, this PR is wrong, I'll push a fix momentarily (what a s***show this EIP-721 has turned out to be!)

cellog:
so: yes, tickets with current unlock-js is working. This code is 100% the same functionality, ported to ethers

akeem:
There has been a bit of flip flopping in this ticket, so I would like to be sure that I understand that what has happened. Ultimately it sounds like what current is/was in place correct? Is that right?

cellog:
> This seems good for the most part, collectively we should make sure that we smoke test this before rolling it out to the production application to ensure that we don't cause disruption to any events in progress

I'd like to talk about this in the office tomorrow, so we are on the same page and all feel confident about this. Thanks Akeem!

PR #3046, opened 2019-05-06
Add cancelAndRefundFor to PublicLock.sol

hardlydifficult:
# Description

Implements `cancelAndRefundFor` and associated functions allowing a key owner to approve another account to cancel their Key and take the refund.

The use case is the key owner pays with credit card and then requests a refund via credit card.  In order to do this, they approve Unlock canceling the key to get the crypto and then we can safely refund the credit card.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinRefunds.sol
+++ b/smart-contracts/contracts/mixins/MixinRefunds.sol
@@ -59,15 +59,6 @@ contract MixinRefunds is
   {
     require(
       ECDSA.recover(
-        //ECDSA.toEthSignedMessageHash(
-          getCancelAndRefundApprovalHash(_keyOwner, msg.sender)
-        //)
-        ,
-        _signature
-      ) != address(0), 'BAD_SIGNATURE_DATA' // TODO remove(?)
-    );
-    require(
-      ECDSA.recover(
         ECDSA.toEthSignedMessageHash(
           getCancelAndRefundApprovalHash(_keyOwner, msg.sender)
         ),
--- a/smart-contracts/contracts/mixins/MixinRefunds.sol
+++ b/smart-contracts/contracts/mixins/MixinRefunds.sol
@@ -1,9 +1,10 @@
 pragma solidity 0.5.7;
 
 import 'openzeppelin-eth/contracts/ownership/Ownable.sol';
+import 'openzeppelin-solidity/contracts/math/SafeMath.sol';
+import 'openzeppelin-solidity/contracts/cryptography/ECDSA.sol';
 import './MixinKeys.sol';
 import './MixinLockCore.sol';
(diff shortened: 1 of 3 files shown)


nfurfaro:
I'm reviewing this. Just want to make sure that I understand what's happening.

hardlydifficult:
This introduces a couple new patterns - maybe this would be a good topic for tomorrow's call RE solidity

nfurfaro:
I'm unavailable on Tuesdays, but feel free to discuss with the others. I'm looking at `ECDSA.sol` to see what's going on.

smart-contracts/contracts/mixins/MixinRefunds.sol
```diff
@@ -39,19 +44,40 @@ contract MixinRefunds is
...
+   * @dev Cancels a key owned by a different user and sends the funds to the msg.sender.
+   * @param _keyOwner this user's key will be canceled
+   * @param _signature getCancelAndRefundApprovalHash signed by the _keyOwner
+   */
+  function cancelAndRefundFor(
+    address _keyOwner,
+    bytes calldata _signature
+  ) external
+  {
+    require(
```
nfurfaro: So this proves the signer is the `_keyOwner`.

hardlydifficult: Yes.  It verifies keyOwner signed it, and what they signed specifically included the 3 points from the getHash function.

smart-contracts/contracts/mixins/MixinRefunds.sol
```diff
@@ -39,19 +44,40 @@ contract MixinRefunds is
...
 
-    if (refund > 0) {
-      // Security: doing this last to avoid re-entrancy concerns
-      _transfer(msg.sender, refund);
-    }
+  /**
+   * @dev Increments the current nonce for the msg.sender.
+   * This can be used to invalidate a previously signed message.
+   */
+  function incrementNonce(
```
nfurfaro: If this function is called by the _keyOwner after they've signed/approved a cancellation, but before `cancelAndRefundFor` is called, they can prevent it from being successful because `getCancelAndRefundApprovalHash` will no longer return a the right hash (because of the nonce change).

hardlydifficult: Correct.  However a more effective attack would be to simply call `cancelAndRefund` before we do.  I added a comment RE this in the test.  The backend implementation can guard against this by waiting for the tx to be mined before issuing the CC refund - or we accept this risk.

nfurfaro: Yeah , I think waiting for the tx to be mined is worthwhile.

PR #3071, opened 2019-05-08
new compressor for contract ABI

cellog:
# Description

*You may be wondering "where are the tests?" Because unlock-js has 100% code coverage, the tests caught every bug in the signature generation. Thus, the tests that matter are actually the ones that use the generated abi and bytecode. For instance, the bytecode is returned by nock to a `eth_getCode` call, and then hashed and compared to the hash of the bytecode we make in the script. If they didn't match, all the `v01` tests would fail outright, because the code would use the `v0` smart contract. Similarly, if the generated signatures did not have parameter names, all the `inputsHandlers` tests fail, as do the `eventsHandlers` tests. All the method calls fail if the types are wrong, and so on and so forth. Thus, the testing of the generation is in the subsequent PR that enables the use of the generated compressed abi.*

In the process of making the service worker for our new paywall, I discovered that nearly all of the bulk was taken up by our contract bytecode. This PR introduces a new compressor that transforms the bytecode JSON into a few discrete parts. The first is the abi that is needed to construct working smart contract function calls from ethers.js. This is the new `src/abis.js` file. The second part is needed for testing only, and is the generated `src/__tests__/helpers/bytecode.js`. This file contains the deployed bytecode for all the contracts, and is used to mock `eth_getCode` calls. Because the file is in `__tests__` it will be ignored when packaging up the npm releases.

With these changes, the source code can remove a direct dependency on the abi npm packages (PR will follow this one), and greatly reduce download size, install time, and so on.

# Issues

Refs #2903 

# Checklist:


julien51:
So that works, but have you thought about how deployment works since it needs to use the bytecode?

cellog:
> So that works, but have you thought about how deployment works since it needs to use the bytecode?

I have. It will work as-is with this PR. Unfortunately, if `deploy.js` is modified to import the abi, it will pull it into the service worker bundle. With the compressed version, we are at 363k for the sw.min.js. This is still too big, but is a lot better than 1.5MB.

Substantial changes will need to be made to the unlock-js output to make it tree-shakeable, namely providing an es module (`module` field in package.json) entrypoint.

cellog:
> > So that works, but have you thought about how deployment works since it needs to use the bytecode?
> 
> I have. It will work as-is with this PR. Unfortunately, if `deploy.js` is modified to import the abi, it will pull it into the service worker bundle. With the compressed version, we are at 363k for the sw.min.js. This is still too big, but is a lot better than 1.5MB.
> 
> Substantial changes will need to be made to the unlock-js output to make it tree-shakeable, namely providing an es module (`module` field in package.json) entrypoint.

I did some (much) more investigating, and figured out a way to make tree shaking work with unlock-js. This very good news means we can modify deploy to accept a version instead of a contract, and then it can use the abi packages. Yay!

It will, however, make sense for that to be separate from this PR. PTAL when you get a chance @julien51

cellog:
@hardlydifficult would you mind taking a look at the generated signatures in `abis.js` to see if there is anything off from what the solidity signature looks like? no need to yes/no on this, just looking for a comment on that specific aspect

hardlydifficult:
Seems okay.  

By storing the signature with variable names, you will need to strip the names before taking the hash to get the correct Solidity signature.  That said, this should be fine cause Web3 and presumably ethers, does that for you.

The bytecode hash is a hash of `deployedBytecode` right?

cellog:
yeah, the bytecode hash is a hash of `deployedBytecode`, because I figured that is what we would actually get back from an `eth_getCode` call (which seems to bear out in practice).

As for the signature, I originally had the variable names stripped out, but it turns out that ethers needs those to generate proper signatures, especially for event parameters and encoding inputs, so I put them back in. Thanks!

Issue #3102, opened 2019-05-09
Create a general-purpose password prompt


cnasc:
Part of #2111

Once a user is "logged in," we'll still need to prompt them for a password to decrypt their private key (to buy a key, change their password, etc.)

I propose using a component like the wallet check overlay to pop up and get a password from the user. In fact, we could generalize the overlay to handle various prompts.

PR #3136, opened 2019-05-10
WalletCheckOverlay now uses new modal implementation

cnasc:
# Description

Another step to #3102. This PR looks bigger than it is. The real action is in `FullScreenModals.tsx`, most of the rest is just updating tests and stories that relied on the wallet status-specific previous implementation.

A follow-up PR will use this infrastructure to implement the password prompt.

# Checklist:


unlock-app/src/middlewares/walletMiddleware.js
```diff
@@ -77,7 +73,7 @@ const walletMiddleware = config => {
       (transactionHash, from, to, input, type, status) => {
         // At this point we know that a wallet was found, because a new transaction
         // cannot be created without it
-        dispatch(gotWallet())
+        dispatch(dismissWalletCheck())
```
cnasc: The original implementation made a distinction between manually dismissed wallet check and automatically dismissed wallet check. The modal doesn't need to know about that, so it just has launch and dismiss actions.

unlock-app/src/pages/_app.js
```diff
@@ -8,7 +8,7 @@ import { createUnlockStore } from '../createUnlockStore'
 import GlobalStyle from '../theme/globalStyle'
 import { ConfigContext } from '../utils/withConfig'
 
-import WalletCheckOverlay from '../components/interface/FullScreenModals'
+import FullScreenModal from '../components/interface/FullScreenModals'
```
cnasc: This is just a name change to keep it consistent with its more general purpose nature now

PR #3143, opened 2019-05-10
Store encrypted private key in Redux

cnasc:
# Description

A piece of #2111. Once we've gotten the user's encrypted private key, it makes sense to cache it in Redux instead of making repeated network calls for it.

Copying lots of people on this, I'd definitely like any thoughts that occur based on my comments below.

# Checklist:

### DIFF
--- a/unlock-app/src/actions/user.ts
+++ b/unlock-app/src/actions/user.ts
@@ -1,3 +1,5 @@
+import { EncryptedPrivateKey } from '../unlockTypes' // eslint-disable-line no-unused-vars
+
 export const LOGIN_CREDENTIALS = 'login/GOT_CREDENTIALS'
 export const LOGIN_SUCCEEDED = 'login/SUCCESS'
 export const LOGIN_FAILED = 'login/FAILED'
@@ -7,6 +9,8 @@ export const SIGNUP_FAILED = 'signup/FAILED'
 export const SIGNUP_SUCCEEDED = 'signup/SUCCESS'
 export const CHANGE_PASSWORD = 'password/CHANGE'
 export const GOT_PASSWORD = 'userCredentials/PASSWORD'
+export const SET_ENCRYPTED_PRIVATE_KEY =
+  'userCredentials/SET_ENCRYPTED_PRIVATE_KEY'
 
 export interface Credentials {
   emailAddress: string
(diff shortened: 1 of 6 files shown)


unlock-app/src/actions/user.ts
```diff
@@ -62,3 +66,12 @@ export const gotPassword = (password: string) => ({
   type: GOT_PASSWORD,
   password,
 })
+
+// This should be dispatched along with setAccount in the storage middleware
+// when a user logs in. This provides a bit of a timing concern, since
+// SET_ACCOUNT usually wipes out all the state. So SET_ACCOUNT must be sent
+// first.
+export const setEncryptedPrivateKey = (key: EncryptedPrivateKey) => ({
```
cnasc: Take note of the comment above. Not sure what the best way to enforce this is.

unlock-app/src/unlockTypes.ts
```diff
@@ -69,3 +69,11 @@ export interface Action {
 
 // TODO: Use this where we have TS files
 export type Dispatch = (action: Action) => any
+
+// This is currrently the way ethers checks the keystore format. TODO:
```
cnasc: See: https://github.com/ethers-io/ethers.js/blob/master/src.ts/utils/json-wallet.ts#L17

We can make this type more concrete, but I'm not certain how solid the other properties of the keystore object are. At any rate, this fulfills what it's meant to, because it should prevent anyone from accidentally using a decrypted key as an encrypted one.

unlock-app/src/actions/user.ts
cellog: What if you also passed the account along with the key? This way you can index them by account, and if the account changes to the account using the key, we don't blow it away?

cnasc: I see, so we’d use SET_ACCOUNT for crypto users, and this one for those with accounts? I’m amenable to that, though it may require some code changes in consumers of the Redux store

cellog: No I mean we pass in the account to this action. But, more to the point, do we need to reset this when account changes? If we get a password, that means we only have one possibility, our account. So maybe we just remove the reset?

cnasc:
For the time being I'm going to merge this as-is with the understanding that we should come back and reevaluate this approach once we've seen how everything works together so that in the meantime we can move forward.

Issue #3140, opened 2019-05-10
Ad Remover Paywall: Data Iframe 


cellog:
This component includes the cache handler and blockchain handlers. It represents the Asserter component of the Asserter-Dispatcher-Checkout triangle

PR #3155, opened 2019-05-12
new caching library to ensure speedy return visits for paywall

cellog:
# Description

As a part of the Forbes ad remover paywall, caching will be important. This implements a caching library. There is a non-trivial chance that we may need to move to a different kind of backend, and all of the other possibilities (`Cache`, `indexeddb`) are async, so this makes the `localStorage` driver async.

Otherwise it's pretty standard cache stuff, with `get`, `put` and `clear` which can be done by user, or by sub-key (like keys, for instance)

Because localStorage can be changed in any open tab, this also provides a facility to listen for changes to the localStorage key for a specific user on a specific network.

# Issues

Refs #3140

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/cache/index.js
+++ b/paywall/src/data-iframe/cache/index.js
@@ -1,5 +1,5 @@
 /**
- * All storrage drivers are async
+ * All storage drivers are async
  */
 export {
   get,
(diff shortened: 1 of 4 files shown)


paywall/src/data-iframe/cache/index.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * All storrage drivers are async
```
cnasc:
```suggestion
 * All storage drivers are async
```

PR #3156, opened 2019-05-12
PROOF-OF-CONCEPT seeking feedback: Web3ProxyProvider

cellog:
While driving today, I was thinking about how we were going to handle sending transactions by constructing them in the iframe and then postMessaging them to the main window.

My initial thought was that we could take advantage of ethers.io `ethers.utils.Interface` (https://docs.ethers.io/ethers.js/html/api-advanced.html#interface) to construct the function call, but this would require substantial rewrites to `walletService` (every single transaction-based method call would need to be touched in some way).

Then, I started thinking about how we are already calling `eth_getAccounts` in the main window, and how we would need to mimic this to pass in `eth_sendTransaction`. Then it occurred to me: what we are trying to do is to intercept the method `sendTransaction` in a provider, and instead of trying to call with the existing provider, we need to instead postMessage it to the main window, call it there, and then send back the result. We need some kind of plugin to providers, in order to catch the call right at the moment it is properly encoded, and then send that to the main window and return the result.

Going further along this path, I realized that a plugin system for ethers already exists! *Providers* are plugins! In fact, we can take advantage of `ethers`'s compatibility with existing web3 wallets to make a very simple provider without needing to implement signing or other complexities. We can simply create a class with 1 method: `sendAsync`, which accepts the already-constructed transaction, and passes it to the main window, and then returns the value.

This is in fact the only requirement for a web3 provider in the `ethers.providers.Web3Provider` class (https://docs.ethers.io/ethers.js/html/api-providers.html#web3provider-inherits-from-jsonrpcprovider) *note: the docs are incorrect, the method needs to accept 2 parameters, `request` and `callback`. `request` contains `method` and `params`*

`walletService` already accepts a web3-like provider, and wraps it in `Web3Provider`, this is how we interface with Metamask and company right now.

So, to communicate with the main window, we really only need to pass in our new proxy as if it were a web3 wallet like Metamask! That's what this draft PR proposes.

Before I go deep and write tests, I want to get an OK on the basic concept.

Here's how it works:

1. at the beginning, the `Web3ProxyProvider` is initialized in an iframe and passed the iframe's `window` object
2. it sets up a listener for `postMessage` events
3. it sends a `ready` event to the main window
4. the main window hears it in the `web3Proxy` file, and connects to the wallet, if any, and sends back information on the wallet (initially just whether the wallet is metamask, which we need to do signing properly)
5. when a user triggers a transaction, let's say `purchaseFor` to buy a key on a lock, ethers constructs the ready-to-go object with `method` and `params`
6. we intercept it in our `sendAsync` function, create a unique ID to track the transaction (they can come back in any order!), and store the callback, indexed by its unique ID
7. we postMessage the transaction to the main window, where it catches it, and passes it to the wallet
8. on return, the main window packages up the id, the error and the result and sends them back up
9. the event listener set up in the constructor of `Web3ProxyProvider` hears this event, and uses the id to find the callback.
10. as a final step, it re-sets the id to the value used by ethers, 42. (when ethers finally implements incrementing ids in v5, we'll use that instead)

This has several important benefits:

1. transactions will always show as initiating from the page the user is actually viewing, instead of from the iframe, which will almost certainly be an unfamiliar URL.
2. we won't need *any* of the baloney I have written for redirecting, passing the account around, etc. The paywall can just work, as-is
3. when we write our own unlock provider, we can drop it in instantly to replace this one.

Let's talk about these ideas on Monday when there is time!


paywall/src/paywall-builder/constants.js
```diff
@@ -7,3 +7,5 @@ export const POST_MESSAGE_GET_PESSIMISTIC = 'pessimistic'
 export const POST_MESSAGE_READY = 'ready'
 export const POST_MESSAGE_CONFIG = 'config'
 export const POST_MESSAGE_ACCOUNT = 'account'
+export const POST_MESSAGE_WEB3 = 'web3'
+export const POST_MESSAGE_WALLET_INFO = 'walletinfo'
```
julien51: `walletInfo` ;)

cellog:
cool, discussion had, closing for now

PR #3171, opened 2019-05-13
Locksmith updated with delegated purchasing

akeem:
# Description

The code has been updated to support purchasing on a users
behalf. At the moment this is written with Ethereum based locks
in mind but shall be updated for others as well.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/__tests__/controllers/purchaseController.test.ts
+++ b/locksmith/__tests__/controllers/purchaseController.test.ts
@@ -1,5 +1,6 @@
 import request from 'supertest'
 import sigUtil from 'eth-sig-util'
+import { Web3Service } from '@unlock-protocol/unlock-js'
 
 const ethJsUtil = require('ethereumjs-util')
 const app = require('../../src/app')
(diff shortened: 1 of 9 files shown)


akeem:
While not standard, this change includes an update to package.json in addition to code. The cut over from the older version of the library create a chicken & egg situation

akeem:
We also have a skipped test in this change for now, it will require a bit more wrangling and will be revisited in short order

locksmith/__tests__/controllers/purchaseController.test.ts
```diff
@@ -57,6 +51,18 @@ describe('Purchase Controller', () => {
     })
   })
 
+  beforeEach(() => {
+    jest.mock('../../src/payment/paymentProcessor', () => {
```
cellog: https://jestjs.io/docs/en/jest-object#jestgenmockfrommodulemodulename plus https://jestjs.io/docs/en/mock-function-api.html#mockfnmockimplementationoncefn may be a better way to do this that fixes the skipped test

locksmith/src/payment/paymentProcessor.ts
```diff
@@ -90,21 +90,24 @@ export class PaymentProcessor {
     return Object.values(itemizedPrice).reduce((a, b) => a + b)
   }
 
-  initiatePurchase(
-    purchaser: ethereumAddress,
+  async initiatePurchase(
+    purchaser: ethereumAddress /** this is the managed user/buyer */,
```
cnasc: I think `purchaser` would make more sense as `recipient`

PR #3175, opened 2019-05-13
Updates lock creation interface to support other backing contracts

akeem:
# Description

Updates lock creation interface to support other backing contracts

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/v02/createLock.test.js
+++ b/unlock-js/src/__tests__/v02/createLock.test.js
@@ -92,7 +92,7 @@ describe('v02', () => {
 
     describe('when providing the address of a denominating currency contract', () => {
       it('should invoke _handleMethodCall with the right params', async () => {
-        expect.assertions(2)
+        expect.assertions(0)
 
         await nockBeforeEach()
 
@@ -113,23 +113,7 @@ describe('v02', () => {
 
         setupSuccess()
 
-        walletService._handleMethodCall = jest.fn(() =>
-          Promise.resolve(transaction.hash)
-        )
-        const mock = walletService._handleMethodCall
-
         await walletService.createLock(lock, owner, testERC20ContractAddress)
-
-        expect(mock).toHaveBeenCalledWith(
-          expect.any(Promise),
-          TransactionTypes.LOCK_CREATION
-        )
-
-        // verify that the promise passed to _handleMethodCall actually resolves
-        // to the result the chain returns from a sendTransaction call to createLock
-        const result = await mock.mock.calls[0][0]
-        await result.wait()
-        expect(result).toEqual(transactionResult)
         await nock.resolveWhenAllNocksUsed()
       })
     })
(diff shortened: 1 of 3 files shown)


unlock-js/src/__tests__/v02/createLock.test.js
```diff
@@ -60,30 +62,76 @@ describe('v02', () => {
...
+          lock.maxNumberOfKeys
+        )
+
+        transaction = testTransaction
+        transactionResult = testTransactionResult
+        setupSuccess = success
+
+        setupSuccess()
+
+        walletService._handleMethodCall = jest.fn(() =>
```
cellog: this test should not mock `_handleMethodCall`, just remove the mock and then the code below

unlock-js/src/__tests__/v02/createLock.test.js
```diff
@@ -60,30 +62,76 @@ describe('v02', () => {
...
+          Promise.resolve(transaction.hash)
+        )
+        const mock = walletService._handleMethodCall
+
+        await walletService.createLock(lock, owner, testERC20ContractAddress)
+
+        expect(mock).toHaveBeenCalledWith(
```
cellog: remove this

unlock-js/src/__tests__/v02/createLock.test.js
```diff
@@ -60,30 +62,76 @@ describe('v02', () => {
...
+          expect.any(Promise),
+          TransactionTypes.LOCK_CREATION
+        )
+
+        // verify that the promise passed to _handleMethodCall actually resolves
```
cellog: remove this:

```js
        // verify that the promise passed to _handleMethodCall actually resolves
        // to the result the chain returns from a sendTransaction call to createLock
        const result = await mock.mock.calls[0][0]
        await result.wait()
        expect(result).toEqual(transactionResult)
```

The test should then be changed to `expect.assertions(0)`

The key line which must be present to actually test the behavior is this one:

```js
await nock.resolveWhenAllNocksUsed()
```

With this line, if the JSON-RPC sent to the chain does not match what we expected, the test will fail. So, we will see (for instance) `ZERO` encoded in the data instead of the ERC20 address, the nocks will not match, and the test will fail because there is an unused nock. As long as you add a comment explaining that the test failure happens this way for future eyes, then all is good.

unlock-js/src/__tests__/v02/createLock.test.js
```diff
@@ -60,30 +62,60 @@ describe('v02', () => {
...
+        const result = await mock.mock.calls[0][0]
+        await result.wait()
+        expect(result).toEqual(transactionResult)
+        await nock.resolveWhenAllNocksUsed()
+      })
+    })
+
+    describe('when providing the address of a denominating currency contract', () => {
+      it('should invoke _handleMethodCall with the right params', async () => {
+        expect.assertions(0)
```
cellog: maybe add a comment explaining how the last line tests this, so assertions are not needed

PR #3178, opened 2019-05-13
Validation page for events

benwerd:
# Description

Loads event, verifies signature, and displays a checkmark. This PR does not include any new designs. This will need to be heavily tested on staging.

# Checklist:


tickets/package.json
```diff
@@ -35,7 +35,7 @@
     "@types/react-redux": "7.0.8",
     "@types/storybook__react": "4.0.1",
     "@types/styled-components": "4.1.14",
-    "@unlock-protocol/unlock-js": "0.0.26",
+    "@unlock-protocol/unlock-js": "0.0.28",
```
julien51: Could you split that into its own pr please?

tickets/src/_server.js
```diff
@@ -21,6 +21,11 @@ function _server(port, dev) {
           } else if (path === 'event') {
             const params = route('/event/:lockAddress')(pathname)
             app.render(req, res, '/event', Object.assign(params, query))
+          } else if (path === 'checkin') {
```
julien51: make sure you add the corresponding changes in `_redirect` for Netlify?

tickets/src/components/content/EventVerify.js
```diff
@@ -51,7 +67,13 @@ export const mapDispatchToProps = dispatch => ({
 })
 
 export const mapStateToProps = ({ router, locks, account, event }) => {
-  const { lockAddress } = lockRoute(router.location.pathname)
```
cellog: This is a rather brittle change, can we instead update the regexp and `lockRoute`?

tickets/src/__tests__/middlewares/walletMiddleware.test.js
```diff
@@ -380,15 +380,16 @@ describe('Wallet middleware', () => {
       signedAddress: 'encrypted sig',
     }
 
-    mockWalletService.recoverAccountFromSignedData = jest.fn(
-      data => data.message.address.publicKey
+    mockWalletService.recoverAccountFromSignedData = jest.fn((data, sa, cb) =>
```
cellog: this is a good test. We also need a separate test that verifies that the real `walletService.recoverAccountFromSignedData` actually does what we expect it, in order to catch API regressions. Otherwise, should a change to `unlock-js` break this, we won't know until it's deployed to staging. The test would need to mock out the json-rpc call using `nock`, I can definitely help with this. We would borrow `nockHelper` from `unlock-js`

The test would also verify that our assumption that `recoverAccountFromSignedData` calls a callback is correct

Issue #3206, opened 2019-05-15
Ad remove paywall: Blockchain handler


cellog:
This is the script that communicates with the blockchain via walletService and web3Service

cellog:
fully implemented

PR #3215, opened 2019-05-15
generic pollForChanges helper to be used in account polling

cellog:
# Description

Initially, I coded all the account polling in this function, but it's very difficult to test an infinite loop that does something. This way, the infinite part can be safely mocked out in the account polling test, and the source for account polling becomes declarative as in:

```js

export async function pollForAccountChange(
  walletService,
  web3Service,
  onAccountChange = () => {}
) {
  await ensureWalletReady()

  pollForChanges(
    async () => await walletService.getAccount() /* getFunc */,
    (before, after) => before !== after /* hasValueChanged */,
    () => true /* continuePolling */,
    async newAccount => {
      // only called when account has changed
      /* changeListener */
      setAccount(newAccount)
      setAccountBalance(
        newAccount ? await web3Service.getAddressBalance(newAccount) : 0
      )
      onAccountChange(account, accountBalance)
    }
  )
}
```

Because there is a continuePolling option, we can also use it to poll for things that have a limit, like transaction confirmations.

# Issues

Refs #3206

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/blockchainHandler/pollForChanges.js
+++ b/paywall/src/data-iframe/blockchainHandler/pollForChanges.js
@@ -15,23 +15,31 @@ import { delayPromise } from '../../utils/promises'
  */
 
 /**
+ * @callback continuePolling
+ * @param {*} currentValue
+ * @returns {bool} true if the polling should continue
+ */
+
+/**
  * Poll for changes every 5 seconds in something, and call the changeListener when it changes
  *
- * @param {Function} getFunc should return the current value(s) we are polling for changes
+ * @param {Function} getCurrentValue should return the current value(s) we are polling for changes
  * @param {compareFunc} hasValueChanged
+ * @param {continuePolling} continuePolling
  * @param {changeListener} changeListener
  * @param {int} delay
  */
 export default async function pollForChanges(
-  getFunc,
+  getCurrentValue,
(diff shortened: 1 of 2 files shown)


paywall/src/data-iframe/blockchainHandler/pollForChanges.js
```diff
@@ -0,0 +1,48 @@
...
+  hasValueChanged,
+  continuePolling,
+  changeListener,
+  delay
+) {
+  let before = await getCurrentValue()
+  while (await continuePolling(before)) {
+    await delayPromise(delay)
+    const after = await getCurrentValue()
+    if (await hasValueChanged(before, after)) {
```
julien51: Does this need to be async? I would assume that the before and after values are actually resolved and then this can be sync?

cellog: In most cases, no it doesn't. However, `await` is not just for async, you can await any value and it will just pretend you wrapped it with `Promise.resolve()` (see the end of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#Description). So, by adding the `await` before this, we lose nothing, but gain flexibility should there be a use case where we need to do some async processing to do the comparison.

PR #3220, opened 2019-05-15
Provides tracing ability for local development environment

akeem:
# Description

* Add default instrumentation 
* Add jaeger to development docker compose file to get things started

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/docker/docker-compose.development.yml
+++ b/docker/docker-compose.development.yml
@@ -7,7 +7,7 @@ services:
        dockerfile: ganache.dockerfile
      ports:
        - 8545:8545
-   jager:
+   jaeger:
       restart: always
       image: jaegertracing/all-in-one:1.6
       ports:
--- a/docker/docker-compose.development.yml
+++ b/docker/docker-compose.development.yml
@@ -7,3 +7,14 @@ services:
        dockerfile: ganache.dockerfile
      ports:
        - 8545:8545
+   jager:
+      restart: always
+      image: jaegertracing/all-in-one:1.6
+      ports:
+        - '5775:5775/udp'
+        - '6831:6831/udp'
+        - '6832:6832/udp'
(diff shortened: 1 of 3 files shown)


akeem:
This is to get the ball rolling, instrumentation the rest of the application will be done a piece meal effort as instrumenting the application will take a while

locksmith/src/app.ts
```diff
@@ -1,6 +1,17 @@
...
 import express from 'express'
+import tracing from '@opencensus/nodejs'
+import { JaegerTraceExporter } from '@opencensus/exporter-jaeger'
+
+const env = process.env.NODE_ENV || 'development'
+const config = require('../config/config')[env]
+
+const exporter = new JaegerTraceExporter(config.jaeger)
+
+if (env != 'test') {
```
akeem: There maybe a slightly better place for this, but reason for being here remains the same. The tracing library intercepts the HTTP library in order to provide details on requests, timings, etc. 

Our test using the testing library `supertest` in order to provide a reasonable way  to assert on request being made during the test. 

To be expected, these together create unexpected results

docker/docker-compose.development.yml
```diff
@@ -6,4 +6,15 @@ services:
        context: ./development
        dockerfile: ganache.dockerfile
      ports:
-       - 8545:8545
\ No newline at end of file
+       - 8545:8545
+   jager:
```
cnasc:
```suggestion
   jaeger:
```

locksmith/src/app.ts
cnasc: I guess there's no real downside to having this not run in the test environment, right?

akeem: I don' think so, but it was a bit weird and unexpected to see this behavior

PR #3243, opened 2019-05-17
Update @types/styled-components

cnasc:
# Description

This PR updates the types for `styled-components` in the apps that currently depends on it.

Note: `unlock-protocol.com` doesn't currently use typescript in any form, except insofar as the build process is copied from an app that does use it. We may want to clean that up, since there are packages in there that we won't need.

`tickets` uses a little bit of TS, so the same might go for that. For the moment I'm keeping the status quo, however.

# Issues

Fixes #3182 

# Checklist:

### DIFF
--- a/tickets/package-lock.json
+++ b/tickets/package-lock.json
@@ -1951,9 +1951,9 @@
       }
     },
     "@types/styled-components": {
-      "version": "4.1.14",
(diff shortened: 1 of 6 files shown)


cnasc:
Whoops, this accidentally set the caret, and I think we don't want that. I will resolve that before merging.

PR #3249, opened 2019-05-17
Start of extracting details for use of environment stand up

akeem:
# Description

A step in getting the standup in a place where we can begin to realistically us it.  The following has occurred in the change:

* Adds token transfer as part of the setup
* use of .env file with use of ENV variable that matches the current approach for token transfer
* extracting details to env file

# Issues

Fixes #
Refs #

# Checklist:


docker/development/.env
```diff
@@ -0,0 +1,8 @@
+BOOTSTRAP_AMOUNT=15.0
+HTTP_PROVIDER_HOST=127.0.0.1
+HTTP_PROVIDER_PORT=8545
+ETHEREUM_ADDRESS=0xE7de59378f3496a5434BBe998c28B71A72888f18
+ERC20_TOKEN_RECIPIENT=0xe29ec42f0b620b1c9a716f79a02e9dc5a5f5f98a
+BOOTSTRAP_SENDER_PRIVATE_KEY=0xfd8abdd241b9e7679e3ef88f05b31545816d6fbcaf11e86ebd5a57ba281ce229
```
julien51: Do we need this? Ganache should have these accounts "unlocked" on the node which means we can trigger transactions from them without their private keys?

akeem: At a quick glance, it didn't look like ethers supported this but it is probably an error my part. I'll look again to confirm

akeem:
@julien51 we should be able to reference the other file from the compose file and cut over to that file but want to make sure that we are cool with this before I make any changes that might break anyone's dev env.

PR #3252, opened 2019-05-17
moving key purchasing logic inside NoKeyLock to improve separtion of concerns an increase re-usaebility

julien51:
# Description

Pretty much everything in the title ;)
Mostly shuffling things around so I can re-use them better in the checkout window!

# Checklist:

### DIFF
--- a/paywall/src/components/lock/NoKeyLock.js
+++ b/paywall/src/components/lock/NoKeyLock.js
@@ -5,64 +5,77 @@ import UnlockPropTypes from '../../propTypes'
 import { LockWrapper, LockHeader, LockBody, LockFooter } from './LockStyles'
 import BalanceProvider from '../helpers/BalanceProvider'
 import Duration from '../helpers/Duration'
+import { UNLIMITED_KEYS_COUNT } from '../../constants'
 
 export const NoKeyLock = ({
+  account,
   lock,
   disabled,
   purchaseKey,
   lockKey,
-  soldOut,
-  tooExpensive,
-}) => (
-  <Wrapper
-    lock={lock}
-    disabled={disabled}
-    onClick={() => {
-      !disabled && purchaseKey(lockKey)
-    }}
-  >
-    <LockHeader>{lock.name}</LockHeader>
-    <BalanceProvider
-      amount={lock.keyPrice}
-      render={(ethPrice, fiatPrice) => (
-        <div>
-          <Body disabled={disabled}>
-            <EthPrice>{ethPrice} Eth</EthPrice>
-            <div>
-              <FiatPrice>${fiatPrice}</FiatPrice>
-              <Separator> | </Separator>
-              <ExpirationDuration>
-                <Duration seconds={lock.expirationDuration} round />
-              </ExpirationDuration>
-            </div>
-            <Footer>
-              {soldOut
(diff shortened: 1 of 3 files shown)


paywall/src/components/lock/NoKeyLock.js
```diff
@@ -5,64 +5,74 @@ import UnlockPropTypes from '../../propTypes'
...
+              <div>
+                <FiatPrice>${fiatPrice}</FiatPrice>
+                <Separator> | </Separator>
+                <ExpirationDuration>
+                  <Duration seconds={lock.expirationDuration} round />
+                </ExpirationDuration>
+              </div>
+              <Footer>
+                {soldOut
+                  ? 'Sold Out'
```
cellog: might be an opportunity to pull this out to the top of the component

paywall/src/components/lock/NoKeyLock.js
```diff
@@ -5,64 +5,74 @@ import UnlockPropTypes from '../../propTypes'
...
+  // When the lock is not disabled for other reasons (pending key on
+  // other lock...), we need to ensure that the lock is disabled
+  // when the lock is sold out or too expensive for the current account
+  const tooExpensive =
+    account && parseFloat(account.balance) <= parseFloat(lock.keyPrice)
+
+  return (
+    <Wrapper
+      lock={lock}
+      disabled={disabled || tooExpensive || soldOut}
```
cellog: another opportunity - perhaps we define this above so all our computation is above the render? This is not super important, but since you're refactoring anyways :)

paywall/src/components/lock/NoKeyLock.js
julien51: not sure I understand. This was _actually_ pulled from the top component in here... because here is the only place where it makes sense to show?

paywall/src/components/lock/NoKeyLock.js
julien51: It's actually more complicated than that.
This component should have the logic of `tooExpensive` and `soldOut`... but a lock can be disabled for other reasons (especially if there is another lock with a pending key for example)

cellog: sorry! I was unclear. I meant something like:

```js
const disableView = disabled || tooExpensive || soldOut

...

    <Wrapper
      lock={lock}
      disabled={disableView}
```

paywall/src/components/lock/NoKeyLock.js
cellog: again, I was unclear, I mean:

```js
let footerMessage
if (soldOut) {
  footerMessage = 'Sold Out'
} else if (tooExpensive) {
  footerMessage = 'Insufficient funds'
} else {
  footerMessage = 'Purchase'
}
```

cellog: and then use `footerMessage` in the component stuff

paywall/src/components/lock/NoKeyLock.js
julien51: oh, ok? Is that really better?

cellog: it isn't better per se, just congregates all the logic in the top, and the display in the bottom, for readability

Issue #3246, opened 2019-05-17
Paywall does not take into account pending transactions


julien51:
**Describe the bug**
When purchasing a key on the paywall, if the user immediately refreshes, will the transaction has not been mined yet, the paywall locks again, rather than showing the the "optimistic unlocked banner"

**To Reproduce**
1. Go to http://ouvre-boite.com/members/
2. Click on 'Purchase'
3. Refresh
4. See the paywall

cellog:
this does not happen with current staging, see https://staging-paywall.unlock-protocol.com/demo/0x39B3D5666fab11D609BaF0e953cef469815f7Fbf

cellog:
On ouvre-boite, it fails while parsing the inputs of the retrieved transaction. What contract version was the lock released under?

cellog:
this is the bug we saw the other day: the transaction fails to retrieve from the chain and so it tries to retrieve from defaults

cellog:
The problem is that the transaction is still pending, see https://etherscan.io/tx/0x272e423ef7e6bc289f6d253be2ce8b6ac94a108b6605a946e62470d66eca874e

Thus, retrieving it from the read-only alchemy provider returns null because it does not see it mined yet.

We need to catch this inside web3Service. Basically, if we refresh, then the defaults are gone and boom, byebye

PR #3253, opened 2019-05-17
respond to ADD_TRANSACTION if it looks like a key purchase

cellog:
# Description

This is the last piece of the fix for #3246. When locksmith passes a pending transaction to the paywall, and the user has refreshed the page, we don't have enough information to do anything with it. `web3Service.getTransaction` will return null, and the page displays "your key is expired" or "purchase a key."

This PR adds some "stuffing" of the action if it looks like a key purchase (from us to the lock on the page). The stuffing is the transaction type, lock and key, which is enough for the UI to respond to it. We also have to make sure the transaction shows up in the key transaction list, which is at the bottom of the file.

Because of this change, it now means we handle `ADD_TRANSACTION`, `NEW_TRANSACTION`, and `UPDATE_TRANSACTION` the same way, so the code refactors the bottom of the file to do this, abstracting the small differences between the structure of each action.

After this PR, a release of unlock-js followed by upgrade to paywall will fix the bug.

# Issues

Refs #3246 

# Checklist:

### DIFF
--- a/paywall/src/middlewares/web3Middleware.js
+++ b/paywall/src/middlewares/web3Middleware.js
@@ -19,6 +19,7 @@ import { lockRoute } from '../utils/routes'
 
 import { SET_PROVIDER } from '../actions/provider'
 import { SET_NETWORK } from '../actions/network'
+import { TRANSACTION_TYPES } from '../constants'
 
 // This middleware listen to redux events and invokes the web3Service API.
 // It also listen to events from web3Service and dispatches corresponding actions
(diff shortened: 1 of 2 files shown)


julien51:
> This PR adds some "stuffing" of the action if it looks like a key purchase (from us to the lock on the page).

Well that is weird because `getTransaction` should NOT return `null`? If the transaction is pending it should actually return some info from the mempool indicating that the transaction has not been mined yet. Web3Service actually has the logic to look at the `data` to identify that it is for sure an actual keey purchase.

Maybe we broke Web3Service?

paywall/src/middlewares/web3Middleware.js
```diff
@@ -105,6 +106,24 @@ const web3Middleware = config => ({ getState, dispatch }) => {
...
+            location: { pathname, hash },
+          },
+        } = getState()
+        const { lockAddress } = lockRoute(pathname + hash)
+        const accountAddress = account && account.address
+        const { to, from } = action.transaction
+        if (from === accountAddress && to === lockAddress) {
+          // this is probably a key purchase transaction, so let's make sure we know it is in the UI
+          // pending key transactions will never see this detail otherwise
+          action.transaction.type = TRANSACTION_TYPES.KEY_PURCHASE
```
julien51: Nah, this is not right we should not do guesswork here. Web3Service **has** the ability to retrieve that info (and used to!), so let's make sure it does.

cellog:
Julien the reason that it has to be here is because the pending transaction cannot be retrieved from the read only provider because it is still pending. We have defaults for this exact case, but the defaults don't exist when the user refreshes. It could be fixed by caching the defaults in localStorage, but that is riskier imo

cellog:
As an alternative, we can leave this unfixed until the work I'm doing for the ad remover paywall is live, because that *will* in fact cache the transaction value, but safely

PR #3264, opened 2019-05-19
key status calculator for the new ad remover blockchain handler

cellog:
# Description

This PR makes it possible to add 2 important fields to all keys: `status` and `confirmations`. Ultimately this will remove all calculation from the checkout and paywall frontend

# Issues

Refs #3206 

# Checklist:


paywall/src/data-iframe/blockchainHandler/keyStatus.js
```diff
@@ -0,0 +1,63 @@
...
+  const account = getAccount()
+  const newKeys = {}
+
+  const transactionsByLock = locks.reduce((indexedTransactions, lock) => {
+    // get the key purchase transactions sorted in reverse chronological order
+    const keyPurchaseTransactions = Object.values(transactions)
+      .filter(transaction => {
+        return transaction.from === account && transaction.to === lock
+      })
+      .sort((a, b) => (a.blockNumber > b.blockNumber ? -1 : 1))
```
julien51: Pending transactions do not have a blockNumber... I want to make sure we handle this case correctly.

cellog: In `web3Service._getSubmittedTransaction`, we do:

```js

    return this.emit('transaction.updated', transactionHash, {
      status: 'submitted',
      confirmations: 0,
      blockNumber: Number.MAX_SAFE_INTEGER, // Asign the largest block number for sorting purposes
    })
```

which covers this situation. pending transactions inherit this huge block number, which finally gets overwritten with the correct one when it gets mined.

cellog:
@julien51 PTAL. I added several more tests, several comments for edge cases, and fixed a couple small bugs

paywall/src/data-iframe/blockchainHandler/keyStatus.js
julien51: Ha! thanks for looking this up.

paywall/src/data-iframe/blockchainHandler/keyStatus.js
```diff
@@ -0,0 +1,88 @@
+import { getAccount } from './account'
+
+/**
+ * Determine whether a key is valid based solely on its expiration
+ *
+ * @oaram {key} key a key object, which must contain an expiration timestamp in seconds
+ * @returns bool
+ */
+export function validKey(key) {
```
julien51: Nit: would you mind using a more explicit `isValidKey` name?

paywall/src/data-iframe/blockchainHandler/keyStatus.js
```diff
@@ -0,0 +1,88 @@
...
+      // its timer only actually starts when the transaction is first confirmed
+      // so users get the pending time for free
+      return lastTransaction.status || 'none'
+  }
+}
+
+/**
+ * Construct the transactions field for each key
+ * @param {object} keys keys, indexed by their lock/owner ID (not the same as smart contract ID)
+ * @param {object} transactions transactions, indexed by hash (submitted transaction is always indexed under "submitted")
```
julien51: Not sure I understand that? submitted transactions should have a hash too?

cellog: There are 3 stages of transaction that we have to manage:

1. submitted - the transaction doesn't exist in the mempool yet, has no hash or block number
2. pending - the transaction exists in the mempool, has a hash, and no block number
3. mined - has a hash and a block number

The first state means we have to index by something to put it in transactions, so I decided (and this is a change from the original work) to index by ``submitted-${lock}-${account}`` so that we can have multiple submitted transactions if the user is weird and does this in many open tabs all at the same time for different locks.

Once the transaction is in the mempool (which should be very fast, but occasionally alchemy doesn't), we have a hash and the submitted placeholder can be removed

Issue #2166, opened 2019-03-18
Allow users to change their passwords


benwerd:
While logged in, a user should be able to change their password. This will

* [ ] prompt for a new password (with validation field)
* [ ] on submission, re-encrypt the user's private key with the new password

cnasc:
I don't know what discussions we've had around password standards, but we may want to include a check against this as part of the validation https://www.troyhunt.com/ive-just-launched-pwned-passwords-version-2/

akeem:
I think this has been resolved for the most part, @cnasc ?

cnasc:
This is almost done, but I think we shouldn’t close it until we have an unlock provider working, since they need to sign the payload for the changed password

PR #3141, opened 2019-05-10
Introducing the Password Prompt

cnasc:
# Description

The final step of the password prompt goal. This introduces a component which prompts for a password and then dispatches it. This PR isn't as big as it seems: 600 lines of it are storyshots updates.

Once this is in place, we can use it for all the user interactions that require them to unlock (:rimshot:) their private keys.

image

# Issues

Fixes #3102 

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [x] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [x] My code follows the style guidelines of this project, including naming conventions
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] If my code adds or changes components, I have written corresponding stories with Storybook
- [x] I have performed a self-review of my own code
- [x] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #3272, opened 2019-05-20
Add a full-screen modal template for password reset prompt

cnasc:
# Description

Another chunk of #2166, this PR introduces the UI by which users can input the new password to re-encrypt their account. I'd like to think of this as just an initial iteration; in particular a next step would be to port over the (slightly) better password validation from account creation. (a next-next step would be to have a better standard password validation for all components to use.)

The component is hooked up to the `FullScreenModals` render path, but nothing yet dispatches an action that can trigger it.

image

# Checklist:

### DIFF
--- a/unlock-app/src/stories/interface/FullScreenModals.stories.js
+++ b/unlock-app/src/stories/interface/FullScreenModals.stories.js
@@ -1,12 +1,25 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
+import { Provider } from 'react-redux'
 import { FullScreenModal } from '../../components/interface/FullScreenModals'
 import {
   WalletCheck,
   PasswordPrompt,
+  ResetPasswordPrompt,
 } from '../../components/interface/modal-templates'
 import { KindOfModal } from '../../unlockTypes'
 import doNothing from '../../utils/doNothing'
+
+const emailState = {
+  userDetails: {
+    email: 'geoff@bitconnect.gov',
+  },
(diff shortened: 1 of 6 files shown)


unlock-app/src/stories/interface/FullScreenModals.stories.js
```diff
@@ -1,13 +1,27 @@
...
   WalletCheck,
   PasswordPrompt,
+  ResetPasswordPrompt,
 } from '../../components/interface/modal-templates'
 import { KindOfModal } from '../../unlockTypes'
 import doNothing from '../../utils/doNothing'
 
+const emailState = {
+  userDetails: {
+    email: 'geoff@bitconnect.gov',
```
julien51: Who's this guy????

julien51: I keep seeing him so now I have to ask!

cnasc:
> Did @smombartz provide the design? I am surprised the button look different than usually?

I don't think we've had any pro-work on the modal designs. These I kept from the password prompt modal as seen here: https://github.com/unlock-protocol/unlock/pull/3141

They are shared components, so swapping them out for something else should be easy

unlock-app/src/stories/interface/FullScreenModals.stories.js
cnasc: A fictional character from the Department of Bitconnect!

PR #3274, opened 2019-05-20
submitted transaction listener

cellog:
# Description

This is the first of 2 steps in the transaction handling flow. It is designed to be called declaratively for all transactions, regardless of prior state. If the newest transaction for the current key is in a later state (confirming, valid, etc.) then this will return immediately.

Otherwise, it waits for the `'transaction.new'` event from walletService and responds by updating the transaction and key state

Probably the final code will look something like:

```js
const submittedTransaction = Promise.all([
  purchaseKey({
    walletService,
    window,
    lockAddress,
    amountToSend
  }),
  submittedListener({
    lockAddress,
    existingTransactions,
    existingKeys,
    walletService,
    requiredConfirmations,
  })
])
// cache the submitted transaction
// use it as defaults for web3Service.getTransaction
```

# Issues

Refs #3206

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/blockchainHandler/keyStatus.js
+++ b/paywall/src/data-iframe/blockchainHandler/keyStatus.js
@@ -42,8 +42,7 @@ export function getKeyStatus(key, requiredConfirmations) {
 /**
  * Construct the transactions field for each key
  * @param {object} keys keys, indexed by their lock/owner ID (not the same as smart contract ID)
- * @param {object} transactions transactions, indexed by hash (submitted transaction is always
- *                              indexed under "submitted-${lock address}-${user account}")
+ * @param {object} transactions transactions, indexed by hash
  * @param {array} locks an array of lock addresses
  * @param {int} requiredConfirmations the number of confirmations needed to ensure a transaction went through
  */
--- a/paywall/src/data-iframe/blockchainHandler/purchaseKey/submittedListener.js
+++ b/paywall/src/data-iframe/blockchainHandler/purchaseKey/submittedListener.js
@@ -36,9 +36,14 @@ export default async function submittedListener({
   const network = getNetwork()
   const keyToPurchase = `${lockAddress}-${account}`
   const key = keys[keyToPurchase]
+  // key.status is one of:
+  // none, submitted, pending, confirming, valid, expired, failed
+  // we can only initiate a new purchase if the current key is not valid, and is
+  // not being purchased. These 3 statuses are the
   if (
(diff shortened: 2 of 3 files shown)


paywall/src/data-iframe/blockchainHandler/keyStatus.js
```diff
@@ -42,8 +42,7 @@ export function getKeyStatus(key, requiredConfirmations) {
 /**
  * Construct the transactions field for each key
  * @param {object} keys keys, indexed by their lock/owner ID (not the same as smart contract ID)
- * @param {object} transactions transactions, indexed by hash (submitted transaction is always
```
cellog: correct a misunderstanding. This is not directly related to this PR, but is indirectly related

paywall/src/data-iframe/blockchainHandler/purchaseKey/submittedListener.js
```diff
@@ -0,0 +1,83 @@
...
+    locks: [lockAddress],
+    requiredConfirmations,
+  })
+  const account = getAccount()
+  const network = getNetwork()
+  const keyToPurchase = `${lockAddress}-${account}`
+  const key = keys[keyToPurchase]
+  if (
+    (key.status !== 'none' && key.status !== 'expired') ||
+    key.status === 'failed'
```
julien51: I really think we should constant-ize the key status here because I have to say I am very blurry on what are the "allowed" values for a status. What does 'failed' mean actually? Is that the transaction status? If so is that the latest transaction or another one?

cellog: The key status is always linked to the most recent transaction.

cellog: You can see how the key status is generated in `keyStatus.js` which is in master

cellog: here: https://github.com/unlock-protocol/unlock/blob/master/paywall/src/data-iframe/blockchainHandler/keyStatus.js#L19

cellog: I pushed a commit, it doesn't "constant-ize" the statuses, as we don't have direct control over some of the statuses that come from transaction status. It does fix the logic inconsistency: we should allow purchasing a new key when the previous transaction failed.

cellog:
ready for a 2nd look @julien51

PR #3289, opened 2019-05-21
pass transaction data as defaults for transactions pulled from locksmith

### LINKED ISSUE #3246: Paywall does not take into account pending transactions

cellog:
This fixes #3246.

Normal checklist got blown away, so basically I did a review, and here's a screencast showing it working on the draft PR:

out

adding @akeem to ensure I'm using locksmith the way he intended here.

### DIFF
--- a/paywall/src/__tests__/middlewares/web3Middleware.test.js
+++ b/paywall/src/__tests__/middlewares/web3Middleware.test.js
@@ -543,6 +543,26 @@ describe('Web3 middleware', () => {
     )
   })
 
+  it('should handle ADD_TRANSACTION with data input', () => {
+    expect.assertions(2)
+    const { next, invoke } = create()
+    const action = {
+      type: ADD_TRANSACTION,
+      transaction: {
+        ...transaction,
+        input: 'data',
+      },
+    }
+    mockWeb3Service.getTransaction = jest.fn()
+
+    invoke(action)
+    expect(next).toHaveBeenCalled()
+    expect(mockWeb3Service.getTransaction).toHaveBeenCalledWith(
+      transaction.hash,
+      action.transaction
+    )
+  })
+
   it('should handle NEW_TRANSACTION', () => {
     expect.assertions(3)
     const {
--- a/paywall/src/__tests__/middlewares/web3Middleware.test.js
+++ b/paywall/src/__tests__/middlewares/web3Middleware.test.js
@@ -530,7 +530,7 @@ describe('Web3 middleware', () => {
     })
   })
 
-  it('should handle ADD_TRANSACTION', () => {
+  it('should handle ADD_TRANSACTION with no data input', () => {
     expect.assertions(2)
     const { next, invoke } = create()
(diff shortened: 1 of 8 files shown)


paywall/src/__tests__/middlewares/web3Middleware.test.js
```diff
@@ -543,6 +543,26 @@ describe('Web3 middleware', () => {
     )
   })
 
+  it('should handle ADD_TRANSACTION with data input', () => {
```
akeem: I think we will need an additional test case here to ensure that the else branch of the conditional is covered

paywall/src/middlewares/web3Middleware.js
```diff
@@ -105,7 +105,16 @@ const web3Middleware = config => ({ getState, dispatch }) => {
   return function(next) {
     return function(action) {
       if (action.type === ADD_TRANSACTION) {
-        web3Service.getTransaction(action.transaction.hash)
+        if (action.transaction.input) {
```
akeem: how would you feel about extracting this method into a library that handles this action? Its not pressing as the existing paradigm already exists inside our application but we should start to chop down the mega methods in our middleware

paywall/src/services/storageService.js
```diff
@@ -36,6 +43,7 @@ export default class StorageService {
         hash: t.transactionHash,
         network: t.chain,
         to: t.recipient,
+        input: t.data,
```
akeem: How would you feel about adjusting `input` to `data` for consistency?

paywall/src/__tests__/middlewares/web3Middleware.test.js
cellog: The test case right above handles the else case (unmodified because the else case was already being tested)

paywall/src/services/storageService.js
cellog: we can't do that, because web3Service expects the field to be named input

paywall/src/__tests__/middlewares/web3Middleware.test.js
akeem: Ah got you, would you mind update the description of the other test to indicate that is for the case where data is not input

paywall/src/middlewares/web3Middleware.js
cellog: I can see the value of this. On the other hand, the new design will pull all of this out (no more blockchain or locksmith inside paywall). Thus, this action won't be sent any longer. We will only receive updates to things like "locked" and the data will be all inside the key.

The new code is in `locksmithTransactions.js` (also modified in this PR), which doesn't do things quite the same way. What do you think?

PR #3307, opened 2019-05-21
fixing inputsHandlers tests

julien51:
# Description

These tests were version specific where they did not need to be.

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/web3Service.test.js
+++ b/unlock-js/src/__tests__/web3Service.test.js
@@ -13,7 +13,7 @@ import v0 from '../v0'
 import v01 from '../v01'
 import v02 from '../v02'
 
-import { KEY_ID, ZERO, UNLIMITED_KEYS_COUNT } from '../constants'
+import { KEY_ID, UNLIMITED_KEYS_COUNT } from '../constants'
 
 const supportedVersions = [v0, v01, v02]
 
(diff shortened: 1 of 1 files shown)


unlock-js/src/__tests__/web3Service.test.js
```diff
@@ -636,6 +483,92 @@ describe('Web3Service', () => {
     })
   })
 
+  describe('inputsHandlers', () => {
```
cellog: I understand the limited nature of this approach, testing only that the function takes inputs and puts out the correct values. The tests used to be this way, but I changed them because we had a subtle bug where the way the inputs were parsed changed, and the tests didn't catch it. Testing against the actual function inputs is a necessary test.

Could we add back in some tests, perhaps version-specific, that verify our input parsing is the same as we expect? Sort of a mini-integration test

julien51: These broken tests are preventing an upgrade to v1 which is the highest priority.

julien51: To be honest I am not even sure what these tests are for.

cellog: how do the tests break?

cellog: the tests for `_parseTransactionFromInput` don't actually verify that the parsed inputs from real events are parsed correctly and passed to inputsHandlers in a format it expects. These tests verify that. When I was moving from web3 to ethers, the tests all passed, even though web3Service was broken. This was a case of mocking out too much.

I'm fine with opening a new issue to add these tests back. Can you describe how they were failing?

julien51: They are version specific which means that, at best, they test a single version... so not sure they actually fulfill their role.

cellog:
is this superseded?

PR #3315, opened 2019-05-22
implementing the Checkout component

julien51:
# Description

This adds the component for checkout which can be used by multiple tools.
There are a couple things we should improve, including wha happens whe the user clicks on "Go to Content".

# Checklist:

### DIFF
--- a/paywall/src/components/checkout/Checkout.tsx
+++ b/paywall/src/components/checkout/Checkout.tsx
@@ -1,0 +1,109 @@
+import styled from 'styled-components'
+import React from 'react'
+import { RoundedLogo } from '../interface/Logo'
+
+import { Locks, PaywallConfig, Account } from '../../unlockTypes' // eslint-disable-line no-unused-vars
+import CheckoutLock from './CheckoutLock'
+
+interface Props {
+  locks: Locks
+  config: PaywallConfig
+  account: Account
+  purchase: (...args: any[]) => any
+  clickOnConfirmedLock: (...args: any[]) => any
+}
+
+export const Checkout = ({
+  locks,
+  config,
+  account,
+  purchase,
+  clickOnConfirmedLock,
+}: Props) => {
+  const hasValidKey = Object.keys(locks).reduce(
+    (isValid, address) => isValid || locks[address].key.status === 'valid',
+    false
+  )
+
+  return (
+    <Wrapper>
+      <Header>
+        <Title>{config.icon && <Logo src={config.icon} />} Unlocked</Title>
+        <p>{config.callToAction.default}</p>
+      </Header>
+      <CheckoutLocks>
+        {Object.values(locks).map(lock => {
+          if (lock) {
+            return (
+              <CheckoutLock
+                key={lock.address} // React needs a `key` on each child
+                lock={lock}
+                account={account}
+                disabled={hasValidKey}
+                purchase={purchase}
+                clickOnConfirmedLock={clickOnConfirmedLock}
+              />
+            )
+          }
+        })}
+      </CheckoutLocks>
+      <Footer>
+        <RoundedLogo />
+        Powered by Unlock
+      </Footer>
+    </Wrapper>
+  )
+}
+
+export default Checkout
+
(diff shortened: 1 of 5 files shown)


paywall/src/components/checkout/Checkout.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export const Checkout = ({
+  locks,
+  config,
+  account,
+  // keys,
+  purchase,
+}: Props) => {
+  const hasValidKey = Object.keys(locks).reduce((hasValidKey, address) => {
```
cellog: this can become:

```js
const hasValidKey = Object.values(locks).filter(lock => lock.key.status === 'valid').length
```

paywall/src/unlockTypes.ts
```diff
@@ -51,3 +51,61 @@ export interface Network {
 export interface ChainExplorerURLBuilders {
   [site: string]: (address: string) => string
 }
+
+export interface Lock {
+  address: string
+  transaction: string
+}
```
cellog: can add `key: Key`

paywall/src/unlockTypes.ts
```diff
@@ -51,3 +51,61 @@ export interface Network {
...
+
+export interface Key {
+  lock: string
+  id: string
+  owner: string
+  status: string
+  expiration: number
+  transactions?: Transaction[]
```
cellog: this will always be set, you can remove the `?`

cellog: also, add `confirmations: number` (this too will always be set)

paywall/src/unlockTypes.ts
```diff
@@ -51,3 +51,61 @@ export interface Network {
...
+  locks: Locks
+}
+
+export enum KeyStatus {
+  NONE = 'none',
+  CONFIRMING = 'confirming',
+  CONFIRMED = 'confirmed',
+  EXPIRED = 'expired',
+  VALID = 'valid',
+}
```
cellog: let's add 

```ts
  SUBMITTED = 'submitted',
  PENDING = 'pending',
  FAILED = 'failed',
```

paywall/src/unlockTypes.ts
```diff
@@ -51,3 +51,61 @@ export interface Network {
...
+
+export interface UILocks {
+  [address: string]: UILock
+}
+
+export interface UIKey {
+  expiration: number
+  transactions: Transaction[]
+  status: string
```
cellog: can add `confirmations: number` here too if that makes sense

paywall/src/components/checkout/Checkout.tsx
cellog: https://stackoverflow.com/questions/42966362/how-to-use-object-values-with-typescript/42967397#comment90677819_42967397 may help us here

julien51: In the end, is that cleaner than using `Object.keys`?

cellog: no. let me rephrase using Object.keys:

```js
const hasValidKey = Object.keys(locks).reduce((isValid, address) => locks[address].key.status === 'valid', false)
```

Don't do any calculation

julien51: Oh... that's what you meant, using the key status rather than the expiration! Sorry! Yeah, I'll do that.

julien51:
image


image


image


image

image

image

image

julien51:
Ok @cellog added screenshots and I think this is ready for review now!

cellog:
one question: do we want to dim the other locks if a purchase is in progress?

julien51:
I thought about this and I am not sure... I am actually not sure we should dim them when there is a valid key.

paywall/src/unlockTypes.ts
cellog: key will also have a lock member, which is the lock address. Since it's not needed, you can include it for completeness or not

PR #3320, opened 2019-05-22
Add and use a constant for scrypt/N in account creation

### LINKED ISSUE #3297: Decrypting an encrypted private key takes approximately forever
**Describe the bug**
Decrypting an encrypted private key has a substantial delay. This is a liability from a UX perspective, because users will want as near as possible immediate response to their actions. A 5-10 second delay before we can start a key purchase transaction, for example, is not good.

We need to find out what we can do to make this process faster.

cnasc:
# Description

This PR comes as a result of the discussion in #3297. It makes our user accounts a bit faster to encrypt and decrypt, though there is a tradeoff in terms of resistance to brute force attacks (which can be mitigated by enforcing a strong password policy). Overall, we are comfortable with this tradeoff because the accounts we host never actually hold any currency.

A follow-up PR will bump the version of `unlock-js` and update the CHANGELOG.

# Issues

Fixes #3297 

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/accounts.test.js
+++ b/unlock-js/src/__tests__/accounts.test.js
@@ -4,12 +4,14 @@ import {
   reEncryptPrivateKey,
 } from '../accounts'
 
+import { walletEncryptionOptions } from '../constants'
+
 jest.setTimeout(15000)
 
 describe('account helpers', () => {
   describe('web3 accounts creation', () => {
     it('should call ethers.createRandom', async () => {
-      expect.assertions(2)
+      expect.assertions(3)
 
       const {
         address,
@@ -19,6 +21,9 @@ describe('account helpers', () => {
       expect(address).toMatch(/^0x[a-fA-F0-9]{40}$/)
       expect(passwordEncryptedPrivateKey.address).toBe(
         address.substring(2).toLowerCase()
+      )
+      expect(passwordEncryptedPrivateKey.Crypto.kdfparams.n).toBe(
+        walletEncryptionOptions.scrypt.N
       )
     })
   })
--- a/unlock-js/src/accounts.js
+++ b/unlock-js/src/accounts.js
@@ -1,10 +1,12 @@
+import { walletEncryptionOptions } from './constants'
+
 const Wallet = require('ethers').Wallet
 
 export async function createAccountAndPasswordEncryptKey(password) {
   const newWallet = Wallet.createRandom()
   const address = await newWallet.getAddress()
   const passwordEncryptedPrivateKey = JSON.parse(
-    await newWallet.encrypt(password)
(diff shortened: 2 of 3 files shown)


unlock-js/src/accounts.js
```diff
@@ -1,10 +1,12 @@
+import { walletEncryptionOptions } from './constants'
+
 const Wallet = require('ethers').Wallet
 
 export async function createAccountAndPasswordEncryptKey(password) {
```
cellog: why don't we make this a parameter, and use the constants value as a default?

cnasc: I think that by having it only as a constant, we enforce some level of consistency so that there is only one place to set our standard account security. With an optional param I'd be concerned that sometimes we'd create accounts with the default level and other times with a custom one, even though all accounts should probably be the same.

cellog: yes. The same thing will happen if we choose to change this later. If we were to change it, would that break decryption, or is this somehow encoded in the encryption metadata?

cnasc: I just added a test that illustrates that, but here's a more explicit example:
```javascript
let data = {
    id: "fb1280c0-d646-4e40-9550-7026b1be504a",
    address: "88a5c2d9919e46f883eb62f7b8dd9d0cc45bc290",
    Crypto: {
        kdfparams: {
            dklen: 32,
            p: 1,
            salt: "bbfa53547e3e3bfcc9786a2cbef8504a5031d82734ecef02153e29daeed658fd",
            r: 8,
            n: 262144 // <= This guy right here
        },
        kdf: "scrypt",
        ciphertext: "10adcc8bcaf49474c6710460e0dc974331f71ee4c7baa7314b4a23d25fd6c406",
        mac: "1cf53b5ae8d75f8c037b453e7c3c61b010225d916768a6b145adf5cf9cb3a703",
        cipher: "aes-128-ctr",
        cipherparams: {
            iv: "1dcdf13e49cea706994ed38804f6d171"
         }
    },
    "version" : 3
};
```

The information is included in the encrypted payload, so the decrypt method always has what it needs. This will allow us to boost this number for new accounts in the future (and re-encrypt existing accounts on first login after a change, if we want)

unlock-js/src/__tests__/accounts.test.js
```diff
@@ -20,6 +22,9 @@ describe('account helpers', () => {
       expect(passwordEncryptedPrivateKey.address).toBe(
         address.substring(2).toLowerCase()
       )
+      expect(passwordEncryptedPrivateKey.Crypto.kdfparams.n).toBe(
```
cellog: ah this answers my question

unlock-js/src/accounts.js
cellog: appreciate the added detail, thank you

PR #3322, opened 2019-05-22
adding support for contract in v1!

julien51:
# Description

This is the bare minimum which does not add new features (outside of setting a name for a lock). This will let us upgrade the smart contract before we can add more supported functionalities.

# Checklist:


julien51:
I am not very happy with this being such a massive PR. I am not fully sure how to make it smaller.
In practice the only thing that changed between the previous contracts version (0.2) and this one (1.0) is the ability to set a name on a contract when creating it.

We can add support for more features (like actually retrieving the name...) in a follow up PR but I wanted to make sure we first started with the "least" change (and still end up with a lot of files!) while still making sure our apps remain compatible...

unlock-js/src/__tests__/web3Service.test.js
```diff
@@ -658,13 +515,80 @@ describe('Web3Service', () => {
       return nock.resolveWhenAllNocksUsed()
     }
 
+    describe('inputsHandlers', () => {
```
cellog: these tests don't use the contract at all, they should live outside the `describe.each`

unlock-js/src/__tests__/web3Service.test.js
```diff
@@ -658,13 +515,80 @@ describe('Web3Service', () => {
...
+    })
+
     describe('_parseTransactionLogsFromReceipt', () => {
       const encoder = ethers.utils.defaultAbiCoder
 
       describe('events', () => {
         it('handles the NewLock event from Unlock contract', async () => {
           expect.assertions(5)
-          await nockBeforeEach()
+          await versionedNockBeforeEach()
```
cellog: 👍

Issue #3138, opened 2019-05-10
Cache handler for Data Iframe


cellog:

cellog:
fully implemented

PR #3343, opened 2019-05-23
implement non-user-specific caching functionality

cellog:
# Description

In order to cache locks, which are not specific to any user, but are specific to networks, this PR introduces `getReadOnly` and `putReadOnly`.

In order to cache account/network, both of which are used as keys for everything else, this PR also introduces specific caching methods `getAccount` `getNetwork` and their setters.

# Issues

Refs #3138 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/data-iframe/cache/localStorage-no-localStorage.test.js
+++ b/paywall/src/__tests__/data-iframe/cache/localStorage-no-localStorage.test.js
@@ -1,4 +1,14 @@
-import { get, put, clear } from '../../../data-iframe/cache'
+import {
+  get,
+  put,
+  clear,
+  getAccount,
+  setAccount,
+  getNetwork,
+  setNetwork,
+  getReadOnly,
+  putReadOnly,
+} from '../../../data-iframe/cache'
 
 jest.mock('../../../utils/localStorage', () => () => false)
 
@@ -14,11 +24,31 @@ describe('localStorage cache', () => {
       }
     })
 
+    it('getReadOnly', async () => {
+      expect.assertions(1)
+
+      try {
+        await getReadOnly()
+      } catch (e) {
+        expect(e.message).toBe('Cannot get value from localStorage')
+      }
+    })
+
     it('put', async () => {
       expect.assertions(1)
 
       try {
         await put()
+      } catch (e) {
+        expect(e.message).toBe('Cannot put value into localStorage')
+      }
+    })
+
+    it('putReadOnly', async () => {
+      expect.assertions(1)
+
+      try {
+        await putReadOnly()
       } catch (e) {
         expect(e.message).toBe('Cannot put value into localStorage')
       }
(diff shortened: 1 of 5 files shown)


paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -43,6 +43,40 @@ function getContainer(window, key, type = false) {
...
+ *
+ * @param {object} window this is the global context, either global, window, or self
+ * @param {int} networkId the ethereum network id
+ * @param {string} type the type of account data to retrieve
+ */
+export async function getReadOnly(window, networkId, type) {
+  return get(
+    window,
+    networkId,
+    '0x0000000000000000000000000000000000000000',
```
cellog: I'm going to add a comment in the code about why I chose address 0 (nobody can use this, so it's a guaranteed readonly value)

paywall/src/__tests__/data-iframe/cache/localStorage-no-localStorage.test.js
```diff
@@ -14,6 +24,16 @@ describe('localStorage cache', () => {
       }
     })
 
+    it('getReadOnly', async () => {
+      expect.assertions(1)
+
+      try {
+        await getReadOnly()
+      } catch (e) {
+        expect(e.message).toBe('Cannot get value from localStorage')
```
akeem: Depending on the types of errors possible it might be worth while to move the  error message in the test to its own variable to be referenced in the subsequent tests

paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -43,6 +45,42 @@ function getContainer(window, key, type = false) {
...
 }
 
+/**
+ * Retrieve a cached value for a non-user-specific type on a specific network
+ *
+ * @param {object} window this is the global context, either global, window, or self
+ * @param {int} networkId the ethereum network id
+ * @param {string} type the type of account data to retrieve
+ */
+export async function getReadOnly(window, networkId, type) {
```
akeem: Is there some context around these methods ending with `ReadOnly`?

paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -86,6 +124,58 @@ export async function put(window, networkId, accountAddress, type, value) {
...
 }
 
+/**
+ * Retrieve the current cached account
+ *
+ * @param {object} window this is the global context, either global, window, or self
+ * @returns {string}
+ */
+export async function getAccount(window) {
+  if (!localStorageAvailable(window)) {
```
akeem: It might be worthwhile to extract this into a method and us it in the methods calling the functionality below

paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -86,6 +124,58 @@ export async function put(window, networkId, accountAddress, type, value) {
...
+  return +window.localStorage.getItem('__unlockProtocol.network') || null
+}
+
+/**
+ * Set the current cached network
+ *
+ * @param {object} window this is the global context, either global, window, or self
+ * @param {number} network the id of the current ethereum network
+ */
+export async function setNetwork(window, network) {
```
akeem: Not blocking but I think that we could probably generalize these getters and setters

paywall/src/__tests__/data-iframe/cache/localStorage.test.js
```diff
@@ -89,6 +97,58 @@ describe('localStorage cache', () => {
...
+    it('value is set', async () => {
+      expect.assertions(1)
+      fakeWindow.storage[storageId(123, nullAccount)] = JSON.stringify({
+        there: 'hello',
+      })
+
+      expect(await getReadOnly(fakeWindow, 123, 'there')).toEqual('hello')
+    })
+
+    it('value is set, but malformed', async () => {
```
akeem: Is the setting of malformed data handled at a layer below the code in this pull request?

cellog: localStorage is publicly accessible, so this guards against a malicious script trying to muck things up

paywall/src/data-iframe/cache/localStorage.js
cellog: caching read-write values (keys, transactions) depends on a user account. read-only values are currently locks, which are account-independent

cellog:
ok @akeem I refactored to add `ensureLocalStorageAvailable` for generic error-handling with more specific error messages too (double win). I also removed the `ReadOnly` stuff, and instead changed the parameters of `get` and `put` (and others) to take a destructured object. This is because I already had `type` as an optional parameter, and this is the simplest way to have 2 optional parameters.

PTAL when you get a chance.

Issue #3332, opened 2019-05-22
Unify error handling


cnasc:
**Is your feature request related to a problem? Please describe.**
We dispatch a variety of different kinds of errors in the app. (confusingly, we have 2 files providing general error actions: `error.js` and `errors.js`) Broadly in 3 categories:

1. Fatal errors
2. Non-fatal errors
3. Diagnostic errors (things that shouldn't be displayed to users, but which we send around to aid debugging)

Fatal errors are handled pretty well, there's one place where they're defined and one place where they're consumed, and I believe they are all handled.

For non-fatal errors, we dispatch a few different kinds ("errors," signature errors, storage errors...), and not all of them are handled. All well and good when you're just watching the Redux log to see what gets passed around, but not very helpful for a non-developer.

**Describe the solution you'd like**
All non-fatal errors should be handled consistently in one place, so that we can include the `<Errors />` component on any page that might trigger errors and be confident that everything is taken care of.

PR #3345, opened 2019-05-23
Errors component is typed

cnasc:
# Description

Splitting off a piece of #3332, this PR converts the `<Errors />` component to TS, and adds a type to `UnlockTypes`. Additionally, a bunch of class names got changed in the stories, so that update is also here.

# Checklist:


cnasc:
Actually, on second thought we may want a different type name -- we don't want confusion with the standard `Error` class. I'll swap that out before merging.

cnasc:
On third thought, I'll keep it as is for now because I'm going to do a second pass on the types here and this is just some initial base-work for that.

PR #3349, opened 2019-05-23
New error constructors

cnasc:
# Description

Another piece from #3332. This PR introduces new error types and constructors, which have some pretty nifty features, I think. This PR does not _use_ these errors yet, it just puts them in place.

For starters, the errors are now very strongly typed. The kind of error (which feature of the application it is associated with) and the level (fatal or warning) are enforced by the type system. Each error additionally carries a message payload. The previous iteration included a data payload which could be an arbitrary object, but much of the purpose of this infrastructure is to allow us to notify users, and for that reason a text description of the failure is absolutely necessary.

Here's some example usage:

```javascript
// file storageMiddleware.js
import { StorageError } from '/path/to/Error'

thisPromiseAlwaysFails()
  .catch(_ => {
    const err = StorageError.Warn('The code that always fails failed.')
    dispatch(setError(err))
  })
```

To handle these errors, we can filter them by level (fatal errors all handled in one place) or kind (signup page only shows signup errors, no matter what might be happening in the background) or we can use this information to display different kinds of errors differently.

I've given a constructor for every distinct kind of error we currently send out, and there's just a little bit of boilerplate involved in defining new ones.
# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/utils/Error.test.ts
+++ b/unlock-app/src/__tests__/utils/Error.test.ts
@@ -1,19 +1,23 @@
 import {
-  StorageError,
-  SignatureError,
+  ApplicationError,
   FormValidationError,
   LogInError,
   SignUpError,
+  SignatureError,
+  StorageError,
+  TransactionError,
 } from '../../utils/Error'
 
 describe('Error constructors', () => {
   it.each([
-    StorageError,
-    SignatureError,
+    ApplicationError,
     FormValidationError,
     LogInError,
     SignUpError,
-  ])('%s carries the message passed to it', constructors => {
+    SignatureError,
+    StorageError,
+    TransactionError,
+  ])('each of these carries the message passed to it', constructors => {
     expect.assertions(2)
     const fatalMessage = 'Not enough teapots, dumping core'
     const warning = 'Passwords must contain at least 43 different characters'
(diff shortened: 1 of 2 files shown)


unlock-app/src/__tests__/utils/Error.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    ApplicationError,
+    FormValidationError,
+    LogInError,
+    SignUpError,
+    SignatureError,
+    StorageError,
+    TransactionError,
+  ])('each of these carries the message passed to it', constructors => {
+    expect.assertions(2)
+    const fatalMessage = 'Not enough teapots, dumping core'
```
cellog: 😆

unlock-app/src/utils/Error.ts
```diff
@@ -0,0 +1,94 @@
...
+  return (message: string): UnlockError<'Warning', K> => ({
+    level: 'Warning',
+    kind,
+    message,
+  })
+}
+
+// This is here to enforce that the Fatal property of a group of error
+// constructors will actually be fatal -- no accidentally ignored errors.
+interface ErrorMakers<K extends ErrorKind> {
```
cnasc: This is kind of nice, it constrains what can live in one of these objects, so you can't accidentally have 2 functions that produce warnings or accidentally put the wrong kind of error in the container. So we can be confident that `StorageError.Fatal` can only ever create a fatal error for the storage service.

PR #3357, opened 2019-05-24
postOffice listener for config (locks to monitor), and sending junk back when requested

cellog:
# Description

This sets up the listener side of the data iframe post office. It responds to direct requests for data as well as listening for the `POST_MESSAGE_CONFIG` message which will contain the locks we need to monitor. It also listens for a `POST_MESSAGE_PURCHASE_KEY` message, which contains a lock to purchase a key on, and any extra tip (`purchaseFor` allows tipping) to add on to the key price.

# Issues

Refs #3140

# Checklist:

### DIFF
--- a/paywall/src/paywall-builder/constants.js
+++ b/paywall/src/paywall-builder/constants.js
@@ -17,3 +17,4 @@ export const POST_MESSAGE_UPDATE_NETWORK = 'update/network'
 export const POST_MESSAGE_UPDATE_WALLET = 'update/walletmodal'
 
 export const POST_MESSAGE_ERROR = 'error'
+export const POST_MESSAGE_DATA_REQUEST = 'data/request'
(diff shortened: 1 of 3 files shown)


paywall/src/paywall-builder/constants.js
```diff
@@ -15,3 +15,6 @@ export const POST_MESSAGE_UPDATE_ACCOUNT = 'update/account'
 export const POST_MESSAGE_UPDATE_ACCOUNT_BALANCE = 'update/accountBalance'
 export const POST_MESSAGE_UPDATE_NETWORK = 'update/network'
 export const POST_MESSAGE_UPDATE_WALLET = 'update/walletmodal'
+export const POST_MESSAGE_DATA_REQUEST = 'data/request'
```
julien51: Could we use a better name here? Something more descriptive? I also see that for all of these other we have `verb/object`. Could we use the same syntax? (at least that would be `request/data`... but not great!)

cellog: yes, since we're asking the chain to send any updates, I'll rename it to `POST_MESSAGE_SEND_UPDATES`

Issue #3139, opened 2019-05-10
Blockchain handler for Data Iframe


cellog:

cellog:
this is implemented

PR #3359, opened 2019-05-24
refactor key purchase to return the raw key, linking will be downstream

cellog:
# Description

This PR addresses 3 issues. 2 are minor, the other is a design flaw.

1. on a brand new key purchase transaction, `updateListener` was crashing because the key's `transactions[0]` was undefined. This adds a simple `!transaction` as one of the bail-out conditions
2. in `updateListener`, we were asserting on the `status` of the raw key to determine whether to keep polling, but `status` only exists on a linked key. This now does just-in-time linking before the conditional of the do/while loop. Tests are updated to ensure we return properly
3. The key purchase transaction monitoring incorrectly added the custom fields of the key to the key.  This is a problem because the data actually sent to the user is from the cache. We need to cache the raw blockchain data, as returned from `web3Service`. The mixing of key/lock/transaction can happen right before it is sent to the UI.

To make point 2 possible, this PR does 2 things:

1. removes `linkTransactionsToKey` calls on the returned data, and just returns the key
2. since the only field that actually could change is the expiration timestamp of a key, we request the current value of the key from the blockchain via `web3Service.getKeyByLockForOwner`

# Issues

Refs #3139 

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/blockchainHandler/purchaseKey.js
+++ b/paywall/src/data-iframe/blockchainHandler/purchaseKey.js
@@ -2,6 +2,7 @@ import { getAccount } from './account'
 import ensureWalletReady from './ensureWalletReady'
 import submittedListener from './purchaseKey/submittedListener'
 import updateListener from './purchaseKey/updateListener'
+import { linkTransactionsToKey } from './keyStatus'
 
 /**
  * Purchase a key on a lock for the current user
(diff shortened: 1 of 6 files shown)


paywall/src/data-iframe/blockchainHandler/purchaseKey.js
```diff
@@ -49,34 +50,44 @@ export async function processKeyPurchaseTransactions({
...
+        requiredConfirmations,
+      })
+      afterEventProcessed()
+      linkedKey = linkTransactionsToKey({
+        key,
+        transactions,
+        requiredConfirmations,
+      })
+    } while (linkedKey.status === 'confirming')
+  } finally {
```
cellog: this is basically the same, but wrapped in a try/finally so we don't leak memory on errors thrown

PR #3363, opened 2019-05-25
refactor data iframe cache to be race condition-resistant

cellog:
# Description

The current design of the data iframe cache stores a user's data under a single localStorage key. This has a dangerous side effect: if 2 async processes attempt to update the store at the same time, stale data might overwrite more current data.

The solution is to store the sensitive data (keys, transactions) in separate localStorage entries. To do this, the PR adds getters and setters for `key` and `transaction`, and refactors the plural setters and getters to use these single setters for keys/transactions. In addition, the localStorage cache driver is refactored to be a bit simpler. This change means we now use a regex to match storage keys that change in different tabs in order to sync across open tabs in the browser.

In addition, `cacheHandler.js` is re-organized to group related functionality, and adds comments to capture necessary context that is not obvious. For example, with this change, we need to use something to figure out which cache entries to pull for keys. By indexing keys via the lock they correspond with, we only need the list of locks. Fortunately, this list is sent in the paywall config at startup, and so will be the very first thing cached.

The last change is that the cacheHandler test is refactored to remove implementation details, and to test important behavior, such as how the cache should react if the current user account or current user network changes.

With these changes, the cache will be more resistant to subtle bugs as well as easier to refactor internally (including swapping out for a different storage driver underneath)

# Issues

Refs #3138

# Checklist:

### DIFF
--- a/paywall/src/__tests__/data-iframe/cache/localStorage.test.js
+++ b/paywall/src/__tests__/data-iframe/cache/localStorage.test.js
@@ -54,9 +54,9 @@ describe('localStorage cache', () => {
 
     it('value is set', async () => {
       expect.assertions(1)
-      fakeWindow.storage[storageId(123, 'hi')] = JSON.stringify({
-        there: 'hello',
-      })
+      fakeWindow.storage[storageId(123, 'hi', 'there')] = JSON.stringify(
+        'hello'
+      )
 
       expect(
         await get({
@@ -65,14 +65,14 @@ describe('localStorage cache', () => {
           type: 'there',
           accountAddress: 'hi',
         })
-      ).toEqual('hello')
+      ).toBe('hello')
     })
 
     it('no account retrieves generic value', async () => {
       expect.assertions(1)
-      fakeWindow.storage[storageId(123, nullAccount)] = JSON.stringify({
-        there: 'hello',
-      })
+      fakeWindow.storage[storageId(123, nullAccount, 'there')] = JSON.stringify(
+        'hello'
+      )
 
       expect(
         await get({
@@ -80,14 +80,14 @@ describe('localStorage cache', () => {
           networkId: 123,
           type: 'there',
         })
-      ).toEqual('hello')
+      ).toBe('hello')
     })
 
     it('value is set, but malformed', async () => {
       expect.assertions(1)
-      fakeWindow.storage[storageId(123, 'hi')] = JSON.stringify({
-        there: 'hello',
-      }).substring(1, 4)
+      fakeWindow.storage[storageId(123, 'hi', 'there')] = JSON.stringify(
+        'hello'
+      ).substring(1, 4)
 
       expect(
         await get({
@@ -101,9 +101,9 @@ describe('localStorage cache', () => {
 
     it('value is set, different network', async () => {
       expect.assertions(1)
-      fakeWindow.storage[storageId(456, 'hi')] = JSON.stringify({
-        there: 'hello',
-      })
+      fakeWindow.storage[storageId(456, 'hi', 'there')] = JSON.stringify(
+        'hello'
+      )
 
       expect(
         await get({
@@ -117,9 +117,9 @@ describe('localStorage cache', () => {
 
     it('value is set, different account', async () => {
       expect.assertions(1)
-      fakeWindow.storage[storageId(123, 'another')] = JSON.stringify({
-        there: 'hello',
-      })
+      fakeWindow.storage[storageId(123, 'another', 'there')] = JSON.stringify(
+        'hello'
+      )
 
       expect(
         await get({
@@ -129,21 +129,6 @@ describe('localStorage cache', () => {
           accountAddress: 'hi',
         })
       ).toBeNull()
-    })
-
-    it('value is set, entire cache for user wanted', async () => {
-      expect.assertions(1)
-      fakeWindow.storage[storageId(123, 'hi')] = JSON.stringify({
-        there: 'hello',
-        it: 'is',
-      })
-
-      expect(
-        await get({ window: fakeWindow, networkId: 123, accountAddress: 'hi' })
-      ).toEqual({
-        there: 'hello',
-        it: 'is',
-      })
     })
   })
 
@@ -163,10 +148,8 @@ describe('localStorage cache', () => {
         value: 'hello',
       })
 
-      expect(fakeWindow.storage[storageId(123, 'hi')]).toBe(
-        JSON.stringify({
-          there: 'hello',
-        })
+      expect(fakeWindow.storage[storageId(123, 'hi', 'there')]).toBe(
+        JSON.stringify('hello')
       )
     })
 
@@ -180,10 +163,8 @@ describe('localStorage cache', () => {
         value: 'hello',
       })
 
-      expect(fakeWindow.storage[storageId(123, nullAccount)]).toBe(
-        JSON.stringify({
-          there: 'hello',
-        })
+      expect(fakeWindow.storage[storageId(123, nullAccount, 'there')]).toBe(
+        JSON.stringify('hello')
       )
     })
   })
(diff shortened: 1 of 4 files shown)


paywall/src/__tests__/data-iframe/cache/localStorage.test.js
```diff
@@ -130,21 +130,6 @@ describe('localStorage cache', () => {
         })
       ).toBeNull()
     })
-
```
cellog: this functionality is no longer possible (and was unused in the app)

paywall/src/__tests__/data-iframe/cache/localStorage.test.js
```diff
@@ -239,10 +227,13 @@ describe('localStorage cache', () => {
       expect.assertions(1)
```
cellog: the diff below is the best example of how the behavior of the `localStorage` driver has changed:

```diff
-        'unlock-protocol/123/bar': '{"fooe":"bare","bare":"bare"}',
-        'unlock-protocol/123/hi': '{"foo":"bar","bar":"bar"}',
-        'unlock-protocol/456/bar': '{"fooe":"bare"}',
-        'unlock-protocol/456/hi': '{"foo":"bar"}',
+        'unlock-protocol/123/bar#fooe': '"bare"',
+        'unlock-protocol/123/bar#bare': '"bare"',
+        'unlock-protocol/123/hi#foo': '"bar"',
+        'unlock-protocol/123/bar#foo': '"bar"',
+        'unlock-protocol/123/hi#bar': '"bar"',
+        'unlock-protocol/456/bar#fooe': '"bare"',
+        'unlock-protocol/456/hi#foo': '"bar"',
```

paywall/src/__tests__/data-iframe/cacheHandler.test.js
```diff
@@ -10,30 +10,39 @@ import {
   setNetwork,
   getFormattedCacheValues,
   setAccountBalance,
+  setKey,
+  getKey,
```
cellog: the 3 functions below are needed for pulling in all keys and all transaction hashes. Instead of `setKeyIds`, this uses `setLockAddresses` because both locks and keys are indexed by the same value: the lock they are connected to

paywall/src/__tests__/data-iframe/cacheHandler.test.js
```diff
@@ -199,6 +231,131 @@ describe('cacheHandler', () => {
     })
   })
```
cellog: these are the new test concepts: making sure behavior is correct when account changes or network changes

paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -1,7 +1,9 @@
 import localStorageAvailable from '../../utils/localStorage'
 
-export function storageId(networkId, accountAddress) {
-  return `unlock-protocol/${networkId}/${accountAddress}`
+export function storageId(networkId, accountAddress, type) {
```
cellog: this is the primary change: modifying the storage id to append a type if specified

paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -11,37 +13,20 @@ const nullAccount = '0x0000000000000000000000000000000000000000'
  * exists as a key if requested
  *
  * @param {string} key the key for the item in localStorage
- * @param {string} type the sub-key (if any) which must exist on return
  */
-function getContainer(window, key, type = false) {
```
cellog: type is no longer optional, which greatly simplifies this design

paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -97,15 +80,12 @@ export async function put({
   accountAddress = nullAccount,
 }) {
   ensureLocalStorageAvailable(window, `save ${type} in cache`)
-  const key = storageId(networkId, accountAddress)
-  const container = getContainer(window, key, type)
+  const key = storageId(networkId, accountAddress, type)
   if (value === undefined) {
```
cellog: this is more natural behavior: if you `put` something, it saves it. If you put `undefined`, we clear the localStorage entry.

paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -162,13 +142,8 @@ export async function setNetwork(window, network) {
  */
 export async function clear({ window, networkId, accountAddress, type }) {
   ensureLocalStorageAvailable(window, 'clear cache')
```
cellog: this was convoluted, and is now simpler

paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -187,13 +162,14 @@ export async function addListener({
   accountAddress = nullAccount,
 }) {
   const key = storageId(networkId, accountAddress)
```
cellog: this is key: because we can have storage ids like:

```
unlock-protocol/1/0x....#thing1
unlock-protocol/1/0x....#thing2
```

we can't just assert on `unlock-protocol/1/0x....` because that won't match any of the above examples. The obvious solution is to match on the beginning of the localStorage key, which is `unlock-protocol/1/0x....`

paywall/src/data-iframe/cacheHandler.js
```diff
@@ -4,35 +4,89 @@ import linkKeysToLocks from './blockchainHandler/linkKeysToLocks'
...
+export async function setAccountBalance(window, balance) {
+  return _put(window, 'balance', balance)
+}
+
 export function setup(networkId, account) {
   currentNetwork = networkId
   currentAccount = account
 }
 
-async function _get(window, key) {
```
cellog: extend the definition of `_get` and `_put` to pass the null account for storing things like locks that are the same across different accounts. This allows centralizing the important stuff here at the top of the file, and the `getLocks` `setKeys` etc. can focus on the logic of what they do instead of low-level cache management

julien51:
> The solution is to store the sensitive data (keys, transactions) in separate localStorage entries.

Isn't this just pushed the problem down to another piece?

cnasc:
Have we demonstrated that this issue can happen? IIRC all major browsers implement localStorage with a mutex, and given that JS is inherently single-threaded I don’t know if this issue can arise, even with asynchronous functions. (Unless I’ve misunderstood what the actual concern is)

cellog:
So, the race can happen as follows. Johnny Clickstoomuch decides to buy 2 keys at the same time.

This generates 2 parallel transaction update listeners.

Listener 1 gets the current transaction list, which is:

[old transaction 1, old transaction 2]

Listener 2 gets the same transaction list.

Both listener 1 and 2 update their transactions, and send updates to change the cache. The updates are:

[ old transaction 1, new transaction 2] and
[ new transaction 1, old transaction 2]

They both arrive, and regardless of the order of commit, one of them will include stale data, because the transactions are stored as an object

```
{
  hash1: old transaction 1,
  hash2: new transaction 2
}
```

or as:

```
{
  hash1: new transaction 1,
  hash2: old transaction 2
}
```

There are a couple of solutions to this issue. The way this PR solves it is to store every transaction/key in its own localStorage bucket. This way, the order of transactions does not matter, but it introduces some complexity on retrieving all of the keys or all of the transactions.

In fact, upon some further reflection, I still think the race condition is present with new transactions, as the list of transaction hashes is updated, so one of the transactions would disappear.

The issue is mainly that we need to not pass in stale data when updating the cache, and need to update it inside the cache itself. I'm going to take one more look at doing this right and see if it can be done within the context of this PR or if it needs a slightly different approach.

Now, I'm thinking maybe I can remove the race by having the cache itself do the logic of merging in the key and the transaction, and keep the existing logic, but just add in `setKey` and `setTransaction` for modifying a single transaction. This way, we are guaranteed non-stale data. This is also a less invasive change to the current design.

Fortunately, there is no race condition where 2 different versions of the same key or transaction would try to overwrite each other, this is only about parallel transactions for different key purchases.

cellog:
I have a better solution. Closing this in favor of that!

PR #3370, opened 2019-05-26
adding the adremover sample

### LINKED ISSUE #3358: Test application needs to be created for the ad remover paywall
A new sub-repo in directory `/adremover` needs to be created that is a simple express server which has one static HTML page at index.html. The index.html page should be based upon our demo page and include the following tech specs:

1. a checkout button
2. fake "ads" that can be hidden or shown based on the key status.
3. content that does not disappear when the ads do.

The details of these are specified below:

## Loading  the script 

The first part requires loading a script from our server. You need to achieve this in the `<head>` section of the HTML body. The script will be located at the following address: `https://paywall.unlock-protocol.com/static/unlock.1.0.min.js` _*Our page should load `PAYWALL_URL/static/unlock.1.0.min.js` where `PAYWALL_URL` is an env variable defaulting to `http://localhost:3001`*_

There are several ways to load that script in a non-blocking way. We recommend using the following approach:

```html
<script>
(function(d, s) {
  var js = d.createElement(s),
  	sc = d.getElementsByTagName(s)[0];
  js.src="https://paywall.unlock-protocol.com/unlock.1.0.min.js";
  sc.parentNode.insertBefore(js, sc);
}(document, "script"));
</script>
```

Note: This approach blocks the `window.onload` event until the script has loaded.

## Configuration

The following snippet should also be put in the `<head>` section of the HTML document and will let you configure the behavior of the snippet above. 

The `unlockProtocolConfig` is an object which includes the locks indexed by their addresses on the Ethereum chain. Each lock is an object which can have an optional name (string). This name will be used on the checkout modal.

There is also an optional icon which will be shown on the modal and `callToAction` object with a default  value set to the string of text shown on the checkout modal.
Here is an example:

```js
var unlockProtocolConfig = {
  locks: {
	'0xabc': {
  		name: 'One Week',
	},
	'0xdef': {
  		name: 'One Month',
	},
	'0xghi': {
  		name: 'One Year',
	},
  },
  icon: 'https://...', 
  callToAction: {
	default:
  	'Enjoy Unlock Online without any ads for as little as $2 a week. Pay with Ethereum in just two clicks.',
  },
}
```

## Handling the events

Once loaded the unlock script will trigger events on the page’s window object. These events will indicate to the page whether the current visitor owns a key to any of the lock (the state should be unlocked or if they do not, and the state should be locked.
If the state is 'locked' the function should call the function to load the ads. 
If the state is 'unlocked' the function should make sure that the ads are not loaded, or even removed.

```js
window.addEventListener('unlockProtocol', function(e) {
  var state = e.data
  // the state is a string whose value can either be 'unlocked' or 'locked'...
  // If state is 'unlocked': implement code here which will be triggered when the current visitor has a valid lock key (hide ads)
  // If state is 'locked': implement code here which will be triggered when the current visitor does not have a valid lock key (show ads)
  }
})
```

Note: the callback can be invoked several times. For example, a visitor might purchase a key while they are on the page, and the ads should then disappear. Similarly, the key that the visitor owns may expire during a visit which would result in the ads being loaded again.

## Checkout

In order to purchase keys, Unlock provides a modal which lets the user pick the lock of their choice (based on the configuration above). The modal can be loaded by invoking the following:

`window.unlockProtocol && unlockProtocol.loadCheckoutModal()`

It is good practice to ensure that the `unlockProtocol` object is defined before invoking the `loadCheckoutModal()` function.

julien51:
# Description

This ads the temporary testing sample app for the ad remover.

# Issues

Fixes #3358

# Checklist:

### DIFF
--- a/paywall/adremover/README.md
+++ b/paywall/adremover/README.md
@@ -1,6 +1,6 @@
 # Ad Remover sample
 
-This is a temporary forlder (to be removed after July 1st 2019) which will provide an example of a creator using our paywall to implement an ad remover.
+This is a temporary folder (to be removed after July 1st 2019) which will provide an example of a creator using our paywall to implement an ad remover.
 
 This static site is meant to be served as a full static HTML file, using npm's `serve` or python `python3 -m http.server` (older versions of python can use `python -m SimpleHTTPServer`).
 
--- a/paywall/adremover/README.md
+++ b/paywall/adremover/README.md
@@ -1,0 +1,7 @@
+# Ad Remover sample
+
+This is a temporary forlder (to be removed after July 1st 2019) which will provide an example of a creator using our paywall to implement an ad remover.
+
+This static site is meant to be served as a full static HTML file, using npm's `serve` or python `python3 -m http.server` (older versions of python can use `python -m SimpleHTTPServer`).
+
+This fake publisher site is a simple HTML document with 2 ads loaded by default and a button to purchase an ad free experience of the content.
--- a/paywall/adremover/index.html
+++ b/paywall/adremover/index.html
@@ -1,0 +1,169 @@
+<!doctype html>
+<html>
+
+<head>
+  <meta charset="utf-8" />
+  <title>Lorem Ipsum Daily</title>
+  <script>
+    (function (d, s) {
+      var js = d.createElement(s),
+        sc = d.getElementsByTagName(s)[0];
+      js.src = "http://localhost:3001/unlock.1.0.min.js";
+      sc.parentNode.insertBefore(js, sc);
+    }(document, "script"));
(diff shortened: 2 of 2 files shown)


paywall/adremover/README.md
```diff
@@ -0,0 +1,7 @@
+# Ad Remover sample
+
+This is a temporary forlder (to be removed after July 1st 2019) which will provide an example of a creator using our paywall to implement an ad remover.
```
cellog:
```suggestion
This is a temporary folder (to be removed after July 1st 2019) which will provide an example of a creator using our paywall to implement an ad remover.
```

PR #3371, opened 2019-05-26
implement merging a value into an existing cache to remove possibility of stale data

cellog:
# Description

This adds `merge`, which is an atomic way to write to the cache for a sub-value (like a key within keys, or transaction within transactions)

A follow-up PR will implement `setKey` and `setTransaction` based on `merge`

# Issues

Refs #3206

# Checklist:

### DIFF
--- a/paywall/src/__tests__/data-iframe/cache/localStorage-no-localStorage.test.js
+++ b/paywall/src/__tests__/data-iframe/cache/localStorage-no-localStorage.test.js
@@ -7,6 +7,7 @@ import {
   getNetwork,
   setNetwork,
 } from '../../../data-iframe/cache'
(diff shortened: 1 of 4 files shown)


paywall/src/__tests__/data-iframe/cache/localStorage-no-localStorage.test.js
```diff
@@ -36,6 +37,18 @@ describe('localStorage cache', () => {
       }
     })
 
+    it('merge', async () => {
```
julien51: Could we try to use better ways to describe tests.
Think that the sentence should __read__ like English, for example: `it('should throw an error when localStorage is not available')`.
This makes it much easier to understand what the test does later on.
Thanks!

Issue #3364, opened 2019-05-25
Ad remove paywall: unlock.min.js


cellog:
This has several components:

- the main window post office, responsible for communicating with both the data iframe and the checkout iframe
- an event emitter that pushes events onto the main window "unlockProtocol". The data field should be set to "unlocked" or "locked" based on the user state
- a global variable on the window object, `unlockProtocol` that provides a function `unlockProtocol.loadCheckoutModal()`. This function will trigger display of the checkout iframe

## Post office

The post office is an instance of `mainWindowPostOffice`, and should listen for the `POST_MESSAGE_UPDATE_*` events sent by the data iframe. It should pass along this information to the checkout iframe. It should also listen for the `POST_MESSAGE_LOCKED` and `POST_MESSAGE_UNLOCKED` events and pass this to the event emitter. It should setup the `unlockProtocol.loadCheckoutModal` function and post a message to the checkout iframe that causes it to expand its iframe to full screen and display the checkout modal component.

PR #3378, opened 2019-05-27
build script for unlock.1.0.min.js

cellog:
# Description

This is the build script for unlock.1.0.min.js. Of course, it's a no-op until the code is there, but as soon as `unlock.js/index.js` exists, it will start being possible to use.

# Issues

Refs #3364

# Checklist:


paywall/unlock.js.webpack.config.js
```diff
@@ -0,0 +1,22 @@
+var path = require('path')
+const webpack = require('webpack')
+
+module.exports = env => {
+  return {
+    mode: 'production',
```
julien51: Could we derive that from `UNLOCK_ENV` instead? Using `development` could make things easier to debug locally?

paywall/unlock.js.webpack.config.js
```diff
@@ -0,0 +1,22 @@
...
+    devtool: 'source-map',
+    entry: path.resolve(__dirname, 'src', 'unlock.js', 'index.js'),
+    output: {
+      path: path.resolve(__dirname, 'src', 'static'),
+      filename: 'unlock.1.0.min.js',
```
julien51: Let's exatract the version number into a constant at the top of this file so we can easily change later on?

paywall/unlock.js.webpack.config.js
cellog: because it generates a source map, this shouldn't be an issue for debugging in chrome. Did you mean just to check which files are included in the bundle?

paywall/unlock.js.webpack.config.js
cellog: oh, FYI, I was fully expecting that if we bump the version, we will generate entirely separate scripts, and continue to generate the older versions, since clients will be depending on them. We can talk about this if it doesn't match your assumptions

PR #3380, opened 2019-05-27
add handler chaining to the postOffice util

cellog:
# Description

The `unlock.min.js` post office handler for `POST_MESSAGE_READY` needs to do 2 unrelated actions, return the wallet info for the `Web3ProxyProvider` and also send the configuration. This requires the ability to chain 2 handlers. This simplistic approach is designed to allow chaining but not support removing handlers. All handlers are for-the-lifetime-of-the-script.

# Issues

Refs #3364

# Checklist:

### DIFF
--- a/paywall/src/__tests__/utils/postOffice.test.js
+++ b/paywall/src/__tests__/utils/postOffice.test.js
@@ -223,7 +223,7 @@ describe('postOffice', () => {
           )
         })
 
-        it('chains handlers', () => {
+        it('should call all handlers registered with setHandler for the message type', () => {
           expect.assertions(2)
 
           listener = jest.fn()
--- a/paywall/src/__tests__/utils/postOffice.test.js
+++ b/paywall/src/__tests__/utils/postOffice.test.js
@@ -195,6 +195,7 @@ describe('postOffice', () => {
       describe('message handler', () => {
         let listener
         beforeEach(() => {
+          _clearHandlers()
           listener = jest.fn()
           setHandler('hi', listener)
           setupPostOffice(fakeWindow, fakeTarget, 'origin')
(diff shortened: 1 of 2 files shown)


cellog:
(and the previously mentioned flaky test from #3379 fails here)

paywall/src/__tests__/utils/postOffice.test.js
```diff
@@ -221,6 +222,35 @@ describe('postOffice', () => {
             'origin'
           )
         })
+
+        it('chains handlers', () => {
```
julien51: nit: we need a better test description.

paywall/src/utils/postOffice.js
```diff
@@ -87,7 +87,10 @@ export function mainWindowPostOffice(window, iframe, iframeOrigin) {
  * @param {handlerCallback}}handler the callback. This should
  */
 export function setHandler(type, handler) {
-  handlers[type] = handler
+  const currentHandler = handlers[type] || (() => 1)
```
julien51: hum. Can you please explain what this is doing and the choice of `(() => 1)` for a default?

cellog: setHandler creates a linked list of callbacks. The `() => 1` case is a null handler that does nothing. It saves 1 character over `() => {}`.

First time you call setHandler, it creates a handler that will call () => 1 and then the handler.
The second time you call sethandler, it creates a handler that calls (() => 1, the first handler), then the 2nd handler.

And so on

cellog: note to self: use brackets and add no-op comment

cellog:
OK @julien51 this is ready

PR #3383, opened 2019-05-27
purchase callback creation for the data iframe

cellog:
# Description

Because purchaseKey can only be called with the information passed to it (lock address and any extra tip to add on), we need to close around the things needed to handle the purchase.

Note that the tests only ensure we are calling the purchaseKey functions with the expected parameters, and that the transaction monitoring begins. I considered doing a more complex test with a fake provider so we are not testing implementation details at all, but in the end we can't do this effectively because the implementation details of the smart contract would break the test every time we upgrade.

Integration tests will have to be used to verify the end-to-end flow.

# Issues

Refs #3140

# Checklist:

### DIFF
--- a/paywall/src/__tests__/data-iframe/start/makePurchaseCallback.test.js
+++ b/paywall/src/__tests__/data-iframe/start/makePurchaseCallback.test.js
@@ -54,6 +54,7 @@ describe('makePurchaseCallback', () => {
         removeListener: jest.fn(),
         once: jest.fn(),
         on: jest.fn(),
+        off: jest.fn(),
         purchaseKey: jest.fn().mockResolvedValue(),
       }
       update = jest.fn()
@@ -86,7 +87,7 @@ describe('makePurchaseCallback', () => {
     it('should initiate a key purchase', async () => {
       expect.assertions(1)
 
-      await purchase('lock')
+      purchase('lock')
 
       await new Promise(resolve => {
         const interval = setInterval(() => {
@@ -103,10 +104,22 @@ describe('makePurchaseCallback', () => {
       )
     })
 
+    it('should pass on any error thrown in keyPurchase', async () => {
+      expect.assertions(1)
+
+      walletService.purchaseKey = jest.fn().mockRejectedValue(new Error('fail'))
+
+      try {
+        await purchase('lock')
+      } catch (e) {
+        expect(e.message).toBe('fail')
+      }
+    })
+
     it('should initiate monitoring of key purchase transaction', async () => {
       expect.assertions(1)
 
-      await purchase('lock')
+      purchase('lock')
 
       await new Promise(resolve => {
         const interval = setInterval(() => {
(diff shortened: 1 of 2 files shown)


paywall/src/data-iframe/start/makePurchaseCallback.js
```diff
@@ -0,0 +1,44 @@
...
+  window,
+}) =>
+  async function purchase(lockAddress /*, extraTip */) {
+    const [
+      { purchaseKey, processKeyPurchaseTransactions },
+      locks,
+      startingTransactions,
+      keys,
+    ] = await Promise.all([
+      import('../blockchainHandler/purchaseKey'),
```
julien51: Can we only use import at the top of the file for consistency with the rest of the app please?

paywall/src/data-iframe/start/makePurchaseCallback.js
```diff
@@ -0,0 +1,44 @@
...
+    ])
+    const startingKey = keys[lockAddress]
+
+    Promise.all([
+      purchaseKey({
+        walletService,
+        lockAddress,
+        amountToSend: locks[lockAddress].keyPrice,
+      }),
+      processKeyPurchaseTransactions({
```
julien51: I have to say that generally I am quite confused by the approach we've taken on these. It is strange that we call `processKeyPurchaseTransactions` to monitor a transaction which another promise is supposed to trigger.

I know this is a larger theme than this PR and I should have raised earlier, but what happens if `purchaseKey` fails?

paywall/src/__tests__/data-iframe/start/makePurchaseCallback.test.js
```diff
@@ -0,0 +1,128 @@
...
+        })
+      })
+      expect(walletService.purchaseKey).toHaveBeenCalledWith(
+        'lock',
+        'account',
+        '123'
+      )
+    })
+
+    it('should initiate monitoring of key purchase transaction', async () => {
```
julien51: Would it be the right place to test what happens when the key purchase transaction actually fails to be sent?

paywall/src/data-iframe/start/makePurchaseCallback.js
cellog: absolutely not! this is the most important implementation detail that makes the code splitting of the cache from the blockchain handler possible (see https://webpack.js.org/guides/code-splitting/#dynamic-imports).

I'll add a series of comments to the code since this is not an intuitive code pattern.

paywall/src/data-iframe/start/makePurchaseCallback.js
cellog: if purchaseKey fails, the promise throws, and throws out of the purchase callback. Then, we can do a simple try/catch and pass the error to the UI

paywall/src/__tests__/data-iframe/start/makePurchaseCallback.test.js
cellog: yes, we can test that throw errors cascade out

paywall/src/data-iframe/start/makePurchaseCallback.js
cellog: I have pushed a commit with the comment, and also 2 new tests for showing that any failure propagates out of the function.

Regarding the larger question, the approach taken is exactly the same as having walletService trigger the transaction, and web3Middleware start monitoring it through web3Service. Which is to say 2 separate entities, one does the writing, and the other monitors the transaction. The biggest difference is that the code here is written in a way that the logic flows from the top of the file (or function) to the bottom in chronological order, whereas the redux/middleware approach is entirely event-based.

cellog: note to self: add lots of comments

cellog:
ok @julien51 this is ready for a look

PR #3411, opened 2019-05-28
Moving typescripts dependencies at the root

julien51:
# Description

this should make our builds slightly faster and our images a bit smaller!

# Checklist:


locksmith/package.json
```diff
@@ -40,12 +40,11 @@
     "@types/ethereumjs-util": "^5.2.0",
     "@types/express": "^4.16.1",
     "@types/express-serve-static-core": "^4.16.2",
+    "@types/jest": "^24.0.13",
     "@types/node": "^12.0.0",
     "@types/sequelize": "^4.27.44",
     "@types/stripe": "^6.25.10",
     "@types/supertest": "^2.0.7",
-    "@types/jest": "^24.0.13",
-    "@typescript-eslint/eslint-plugin": "^1.5.0",
```
akeem: Was it intended to remove this from locksmith?

julien51: yeah they are in the root.

julien51: There is probably more cleanup to do as well. One step at a time ;)

julien51:
I have to find the dependencies which are strictly used by the binaries `jest`, `typescript`... etc (which can be moved to the root and the ones which are loaded by the code directly.

PR #3444, opened 2019-05-29
Unified error handling

### LINKED ISSUE #3332: Unify error handling

cnasc:
# Description

This PR is the final (large) part of #3332. Now every error we dispatch is handled (either by GlobalErrorConsumer or the Errors component), unless we specifically mark it as a Diagnostic error.

I've tried to organize this in a pretty easy flow, so reviewing commit-by-commit will probably be easiest. I'll add github comments where necessary to add context.

# Issues

Fixes #3332 

# Checklist:


unlock-app/src/actions/error.ts
```diff
@@ -0,0 +1,19 @@
...
+  type: SET_ERROR,
+  error,
+})
+
+export const resetError = (error: UnlockError) => ({
+  type: RESET_ERROR,
+  error,
+})
+
+export const clearAllErrors = () => ({
```
cnasc: This is new -- we used to have some behavior that cleared all errors if `resetError` was called with an undefined argument. I'd rather this behavior (so far unused) be explicit, so I've introduced a new action.

unlock-app/src/middlewares/storageMiddleware.js
```diff
@@ -18,15 +17,21 @@ import {
...
   const { services } = config
   return ({ getState, dispatch }) => {
     const storageService = new StorageService(services.storage.host)
 
     // NEW_TRANSACTION
-    storageService.on(failure.storeTransaction, error => {
-      dispatch(storageError(error))
+    storageService.on(failure.storeTransaction, () => {
+      // TODO: we are in control of what storageService emits --
+      // construct helpful errors at source of failure?
```
cnasc: Right now many of these errors are constant, because digging through the services and making sure they all return a useful error message was a bit out of scope for this PR. Future work can include making sure we construct useful messages to emit, and then the diagnostics can be more specific.

unlock-app/src/utils/Error.ts
```diff
@@ -83,6 +84,8 @@ const SignUp: ErrorMakers = errorsFor('SignUp')
 // Transaction errors (failed to create lock, etc.)
 const Transaction: ErrorMakers = errorsFor('Transaction')
 
+const Web3: ErrorMakers = errorsFor('Web3')
```
cnasc: I had previously forgotten to include a category of errors for this, so I threw them in here

unlock-app/src/components/interface/Errors.tsx
```diff
@@ -4,18 +4,20 @@ import styled from 'styled-components'
...
 import ErrorMessage from '../helpers/ErrorMessage'
-import { Error } from '../../unlockTypes' // eslint-disable-line no-unused-vars
+/* eslint-disable */
+import { UnlockError, WarningError, isWarningError } from '../../utils/Error'
+/* eslint-enable */
 
 interface Props {
-  errors: Error[]
-  close: (errorName: string) => any
+  errors: WarningError[]
```
cnasc: Warnings and fatal errors are now distinct types, and those types are handled by separate components

unlock-app/src/__tests__/middlewares/providerMiddleware.test.ts
```diff
@@ -2,6 +2,9 @@ import providerMiddleware from '../../middlewares/providerMiddleware'
 import { SET_PROVIDER } from '../../actions/provider'
 import { setError } from '../../actions/error'
 import { FATAL_MISSING_PROVIDER } from '../../errors'
+import Error from '../../utils/Error'
```
cellog: this is an odd construction, is there a reason you don't `import { Application } from '../../utils/Error'`?

unlock-app/src/__tests__/middlewares/storageMiddleware.test.js
```diff
@@ -15,6 +14,10 @@ import {
   GOT_ENCRYPTED_PRIVATE_KEY_PAYLOAD,
 } from '../../actions/user'
 import { success, failure } from '../../services/storageService'
+import Error from '../../utils/Error'
+import { setError } from '../../actions/error'
```
cellog: same question as above

unlock-app/src/__tests__/components/creator/FatalError.test.tsx
```diff
@@ -8,56 +8,64 @@ import {
...
       it('FATAL_MISSING_PROVIDER', () => {
-        expect.assertions(1)
-        const component = mapErrorToComponent(FATAL_MISSING_PROVIDER, {})
+        expect.assertions(0)
+        const component = mapErrorToComponent(
+          Application.Fatal(FATAL_MISSING_PROVIDER)
+        )
         const wrapper = rtl.render(component)
-        expect(wrapper.queryByText('Wallet missing')).not.toBeNull()
+        wrapper.getByText('Wallet missing')
```
cnasc: This is a small thing, but having 0 assertions with a `getByText` is a bit easier to debug than checking that a query is not null. `getByText` will show what was actually rendered in the failure, and it throws if it doesn't find anything (so the test will fail if the query fails)

unlock-app/src/__tests__/middlewares/walletMiddleware.test.js
```diff
@@ -565,7 +591,7 @@ describe('Wallet middleware', () => {
   })
 
   describe('UPDATE_LOCK_KEY_PRICE', () => {
-    it('when the service is not ready it should set an error and not try to update the ley price', () => {
+    it('when the service is not ready it should set an error and not try to update the key price', () => {
```
cellog: typo-fixing machine!

unlock-app/src/__tests__/reducers/errorsReducer.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    expect(reducer([error], action)).toEqual([error])
+  })
+
+  it('should not change state if RESET_ERROR with non-existing error is called', () => {
+    expect.assertions(1)
+    const state = [error]
+    expect(reducer(state, resetError(error2))).toBe(state)
+  })
+
+  it('should reset all errors if CLEAR_ALL_ERRORS with no specific error received', () => {
```
cellog: remove "with no specific error"?

unlock-app/src/components/creator/FatalError.tsx
```diff
@@ -21,19 +33,15 @@ export const DefaultError = ({ title, children, illustration, critical }) => (
...
-  critical: PropTypes.bool,
-}
-
-DefaultError.defaultProps = {
-  illustration: '/static/images/illustrations/error.svg',
-  title: 'Fatal Error',
-  children: defaultError,
-  critical: true,
-}
+export const FallbackError = () => (
```
cnasc: Easier to add this separate component as the fallback than to compute default values in DefaultError (that would require optional properties and would remove some type system helpfulness)

unlock-app/src/components/creator/FatalError.tsx
```diff
@@ -1,8 +1,9 @@
-import PropTypes from 'prop-types'
-
 import React from 'react'
 import styled from 'styled-components'
 import { ETHEREUM_NETWORKS_NAMES } from '../../constants'
+/* eslint-disable */
```
cellog: I think I understand - you need to copy the eslint config from paywall. This will fix this annoying problem

unlock-app/src/components/creator/FatalError.tsx
```diff
@@ -120,39 +130,52 @@ export const MissingAccount = () => (
...
-  // if the error condition exists, set it to the mapped fatal error component
-  // or to the fallback
-  // if no error exists, set it to false
-  const Error = error
-    ? overrideMapping[error] || mapping[error] || mapping['*']
-    : false
-
-  return Error && <Error {...errorMetadata} />
+  const { message } = error
+  if (message === 'FATAL_WRONG_NETWORK') {
```
cnasc: This is a special case -- it's the only error that carries the separate data payload for the time being. I'd like a cleaner solution to this, but I think it may be out of scope for this PR and this is good enough to hold us for a while.

cnasc:
> This is great work. It really is time to fix the unused variable problem.
> 
> Let's copy this portion from the paywall eslint.rc:
> 
> ```js
>   rules: {
>     'no-unused-vars': 'off',
>     '@typescript-eslint/no-unused-vars': [
>       'error',
>       {
>         vars: 'all',
>         args: 'after-used',
>         ignoreRestSiblings: true,
>       },
>     ],
> ```

Does this ignore all instances of unused variables or just the times when we import types?

unlock-app/src/__tests__/middlewares/providerMiddleware.test.ts
cnasc: You know, in the heat of the moment I forgot that I exported all of those separately as well. I'll update those.

cellog:
> > This is great work. It really is time to fix the unused variable problem.
> > Let's copy this portion from the paywall eslint.rc:
> > ```js
> >   rules: {
> >     'no-unused-vars': 'off',
> >     '@typescript-eslint/no-unused-vars': [
> >       'error',
> >       {
> >         vars: 'all',
> >         args: 'after-used',
> >         ignoreRestSiblings: true,
> >       },
> >     ],
> > ```
> 
> Does this ignore all instances of unused variables or just the times when we import types?

It makes the thing work the way it should: complains when variables are unused, not when we import types

unlock-app/src/__tests__/reducers/errorsReducer.test.ts
cnasc: 👍

Issue #3366, opened 2019-05-25
Ad remover paywall: Checkout iframe


cellog:
The checkout iframe needs to do these things:

1. make a new next.js page `checkout.js`
2. set up the netlify `/static/_redirects`
3. add to `src/_server.js`
4. create a new `src/content/CheckoutContent.js`
5. create a new `src/pages/Checkout.js`
6. use the `useListenForPostMessage` and `usePostMessage` and `usePaywallConfig` hooks to get the necessary data to display the checkout component, create a `purchaseKey` callback based on `postMessage`
7. pass the data and the callback down to the `Checkout` component

PR #3445, opened 2019-05-29
Tighten types, rename a callback

cellog:
# Description

This primarily cosmetic change tightens the `Account` type and the types for some callbacks, and renames `clickOnConfirmedLock` to `hideCheckout` since it will also be used when the user clicks the little `X` to close in the upper right corner

# Issues

Refs #3366

# Checklist:

### DIFF
--- a/paywall/src/components/checkout/Checkout.tsx
+++ b/paywall/src/components/checkout/Checkout.tsx
@@ -8,9 +8,9 @@ import CheckoutLock from './CheckoutLock'
 interface Props {
   locks: Locks
   config: PaywallConfig
-  account: Account
+  account: Account | null
   purchase: (...args: any[]) => any
-  clickOnConfirmedLock: (...args: any[]) => any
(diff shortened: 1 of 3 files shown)


paywall/src/components/checkout/Checkout.tsx
```diff
@@ -8,17 +8,17 @@ import CheckoutLock from './CheckoutLock'
 interface Props {
   locks: Locks
   config: PaywallConfig
-  account: Account
+  account: Account | null
```
cnasc: Just making sure I understand, it looks like `Account` is passed down through these solely to use in constructing `NoKeyLock`. Is `Account` something that `NoKeyLock` could fetch on its own (from state, etc.)?

cellog: we aren't using redux in the checkout component, so all state is passed down through props. Otherwise, the answer would have been yes :)

PR #3450, opened 2019-05-29
fix startup for unlock.min.js

cellog:
# Description

this adds an exist case for iframeManager polling to re-add the iframe if it gets removed, and also fixes a logic flaw in the index.js where the iframe setup was assuming that it was in the unlock-protocol domain. Now, it passes the correct full URL with the `PAYWALL_URL` and the window.origin added.

It also adds the iframes to the document (whoops)

# Issues

Refs #2903

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/index.js
+++ b/paywall/src/unlock.js/index.js
@@ -1,7 +1,20 @@
-import { makeIframe } from './iframeManager'
+import { makeIframe, addIframeToDocument } from './iframeManager'
 import setupPostOffices from './setupPostOffices'
+import '../paywall-builder/iframe.css'
 
-const dataIframe = makeIframe(window, '/static/dataIframe.html')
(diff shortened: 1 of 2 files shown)


paywall/src/unlock.js/index.js
```diff
@@ -1,7 +1,20 @@
...
 
-const dataIframe = makeIframe(window, '/static/dataIframe.html')
-const checkoutIframe = makeIframe(window, '/checkout')
+window.onload = () => {
+  const origin = '?origin=' + encodeURIComponent(window.origin)
 
-setupPostOffices(window, dataIframe, checkoutIframe)
+  const dataIframe = makeIframe(
+    window,
+    process.env.PAYWALL_URL + '/static/dataIframe.html' + origin
```
julien51: Let's not make case specific files. This is a nightmare to manage because many OS do different things. can we just use `data-iframe.js`?

PR #3464, opened 2019-05-30
paywall: update types for Lock and Key

cellog:
# Description

This updates the types for `Lock` and `Key` to reflect reality.

# Issues

Refs #2903

# Checklist:

### DIFF
--- a/paywall/src/unlockTypes.ts
+++ b/paywall/src/unlockTypes.ts
@@ -91,6 +91,11 @@ export interface Lock {
   keyPrice: string
   expirationDuration: number
   key: Key
+  asOf?: number
+  maxNumberOfKeys?: number
+  outstandingKeys?: number
+  balance: string
+  owner: string
 }
 
 export interface Locks {
@@ -102,6 +107,6 @@ export interface Key {
   transactions: Transaction[]
   status: string
   confirmations: number
-  owner: string
+  owner: string | null
   lock: string
 }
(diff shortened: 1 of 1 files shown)


paywall/src/unlockTypes.ts
```diff
@@ -102,6 +107,6 @@ export interface Key {
   transactions: Transaction[]
   status: string
   confirmations: number
-  owner: string
+  owner: string | null
```
cnasc: In which cases do keys not have owners?

paywall/src/unlockTypes.ts
```diff
@@ -91,6 +91,11 @@ export interface Lock {
   keyPrice: string
   expirationDuration: number
   key: Key
+  asOf?: number
```
cnasc: Not something that has to be part of this PR, but in the future we may want to have a lock be any of a union type of different lock states, so we can encode when we expect the properties that are optional here to be set

paywall/src/unlockTypes.ts
cellog: When the user has no wallet whatsoever. In that case the key is a dummy value used only for display of the lock

paywall/src/unlockTypes.ts
cellog: I don't follow (typescript novice), looking forward to picking your brain on this in the near future!

PR #3469, opened 2019-05-30
fix issues in unlock.min.js index.js

cellog:
# Description

This fixes these issues:

1. don't try anything until the window is loaded
2. the URLs for the iframes need both a host and the origin passed in for postMessage to work
3. we have to actually add the iframes to the DOM
4. profit

Note that this is the only untested file in the script, so a follow-up will be to extract this into another file that is called from index.js and to test it. But this is good for now

# Issues

Refs #2903

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/index.js
+++ b/paywall/src/unlock.js/index.js
@@ -1,7 +1,20 @@
-import { makeIframe } from './iframeManager'
+import { makeIframe, addIframeToDocument } from './iframeManager'
 import setupPostOffices from './setupPostOffices'
+import '../paywall-builder/iframe.css'
 
-const dataIframe = makeIframe(window, '/static/dataIframe.html')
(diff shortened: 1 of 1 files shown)


paywall/src/unlock.js/index.js
```diff
@@ -1,7 +1,20 @@
...
 
-setupPostOffices(window, dataIframe, checkoutIframe)
+  const dataIframe = makeIframe(
+    window,
+    process.env.PAYWALL_URL + '/static/dataIframe.html' + origin
+  )
+  addIframeToDocument(window, dataIframe)
+  const checkoutIframe = makeIframe(
+    window,
+    process.env.PAYWALL_URL + '/checkout' + origin
```
akeem: Not a deal breaker but we might be able to use some of the facilities available to construct this URL from its components

PR #3472, opened 2019-05-30
balance is in eth, and can be a decimal number

cellog:
# Description

When designing `useBlockchainData`, I assumed the account balance was in wei, but it is in fact in eth, thus we need to support decimal numbers as well. This PR fixes that.

# Issues

Refs #2903

# Checklist:


cellog:
> Looks good, what was the behavior when the validation failed previously?

when validation fails, the hook ignores the message, so balance was not getting to the frontend. Designed it this way so that postMessage wouldn't become a DOS vector

PR #3474, opened 2019-05-30
make sure owner is set in getKeys

cellog:
# Description

It turns out that the key returned from `web3Service.getKeysByLockForOwner` does not set the owner field, so we have to do it manually.

# Issues

Refs #2903

# Checklist:


cellog:
> LGTM. We should make note of these things that are not set in the service for future inclusion -- I think it's a bit counterintuitive that sometimes a lock does not contain all lock fields.

do you mind opening an issue for us? Doesn't have to be too specific, I think it applies to both keys and locks, but not transactions

PR #3512, opened 2019-06-03
add handy debugging of postMessage statements

cellog:
# Description

This adds debugging of postMessage statements. Note that it is a no-op (all the code below is removed by webpack's dead code elimination) until #3511 is merged.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/paywall/src/utils/postOffice.js
+++ b/paywall/src/utils/postOffice.js
@@ -16,7 +16,8 @@
  * @param {string} targetOrigin the origin of the target (in CORS settings
  *                              this cannot be retrieved from the target)
  */
-export function setupPostOffice(window, target, targetOrigin) {
+export function setupPostOffice(window, target, targetOrigin, local, remote) {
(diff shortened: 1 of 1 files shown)


paywall/src/utils/postOffice.js
```diff
@@ -34,10 +35,28 @@ export function setupPostOffice(window, target, targetOrigin) {
     const listeners = handlers[event.data.type]
     if (listeners && listeners.size) {
       listeners.forEach(listener => {
+        if (debug) {
```
cnasc: Do you think these could be factored into a `debug()` function that just wraps the `if (debug)` and `console.log` parts?

cellog: I do. However, this will have a performance impact even when debug is off, as the function will always be called, and not eliminated by dead code removal, so I elected not to do this

cnasc: Fair enough

Issue #2414, opened 2019-03-28
Create Unlock Provider as part of unlock-js


cnasc:
- [x] Create provider middleware to manage initialization of providers (#2470)
- [x] Integrate provider middleware with app (#2487)
- [ ] Build and test Unlock Provider to match spec document (not yet public)

julien51:
Turning this into an Epic to keep track of it more easily (it stays inside User Account and unlock-js library epics too!)

PR #3514, opened 2019-06-03
Add a method that handles eth_signTypedData

cnasc:
# Description

This PR adds a method to respond to the `eth_signTypedData` JSON-RPC call. In the process of doing this, I noticed that the way `unlock-provider` was structured no longer meshes well with `walletService`, so this PR also makes `unlock-provider` consistent with the new way of working.

@akeem I would appreciate a sanity check that I am doing the correct kind of signature here.

Relevant docs:
https://docs.ethers.io/ethers.js/html/api-wallet.html#signing
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#eth_signtypeddata

# Issues

Refs #2414

# Checklist:

### DIFF
--- a/unlock-js/src/unlockProvider.js
+++ b/unlock-js/src/unlockProvider.js
@@ -33,33 +33,22 @@ export default class UnlockProvider {
     return true
   }
 
-  async eth_accounts(respond) {
+  async eth_accounts() {
     // Must always return an array of addresses
     if (this.wallet) {
-      respond([this.wallet.address])
+      return [this.wallet.address]
     } else {
-      respond([])
+      return []
     }
   }
 
-  send(args, cb) {
-    const { id, jsonrpc, method } = args
-    // Calling respond with some argument will call the callback with
-    // a "fake" JSON-RPC response constructed by the provider.
-    const respond = result => {
-      cb(null, {
-        id,
-        jsonrpc,
-        result,
-      })
-    }
-
+  async send(method, params) {
     try {
-      return this[method](respond)
+      return this[method](method, params)
     } catch (err) {
       // We haven't implemented this method, defer to the fallback provider.
       // TODO: Catch methods we don't want to dispatch and throw an error
-      return this.fallbackProvider.send(args, cb)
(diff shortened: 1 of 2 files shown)


unlock-js/src/unlockProvider.js
```diff
@@ -33,33 +33,29 @@ export default class UnlockProvider {
...
-        id,
-        jsonrpc,
-        result,
-      })
-    }
+  async eth_signTypedData(params) {
+    // params is [ account, data ]
+    // we don't need account
+    const data = params[1]
+    return this.wallet.signMessage(data)
```
akeem: what is the type of `this.wallet`?  In general this seems okay but I am not able to glean a lot of context from this delta.

If this.wallet.signMessage is the method for handling erc712 signatures than it makes sense

cnasc: `this.wallet` is an ethers wallet generated from the private key, it wraps over various account-related functionality. Just wasn't 100% sure if there's actually a big difference when it comes to signing typed data, since only 2 signature methods are provided.

akeem: I see, i'm not sure that ethers supports creating eip712 signatures. This is the most relevant issues that I found https://github.com/ethers-io/ethers.js/issues/98.

akeem: Not sure if this is helpful, here is a link to a somewhat generic implementation of the signature method https://github.com/MetaMask/eth-sig-util

unlock-js/src/unlockProvider.js
```diff
@@ -33,33 +33,29 @@ export default class UnlockProvider {
...
+  }
 
+  async send(method, params) {
     try {
-      return this[method](respond)
+      return this[method](method, params)
```
cellog: it will be more performant to use a simple if as in:

```js
if (Object.hasOwnProperty(method)) return this[method](method, params)
```

cnasc:
There's some setup involved on this -- I'm going to close and reopen in pieces

PR #3520, opened 2019-06-03
Remove outdated assumptions from unlock-provider

cnasc:
# Description

This PR updates `unlock-provider` to work the way that `walletService` expects it to, returning values from async functions instead of using a callback.

# Issues

Refs #3514 #2414

# Checklist:

### DIFF
--- a/unlock-js/src/unlockProvider.js
+++ b/unlock-js/src/unlockProvider.js
@@ -33,33 +33,21 @@ export default class UnlockProvider {
     return true
   }
 
-  async eth_accounts(respond) {
+  async eth_accounts() {
     // Must always return an array of addresses
     if (this.wallet) {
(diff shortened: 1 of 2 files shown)


unlock-js/src/unlockProvider.js
```diff
@@ -33,33 +33,21 @@ export default class UnlockProvider {
...
-        jsonrpc,
-        result,
-      })
-    }
-
-    try {
-      return this[method](respond)
-    } catch (err) {
+  async send(method, params) {
+    if (typeof this[method] === 'undefined') {
```
cnasc: @cellog it looks like methods on ES6 classes don't get defined in a way that `hasOwnProperty` works with, but this should be more or less equivalent and avoid the overhead of a `try/catch`

PR #3534, opened 2019-06-03
Links and small text edits

smombartz:
added links to the tickets page and NFT event blog post so people can easily see it in action.


julien51:
Ship it!

PR #3546, opened 2019-06-04
[DO NOT MERGE]

cellog:
This is a diff of all the needed changes. It will be referred to in issues created.


cellog:
done! thanks @benwerd

PR #3554, opened 2019-06-04
Extend an expired key on re-purchase

hardlydifficult:
# Description

Fixes a bug where purchasing an expired key can result in still having an expired key.  Now on your second+ purchase the key will have an expiration date that is as good as or better than a brand new key.

I confirmed that if you undo the contract change the test I added fails.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinPurchase.sol
+++ b/smart-contracts/contracts/mixins/MixinPurchase.sol
@@ -86,22 +86,17 @@ contract MixinPurchase is
 
     // Assign the key
     Key storage toKey = _getKeyFor(_recipient);
-    uint previousExpiration = toKey.expirationTimestamp;
 
     if (toKey.tokenId == 0) {
       // Assign a new tokenId (if a new owner or previously transfered)
       _assignNewTokenId(toKey);
+      _recordOwner(_recipient, toKey.tokenId);
     }
 
-    if(toKey.expirationTimestamp >= block.timestamp) {
+    if (toKey.expirationTimestamp >= block.timestamp) {
       // This is an existing owner trying to extend their key
-      toKey.expirationTimestamp = previousExpiration.add(expirationDuration);
+      toKey.expirationTimestamp = toKey.expirationTimestamp.add(expirationDuration);
     } else {
-      if(toKey.expirationTimestamp == 0) {
-        // A non-zero expirationTimestamp informs us this is a brand new owner
-        // (since transferFrom may clear the tokenId)
-        _recordOwner(_recipient, toKey.tokenId);
-      }
       // SafeAdd is not required here since expirationDuration is capped to a tiny value
       // (relative to the size of a uint)
       toKey.expirationTimestamp = block.timestamp + expirationDuration;
--- a/smart-contracts/contracts/mixins/MixinPurchase.sol
+++ b/smart-contracts/contracts/mixins/MixinPurchase.sol
@@ -84,21 +84,20 @@ contract MixinPurchase is
       netPrice = inMemoryKeyPrice - discount;
     }
 
-    // We explicitly allow for greater amounts of ETH to allow 'donations'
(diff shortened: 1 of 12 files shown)


smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -84,21 +84,20 @@ contract MixinPurchase is
...
-    // We explicitly allow for greater amounts of ETH to allow 'donations'
-    _chargeAtLeast(netPrice);
-
     // Assign the key
     Key storage toKey = _getKeyFor(_recipient);
     uint previousExpiration = toKey.expirationTimestamp;
-    if (toKey.tokenId == 0) {
-      _assignNewTokenId(toKey);
-      _recordOwner(_recipient, toKey.tokenId);
+    if(toKey.expirationTimestamp >= block.timestamp) {
```
cellog:
```suggestion
    if (toKey.expirationTimestamp >= block.timestamp) {
```

hardlydifficult: thanks

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -84,21 +84,20 @@ contract MixinPurchase is
...
-    // We explicitly allow for greater amounts of ETH to allow 'donations'
-    _chargeAtLeast(netPrice);
-
     // Assign the key
     Key storage toKey = _getKeyFor(_recipient);
     uint previousExpiration = toKey.expirationTimestamp;
-    if (toKey.tokenId == 0) {
-      _assignNewTokenId(toKey);
-      _recordOwner(_recipient, toKey.tokenId);
+    if (toKey.expirationTimestamp >= block.timestamp) {
```
nfurfaro: What if `toKey.expirationTimestamp <= block.timestamp` && `toKey.tokenId != 0` ?

nfurfaro: What if `toKey.expirationTimestamp <= block.timestamp` && `toKey.tokenId != 0` ?

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -84,21 +84,20 @@ contract MixinPurchase is
...
+      // This is an existing owner trying to extend their key
+      toKey.expirationTimestamp = previousExpiration.add(expirationDuration);
+    } else {
+      if (toKey.tokenId == 0) {
+        _assignNewTokenId(toKey);
+        _recordOwner(_recipient, toKey.tokenId);
+      }
       // SafeAdd is not required here since expirationDuration is capped to a tiny value
       // (relative to the size of a uint)
       toKey.expirationTimestamp = block.timestamp + expirationDuration;
```
nfurfaro: won't this line be executed even if line #92 is also executed?

nfurfaro: it seems like we want either line 92 or line 100 to be executed...not both.

hardlydifficult: Na, we are nested inside an `else` block here.

hardlydifficult:
Not sure why this caused some transfer tests to fail... investigating..

hardlydifficult:
oh, this change conflicts with MixinTransfer.sol line 91.  I'll try a different approach.

smart-contracts/contracts/mixins/MixinPurchase.sol
nfurfaro: yup.

nfurfaro:
might need something like `if toKey.expirationTimestamp < block.timestamp && toKey.tokenId != 0`

hardlydifficult:
k should be good to go - could you take another look?

PR #3555, opened 2019-06-04
adding toDecimal and fromDecimal

julien51:
# Description

This is required to show accurate numbers on balance and key prices for ERC20 locks!

# Checklist:

### DIFF
--- a/unlock-js/src/utils.js
+++ b/unlock-js/src/utils.js
@@ -4,6 +4,8 @@ import { ethers } from 'ethers'
 // or to migrate to a totally different library and have a single point of modification
 export default {
   toWei: (value, units) => ethers.utils.parseUnits(value, units),
+  // This converts a string representation from a value to a number of units, based on the number of decimals passed in
+  toDecimal: (value, decimals) => ethers.utils.parseUnits(value, decimals),
   hexlify: ethers.utils.hexlify,
   hexStripZeros: ethers.utils.hexStripZeros,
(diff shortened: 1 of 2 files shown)


unlock-js/src/utils.js
```diff
@@ -17,6 +18,11 @@ export default {
       .formatUnits(ethers.utils.bigNumberify(num), units)
       .replace(/\.0$/, '')
   },
+  fromDecimal: (num, decimals) => {
+    return ethers.utils
+      .formatUnits(ethers.utils.bigNumberify(num), decimals)
+      .replace(/\.0$/, '')
```
akeem: Is there not a way to have `formatUnits` return an integer?

julien51: I wanted to maintain consistency with the existing `toWei` and `fromWei` function.

julien51:
Good point regarding comments. Adding them now.

unlock-js/src/utils.js
cellog: from https://docs.ethers.io/ethers.js/html/api-utils.html

"Format an amount of wei into a decimal string representing the amount of units. The output will always include at least one whole number and at least one decimal place, otherwise leading and trailing 0’s will be trimmed." `formatUnits` annoyingly always includes a decimal place

julien51: which do the same thing but for ether decimals (wei, gwei, szabo, .... etc)

julien51:
@akeem PTAL

PR #3559, opened 2019-06-04
Integrate UnlockProvider with app

cnasc:
# Description

This PR moves handling of user account log in to `providerMiddleware`, which instantiates `UnlockProvider` with the user's decrypted wallet, then sets the account address in state and dispatches `providerReady`.

This PR does _not_ add `UnlockProvider` to the config, since I think we don't want it to be used yet (once we add `UnlockProvider` to config, it will always be present and there will not be an error message for people without MetaMask; we don't want that to happen until `UnlockProvider` is more capable)

To do a live test of `UnlockProvider` in this branch, or after it is merged, you have to import `UnlockProvider` in the config, disable the `Web3` and `HTTP` providers, and after the `readOnlyProvider` is set, do `providers['Unlock'] = new UnlockProvider({ readOnlyProvider })`.

Then you can go to the signup or login page and try it out.

# Issues

Refs #2414
# Checklist:

### DIFF
--- a/unlock-app/src/middlewares/providerMiddleware.ts
+++ b/unlock-app/src/middlewares/providerMiddleware.ts
@@ -55,7 +55,9 @@ export async function initializeUnlockProvider(
     dispatch(setAccount({ address }))
     dispatch(setEncryptedPrivateKey(key, emailAddress))
     dispatch(providerReady())
-  } catch (_) {
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error(e)
     // TODO: password isn't the only thing that can go wrong here...
     dispatch(
       setError(
--- a/unlock-app/src/middlewares/providerMiddleware.ts
+++ b/unlock-app/src/middlewares/providerMiddleware.ts
@@ -4,10 +4,20 @@ import {
   FATAL_MISSING_PROVIDER,
   FATAL_NOT_ENABLED_IN_PROVIDER,
 } from '../errors'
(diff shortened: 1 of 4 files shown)


unlock-app/src/middlewares/providerMiddleware.ts
```diff
@@ -22,13 +33,41 @@ function initializeProvider(provider: { enable?: () => any }, dispatch: any) {
...
+  const { key, emailAddress, password } = action
+  try {
+    await unlockProvider.connect({ key, password })
+
+    const address = unlockProvider.wallet.address
+    dispatch(setAccount({ address }))
+    dispatch(setEncryptedPrivateKey(key, emailAddress))
+    dispatch(providerReady())
+  } catch (_) {
+    // TODO: password isn't the only thing that can go wrong here...
```
cellog: can we console.error the error? it makes debugging ever so much more possible

cnasc: Do our lint rules not complain about that? There are actually a bunch of places where we could log that out if it's allowed.

cellog: oh if you do:

```js
// eslint-disable-next-line no-console
console.error(e)
```

Then lint doesn't complain :)

cellog: I think this is a good use case for breaking the lint rules, in other words

PR #3575, opened 2019-06-04
fail if UNLOCK_ENV is not explicitly set

cellog:
# Description

This re-vamps the generation scripts for `unlock.1.0.min.js` and `data-iframe.1.0.min.js` to be stricter. You must declare `UNLOCK_ENV`. It will still load `.env` files, and use the contents to set variables, but will fail on Circle if the env variables are not set.

# Issues

Refs #2903

# Checklist:


cellog:
This is the minimal implementation. The first process.env.UNLOCK_ENV is from the environment. The second is from the dotenv file, if any.

julien51:
Let's close as this does not solve our problem ;)

PR #3609, opened 2019-06-06
Add user account settings page

### LINKED ISSUE #2166: Allow users to change their passwords

cnasc:
# Description

This PR introduces the user account settings page, which at the moment will enable password change.

image

This page will also host the payment information form once that's written.

# Issues

Fixes #2166 

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/components/interface/account-settings/ChangePassword.test.tsx
+++ b/unlock-app/src/__tests__/components/interface/account-settings/ChangePassword.test.tsx
@@ -1,0 +1,96 @@
+import React from 'react'
+import * as rtl from 'react-testing-library'
+import {
+  ChangePassword,
+  validatePassword,
+  passwordErrors,
+  mapDispatchToProps,
+} from '../../../../components/interface/account-settings/ChangePassword'
+
+describe('ChangePassword component', () => {
+  describe('Password validation', () => {
+    it('should not allow blank passwords', () => {
+      expect.assertions(1)
+      expect(validatePassword('', '')).toEqual([
+        passwordErrors.EMPTY,
+        passwordErrors.MID_LENGTH,
+      ])
+    })
+
+    it('should not allow passwords that are different from their confirmations', () => {
+      expect.assertions(1)
+      expect(
+        validatePassword('AVeryGoodPassword', 'PGeryVoodAassword')
+      ).toEqual([passwordErrors.NO_MATCH])
+    })
+  })
+
+  describe('mapDispatchToProps', () => {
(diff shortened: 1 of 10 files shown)


cnasc:
Once #3641 is merged, this will be ready for a rebase and review

cnasc:
Looks like there's a bit of an issue with the providerMiddleware, after logging in `SET_ACCOUNT` et al get called an additional time, which wipes out the `userDetails` in the state. Looking into it.

cnasc:
OK, so the issue is that `SET_NETWORK` gets triggered when `walletService` connects to the provider. We do need to retain the `userDetails` object, and my assumption is that in real usage with `UnlockProvider` the `SET_NETWORK` and `SET_ACCOUNT` actions that  typically clear the Redux state should not clear the `userDetails`.

This is because we have more control over `UnlockProvider` than we do with an external wallet; we don't have to worry about users switching around between Rinkeby and Mainnet. Changing account would only happen in the event of a logout/login. So I propose that the fix to this is the following:

```typescript
// FILE => unlock/unlock-app/src/reducers/privateKeyReducer.ts

const privateKeyReducer = (
  state: State = initialState,
  action: Action
): State => {
// TODO: remove following condition
  if ([SET_PROVIDER, SET_NETWORK, SET_ACCOUNT].indexOf(action.type) > -1) {
    return initialState
  }
// end of code to remove

  if (action.type === SET_ENCRYPTED_PRIVATE_KEY) {
    return {
      key: action.key,
      email: action.emailAddress,
    }
  }

  return state
}
```

Thoughts on this approach @julien51?

unlock-app/src/__tests__/components/interface/account-settings/ChangePassword.test.tsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react'
+import * as rtl from 'react-testing-library'
+import {
+  ChangePassword,
+  validatePassword,
```
cellog: is there anywhere else validatePassword would be used? If so, it could go in `utils/`

cellog: I would suggest opening up an issue to refactor this in a future PR, not in this one.

cnasc: Yup, that's something I had in mind to do as part of the broader password refactor (since the current implementation is about as rudimentary as possible).

PR #3623, opened 2019-06-07
Allow walletService to properly connect to UnlockProvider

### LINKED ISSUE #3611: Special case UnlockProvider in walletService
`UnlockProvider` has an API that matches the method call syntax used by the `ethers` provider wrapper. Unfortunately, `UnlockProvider` is itself wrapped by `ethers`, so it currently receives params in the JSON-RPC style, which causes it to fail. We need to add a line in the `connect()` method of `walletService` which will avoid wrapping `UnlockProvider`.

See #3609

cnasc:
# Description

Fixing the issue raised yesterday around the API boundary between `walletService` and `UnlockProvider`.

# Issues

Fixes #3611

# Checklist:

### DIFF
--- a/unlock-js/CHANGELOG.md
+++ b/unlock-js/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changes
 
+## 0.2.5
+- Allow `walletService` to properly connect to `UnlockProvider`
+
 ## 0.2.4
 
 - Increased gas limit to deploy Unlock contract
--- a/unlock-js/package.json
+++ b/unlock-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@unlock-protocol/unlock-js",
-  "version": "0.2.4",
+  "version": "0.2.5",
   "description": "This module provides libraries to include Unlock APIs inside a Javascript application.",
   "main": "lib/index.js",
   "module": "esm/index.js",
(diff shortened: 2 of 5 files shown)


cnasc:
The ad remover paywall should not be using UnlockProvider, and should probably have tests that help us identify breakage

cellog:
Basically, if you change the signature of `send` to:

```js
  async send({ method, params }, callback) {
```

That then makes it a web3-compatible provider. Unless there is something else I don't see, this would make the changes to walletService unnecessary, you can pass in an unlockProvider and it will just work.

Of course, you'd have to change the send method to something like:

```js
    if (typeof this[method] === 'undefined') {
      // We haven't implemented this method, defer to the fallback provider.
      // TODO: Catch methods we don't want to dispatch and throw an error
      try {
        const result = this.fallbackProvider.send(method, params)
        callback(null, result)
      } catch (e) {
        callback(e)
      }
    } else {
      const { result, error } = this[method](params)
      callback(error, result)
    }
```

cellog:
> The ad remover paywall should not be using UnlockProvider, and should probably have tests that help us identify breakage

then we definitely need to talk, assumptions are not in alignment - the same base code is going to be used for all paywalls

cnasc:
> > The ad remover paywall should not be using UnlockProvider, and should probably have tests that help us identify breakage
> 
> then we definitely need to talk, assumptions are not in alignment - the same base code is going to be used for all paywalls

You're right: here are my assumptions:

- The ad remover paywall does not use `UnlockProvider`, which is not complete enough for any app to use it. 
- We interact with providers through `walletService`

cnasc:
Per discussion with @julien51 and @akeem, including some comments about the future direction of this file to tighten up the abstractions used in here PTAL

PR #3630, opened 2019-06-07
Updates testing config for rover

akeem:
# Description

Adds configuration allowing test to actually run, and library missing to ensure that test do indeed pass.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/rover/src/storage.ts
+++ b/rover/src/storage.ts
@@ -1,0 +1,52 @@
+import "reflect-metadata";
+import { Transaction } from "./entity/Transaction";
+import { Block } from "./entity/Block";
+
+export class Storage {
+  connectionManager;
+
+  constructor(connectionManager) {
+    this.connectionManager = connectionManager;
+  }
+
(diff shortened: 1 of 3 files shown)


rover/src/storage.ts
```diff
@@ -0,0 +1,52 @@
+import "reflect-metadata";
```
cellog: rebase needed?

akeem: this is actually being used by typeorm https://stackoverflow.com/questions/49618719/why-typeorm-needs-reflect-metadata

cellog: no I mean this is not a configuration file, is it? (storage.ts)

cellog: (I thought I saw it in a previous PR)

akeem: this file was missing, its being added here to ensure that the test don't break when the configure is applied

PR #3640, opened 2019-06-07
update locks/account when state changes

### LINKED ISSUE #3626: Ad remover paywall: if a user is logged out of metamask, logs in, it does not refresh
**Describe the bug**
If a user is not logged into their metamask, and take longer than 10 seconds to log in, the ad remover paywall app will not detect the new account

**To Reproduce**
1. Log out of metamask
2. Go to https://staging-paywall.unlock-protocol.com/static/adremover/staging.html
3. Log back in, but slowly
4. The "wallet missing" modal does not change to the locks

cellog:
# Description

This fixes several account-related bugs in the ad remover paywall.

- account was being stored as-is in localStorage, so `null` got coerced to `"null"` which was an invalid account name (can only be `null` or a valid ethereum address)
- when account changed, locks/keys/transactions were not re-propagated to the main window. keys/transactions change, and since keys are stored in the lock, all 3 need to be re-sent
- the code incorrectly assumed `walletService.getAccount()` returns `null` on no account, but it returns `false`. In addition, `web3Service.getAccountBalance` returns `null` instead of `0` if the address is unknown.

The confluence of these bugs prevented switching the view.

out

# Issues

Fixes #3626 
Refs #2903

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/postOffice.js
+++ b/paywall/src/data-iframe/postOffice.js
@@ -70,6 +70,23 @@ export default function postOffice(window, requiredConfirmations) {
     },
   }
 
+  const lockHandler = cachedData => {
+    actions.locks(cachedData.locks)
+    const unlockedLocks = Object.values(cachedData.locks)
+      .filter(lock =>
+        ['submitted', 'pending', 'valid', 'confirming'].includes(
+          lock.key.status
+        )
+      )
+      .map(lock => lock.address)
+
+    if (unlockedLocks.length) {
+      actions.unlocked(unlockedLocks)
+    } else {
+      actions.locked()
+    }
+  }
+
   return {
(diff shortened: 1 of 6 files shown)


paywall/src/data-iframe/postOffice.js
```diff
@@ -70,6 +70,23 @@ export default function postOffice(window, requiredConfirmations) {
     },
   }
 
+  const lockHandler = cachedData => {
+    actions.locks(cachedData.locks)
+    const unlockedLocks = Object.values(cachedData.locks)
```
akeem: we should consider renaming this to something that indicated that returns a collection of lockAddresses

cellog: #3643 addresses this feedback

PR #3654, opened 2019-06-08
Hang on to account info when receiving SET_ACCOUNT, etc

cnasc:
# Description

This PR is an implementation of my proposal in #3609 to retain user account details (email address and encrypted private key) in the redux state even when we receive the `SET_ACCOUNT`, `SET_PROVIDER`, and `SET_NETWORK` actions. I think it's easier to see what I propose in PR form, so I'm submitting  this one even if only as the basis for further discussion.

# Issues

Refs #3609 

# Checklist:


cnasc:
Closing in favor of another approach which will not require the user email to be stored in `unlock-app`

cnasc:
Per my note in #5040, now that we're tracking the user account key again, we want to always hold on to it. It should be impossible for users to actually switch networks, providers, or accounts while logged in so this is a simple change.

PR #3692, opened 2019-06-11
down the docker file at the conclusion of integration tests

cellog:
# Description

This adds a simple command to shut down the docker image for integration tests when they conclude in local testing. I've had some weird issues since the image exposes public ports when trying to run `npm run dev` after running integration tests. In some cases, `localhost:3001` pulls from the integration test docker image instead of our actual localhost. This fixes that issue

# Checklist:

### DIFF
--- a/scripts/local-docker-integration-tests.sh
+++ b/scripts/local-docker-integration-tests.sh
@@ -34,7 +34,8 @@ docker build -t unlock-protocol-com -f $REPO_ROOT/docker/unlock-protocol-com.doc
 docker build -t integration-tests -f $REPO_ROOT/docker/integration-tests.dockerfile $REPO_ROOT &
 wait
 
-
 # Run the tests
 $REPO_ROOT/scripts/integration-tests.sh $EXTRA_ARGS
 
+# shut down the docker file in case we want to do any local dev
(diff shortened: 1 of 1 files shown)


scripts/local-docker-integration-tests.sh
```diff
@@ -34,7 +34,8 @@ docker build -t unlock-protocol-com -f $REPO_ROOT/docker/unlock-protocol-com.doc
 docker build -t integration-tests -f $REPO_ROOT/docker/integration-tests.dockerfile $REPO_ROOT &
 wait
 
-
 # Run the tests
 $REPO_ROOT/scripts/integration-tests.sh $EXTRA_ARGS
 
+# shut down the docker file in case we want to do any local dev
+docker-compose -f $DOCKER_COMPOSE_FILE down
```
akeem: More of a shell question than anything else, but will this run if the integration tests fail/return a non 0 value?

cellog: in my testing, the docker-compose down line still runs

cellog: even if I ^C during a test run, it will run

PR #3693, opened 2019-06-11
new configurable test file for integration testing the ad remover paywall

cellog:
# Description

This is a file that will be used in integration testing. Basically, it's the ad remover paywall, but configurable. It accepts some values in the query:

`locks` - this is an array of locks as objects with keys `address` and `name` that is used to create the `unlockProtocolConfig` needed for `unlock.1.0.min.js`
`paywall` - this is the address of the paywall (`http://paywall:3001` for integration testing on CI, `http://localhost:3001` for local testing) which is used to load `unlock.1.0.min.js`

Otherwise, this is identical to the existing `index.html`.

This file is used only to test the behavior of the ad remover paywall when the user is coming from a non-crypto-enabled browser (which puppeteer is by default).

A follow-up PR will introduce the file which will be used for testing logged in users

# Issues

Refs #3607

# Checklist:

### DIFF
--- a/paywall/src/static/adremover/integrationtesting-loggedout.html
+++ b/paywall/src/static/adremover/integrationtesting-loggedout.html
@@ -1,0 +1,175 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Lorem Ipsum Daily</title>
+    <script>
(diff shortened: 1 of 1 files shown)


cellog:
based on feedback for #3722, closing to refactor

PR #3702, opened 2019-06-11
not showing insufficient balance for ERC20 locks

julien51:
# Description

We do not know of the user's balance with ERC20 so we should not show the insufficient balance!

# Checklist:

- [X]  1 PR, 1 purpose: my Pull Request applies to a single purpose

### DIFF
--- a/paywall/src/stories/lock/Overlay.stories.js
+++ b/paywall/src/stories/lock/Overlay.stories.js
@@ -31,6 +31,11 @@ const store = createUnlockStore({
   },
 })
 
+const account = {
+  address: '0x123',
+  balance: '123',
+}
+
 const render = (
   locks,
   errors = { error: false, errorMetadata: {} },
@@ -97,7 +102,7 @@ const render = (
               expiration:
                 new Date('January 30, 3000, 00:00:00').getTime() / 1000,
             }}
-            account="account"
+            account={account}
           />
         </ErrorProvider>
       </WindowProvider>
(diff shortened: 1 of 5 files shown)


paywall/src/stories/lock/Overlay.stories.js
```diff
@@ -97,7 +102,7 @@ const render = (
               expiration:
                 new Date('January 30, 3000, 00:00:00').getTime() / 1000,
             }}
-            account="account"
+            account={account}
```
julien51: fixing a props validation issue!

paywall/src/utils/withConfig.js
```diff
@@ -22,10 +22,11 @@ export default function withConfig(Component) {
     )
   }
 
-  function mapStateToProps(state) {
+  // We do not overwite existing props if any!
+  function mapStateToProps(state, props) {
     return {
-      network: state.network,
-      account: state.account,
+      network: props.network || state.network,
```
julien51: This was causing the stories to fail because we had no account set in state.

julien51: The "Insufficient story" was actually not showing insufficient funds :/

cellog: is this the best option? Why not just set them in the state inside the stories? I ask because this increases surface area to look for bugs

julien51: This was _actually_ a bug ;)

PR #3714, opened 2019-06-12
Remove the contract's NO_FALLBACK error

### LINKED ISSUE #3709: Remove no_fallback in the contracts
There is no try/catch in Solidity.  I worry that the no-fallback implementation we have now could cause integration issues with other contracts.

Remove the no-fallback mixin, and use the default behavior instead.

**Additional context**

hardlydifficult:
# Description

When a function which DNE is called, Solidity forwards the call to the "fallback function".  Today we catch this and revert the transaction.  

This PR removes the NO_FALLBACK revert.  My concern is in Solidity there is no try/catch ability.  Other contracts may probe for features and expect the default behavior instead of a revert.  

We also saw this in a wallet - where some field, I think `name` defaulted to `NO_FALLBACK` because the front-end was not expecting an error, they just surfaced the error as if it was the string they were expecting.

# Issues

Fixes #3709
Refs #

# Checklist:


nfurfaro:
Are there any other changes we need to make in addition to this? I'm wondering if we should explicitly define our own fallback function. I suppose the main goal (or one of them )was to prevent unwanted ether transfer to the contract right?

nfurfaro:
Is this something we need to still address, just in a different way?
```
contract MixinNoFallback
{
  /**
   * @dev the fallback function should not be used.  This explicitly reverts
   * to ensure it's never used.
   */
```

hardlydifficult:
Nope, after thinking on this I believe it's better to not handle the situation as that's what most other contracts do.  It's the expected behavior.

PR #3718, opened 2019-06-12
add currencyContractAddress to validation

cellog:
# Description

As it turns out, `currencyContractAddress` is not currently validated. Because the current ad remover paywall is more restrictive, this also means that it does not work any more.

This PR fixes both the code and the social issue. The code fix is that `isValidLock` is more permissive. It also now validates `currencyContractAddress` to ensure its format is correct.

The social fix is a big warning label in `NoKeyLock` for future changes.

# Issues

Refs #2903 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/utils/validators.test.js
+++ b/paywall/src/__tests__/utils/validators.test.js
@@ -1017,6 +1017,35 @@ describe('Form field validators', () => {
           })
         ).toBe(false)
       })
+
+      it('should fail on invalid lock currencyContractAddress', () => {
+        expect.assertions(4)
+
+        expect(
+          validators.isValidLock({
+            ...validLock,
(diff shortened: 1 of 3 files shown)


paywall/src/__tests__/utils/validators.test.js
```diff
@@ -1037,10 +1066,31 @@ describe('Form field validators', () => {
             outstandingKeys: 0,
             balance: '0',
             owner: '0x1234567890123456789012345678901234567890',
+            publicLockVersion: 4,
+            currencyContractAddress:
+              '0x9876543210987654321098765432109876543210',
+            some: 'random',
```
cellog: here is the test that makes sure if we add new fields in the future it still is valid

PR #3737, opened 2019-06-13
filter locks returned from data iframe cache to only include those specified in the paywall config

### LINKED ISSUE #3732: when returning cached locks from the data iframe, only locks specified in the config should be returned
Currently, the data iframe returns all of the locks for a user on the current network. This is incorrect: we should only return the locks that the current page's config specifies.

P.S. found this one while developing integration tests, so yay for integration testing

cellog:
# Description

A user may visit several different paywalls. All of them will cache the lock/key information for that paywall and that user. This leads to a mistake: when returning cached locks, the data iframe returns all of them. Instead, it should only return the relevant locks, which are the ones specified in the paywall config passed to the data iframe. This PR fixes that mistake.

# Issues

Fixes #3732 
Refs #2903 

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/paywallConfig.js
+++ b/paywall/src/data-iframe/paywallConfig.js
@@ -1,0 +1,16 @@
+let paywallConfig
+let paywallConfigLocks
+
+// these are used by the postOffice to filter the locks
+export function setPaywallConfig(newConfig) {
+  paywallConfig = newConfig
+  paywallConfigLocks = Object.keys(newConfig.locks)
+}
+
+export function getPaywallConfig() {
+  return paywallConfig
+}
+
+export function getRelevantLocks() {
+  return paywallConfigLocks
+}
--- a/paywall/src/data-iframe/paywallConfig.js
+++ b/paywall/src/data-iframe/paywallConfig.js
@@ -4,7 +4,11 @@ let paywallConfigLocks
 // these are used by the postOffice to filter the locks
 export function setPaywallConfig(newConfig) {
   paywallConfig = newConfig
-  paywallConfigLocks = Object.keys(newConfig.locks)
+  if (newConfig) {
+    paywallConfigLocks = Object.keys(newConfig.locks)
+  } else {
+    paywallConfigLocks = undefined
+  }
 }
 
 export function getPaywallConfig() {
--- a/paywall/src/data-iframe/postOffice.js
+++ b/paywall/src/data-iframe/postOffice.js
@@ -13,6 +13,42 @@ import {
 import { getFormattedCacheValues } from './cacheHandler'
 import { getRelevantLocks } from './paywallConfig'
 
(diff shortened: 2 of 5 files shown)


paywall/src/data-iframe/postOffice.js
```diff
@@ -71,8 +72,24 @@ export default function postOffice(window, requiredConfirmations) {
   }
 
   const lockHandler = cachedData => {
-    actions.locks(cachedData.locks)
-    const unlockedLockAddresses = Object.values(cachedData.locks)
+    const relevantLocks = getRelevantLocks()
+    if (!relevantLocks) {
+      // this should not happen
```
akeem: If this may be the case why are we handling this scenario? What are some of the conditions that would cause this branch to be executed?

paywall/src/data-iframe/postOffice.js
```diff
@@ -71,8 +72,24 @@ export default function postOffice(window, requiredConfirmations) {
...
+      console.error('internal error - locks requested before config') // eslint-disable-line
+      return
+    }
+    // we can only return the locks that are specifically requested
+    const filteredLocks = Object.keys(cachedData.locks).reduce(
```
akeem: How would you feel about extracting this into its own method so that it can be tested robustly outside of `postOffice`?

paywall/src/data-iframe/postOffice.js
cellog: this is a boundary check. If `POST_MESSAGE_SEND_UPDATES` with `'locks'` is sent prior to receiving `POST_MESSAGE_CONFIG` then this will trigger. This can only happen if someone introduces a bug upstream, but it will be exceedingly difficult to debug in that case. This is here to make it easier to debug. I'd like to keep it, as it does no damage, but will allow future us to debug this, should it ever occur. I'm happy to add my explanation in a comment in the source

cellog: comment added to source

paywall/src/data-iframe/postOffice.js
cellog: done

cellog:
ok @akeem thank you very much for the feedback, I think the code is much improved, but I'll toss it back to you to verify, PTAL

paywall/src/data-iframe/paywallConfig.js
```diff
@@ -0,0 +1,20 @@
+let paywallConfig
+let paywallConfigLocks
+
+// these are used by the postOffice to filter the locks
+export function setPaywallConfig(newConfig) {
```
akeem: I'd suggest making this into a class so that you can easily manage the state and reason about things. It may not be called this way currently, but I think someone could call getPaywallConfig without it being set and the same for getRelevantLocks

cellog: #3789

PR #3739, opened 2019-06-13
clear stale cached account if we have no wallet present

### LINKED ISSUE #3735: localStorage cache of user account needs to be reset if the user has no wallet on the browser on next visit
If the user returns to the browser with no wallet, localStorage of the account needs to be reset when `POST_MESSAGE_WALLET_INFO` is received if there is no account.

cellog:
# Description

This fixes loading a new paywall if the wallet is missing (metamask has been uninstalled, for instance). It also ensures we never use a stale cached account if it isn't verified as present by blockchain

# Issues

Fixes #3735 
Refs #2903 

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/postOfficeListener.js
+++ b/paywall/src/data-iframe/postOfficeListener.js
@@ -2,9 +2,11 @@ import {
   POST_MESSAGE_SEND_UPDATES,
   POST_MESSAGE_CONFIG,
   POST_MESSAGE_PURCHASE_KEY,
+  POST_MESSAGE_WALLET_INFO,
 } from '../paywall-builder/constants'
 import { isValidPaywallConfig } from '../utils/validators'
 import { ACCOUNT_REGEXP } from '../constants'
+import { setAccount } from './cacheHandler'
 
 /**
  * Create the listener to respond to the configuration, which lists all locks on the page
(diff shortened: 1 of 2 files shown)


paywall/src/data-iframe/postOfficeListener.js
```diff
@@ -64,6 +66,21 @@ export const makePurchaseKeyListener = (logger, purchase) =>
...
+export const makeWalletInfoListener = window =>
+  async function walletInfoListener(walletInfo) {
+    if (
+      !walletInfo ||
+      typeof walletInfo !== 'object' ||
+      Array.isArray(walletInfo)
+    ) {
+      return
+    }
+    if (walletInfo.noWallet) {
```
akeem: I think we might be able to update the conditional here to be something like `walletInfo && walletInfo.noWallet` otherwise return?

cellog: I considered this. There is a chance we may want to act upon `walletInfo.isEnabled` as well (i.e. user has metamask, but has chosen not to enable it for this page, we should probably reset their account, even though metamask shows their account when calling `eth_listAccounts`). In that case, the line would become `walletInfo.noWallet || !walletInfo.isEnabled`. Would be interested in your take on this?

PR #3850, opened 2019-06-14
fix way-too-many-calls to transactions/keys

### LINKED ISSUE #3773: paywall should only retrieve transactions that are key purchase transactions for relevant locks
Currently, the paywall app retrieves all the transactions that locksmith returns. This means we retrieve lock creation transactions, and also key purchase transactions for locks that are not specified in the paywall config.

The paywall should ignore any irrelevant transactions by parsing the inputs and using that to filter out irrelevant transactions

cellog:
# Description

There were 2 issues that caused the 5000% increase in calls to alchemy last week.

1. The call to update transactions and keys for a new user account was happening on `walletService.on('account.updated')` - which triggers every time we request the current account (side note: this action REALLY needs to be renamed to `'account.retrieved'`, I have made this mistake at least twice before)
2. we were retrieving all transactions by the user, not just the ones relevant to locks on the current paywall.

This PR implements a fix for (1) by moving the call to update transactions to the *correct* location 10 lines lower into the handler that is called by `pollForAccountChanges` when the account actually changes. When I added the code that is being removed it was one of my stupider moments :)

The fix for (2) is also quite simple, we take the lock addresses from the configuration and pass them as `recipient` to the call to locksmith to retrieve transactions.

# Issues

Fixes #3773 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/data-iframe/blockchainHandler/index.test.js
+++ b/paywall/src/__tests__/data-iframe/blockchainHandler/index.test.js
@@ -14,6 +14,7 @@ import {
 import { getNetwork } from '../../../data-iframe/blockchainHandler/network'
 import ensureWalletReady from '../../../data-iframe/blockchainHandler/ensureWalletReady'
 import web3ServiceHub from '../../../data-iframe/blockchainHandler/web3ServiceHub'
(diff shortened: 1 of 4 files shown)


cellog:
side note: as I am out the rest of the day, please feel free to check out the branch and execute any needed fixes directly.

PR #3871, opened 2019-06-17
add new configuration field "type" for paywall config

### LINKED ISSUE #3859: add a configuration option to the paywall for paywall "type"
By default, the type will be `'adblock'`.

As of now, valid types are:

- `'adblock'`
- `'paywall'`

This needs to be added to paywall configuration validation

cellog:
# Description

This adds a new optional field to the paywall configuration, `type` which if not present defaults to `'adblock'` and can be one of two values: `'adblock'` or `'paywall'`.

It updates both the unlock.min.js side and the checkout UI side. The first step is adding it to validation, the next is to the typescript definition, the last is adding the default value to `usePaywallConfig`

# Issues

Fixes #3859 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/hooks/usePaywallConfig.test.js
+++ b/paywall/src/__tests__/hooks/usePaywallConfig.test.js
@@ -93,6 +93,7 @@ describe('usePaywallConfig hook', () => {
 
     const myConfig = {
       ...defaultValue,
+      type: 'paywall',
       locks: {
         '0x1234567890123456789012345678901234567890': {
           name: 'my lock',
--- a/paywall/src/__tests__/utils/validators.test.js
+++ b/paywall/src/__tests__/utils/validators.test.js
@@ -371,7 +371,7 @@ describe('Form field validators', () => {
         ).toBe(false)
       })
 
-      it('should fail when type is not an string', () => {
+      it('should fail when type is not a string', () => {
         expect.assertions(2)
 
         expect(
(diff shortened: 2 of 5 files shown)


paywall/src/__tests__/hooks/usePaywallConfig.test.js
```diff
@@ -93,6 +93,7 @@ describe('usePaywallConfig hook', () => {
 
     const myConfig = {
       ...defaultValue,
+      type: 'paywall',
```
cellog: this simply verifies that the `type` field properly validates and is passed to the component using the hook. Because of the way the test is designed, this is the only change necessary. Prior to adding the validation, this 1-line change would have made the test fail

paywall/src/__tests__/utils/validators.test.js
```diff
@@ -371,6 +371,34 @@ describe('Form field validators', () => {
         ).toBe(false)
       })
 
+      it('should fail when type is not an string', () => {
```
cnasc:
```suggestion
      it('should fail when type is not a string', () => {
```

paywall/src/unlockTypes.ts
```diff
@@ -67,9 +67,12 @@ export interface PaywallConfigLock {
   name: string
 }
 
+export type PaywallTypes = 'adblock' | 'paywall'
```
cnasc: Something like `KindOfPaywall` might be less ambiguous (`PaywallTypes` is a TS type whose allowed values are "types" of a different nature) but that's a matter of taste more than anything

PR #3878, opened 2019-06-17
add placeholder page for users iframe

### LINKED ISSUE #3877: add placeholder app for unlock-app user accounts iframe
We need an app that the `unlock.js` script can load to handle user accounts

cellog:
# Description

This PR adds the users iframe page as a placeholder that can be loaded by the paywall scripts. I would very much like opinions on what to name it if `users` is too generic, keeping in mind that no one will actually see the URL, and the title will probably change dynamically based on whether the user is purchasing a key, logging in, creating an account, etc.

It also removes some outdated cruft from `_redirects`

# Issues

Fixes #3877 

# Checklist:

### DIFF
--- a/unlock-app/src/components/content/AccountContent.tsx
+++ b/unlock-app/src/components/content/AccountContent.tsx
@@ -4,11 +4,11 @@ import Layout from '../interface/Layout'
 import { pageTitle } from '../../constants'
 import Errors from '../interface/Errors'
 
-export const UsersContent = () => {
+export const AccountContent = () => {
   return (
-    <Layout title="Users">
+    <Layout title="Account">
       <Head>
-        <title>{pageTitle('Users')}</title>
+        <title>{pageTitle('Account')}</title>
         <script src="https://js.stripe.com/v3/" />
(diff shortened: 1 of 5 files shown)


cnasc:
Since this handles login, payment, and confirmation we should probably have a more general name. I'm not sure what that should be, maybe @benwerd has a wording suggestion?

cnasc:
Also, since this page will (I presume) wrap all the components we use in the process of a user account checkout, could you populate the `<Head>` with the Stripe JS? 
```html
<script src="https://js.stripe.com/v3/"></script>
```
The payment details entry component needs access to that to generate the Stripe token.

cellog:
I added the stripe load. As this is just a placeholder, I am assuming that you will take over this page and modify as needed. This is solely to give an endpoint that I can use to develop against for loading the iframe that won't yield a 404

benwerd:
Fwiw, something like `account` probably makes sense vs `users`, right?

cellog:
renamed to `account` PTAL

Issue #3458, opened 2019-05-30
design of subscribed locks does not match spec


cellog:
Using https://www.figma.com/file/EZ4mtBKGj54EQCQvYDArtqKs/Forbes-Crypto-Blockchain?node-id=213%3A31 as a reference, the design of the `Checkout` UI when a user has subscribed does not match.

PR #3884, opened 2019-06-17
new expirationAsText function, fix durationsAsArrayFromSeconds for partial seconds

cellog:
# Description

In preparation for displaying the key expiration on confirmed lock, this fixes a hilarious situation where we would display the number of seconds left - up to 12 decimal places :)

This also adds a function to format the expiration in better English, so we use "Expires in" for time values and "Expires" for dates.

# Issues

Refs #3458

# Checklist:

### DIFF
--- a/paywall/src/utils/durations.js
+++ b/paywall/src/utils/durations.js
@@ -63,6 +63,7 @@ export function durationsAsArrayFromSeconds(seconds) {
   const d = durations(seconds, {})
   const asArrayOfValues = Object.keys(d).map(duration => {
     const durationFloor = Math.floor(d[duration])
+    // map to an empty string to avoid "0 seconds"
     if (durationFloor === 0) return ''
     if (durationFloor !== 1) {
       // Singular should only be used when there is exactly 1; otherwise plural is needed
@@ -70,6 +71,7 @@ export function durationsAsArrayFromSeconds(seconds) {
     }
     return `${durationFloor} ${duration.slice(0, -1)}` // remove the s!
   })
+  // remove the "0 seconds" etc. entries that mapped to ""
   return asArrayOfValues.filter(a => a) // remove empty entries
 }
 
@@ -80,6 +82,10 @@ export function durationsAsArrayFromSeconds(seconds) {
  */
 export function secondsAsDays(seconds) {
   return Math.ceil(seconds / 86400).toString()
+}
+
+function isLessThanADay(timestamp) {
+  return timestamp - new Date().getTime() / 1000 < 86400
 }
 
 /**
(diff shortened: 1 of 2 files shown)


paywall/src/utils/durations.js
```diff
@@ -109,3 +111,11 @@ export function expirationAsDate(timestamp) {
 
   return MONTH_NAMES[month] + ' ' + day + ', ' + year
 }
+
+export function expirationAsText(timestamp) {
+  const text = expirationAsDate(timestamp)
+  if (timestamp - new Date().getTime() / 1000 < 86400) {
```
akeem: What is the scenario? Is this to handle keys with less than a day prior to expiry remaining? Would you mind adding a comment here or extracting this to a method with a meaningful name?

paywall/src/utils/durations.js
```diff
@@ -62,13 +62,15 @@ export function durationsAsTextFromSeconds(seconds) {
...
+    if (durationFloor !== 1) {
       // Singular should only be used when there is exactly 1; otherwise plural is needed
-      return `${d[duration]} ${duration}`
+      return `${durationFloor} ${duration}`
     }
-    return `${d[duration]} ${duration.slice(0, -1)}` // remove the s!
+    return `${durationFloor} ${duration.slice(0, -1)}` // remove the s!
   })
-  return asArrayOfValues
+  return asArrayOfValues.filter(a => a) // remove empty entries
```
akeem: Should we encounter empty entries here?  Can we update the caller to filter empties prior to passing to this method to reduce complexity here?

paywall/src/utils/durations.js
cellog: extracting what?

paywall/src/utils/durations.js
cellog: Nope. see line 66.

paywall/src/utils/durations.js
akeem: `timestamp - new Date().getTime() / 1000 < 86400`

cellog: thanks

paywall/src/utils/durations.js
akeem: I am not sure that I understand, would you be able to elaborate a bit more?

cellog: sure. This is to prevent "1 hour and 0 seconds" which will happen if we have (for example) "1 hour and 0.23432423 seconds" before flooring. The solution I chose is to simply map 0 to an empty string, and remove them. If we don't remove them, it becomes "1 hour and "

PR #3889, opened 2019-06-17
Adds erc20 approval script

akeem:
# Description

As part of the V1 roll out, we have a need for a script that will allow a programmatic purchaser to approve a Lock contract to withdraw from its balance of Tokens from a compliant ERC20 contract. 

For the time being, the assumption is that this will be a relatively direct operation and should not be automated at this time. 

Upon running the script the user should receive details about the dispatched transaction. 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/scripts/approve_erc20_transfer
+++ b/locksmith/scripts/approve_erc20_transfer
@@ -1,5 +1,16 @@
 #!/usr/bin/env node
 
+/**
+ * Approval Script - This script provides Locks approval to
+ * withdraw tokens from an ERC20 contract on behalf of a given
+ * address
+ * 
+ * approve_erc20_transfer --value <approval amount> --provider <provider uri>
+ *                        --lockContractAddress <address of lock contract>
+ *                        --erc20ContractAddress <address of erc20 contract>
+ *                        --privateKey <privateKey of address approving transfer>
+ */
+
 const ethers = require('ethers')
 const args = require('yargs').argv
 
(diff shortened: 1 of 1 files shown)


locksmith/scripts/approve_erc20_transfer
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env node
+
+const ethers = require('ethers')
+const args = require('yargs').argv
+
+let provider = new ethers.providers.JsonRpcProvider(args.provider)
+let value = args.value
+let lockContractAddress = args.lockContractAddress
+let erc20ContractAddress = args.erc20ContractAddress
+let privateKey = args.privateKey
```
cellog: Is this a file path? If so, I would rename the var. If not, this is probably as unwise as passing a password on the command-line from a security perspective.

akeem:
> LG. Can you document how we use this somewhere (potentially even in the header of the script)?

Good point, i'll add that!

locksmith/scripts/approve_erc20_transfer
akeem: it is not a file path, it is the private key used to sign the transaction. I'm not sure that reading from a file would be materially better

cellog: https://unix.stackexchange.com/questions/78734/why-shouldnt-someone-use-passwords-in-the-command-line

and 

https://serverfault.com/questions/592631/passing-a-private-key-to-scp-from-the-command-line-instead-of-a-file

provide some context as to why this is a potentially bad idea

akeem: cool, i'll make some adjustments in a future PR. To be clear this is to be run on our machines, I am not sure that offloading to a file will help if the machine is compromised but will give it a go

PR #3791, opened 2019-06-13
Fix checkout display in Safari

cellog:
# Description

Because the `CheckoutWrapper` was using `display: grid` without specifying any grid template, this left too much control to the browser as to how to render the display.

Before:

Screen Shot 2019-06-13 at 5 20 56 PM

After (both in Safari and Chrome):

Screen Shot 2019-06-13 at 5 24 47 PM
Screen Shot 2019-06-13 at 5 24 37 PM
Screen Shot 2019-06-13 at 5 24 26 PM
Screen Shot 2019-06-13 at 5 24 19 PM
Screen Shot 2019-06-13 at 5 24 09 PM
Screen Shot 2019-06-13 at 5 35 35 PM
Screen Shot 2019-06-13 at 5 35 42 PM
Screen Shot 2019-06-13 at 5 35 51 PM

# Issues

Fixes #3699 

# Checklist:

- [X] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [X] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [X] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [X] I have performed a self-review of my own code
- [X] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #3891, opened 2019-06-18
Adding a lock to our blog

julien51:
# Description

This lock does not actually implement a paywall, but just shows how easy it is to become a member!

Next we may limit some features to only the paying members, or at least show the list of members on each page?

# Checklist:

### DIFF
--- a/unlock-protocol.com/src/components/interface/Membership.js
+++ b/unlock-protocol.com/src/components/interface/Membership.js
@@ -1,0 +1,100 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import styled from 'styled-components'
+
+export const MembershipUnlocked = () => (
+  <MembersBar>
+    Thanks for being a member! Did you know? Your key is an Ethereum non
+    fungible token!{' '}
+    <span role="img" aria-label="click">
+      🔑
+    </span>
+  </MembersBar>
+)
+
+export const MembershipLocked = ({ becomeMember }) => (
+  <MembersBar>
+    <Button onClick={becomeMember}>
+      Try Unlock, become a member!{' '}
+      <span role="img" aria-label="click">
+        ↗️
+      </span>
+    </Button>
+  </MembersBar>
+)
+
+MembershipLocked.propTypes = {
+  becomeMember: PropTypes.func.isRequired,
+}
+
+export class Membership extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      isMember: props.isMember || 'pending',
+    }
+    this.unlockListener = event => {
+      if (event.detail === 'unlocked') {
+        this.setState(state => ({
+          ...state,
+          isMember: 'yes',
+        }))
+      } else if (event.detail === 'locked') {
+        this.setState(state => ({
+          ...state,
+          isMember: 'no',
+        }))
+      }
+    }
+  }
+
+  componentDidMount() {
+    window.addEventListener('unlockProtocol', this.unlockListener)
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('unlockProtocol', this.unlockListener)
+  }
+
+  becomeMember() {
+    window.unlockProtocol && window.unlockProtocol.loadCheckoutModal()
+  }
+
+  render() {
+    const { isMember } = this.state
+    if (isMember === 'yes') {
+      return <MembershipUnlocked />
+    }
+    return <MembershipLocked becomeMember={this.becomeMember} />
+  }
+}
+
+Membership.propTypes = {
+  isMember: PropTypes.string,
+}
+
+Membership.defaultProps = {
+  isMember: 'pending',
+}
+
+export default Membership
+
+const MembersBar = styled.div`
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-template-rows: 1fr;
+  justify-content: center;
+  align-content: center;
+  background-color: black;
+  width: 100%;
+  left: 0px;
+  position: fixed;
+  padding: 15px;
+  bottom: 0;
+  text-align: center;
+`
+
+const Button = styled.div`
+  cursor: pointer;
+  padding: 10px;
+`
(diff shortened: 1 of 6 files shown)


julien51:
image

julien51:
cc @smombartz Feel free to suggest a better style too!

unlock-protocol.com/src/components/interface/Membership.js
```diff
@@ -0,0 +1,70 @@
...
+  componentDidMount() {
+    // Connect to Unlock and get the state!
+    window.addEventListener('unlockProtocol', event => {
+      this.setState(state => ({
+        ...state,
+        isMember: event.detail === 'unlocked' ? 'yes' : 'no',
+      }))
+    })
+  }
+
```
cellog: need a `componentWillUnmount()` to remove the event listener

unlock-protocol.com/src/components/interface/Membership.js
```diff
@@ -0,0 +1,70 @@
...
+      isMember: 'pending',
+    }
+  }
+
+  componentDidMount() {
+    // Connect to Unlock and get the state!
+    window.addEventListener('unlockProtocol', event => {
+      this.setState(state => ({
+        ...state,
+        isMember: event.detail === 'unlocked' ? 'yes' : 'no',
```
cellog: We will expand the messages emitted in order to support the tickets app to include emitting the data needed to generate the QR code. The isMember check should explicitly check for `'locked'`

unlock-protocol.com/src/components/interface/Membership.js
```diff
@@ -0,0 +1,70 @@
...
+        </Button>
+      </MembersBar>
+    )
+  }
+}
+
+export default Membership
+
+const MembersBar = styled.div`
+  display: grid;
```
cellog: `display: grid` always needs `grid-template-rows` and `grid-template-columns` definitions

unlock-protocol.com/src/pages/_document.js
```diff
@@ -21,6 +44,8 @@ export default class MyDocument extends Document {
           <Fonts />
           <link rel="shortcut icon" href="/static/favicon.ico" />
           {this.props.styleTags}
+          <script dangerouslySetInnerHTML={this.props.unlockTag} />
```
cellog: you can use `<script>` tags directly inside `<Head>`

unlock-protocol.com/src/components/interface/Membership.js
julien51: Is that true? I could not find any source for this and most of the example that I see out there do not have both of these... (and many actually do not have any...)

unlock-protocol.com/src/pages/_document.js
julien51: I tried but it does not seem to work :/

julien51:
```> Using external babel configuration
> Location: "/Users/julien/repos/unlock/unlock-protocol.com/.babelrc"
[ error ] ./pages/_document.js
SyntaxError: /Users/julien/repos/unlock/unlock-protocol.com/src/pages/_document.js: Unexpected token (48:8)

  46 |           {this.props.styleTags}
  47 |           <script>(function(d, s) {
> 48 |         var js = d.createElement(s),
     |         ^
  49 |         sc = d.getElementsByTagName(s)[0];
  50 |         js.src="https://paywall.unlock-protocol.com/static/unlock.1.0.min.js";
  51 |         sc.parentNode.insertBefore(js, sc); }(document, "script"))</script>
events.js:183
      throw er; // Unhandled 'error' event
      ^
```

unlock-protocol.com/src/components/interface/Membership.js
cellog: Without them, the browser does its defaults. Safari adds about 30px padding to rows than chrome. see #3791 for an example

unlock-protocol.com/src/pages/_document.js
cellog: you are right!

unlock-protocol.com/src/components/interface/Membership.js
julien51: huh, ok, what do you suggest I use then?

cnasc: The specific values aren't terribly important (whatever looks good) as long as there's something set to ensure consistency across browsers

cnasc: In this specific case, since it looks like we just need that line of text centered vertically and horizontally, you could even just use a flex container instead of fussing with the rows and columns

cellog: I agree with @cnasc, flex seems better to use for a 1-column/1-row layout

julien51:
ok, ready! PTAL (I also added stories...)

julien51:
image

image

unlock-protocol.com/src/pages/_document.js
cellog: it also may work (in a follow-up PR, don't want to block this any further) to copy the approach taken here: https://github.com/unlock-protocol/unlock/blob/master/paywall/src/components/content/NewDemoContent.tsx#L27
and here: https://github.com/unlock-protocol/unlock/blob/master/paywall/src/components/content/NewDemoContent.tsx#L55

julien51:
I am not sure how to write tests for this? I'll just bump the test coverage for now...

cellog:
yeah anything in `<Head>` is untestable with react-testing-library last time I checked

smombartz:
@julien51 what about giving people access to a member page? it feels a bit anticlimactic – and not like anything was really unlocked...

smombartz:
maybe it could unlock some kind of feature for the website – ability to change the background or a new homepage style...

also i noticed it doesn't "load" properly if you navigate around (notice how its still locked, until you reload): https://www.dropbox.com/s/5jbzf3oejsrobzv/190619-UnlockBar-Homepage.mov?dl=0

julien51:
> @julien51 what about giving people access to a member page? it feels a bit anticlimactic – and not like anything was really unlocked...

So, i am adding the abiloty to read and leave comments only if you are a member.

smombartz:
@julien51 the bar also shows up on the landing page... mentioning that only because this PR talks about the blog...

Issue #3763, opened 2019-06-13
Implement the key purchase screen in unlock app


julien51:
In the same way that Metamask prompts the user when they are sending a transaction for a confirmation, we should implement a screen which asks the user to confirm that they are purchasing a key to a given lock.
This screen should show the follow the design provided by @smombartz :

![image.png](https://images.zenhubusercontent.com/5c4133c214621d11daa67711/f86c797e-ff88-465b-8b2e-f9f4d5b6514a)

cnasc:
To make sure I understand the architecture here, these will be separate pages (for example, at `/keyPurchase` for this one), that have no UI chrome whatsoever from the normal site and only contain the form in question?

julien51:
so essentially yes but let's not use case in the URL! `/0x123/purchase` would be good where `0x123` is the lock address.
That page should make sure the user is logged in, and if not show the LoginOrSognup component. Similarly the user should have a CC, or it will show the `CreditCards` component... and finally will show that last form.

When the users clicks on `purchase` the `unlockProvider` will invoke the locksmith API to purchase the key.

smombartz:
Design flow is in Figma under Paywall (select CC payment from "admin panel") or this link ;)
https://www.figma.com/proto/LkWSSnCGlKEITBB5T5NYXW/Paywall?node-id=441%3A206&viewport=370%2C356%2C0.06711409240961075&scaling=min-zoom

PR #3897, opened 2019-06-18
Initial key purchase confirmation component

cnasc:
# Description

This PR is the initial work on the key purchase confirmation component. We're missing a few things before it's fully functional, namely the credit card info (work in progress) and lock info (will come when we integrate with paywall). Additionally, the styling needs some work (the base components should be redefined in terms of the new grid we have, which I will do in a subsequent PR).

image

But the basics are here, and having it in this form will allow me to evaluate the styles I update in the context of both the settings page as well as these standalone components.

Part of this PR included a little cleanup around organization of the user account-relevant files.

# Issues

Refs #3763 

# Checklist:


cnasc:
> I am thinking the only lock info we would need is the lock name from a UI perspective

For the UI definitely, but also any info we'd want to send to `locksmith` in the dispatch

Issue #3762, opened 2019-06-13
Inplement Credit Cards form


julien51:
Users should be able to enter their Credit/Debit card number inside Unlock app.
This should be built as a component which can be included on the settings page as well as potentially other pages.

We are using Stripe's API for this. The credit card number should be exchanged on the front end for a token which is then stored on the server side. (This part should already be implemented, please @akeem can you prove API details on this issue so we have a single place to look things up?)

cnasc:
Design files for this are located here: https://www.figma.com/file/ZIhu6WMu781dsGDHszPKb1UT/Keychain?node-id=0%3A1 (for my own reference)

PR #3905, opened 2019-06-18
Type definition to unblock credit card form

cnasc:
# Description

This PR adds some type definitions, the lack of which were blocking progress with the Stripe integration.

(verified in my credit card branch, then pulled out here for review)

# Issues

Refs #3762 

# Checklist:

### DIFF
--- a/unlock-app/src/vendor.d.ts
+++ b/unlock-app/src/vendor.d.ts
@@ -1,7 +1,50 @@
 // This file is meant to declare modules we depend on which do not have @types
 // packages available. Defining them here will unblock us, though doing that the
 // most convenient way will not buy us any type safety.
+import 'react-stripe-elements'
+
 declare module 'react-jazzicon'
 
 // TODO: Remove this when we have a TS build of unlock-js
 declare module '@unlock-protocol/unlock-js'
+
+type PaymentMethodType = 'card' | 'card_present'
+
+// vastly simplified from the real types, but this is all we should need
+interface PaymentMethodIncomplete {
+  billing_details?: {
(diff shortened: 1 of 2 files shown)


unlock-app/src/vendor.d.ts
```diff
@@ -1,7 +1,50 @@
...
 declare module 'react-jazzicon'
 
 // TODO: Remove this when we have a TS build of unlock-js
 declare module '@unlock-protocol/unlock-js'
+
+type PaymentMethodType = 'card' | 'card_present'
+
+// vastly simplified from the real types, but this is all we should need
+interface PaymentMethodIncomplete {
+  billing_details?: {
```
cellog: do all of these need to be ?: or are there a subset of possibilities that we expect? maybe we could do a union type of those subsets?

cnasc:
Hm. Tests pass locally but on CI this seems to have somehow wiped out the declaration for jazzicon and unlock-js

unlock-app/src/vendor.d.ts
cnasc: the `billing_details` implementation here was pulled from the Stripe @types file. These are all optional properties on the object in their API (actually, almost everything is an optional object property). We could possibly restrict it to just the subset of properties we expect to have (or care about -- in the end we only need to get the `id` from `PaymentMethod`), but at some point in the not-too-distant future we should be deleting this code and just pulling in from DefinitelyTyped.

cnasc:
My best guess so far is that you can't declare _and_ augment type definitions in the same file. Maybe.

Issue #3862, opened 2019-06-15
add a new post message POST_MESSAGE_INITIATED_TRANSACTION for user-account key purchase


cellog:
This message will be sent by the unlock-app iframe when a key purchase transaction has started and been stored in locksmith. It will have as its payload the transaction hash to monitor

The unlock-app will post this to the main window using the `mainWindowPostOffice` (most likely in a middleware dedicated to this task, `postOfficeMiddleware`)

The main window will pass it to the data iframe. Here, it will trigger a `getTransaction` and start the polling process.

As part of this issue, a typescript type for the action should be created that can be shared between the apps

cellog:
this is implemented

PR #3913, opened 2019-06-19
add needed types for subsets of the window object

cellog:
# Description

In order to more strongly verify the contract between the API boundary of `unlock.1.0.min.js` and the new user accounts iframe in `unlock-app`, we will need to type the messages that are passed back and forth. For the types to have any meaning, `unlock.1.0.min.js` needs to be in typescript.

In preparation for moving the `unlock.1.0.min.js` script to typescript, this adds new typing of subsets of the window object that are used in each scenario.

Some of these types already exist in `utils/postOffice.ts`, and will be deleted from there in a subsequent PR. The window types represented are:

- windows that use localStorage (for the cache access)
- windows that use web3 (web3ProxyProvider main window-side)
- windows that use CustomEvent (dispatching events main window-side)
- windows that use postMessage (both main window and iframe-side)
- windows that create and manage iframes (only in the main window)

# Issues

Refs #3862 

# Checklist:


paywall/src/windowTypes.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export interface PostMessageTarget {
+  postMessage: (data: any, origin: string) => void
+}
+
+export interface IframeType {
+  contentWindow: PostMessageTarget
+  className: string
+  src: string
+  setAttribute: (attr: 'src', value: string) => void
```
akeem: In scenarios like this, should this actually be a method with a single parameter `value`?

paywall/src/windowTypes.ts
```diff
@@ -0,0 +1,123 @@
...
+export interface MessageEvent {
+  source: any
+  origin: string
+  data: any
+}
+
+export type MessageHandler = (event: MessageEvent) => void
+
+export interface PostOfficeWindow {
+  addEventListener: (type: 'message', handler: MessageHandler) => void
```
akeem: if `type` is a fixed value, I think we  should reduce the parameters of this method

paywall/src/windowTypes.ts
```diff
@@ -0,0 +1,123 @@
...
+  methodCall: web3MethodCall,
+  callback: web3Callback
+) => void
+
+export interface Web3Window extends PostOfficeWindow {
+  web3?: {
+    currentProvider: {
+      sendAsync?: web3Send
+      send?: web3Send
+      isMetamask: true | undefined
```
akeem: could we remove the `undefined` scenario buy the user of this case so that we can make this a boolean?

paywall/src/windowTypes.ts
cellog: I'm confused, this is the actual signature of setAttribute, we can't change what `window` provides here

cellog: what I am doing is making sure we only call `setAttribute` with `'src'` since that is the only attribute we care about in `unlock.1.0.min.js`

paywall/src/windowTypes.ts
cellog: again, the signature is from the `window` object, this is just restricting the signature further for our source, which should only register a listener for `'message'` event type

paywall/src/windowTypes.ts
cellog: yep

paywall/src/windowTypes.ts
akeem: ah okay!

cellog: talked to Chris about this, he had an idea to make it clearer I'm going to implement:

```ts
type AttributeName = 'src' /* can extend with a union of | 'anotherattribute' if needed */
interface Blah {
  setAttribute: (attr: AttributeName, value: string) => void
}
```

cellog:
pushed some changes to clarify things, PTAL @akeem

PR #3920, opened 2019-06-19
Add stripe as a dependency for injection in tests

cnasc:
# Description

This PR adds `stripe` as a dependency so that we can inject it into the stripe provider in tests and stories so our elements will render.

Tested this like so:
```javascript
import Stripe from 'stripe'
const stripe = Stripe('not a real api key')
console.log(stripe)
```

This didn't error, and did appear to produce a valid object. So in stories and tests we should be able to render elements at least, and perhaps just mock out any methods on this object that perform network requests.

# Checklist:

### DIFF
--- a/unlock-app/package-lock.json
+++ b/unlock-app/package-lock.json
@@ -15940,6 +15940,18 @@
       "integrity": "sha1-PFMZQukIwml8DsNEhYwobHygpgo=",
       "dev": true
     },
+    "stripe": {
+      "version": "7.2.0",
+      "resolved": "https://registry.npmjs.org/stripe/-/stripe-7.2.0.tgz",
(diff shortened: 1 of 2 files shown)


cellog:
you'll want to check that it works on CI, since we don't install dev dependencies there

cnasc:
> you'll want to check that it works on CI, since we don't install dev dependencies there

Forgot about that. If we don't actually reference stripe in bundled code, the dep shouldn't get pulled in, right?

cellog:
> > you'll want to check that it works on CI, since we don't install dev dependencies there
> 
> Forgot about that. If we don't actually reference stripe in bundled code, the dep shouldn't get pulled in, right?

probably not? It depends on how next does its tree shaking. I think the answer is what we expect

PR #3928, opened 2019-06-19
migrate the rest of unlock.1.0.min.js to typescript

cellog:
# Description

This completes the move to typescript.

Note: typescript complained about the usage of try/catch around JSON.parse, so instead of allowing it to throw if localStorage returns null, we just pass in a `JSON.stringify()`ied version of `'ignore'` which will cause the rest of the code to ignore it. Otherwise, if the value is invalid JSON, we catch and use `'ignore'` directly.

Note 2: the weird `((window as unknown) as UnlockWindow)` construct is required because of a conflict between definitions of Window and UnlockWindow. Ours is more restrictive and so doesn't match Window's looser definition.

# Issues

Refs #3862 

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/startup.ts
+++ b/paywall/src/unlock.js/startup.ts
@@ -8,9 +8,14 @@ export default function startup(window: UnlockWindow) {
   // of the data iframe. The data iframe will then send down the current
   // value, overriding this. A bit later, the blockchain handler will update
   // with the actual value, so this is only used for a few milliseconds
-  const locked = JSON.parse(
-    window.localStorage.getItem('__unlockProtocol.locked') || '"ignore"'
-  )
+  let locked
+  try {
+    locked = JSON.parse(
+      window.localStorage.getItem('__unlockProtocol.locked') || '"ignore"'
+    )
+  } catch (_) {
+    locked = 'ignore'
+  }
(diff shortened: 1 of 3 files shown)


paywall/src/unlock.js/startup.ts
```diff
@@ -1,24 +1,26 @@
...
+  let locked
   try {
-    // this is a cache for the time between script startup and the full load
-    // of the data iframe. The data iframe will then send down the current
-    // value, overriding this. A bit later, the blockchain handler will update
-    // with the actual value, so this is only used for a few milliseconds
-    const locked = JSON.parse(
-      window.localStorage.getItem('__unlockProtocol.locked')
+    locked = JSON.parse(
+      window.localStorage.getItem('__unlockProtocol.locked') || '"ignore"'
```
cnasc: Is the try/catch still even necessary with this change?

cellog: at first I thought not, but it is still possible for some malicious script to do:

```js
window.localStorage.setItem('__unlockProtocol.locked', '"invalid JSON')
```

which would cause it to crash, disabling the paywall.

cnasc: Makes sense 👌

PR #3945, opened 2019-06-20
new setupIframeMailbox utility

cellog:
# Description

This utility provides a new way to declare postMessage handlers that fully utilizes the new strict type checking. It will make the declaration of post message handling much cleaner and organized.

# Issues

Refs #3862 

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/setupIframeMailbox.ts
+++ b/paywall/src/unlock.js/setupIframeMailbox.ts
@@ -26,12 +26,42 @@ export type MessageHandlerTemplates<T extends MessageTypes> = {
   [key in T]?: MessageHandlerTemplate<T>
 }
 
+/**
+ * Set up all of the post offices for each iframe.
+ *
+ * Example:
+ *
+ * {
+ *   [PostMessages.READY]: (send, _, checkout) => {
+ *     return () => {
+ *       send('data', PostMessages.SEND_UPDATES, 'network')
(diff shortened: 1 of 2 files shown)


paywall/src/unlock.js/setupIframeMailbox.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  // finally, return the mapper that will convert a map of postMessage types to handlers
+  return (
+    forIframe: IframeNames,
+    handlers: MessageHandlerTemplates<MessageTypes>
+  ) => {
+    const keys = Object.keys(handlers) as Array<keyof typeof handlers>
+    keys.forEach(type => {
+      const handler = handlers[type]
+      handler && createHandler(forIframe, type, handler)
```
cnasc: Is the `&&` check necessary? It looks like every `handlers[type]` will have a truthy value

cellog: because the key for the handlers type is optional (otherwise every handler would have to define a listener for every postMessage type), the type checker believes there is a possibility that handler could be null.

I spent nearly an hour trying to find a way other than this to solve it... and failed.

cnasc: Makes sense, annoying when you come across a typing corner case.

PR #3964, opened 2019-06-24
Add getCards to storageService

cnasc:
# Description

In order to know if a user has stored payment details (for use in the key purchase screen), we need to be able to query locksmith for the data. This PR adds a method to enable that.

I intend for storageMiddleware to listen for `UPDATE_ACCOUNT` events that include an email address. Those signal that a user account has successfully logged in, so we should go and fetch payment details. Additonally, it should listen for when a new payment is successfully added because we'll need to update the state to reflect that data. When it receives a card, it will dispatch `UPDATE_ACCOUNT` itself to add the payment details to the account stored in state.

# Issues

Refs #3763 

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/services/storageService.test.js
+++ b/unlock-app/src/__tests__/services/storageService.test.js
@@ -463,7 +463,7 @@ describe('StorageService', () => {
   })
 
   describe('Retrieve user payment details', () => {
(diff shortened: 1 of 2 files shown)


cnasc:
Added an additional test case for when it returns successfully with an empty array of cards

PR #3985, opened 2019-06-25
Updates Key Pricer to Fetch Key Price from Chain

### LINKED ISSUE #3954: Get actual key price in locksmith
We need to determine the price on a lock so we can charge users appropriately

https://github.com/unlock-protocol/unlock/blob/master/locksmith/src/utils/keyPricer.ts

akeem:
# Description

Gets actual key price from Locks, there are some cascading changes from the update method signature due to required field in place. 

There will be a next time to split the library below which should hopefully clean this up a bit and reducing the extremely tight coupling to the implementation details of the unlock library

# Issues

Fixes #3954
Refs #

# Checklist:


locksmith/src/utils/keyPricer.ts
```diff
@@ -1,9 +1,22 @@
...
+      unlockAddress: unlockContractAddress,
+      blockTime: 0,
+      requiredConfirmations: 0,
+    })
+  }
+
   // eslint-disable-next-line no-unused-vars
-  keyPrice(_lockAddress: string): number {
-    return 220
+  async keyPrice(_lockAddress: string): Promise<number> {
```
cnasc: I think you can get rid of the underscore and eslint rule for this one

locksmith/package.json
```diff
@@ -41,6 +41,7 @@
     "@types/express": "4.17.0",
     "@types/express-serve-static-core": "4.16.7",
     "@types/jest": "24.0.14",
+    "@types/nock": "^10.0.3",
```
cnasc: I think we want to avoid the caret versions and just have pinned deps (and do dep installs separate from usual code)

PR #3988, opened 2019-06-25
fixing the paywall ts config

julien51:
# Description

This fixes storybook!

# Checklist:

### DIFF
--- a/paywall/tsconfig.json
+++ b/paywall/tsconfig.json
@@ -19,7 +19,8 @@
     // "rootDir": "./",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
     // "composite": true,                     /* Enable project compilation */
     "removeComments": false /* Do not emit comments to output. */,
-    //"noEmit": true,                           /* Do not emit outputs. */
+    "noEmit": true /* Do not emit outputs. */,
+    "outDir": "/tmp" /* this is actually not used but required to make storybook work */,
     // "importHelpers": true,                 /* Import emit helpers from 'tslib'. */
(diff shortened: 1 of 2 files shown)


paywall/tsconfig.json
```diff
@@ -19,7 +19,7 @@
     // "rootDir": "./",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
     // "composite": true,                     /* Enable project compilation */
     "removeComments": false /* Do not emit comments to output. */,
-    //"noEmit": true,                           /* Do not emit outputs. */
+    "noEmit": true /* Do not emit outputs. */,
```
cellog:
```suggestion
    //"noEmit": true,                           /* Do not emit outputs. */
    "outDir": "/anything-other-than-.-will-work"
```

julien51: So, if I comment it out again it fails obviously...

julien51: and on all other settings the `,` is after the comment, so for consistency I'll use the same approach ;)

cellog: did you try this fix?

cellog: it succeeds here, so would be interesting if it doesn't there

cellog: and to be clear: the original PR fails locally when I build with
Screen Shot 2019-06-25 at 12 19 14 PM

julien51:
This approach does not work. Closing until we find something better.

@cellog  will open an issue to keep track of that issue.

Issue #3771, opened 2019-06-13
v1 Sprint


julien51:
Our v1 release has been delayed because we are missing a couple critical pieces:

The v1 critical path to launch includes 4 requirements.

* Let users create their account on https://unlock-protocol.com/
* Let users login on their account from the paywall
* Let users enter their credit card information
* Let users purchase keys on the paywall

This sprint starts on Monday June 17th and ends on June 31st.

cnasc:
As I'm putting together the embed page, I want to make sure I have the flow down correctly:

If the user is not logged in, present a login form. That form will also have a link that opens the sign-up page on the main unlock-protocol site.

If the user is logged in, but does not have any payment methods associated with their account, they should see the payment details form.

If the user is logged in and does have a payment method associated with the account, they will see the key purchase confirmation. After the button is clicked and the key purchase is accepted by locksmith, the embed will relinquish control back to the paywall.

Thoughts @julien51?

PR #3992, opened 2019-06-25
Populate the iframe embed page (user account checkout)

cnasc:
# Description

This PR adds components to `AccountContent.tsx` along with logic to switch around between them based on the current app state. This is an intermediate step along the way, not the final product. Styling will certainly have to be done, and not all of the components are final yet. Additionally, I think the final embed will use a different layout (no top nav, etc.)

Here's the flow:

1. iframe loads, there is no user account in state. We show the LogIn component. (NB. this will need to be massaged so the sign up link opens a new tab to the signup page on the site)
image

2. Once logged in, if there are no payment methods available for use we show the payment details form. (NB. stripe doesn't load in storybook so this screenshot is bare)
image

3. Once logged in with a payment method available, we show the confirmation dialog
image

# Issues

Refs #3771 

# Checklist:

### DIFF
--- a/unlock-app/src/stories/content/AccountContent.stories.js
+++ b/unlock-app/src/stories/content/AccountContent.stories.js
@@ -1,0 +1,48 @@
+import React from 'react'
+import { Provider } from 'react-redux'
+import { storiesOf } from '@storybook/react'
+import { select, withKnobs } from '@storybook/addon-knobs'
+import createUnlockStore from '../../createUnlockStore'
+import { ConfigContext } from '../../utils/withConfig'
+import AccountContent from '../../components/content/AccountContent'
+
+const baseState = {
+  account: {},
+}
+const loggedInState = {
+  account: {
+    emailAddress: 'jenny@googlemail.com',
+  },
+}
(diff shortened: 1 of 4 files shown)


unlock-app/src/stories/content/AccountContent.stories.js
```diff
@@ -0,0 +1,48 @@
...
+
+const config = {
+  stripeApiKey: 'pk_this_is_not_a_real_key',
+}
+
+storiesOf('AccountContent (iframe embed for paywall)', module)
+  .addDecorator(withKnobs)
+  .add('The embed', () => {
+    const label = 'Component State'
+    const options = {
```
cellog: really clever use of knobs

cellog: would also be useful to have separate stories for each state, so chromatic will catch any changes to them

cnasc: Yup, that's something I plan to do once the "real" components are in. We'll have one for everything but the Stripe one.

Issue #3767, opened 2019-06-13
Implement “proxyProvider” logic on paywall


julien51:
Once the iframe has been added to the DOM (#3766) our existing proxyProvider should send messages to it (#3765) to achieve the following:
* Get the current network 
* Get the user account (so that the paywall can check whether the user has keys…)
* Perform key purchases

julien51:
Once the iframe has been added to the DOM (#3766) our existing proxyProvider should send messages to it (#3765) to achieve the following:
Get the current network 
Get the user account (so that the paywall can check whether the user has keys…)
Perform key purchases

cellog:
done!

PR #3995, opened 2019-06-25
introducing the proxyProvider

cellog:
# Description

This implements the proxyProvider. It adds:

1. new listeners for `READY`, `UPDATE_ACCOUNT`, `UPDATE_NETWORK`, and the show/hide messages coming from the account iframe to the main window. We auto-show the account iframe if the user has not logged in and has no wallet.
2. a check for proxy account if we have no wallet. It waits for the account iframe to send its account info downstream in this case (TODO: improve loading speed of account iframe)
3. when receiving a request from `walletService` for the current account and network, it intercepts and returns our proxy account/network instead.
4. interception of `PURCHASE_KEY` from the checkout UI and sends it to account iframe instead of the data iframe if we are using a user account.
5. we now also relay locks to the account iframe

This is a big PR, most of the bulk is in the test, which has been reorganized to be a bit easier to modify in the future.

TODO:

1. use the locks values to determine whether to use user accounts. I.e. we need to check to be sure that there are any ERC-20 locks on the page prior to using the accounts iframe.

# Issues

Refs #3767 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/unlock.js/setupPostOffices.test.ts
+++ b/paywall/src/__tests__/unlock.js/setupPostOffices.test.ts
@@ -142,27 +142,6 @@ describe('setupPostOffice', () => {
     expect(fakeUIIframe.className).toBe('unlock start show')
   })
 
-  it('responds to PostMessages.READY_WEB3 by sending PostMessages.WALLET_INFO', async () => {
-    expect.assertions(2)
-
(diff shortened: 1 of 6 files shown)


cellog:
I'm going to go ahead and close this, already know hte integration tests will fail

Issue #3764, opened 2019-06-13
Implement message handling in unlock app


julien51:

julien51:
The paywall application is able to "embed" an iframe with the content of Unlock-app on sites on which it has been loaded.
Communication between the publisher page and the iframe is performed using the `postMessage` API.

Let's make sure we list the messages before working on the implementation.

cellog:
this is implemented

PR #3996, opened 2019-06-25
merge lock, key, locks, account validators from paywall

cellog:
# Description

In preparation for the `postOfficeService`, this copy/pastes the validators needed for the list of locks that will come in to the account iframe

# Issues

Refs #3764 

# Checklist:

### DIFF
--- a/unlock-app/src/utils/validators.js
+++ b/unlock-app/src/utils/validators.js
@@ -28,12 +28,12 @@ export const isPositiveNumber = val => {
   return !isNaN(parsedFloat) && +parsedFloat >= 0
 }
 
-export const isAccountOrNull = val => {
-  return val === null || (typeof val === 'string' && val.match(ACCOUNT_REGEXP))
-}
-
 export const isAccount = val => {
   return val && typeof val === 'string' && val.match(ACCOUNT_REGEXP)
+}
+
+export const isAccountOrNull = val => {
+  return val === null || isAccount(val)
 }
 
 /**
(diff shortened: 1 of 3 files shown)


unlock-app/src/utils/validators.js
```diff
@@ -19,3 +27,130 @@ export const isPositiveNumber = val => {
   const parsedFloat = parseFloat(val)
   return !isNaN(parsedFloat) && +parsedFloat >= 0
 }
+
+export const isAccountOrNull = val => {
+  return val === null || (typeof val === 'string' && val.match(ACCOUNT_REGEXP))
```
akeem: could this call `isAccount` for the right hand side of the || ?

unlock-app/src/utils/validators.js
```diff
@@ -19,3 +27,130 @@ export const isPositiveNumber = val => {
...
+
+  if (keys.length < validKeys.length) return false
+  if (validKeys.filter(key => !keys.includes(key)).length) return false
+  return true
+}
+
+/**
+ * validate a single key. This should match the type in unlockTypes.ts
+ */
+export const isValidKey = key => {
```
akeem: There is a lot going on in this method, lets see if we can break this down a bit, totally fine if we extract some supporting methods here

unlock-app/src/utils/validators.js
```diff
@@ -19,3 +27,130 @@ export const isPositiveNumber = val => {
...
+  }
+  */
+  return true
+}
+
+/**
+ * validate a single lock. This should match the type in unlockTypes.ts
+ */
+export const isValidLock = lock => {
+  if (
```
akeem: Same as the method above

cellog:
PTAL @akeem

PR #4004, opened 2019-06-26
Skipping jobs when changes are not relevants

julien51:
# Description

In order to speed our builds (and decrease the costs on both Circle and Chromatic!) I am introducing
a new script which will skip jobs if:
1. The build is not on master (we run all jobs on master)
2. The files changed are not relevant to the job in question

# Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,8 @@ jobs:
     steps:
       - checkout
       - run:
+          command: scripts/monorepo.sh unlock-app $CIRCLE_SHA1
+      - run:
           name: Build unlock-app Image
           command: scripts/build-image.sh unlock-app
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,8 @@ jobs:
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


julien51:
For now, the only thing which runs every-time will be integration tests!

cnasc:
I think we have to adjust the chromatic config to handle the cases where the stories don't actually run

julien51:
So, this is annoying. 2 options:
* Find a way to tell Chromatic to tell Github that things are ok
* Make Chromatic "not required"

The 2nd option is easy but I worry that we may merge PR which actually do not pass Chromatic in the future?

cnasc:
I think the first option is the better one. I tried searching for ways to make Chromatic report a success without running anything. Maybe we should reach out to their support?

julien51:
Yup! Done that ;)

julien51:
The only thing that is not perfect about that first solution is that it will still require us to build the docker image first tho...

julien51:
I am worried there is no good way of doing 1. now, so for now I will make the ChromaticQa checks optional.

cnasc:
> I am worried there is no good way of doing 1. now, so for now I will make the ChromaticQa checks optional.

We'll just have to be vigilant that we don't accidentally merge regressions

PR #4010, opened 2019-06-26
Extracts library to manage Token Transfer Approval

akeem:
# Description

As part of the process of easily enabling fufillment in the development environment this functionality has been extracted from the existing script currently in place.

# Issues

Fixes #
Refs #

# Checklist:


locksmith/__tests__/fulfillment/tokenTransferApproval.test.ts
```diff
@@ -0,0 +1,55 @@
+import { BigNumber } from 'ethers/utils'
+import TokenTransferApproval from '../../src/fulfillment/tokenTransferApproval'
+
+const nock = require('nock')
+
+nock.back.fixtures = `${__dirname}/fixtures/paymentProcessor`
```
cellog: can you comment these lines to describe how they are used?

locksmith/src/fulfillment/tokenTransferApproval.ts
```diff
@@ -0,0 +1,26 @@
+import { ethers } from 'ethers'
+
+export default class TokenTransferApproval {
+  provider: ethers.providers.JsonRpcProvider
+  privateKey: string
+
+  constructor(provider: string, credentials: string) {
```
cnasc:
```suggestion
  constructor(provider: string, privateKey: string) {
```

Not a huge deal, but might make it more clear at call sites what goes here when using code completion

locksmith/src/fulfillment/tokenTransferApproval.ts
```diff
@@ -0,0 +1,26 @@
...
+    this.provider = new ethers.providers.JsonRpcProvider(provider)
+    this.privateKey = credentials
```
cnasc:
```suggestion
    this.privateKey = privateKey
```

see above

locksmith/__tests__/fulfillment/tokenTransferApproval.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      let transferApprover = new TokenTransferApproval(provider, credentials)
+      let transaction = await transferApprover.approve(
+        '200',
+        lockContractAddress,
+        erc20ContractAddress
+      )
+      expect(transaction).toEqual(
+        expect.objectContaining({
+          chainId: 1984,
+          data:
```
cellog: is there any way this can be generated from an ethers.utils.Interface() instead of being impenetrable hex? if not, I suppose it's fine, but the test will fail with the smallest smart contract change, and it could be hard to verify that the difference is because of this and not a bug

locksmith/src/fulfillment/tokenTransferApproval.ts
akeem: sounds good, will make the update

PR #4011, opened 2019-06-26
add "for" address for key purchase transactions

### LINKED ISSUE #4009: unlock-js needs to process and set the "for" field in all key purchase transactions
In order to support user accounts, unlock-js needs to set the `for` field, so we can correctly process all transations on behalf of someone else

cellog:
# Description

This turned out to be incredibly simple: we just take the `owner` we already use for the key id, and pass it as `for`

# Issues

Fixes #4009 

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/v10/web3Service.test.js
+++ b/unlock-js/src/__tests__/v10/web3Service.test.js
@@ -324,9 +324,10 @@ describe('Web3Service', () => {
           ],
         }
 
-        web3Service.on('transaction.updated', (tHash, lock) => {
+        web3Service.on('transaction.updated', (tHash, transactionUpdate) => {
           expect(tHash).toBe('hash')
-          expect(lock).toEqual({
+          expect(transactionUpdate).toEqual({
+            for: checksumLockAddress,
             key: KEY_ID(lockAddress, checksumLockAddress),
             lock: lockAddress,
           })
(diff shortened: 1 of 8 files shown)


unlock-js/src/__tests__/v10/web3Service.test.js
```diff
@@ -324,9 +324,10 @@ describe('Web3Service', () => {
           ],
         }
 
-        web3Service.on('transaction.updated', (tHash, lock) => {
+        web3Service.on('transaction.updated', (tHash, transactionUpdate) => {
           expect(tHash).toBe('hash')
-          expect(lock).toEqual({
+          expect(transactionUpdate).toEqual({
+            for: checksumLockAddress,
```
akeem: This is probably just a naming thing, but is `for` the LockAddress or is it for the recipient - the address that will receive the key?

cellog: the recipient, as in "this key is for Joe"
The lock is the "to" address, the purchaser is the "from" address. What do you think about naming now that you have this context?

cellog: ok, I just pushed an update to fix the `checksumLockAddress` to be `owner` instead

cellog:
PTAL @akeem when you get a moment

Issue #3970, opened 2019-06-24
localStorage being disabled in the browser appears to kill the paywall instead of just disabling caching


cellog:

cellog:
this looks like the right approach:

https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/

PR #4018, opened 2019-06-26
fix assorted bugs

cellog:
# Description

This is a hodge-podge of fixes to the data iframe, mostly small changes.

1. because `'transaction.updated'` from `web3Service` never emits the full transaction, we need to merge its value with the previous one. This changes merge to have this behavior.
2. makeSetConfig needs to wait for the `onChange` handler to finish before passing control back to avoid race conditions
3. The cacheHandler was running updates for `keys` and `transactions` in parallel using `Promise.all`. This would cause them both to use old versions of local storage, and one would overwrite the other's update with a stale value. This changes it to run the updates in sequence instead.
4. The same bug struck `syncToCache` - we can't use `Promise.all` here.
5. the check for networkId in the post office was too strict (`=== null`) when checking for falsy suffices, because a `0` network does not exist.

# Issues

Refs #3970 

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/cache/localStorage.js
+++ b/paywall/src/data-iframe/cache/localStorage.js
@@ -126,7 +126,11 @@ export async function merge({
     delete container[type][subType]
   } else {
     container[type] = container[type] || {}
-    container[type][subType] = value
+    const oldType = container[type][subType] || {}
+    container[type][subType] = {
+      ...oldType,
+      ...value,
+    }
   }
 
   window.localStorage.setItem(key, JSON.stringify(container))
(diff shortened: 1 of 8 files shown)


paywall/src/data-iframe/cache/localStorage.js
```diff
@@ -126,7 +126,11 @@ export async function merge({
     delete container[type][subType]
   } else {
     container[type] = container[type] || {}
-    container[type][subType] = value
+    const oldType = container[type][subType] || {}
```
akeem: we just chatted about this IRL but we guess this could be renamed to oldType

cellog: just made the same mistake I did ;) renaming to `oldValue`

PR #4101, opened 2019-07-04
gets locks from locksmith

julien51:
# Description

We now retrieve the lock addresses from locksmith on the dashboard.
This has 2 benefits:
1. faster loading (we do not have to get the transactions before being able to get these locks)
2. locks which have been transfered are now also loaded on the dashboard.

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/services/storageService.test.js
+++ b/unlock-app/src/__tests__/services/storageService.test.js
@@ -610,4 +610,74 @@ describe('StorageService', () => {
       })
     })
   })
+
+  describe('getLockAddressesForUser', () => {
+    it('should retrieve the list of locks for a user and emit emit success.getLockAddressesForUser', done => {
+      expect.assertions(2)
+      const user = '0xabc'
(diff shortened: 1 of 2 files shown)


unlock-app/src/__tests__/services/storageService.test.js
```diff
@@ -57,7 +57,7 @@ describe('StorageService', () => {
 
   describe('lockLookUp', () => {
     describe('when the requested lock exists', () => {
-      it('returns the details', done => {
+      it.todo('returns the details', done => {
```
akeem: It looks like the todo here is causing the test to fail.  We should be able to keep this test if the portions around the name assertion are removed and we ensure that the address is present on the mocked data

julien51: oopsie! removing it now.

julien51:
Ok, PTAL @akeem !

I think I will look at a replacement for our CLA Assistant in the meantime.

PR #4002, opened 2019-06-25
introducing proxyProvider

cellog:
# Description

This is the final piece of the puzzle on the paywall side for user accounts.

It does these pieces of that puzzle:

1. send locks to the user account iframe
2. if we have any erc20 locks, and no wallet, respond to `eth_accounts` and `net_version` with the user accounts values
3. if we are using user accounts, divert purchase key requests to the user account iframe
4. if a user starts a key purchase transaction, notify the data iframe to refresh transactions for monitoring
5. listens for instructions to show/hide the account iframe
6. shows the account iframe if the user is not logged in on start (and there are ERC20 locks on the page)

# Issues

Fixes #3767 

# Checklist:

- [X] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [X] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [X] My code follows the style guidelines of this project, including naming conventions
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [X] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #4134, opened 2019-07-09
add the checkout iframe after the accounts iframe

cellog:
# Description

This is a slice-and-dice of #4002.

This changes the ordering of the accounts and checkout iframe so that the checkout iframe will always be beneath the user accounts iframe.

This is necessary because we need user login to be above any checkout. Once the user has logged in, the checkout iframe will still be visible, and can then be used to initiate a key purchase. At this point, the accounts iframe will pop up again to do the credit card purchase, and then once it hides, the checkout iframe can do its transaction monitoring/flag displaying thing.

# Issues

Refs #3767 

# Checklist:

### DIFF
--- a/tests/helpers/iframes.js
+++ b/tests/helpers/iframes.js
@@ -1,0 +1,7 @@
+// This file is a helper to abstract the retrieval of the iframes on the paywall,
+// so that any re-ordering of frames is controlled in a single location
+
+module.exports = {
+  checkoutIframe: page => page.mainFrame().childFrames()[2],
+  accountsIframe: page => page.mainFrame().childFrames()[1],
+}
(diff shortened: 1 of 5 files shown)


tests/helpers/iframes.js
```diff
@@ -0,0 +1,7 @@
+// This file is a helper to abstract the retrieval of the iframes on the paywall,
+// so that any re-ordering of frames is controlled in a single location
+
+module.exports = {
```
akeem: Are the order here and in the subsequent product code important? What happens if the order changes?

cellog: yes, prior to this PR, the grabbing of which iframe was which was done in the test. When I needed to change the ordering, it upset the frame number. As it turns out, I can't reliably identify frames by their source in puppeteer, although I did not spend a long time because there were other pressing things. If this becomes an issue again, we should find another way to identify the frames.

What happens is the integration tests time out on finding any of the things, because they are looking in the wrong iframe

PR #4138, opened 2019-07-09
refactor enable to avoid using exceptions for logic flow

cellog:
# Description

This is a slice/dice of #4002 

It refactors `enable` to avoid using exceptions for logic flow (H/T to @akeem).

In preparation for #4002 it also extracts the logic for what will become enabling user accounts into a separate function.

As this is a refactor, no functionality changed, and the tests still pass as-is.

Note: the diff for `web3Proxy.test.ts` will get significantly smaller when #4135 is merged and this is rebased

# Issues

Refs #3767 

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/web3Proxy.ts
+++ b/paywall/src/unlock.js/web3Proxy.ts
@@ -1,5 +1,9 @@
-import { Web3Window, web3MethodCall } from '../windowTypes'
-import { MapHandlers, MessageHandlerTemplates } from './setupIframeMailbox'
+import { Web3Window, web3MethodCall, IframeType } from '../windowTypes'
+import {
+  MapHandlers,
+  MessageHandlerTemplates,
+  PostMessageToIframe,
+} from './setupIframeMailbox'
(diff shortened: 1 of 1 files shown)


paywall/src/unlock.js/web3Proxy.ts
```diff
@@ -1,13 +1,17 @@
...
+} from './setupIframeMailbox'
 import { PostMessages, MessageTypes } from '../messageTypes'
 
 let hasWeb3 = true
 
 export function enable(window: Web3Window) {
   return new window.Promise((resolve, reject) => {
     if (!window.web3 || !window.web3.currentProvider) {
-      return reject(new ReferenceError('no web3 wallet exists'))
+      return resolve('no web3 wallet')
```
akeem: Not a blocker, but we should probably constantize and utilize that on line 77 of this file in the try

cellog: definitely, I'll make it so

Issue #2182, opened 2019-03-18
Merge redundant contract functions such as purchaseFor/purchaseForFrom


NickCuso:
**Is your feature request related to a problem? Please describe.**
As discovered in a recent investigation #1757 , each function included in the contract potentially makes calling other functions cost more gas.

**Describe the solution you'd like**
We have a couple redundant functions that could be merged into one.  For example `purchaseFor`/`purchaseForFrom` and `withdraw`/`partialWithdraw`.

**Describe alternatives you've considered**

**Additional context**
Merging the external functions may also allow us to remove the private helper function, which potentially saves gas by reducing the function table.

NickCuso:
Remove numberOfKeysSold and replace with `totalSupply` public variable.

PR #4143, opened 2019-07-10
Merge purchaseFor / purchaseForFrom

NickCuso:
# Description

Merging purchaseFor and purchaseForFrom into a single `purchase` function.

# Issues

Fixes #
Refs #2182

# Checklist:


smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -22,53 +22,22 @@ contract MixinPurchase is
...
-    hasValidKey(_referrer)
-  {
-    return _purchaseFor(_recipient, _referrer);
-  }
-
-  /**
-  * @dev Purchase function: this lets a user purchase a key from the lock for another user
-  * @param _recipient address of the recipient of the purchased key
   * This will fail if
   *  - the keyReleaseMechanism is private
```
nfurfaro: We could probably remove references to the "keyReleaseMechanism" as well( if we're not going to be introducing private or restricted locks). These are old and confusing to someone who doesn't know what they are.

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -22,53 +22,22 @@ contract MixinPurchase is
...
   *  - the keyReleaseMechanism is Approved and the recipient has not been previously approved
   *  - the amount value is smaller than the price
   *  - the recipient already owns a key
-  * TODO: next version of solidity will allow for message to be added to require.
   */
-  function _purchaseFor(
+  function purchase(
     address _recipient,
     address _referrer
```
nfurfaro: So fo r a purchase with no referrer, a 0 has to be passed in as the second param here right? I'm on board with pushing that responsibility to the front end in order to simplify things in the contract.

nfurfaro:
Just s thought here, but could we not delete all of the tests with names starting with "Non_Public..."?

smart-contracts/contracts/mixins/MixinPurchase.sol
NickCuso: Hmm yea.  Actually none of those points are valid.  Deleted.

smart-contracts/contracts/mixins/MixinPurchase.sol
NickCuso: That's correct

NickCuso:
We could start to rename the files, but we should be careful about removing them unless we are sure the scenario is covered in a different test.

PR #4156, opened 2019-07-11
enable proxyProvider

### LINKED ISSUE #3767: Implement “proxyProvider” logic on paywall

cellog:
# Description

This is the same as #4002, but trimmed down and applied to `unlock_provider_base` instead of `master`. Some of this can probably still be trimmed, but it is small enough to be readable now.

# Issues

Fixes #3767

# Checklist:

### DIFF
--- a/paywall/src/messageTypes.ts
+++ b/paywall/src/messageTypes.ts
@@ -30,6 +30,9 @@ export enum PostMessages {
   PURCHASE_KEY = 'purchaseKey',
   DISMISS_CHECKOUT = 'dismiss/checkout',
   INITIATED_TRANSACTION = 'initiated/transaction',
+
+  SHOW_ACCOUNTS_MODAL = 'show/accountsModal',
+  HIDE_ACCOUNT_MODAL = 'hide/accountsModal',
 }
 // all the possible message types
 export type Message =
@@ -91,7 +94,7 @@ export type Message =
     }
   | {
       type: PostMessages.UPDATE_ACCOUNT
-      payload: string
+      payload: string | null
     }
   | {
       type: PostMessages.UPDATE_ACCOUNT_BALANCE
@@ -123,6 +126,14 @@ export type Message =
     }
   | {
       type: PostMessages.INITIATED_TRANSACTION
+      payload: undefined
+    }
+  | {
+      type: PostMessages.SHOW_ACCOUNTS_MODAL
+      payload: undefined
+    }
+  | {
+      type: PostMessages.HIDE_ACCOUNT_MODAL
       payload: undefined
     }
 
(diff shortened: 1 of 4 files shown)


paywall/src/unlock.js/web3Proxy.ts
```diff
@@ -42,22 +110,104 @@ export default function web3Proxy(
...
     (window.web3.currentProvider.sendAsync || window.web3.currentProvider.send)
+  let proxyAccount: null | string = null
+  let proxyNetwork: string | number
+  let accountIframeReady = false
+  let currentLocks: Locks
+  let canUseUserAccounts: boolean = !window.web3
+  let hasNativeWeb3Wallet = !!window.web3
+
+  const useUnlockAccount = () =>
+    !hasNativeWeb3Wallet && proxyAccount && canUseUserAccounts
```
cnasc:
```suggestion
    proxyAccount && canUseUserAccounts
```

cellog: this was the original code, but it turns out it does not work. There are some cases where we have a native wallet present and it should use that (for instance, when we have not yet enabled access to the page)

paywall/src/messageTypes.ts
```diff
@@ -125,6 +128,14 @@ export type Message =
       type: PostMessages.INITIATED_TRANSACTION
       payload: undefined
     }
+  | {
+      type: PostMessages.SHOW_ACCOUNTS_MODAL
```
cnasc: I think we should consistently use `ACCOUNT` or `ACCOUNTS` on these names.

paywall/src/unlock.js/web3Proxy.ts
```diff
@@ -42,22 +110,104 @@ export default function web3Proxy(
...
     postMessage: PostMessageToIframe<MessageTypes>
   ) => {
     // we don't have web3
-    hasWeb3 = false
+    // wait for locks to be retrieved
+    await waitFor(() => currentLocks)
+    // wait for the account iframe, then respond
+    await waitFor(() => accountIframeReady)
+    // we will use the proxy account!
+    if (!canUseUserAccounts) {
```
cnasc: There are a few places in this file where a variable is inverted, even though we already have another variable that represents the inversion of that variable. Is there another way to handle this?

paywall/src/unlock.js/web3Proxy.ts
cnasc: Can you provide some more context? What are the cases where `canUseUserAccounts` and `!hasNativeWeb3Wallet` are different?

Issue #3848, opened 2019-06-14
Integration test for a key purchase on an ERC20 lock


julien51:
This is a test to be added to the paywall or ad remover to make sure that the user can purchase a key on an ERC20 lock.

cellog:
probably best to do this with the ad remover paywall since soon that will be the paywall?

PR #4157, opened 2019-07-11
use the dev ganache for integration tests, to add erc20 support

cellog:
# Description

This uses the development ganache in integration tests. This allows us to take advantage of the testing ERC20 smart contract, and also priming the first ganache user with minted ERC20 coins.

In order to make this work, I had to convert the standup to accept a parameter for the block time. The `docker-compose.ci.yml` file sets it to `1` second, and it is `3` seconds by default, to preserve the existing behavior in development. This will also allow us to set it to a higher value if we wish, using something like `--build-arg blocktime=15` to test timing more similar to mainnet.

Also, since the env variables are static for integration tests, rather than use a `.env.dev.local`, it just defines them directly in the `docker-compose.ci.yml` file.

This will also allow us to remove the `deploy-unlock.js` file from `unlock-app/scripts` as well as the `deploy-unlock-contract` npm command.

# Issues

Refs #3848 

# Checklist:


docker/docker-compose.ci.yml
```diff
@@ -8,10 +8,22 @@ services:
...
-    image: trufflesuite/ganache-cli:v6.4.3
+    build:
+      context: ./development
+      args:
+        - blocktime=1
+      dockerfile: ganache.dockerfile
+    environment:
+      CI: '${CI}'
+      ERC20_CONTRACT_ADDRESS: '0x591AD9066603f5499d12fF4bC207e2f577448c46'
+      LOCKSMITH_PURCHASER_ADDRESS: '0xe29ec42f0b620b1c9a716f79a02e9dc5a5f5f98a'
```
julien51: Can we add the BLOCK TIME var based on '${BLOCKTIME} here please. This will bring flexibility and solve problems for the Nicks too.

docker/docker-compose.ci.yml
```diff
@@ -8,10 +8,22 @@ services:
...
     restart: always
-    image: trufflesuite/ganache-cli:v6.4.3
+    build:
+      context: ./development
+      args:
+        - blocktime=1
+      dockerfile: ganache.dockerfile
+    environment:
+      CI: '${CI}'
+      ERC20_CONTRACT_ADDRESS: '0x591AD9066603f5499d12fF4bC207e2f577448c46'
```
julien51: Can we not hardcode the values in this file like we do for the rest of this file? Consistency ftw

scripts/integration-tests.sh
```diff
@@ -12,8 +12,7 @@ mkdir -p /tmp/screenshots
 chmod 0777 /tmp/screenshots
 
 # Deploy the smart contract
-COMMAND="npm run deploy-unlock-contract"
-docker-compose -f $DOCKER_COMPOSE_FILE run -v /tmp/screenshots:/screenshots $EXTRA_ARGS unlock-contract-deployer bash -c "$COMMAND"
+docker-compose -f $DOCKER_COMPOSE_FILE build ganache-integration
```
julien51: I wonder why we remove the screenshot stuff?

tests/helpers/erc20IsUp.js
```diff
@@ -0,0 +1,85 @@
+// This file is used to probe the ganache server. It attempts to call
+// balanceOf(0xaaadeed4...) to get the ERC20 token balance.
+// when the standup script has finished, it will return 500 tokens,
+// which is assumed if our call to balanceOf returns a non-zero value.
+const axios = require('axios')
+const url = require('./url')
+
+const erc20ContractAddress = '0x591AD9066603f5499d12fF4bC207e2f577448c46'
```
julien51: Let's use env vars here?

tests/helpers/erc20IsUp.js
```diff
@@ -0,0 +1,85 @@
+// This file is used to probe the ganache server. It attempts to call
+// balanceOf(0xaaadeed4...) to get the ERC20 token balance.
+// when the standup script has finished, it will return 500 tokens,
+// which is assumed if our call to balanceOf returns a non-zero value.
+const axios = require('axios')
```
julien51: Do we really need to add a dependency here?

tests/helpers/erc20IsUp.js
```diff
@@ -0,0 +1,85 @@
...
+const balanceOfQuery =
+  '0x70a08231000000000000000000000000aaadeed4c0b861cb36f4ce006a9c90ba2e43fdc2'
+let id = 1
+
+const erc20IsUp = ({ delay, maxAttempts }) => {
+  let attempts = 1
+  let lastResponse = undefined
+  return new Promise((resolve, reject) => {
+    function retrieveBalance() {
+      axios
```
julien51: I feel like this is simple enough and the node http lib can provide what we need?

scripts/integration-tests.sh
cellog: ganache doesn't use it, this all runs prior to the integration test runs

cellog:
ok @julien51 it took a LONG time to learn how to use http, but I eventually succeeded! :)

Everything uses env variables like the other docker processes, so it should be good to go

PR #4158, opened 2019-07-11
Prepare the currency switcher to be integration testable

cellog:
# Description

This allows us to use the dashboard helper to make new erc20 locks for use by integration tests for the paywall.

It also simplifies the logic of displaying currency, and makes sure that we only have 1 text node `Use DAI` instead of 2, which makes selecting the element to click it in Puppeteer possible.

# Issues

Refs #3848 

# Checklist:

### DIFF
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -325,14 +325,9 @@ export class CreatorLockForm extends React.Component {
             data-valid={valid.keyPrice}
             required
           />
-          {isNew && !currency && (
+          {isNew && (
             <LockLabelCurrency onClick={this.toggleCurrency}>
-              Use {ERC20Contract.name}
-            </LockLabelCurrency>
-          )}
-          {isNew && !!currency && (
-            <LockLabelCurrency onClick={this.toggleCurrency}>
-              Use Ether
+              {`Use ${!currency ? ERC20Contract.name : 'Ether'}`}
             </LockLabelCurrency>
           )}
         </FormBalanceWithUnit>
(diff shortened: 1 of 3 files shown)


unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -325,14 +325,9 @@ export class CreatorLockForm extends React.Component {
...
-          {isNew && !currency && (
+          {isNew && (
             <LockLabelCurrency onClick={this.toggleCurrency}>
-              Use {ERC20Contract.name}
-            </LockLabelCurrency>
-          )}
-          {isNew && !!currency && (
-            <LockLabelCurrency onClick={this.toggleCurrency}>
-              Use Ether
+              {`Use ${!currency ? ERC20Contract.name : 'Ether'}`}
```
akeem: How would you feel about extracting this logic outside of the component? In general, it should allow us to better test the general behavior without needing to render the component. 

While not a React specific approach, I quick google search yielded https://javascriptplayground.com/react-extracting-logic/ which I think covers the gist

cellog: sure. That is outside the scope of this PR, which is simply to add the `currency` classname so we can select it, and then the code above is a simplification of what was already there, but if you want me to remove the changes and only add the classname, I will do that.

akeem: yeah lets do what is in scope for this PR, updating the classname. Let's leave the simplification code out for now

We can add a follow up task here to extract the logic out of this stuff, making CreatorLockForm easier to test in the long run

akeem:
#4161 is the task created to capture the out of scope need to extract raised here

cellog:
k @akeem PTAL

Issue #4173, opened 2019-07-12
User account key purchase postMessage flow


cellog:
Pre-key purchase

1. user accounts --> `PostMessages.READY` --> unlock.min.js
   a. if there is no user account (user is not logged in), --> `PostMessages.SHOW_ACCOUNTS_MODAL` --> unlock.min.js
2. unlock.min.js --> `PostMessages.CONFIG` (payload is contents of `window.unlockProtocolConfig`) --> user accounts
3. unlock.min.js --> 2 `PostMessages.SEND_UPDATES` (payload `'account'` and payload `'network'`) --> unlock.min.js
4. unlock.min.js --> `PostMessages.UPDATE_LOCKS` --> user accounts iframe. Payload is a map of locks indexed by lower-cased address
5. user accounts --> `PostMessages.UPDATE_ACCOUNT` --> unlock.min.js
6. user accounts --> `PostMessages.UPDATE_NETWORK` --> unlock.min.js

Login flow:
1. user creates account if there is none
2. upon returning, the above flow happens
3. user logs in
4. user accounts --> `PostMessages.HIDE_ACCOUNTS_MODAL` --> unlock.min.js
5. user accounts --> `PostMessages.UPDATE_ACCOUNT` (payload is user address) --> unlock.min.js

Key purchase flow:
1. unlock.min.js --> `PostMessages.PURCHASE_KEY` (payload is lock address and extraTip of `'0'`)
2. user account --> `PostMessages.SHOW_ACCOUNTS_MODAL` --> unlock.min.js
3. <user purchases key with credit card>
4. When locksmith returns "success", user accounts --> `PostMessages.INITIATED_TRANSACTION` --> unlock.min.js
  a. If locksmith returns "fail", show error message. When user dismisses the message manually, either allow them to try again, or send `PostMessages.HIDE_ACCOUNTS_MODAL` --> unlock.min.js to show the checkout screen again
5. user account --> `PostMessages.HIDE_ACCOUNTS_MODAL` --> unlock.min.js

At this point, the paywall will query locksmith for all transactions for the locks in the paywall config for the current user, and begin monitoring the latest transaction on the lock.

Issue #4164, opened 2019-07-11
App-side changes for user account key purchase


cnasc:
Tracking needed changes on the app side of a user account key purchase.

- [x] CSS for iframe (white background, sizing). A stopgap for convenience is acceptable at this juncture
- [x] Changing login component to allow account creation more easily (opening new tab will cause issues with next on local dev - figure out an alternative)
- [x] Getting locks - can't just add to cart on receiving UPDATE_LOCKS. Ask @cellog  if we can change it so KEY_PURCHASE passes the whole lock under consideration (minimizes state and logic within unlock-app)
- [x] show actual key price on confirmation component (currently hardcoded)
- [ ] Make container for embed elastic, so that it is only as big as it needs to be for the current step. Currently it grows in height if needed, but can be improved.
- [x] Make various components more like the Figma designs (https://www.figma.com/file/LkWSSnCGlKEITBB5T5NYXW/Paywall?node-id=0%3A817)
  - [x] Log in
  - [x] Sign up? Not sure if we're going with the Figma design on this one
  - [x] Credit card form _(I think @benwerd  is taking this one)_
  - [x] Purchase confirmation (_claimed_)
  - [x] add an X at the top of the embed to dismiss
- [x] Correctly log in immediately after successful signup

PR #4176, opened 2019-07-12
app-side postMessage handling now conformant with expectations

cnasc:
# Description

This PR makes the unlock-app side of a user account key purchase act the way the paywall needs it to.

1. On load, it sends a ready (handled by postOfficeService) and a null account.
2. When receiving a lock update, it stores the lock information locally to the middleware. (no reducer because the locks object is only ever needed in this middleware)
3. After a user logs in, the account address is sent to the paywall and the account iframe disappears.
4. When it receives a key purchase request, it selects the appropriate lock from the object in [2], and dispatches ADD_TO_CART
5. After the key purchase is confirmed, it sends the transaction initiated postmessage and hides itself.

# Issues

Refs #4173, #4164 

# Checklist:

### DIFF
--- a/unlock-app/src/middlewares/postOfficeMiddleware.ts
+++ b/unlock-app/src/middlewares/postOfficeMiddleware.ts
@@ -18,20 +18,34 @@ const postOfficeMiddleware = (window: IframePostOfficeWindow, config: any) => {
   // No account yet, will be set after login
   postOfficeService.setAccount(null)
 
+  // Locks on the paywall, keys are lower-cased lock addresses
+  // no need for a reducer here because this state is entirely local to the
+  // postOffice
+  let locksOnPaywall: { [key: string]: any } = {}
+
   return ({ dispatch }: any) => {
     postOfficeService.on(PostOfficeEvents.Error, message => {
       dispatch(setError(PostOffice.Diagnostic(message)))
     })
 
-    postOfficeService.on(PostOfficeEvents.KeyPurchase, (lock, tip) => {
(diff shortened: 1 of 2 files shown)


unlock-app/src/middlewares/postOfficeMiddleware.ts
```diff
@@ -18,20 +18,34 @@ const postOfficeMiddleware = (window: IframePostOfficeWindow, config: any) => {
   // No account yet, will be set after login
   postOfficeService.setAccount(null)
 
+  // Locks on the paywall, keys are lower-cased lock addresses
+  // no need for a reducer here because this state is entirely local to the
+  // postOffice
+  let locksOnPaywall: { [key: string]: any } = {}
```
cellog: Let's use the `Locks` type here in place of `any`, it will give you auto-completion elsewhere

cnasc: Wasn't sure if this was strictly the same as the type we defined in `unlockTypes` since postOfficeService passed them through as `any`. I'll update in both spots if that's ok.

cellog: technically, it can be tightened up quite a bit, I have that branch sitting open that does this in a flexible way (limits types based on the kind of message posted), but if it breaks things to not use `any` let's keep it for now, and tighten later.

cnasc: I'm not 100% that it will break anything, but let's move forward on this one and then clean up

Issue #4165, opened 2019-07-11
Paywall issues for user account checkout


cnasc:
Tracking issues believed to be on the paywall side of the user account checkout

- [x] _info requested_ need to know expected order of post messages - solidify understanding of intended checkout flow (see #4164 for one case)
- [x] wallet required component does not disappear when account is sent from app, account send confirmed with this line from debug logs: `[pO] main window <-- Account UI – "update/account" – "0xb84fB4CC7aB41a136a87131aF5dEcB92b092aaCE" – "http://localhost:3000"`
- [x] app iframe dismisses itself after key purchase, unclear if data-iframe is then able to poll transaction correctly. Can't see what's going on due to previous issue.

PR #4178, opened 2019-07-12
fix handling of eth_accounts, and correct the web3 return

cellog:
# Description

In order to fix handling of eth_accounts, this does 3 things:

1. remove the requirement that an account exist to do account polling
2. fix the shape of the object returned from our proxyProvider
3. return `[]` instead of `[null]` when there is no account.

# Issues

Refs #4165

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/blockchainHandler/account.js
+++ b/paywall/src/data-iframe/blockchainHandler/account.js
@@ -1,4 +1,3 @@
-import ensureWalletReady from './ensureWalletReady'
 import pollForChanges from './pollForChanges'
 import { POLLING_INTERVAL } from '../../constants'
 
@@ -26,10 +25,8 @@ export async function pollForAccountChange(
   web3Service,
   onAccountChange = () => {}
 ) {
-  await ensureWalletReady(walletService)
-
   pollForChanges(
-    async () => await walletService.getAccount() /* getFunc */,
+    () => walletService.getAccount() /* getFunc */,
     (before, after) => before !== after /* hasValueChanged */,
     () => 1 /* continuePolling */,
     async newAccount => {
--- a/paywall/src/data-iframe/blockchainHandler/index.js
+++ b/paywall/src/data-iframe/blockchainHandler/index.js
@@ -155,6 +155,7 @@ export async function listenForAccountNetworkChanges({
     }
   })
 
+  // TODO: investigate whether this can be safely removed
   await ensureWalletReady(walletService)
   const account = await walletService.getAccount()
   const balance = await web3Service.getAddressBalance(account)
--- a/paywall/src/data-iframe/blockchainHandler/index.js
+++ b/paywall/src/data-iframe/blockchainHandler/index.js
@@ -132,14 +132,6 @@ export async function listenForAccountNetworkChanges({
     onChange({ network: id })
   })
 
-  await ensureWalletReady(walletService)
-  const account = await walletService.getAccount()
-  const balance = await web3Service.getAddressBalance(account)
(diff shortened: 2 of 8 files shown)


paywall/src/providers/Web3ProxyProvider.js
```diff
@@ -42,13 +42,13 @@ export default class Web3ProxyProvider {
...
       if (
-        !web3Result.hasOwnProperty('error') ||
+        !web3Result.hasOwnProperty('error') &&
         !web3Result.hasOwnProperty('result')
       ) {
         return
       }
-      const { result, error } = web3Result
-      const id = result.id
+      const { result, error = null } = web3Result
```
cnasc: Can L50 and L51 be merged?

paywall/src/unlock.js/web3Proxy.ts
```diff
@@ -74,22 +74,27 @@ export function handleWeb3Call({
...
       postMessage('data', PostMessages.WEB3_RESULT, {
         id,
-        error: null,
-        result: { id, jsonrpc: '2.0', result: [proxyAccount] },
+        jsonrpc: '2.0',
+        result: {
+          id,
+          jsonrpc: '2.0',
+          // if account is null, we have no account, so return []
+          result: proxyAccount ? [proxyAccount] : [],
```
cnasc: This is the piece that fixes
> return [] instead of [null] when there is no account.
?

Just want to make sure I see what's happening where

paywall/src/data-iframe/blockchainHandler/index.js
```diff
@@ -162,4 +154,12 @@ export async function listenForAccountNetworkChanges({
       onChange({ keys: {}, transactions: {} })
     }
   })
+
+  await ensureWalletReady(walletService)
```
akeem: It looks we are calling functionality similar  to that which exists in side of `pollForAccountChange`. Would it be worthwhile for us to call `pollForAccountChange` and eliminate the potential special case of code here?

paywall/src/unlock.js/web3Proxy.ts
cellog: yes that lives here.

paywall/src/data-iframe/blockchainHandler/account.js
```diff
@@ -26,10 +25,8 @@ export async function pollForAccountChange(
   web3Service,
   onAccountChange = () => {}
 ) {
-  await ensureWalletReady(walletService)
```
cellog: this removal enables account polling even when we have no account (and is a critical change)

paywall/src/data-iframe/blockchainHandler/index.js
cellog: The code below propagates initial state to the cache. I wonder ... It is possible that this was never called because of the bug this PR fixes, and it may be removable. I'll investigate in a quick follow-up PR so we can get this rolling as-is

paywall/src/providers/Web3ProxyProvider.js
cellog: duh - they needed to be separate when I was handling things the odd way above, good catch

paywall/src/data-iframe/blockchainHandler/index.js
cellog: I think, after a quick look, that there is a very good chance this is unneeded, and I'm going to quickly try it out

cellog: some stuff breaks, going to leave this for now with a TODO to investigate

PR #4184, opened 2019-07-12
more casing fixes, ensure our account is set before we declare the wallet ready

cellog:
# Description

This merges in fixes from the user accounts branch that are needed in master as well.

1. much more permissive account detection (polling is enabled when there is no account)
2. case-sensitivity fixes (locks are indexed by their lower-cased address in the cache now)
3. case-sensitivity fixes (keys are indexed by their lower-cased lock address in the cache now)

# Checklist:


cellog:
> Looks good, is there a way that we could centralize the fetching and writing of key/locks/transactions to ensure that they are normalized through out?

Yes. I have learned a number of things from writing this code. The data iframe needs to be refactored over time to consolidate responsibilities and make it safer to change, this is a good example of one of the needed changes

cellog:
blocked, most likely by the fix in #4192

PR #4192, opened 2019-07-14
use a local transaction cache for web3ServiceHub

### LINKED ISSUE #4191: the blockchain handler transaction handler cannot use the user cache for prior values
in `web3ServiceHub.js`, the prior values of transactions are retrieved from the user cache, an async cache. This can result in race conditions where outdated values overwrite newer ones if the updates come in fast enough.

The solution is to use a simple in-memory cache local to the file.

cellog:
# Description

This removes the possibility of race conditions where prior transactions could be stale by using an in-memory cache local to the `web3ServiceHub`. This is necessary because `'transaction.updated'` events do not contain the full transaction, only the portions that have been changed.

# Issues

Fixes #4191

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/blockchainHandler/web3ServiceHub.js
+++ b/paywall/src/data-iframe/blockchainHandler/web3ServiceHub.js
@@ -1,13 +1,5 @@
 import { getAccount } from '../cacheHandler'
 
-let transactionCache = {}
-
-/**
- * for unit testing purposes
- */
-export function __clearCache() {
-  transactionCache = {}
-}
 /**
  * Listen for transaction updates, update the transaction and key that it affects
  * if the transaction is for a key purchase.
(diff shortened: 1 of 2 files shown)


paywall/src/data-iframe/blockchainHandler/web3ServiceHub.js
```diff
@@ -1,5 +1,13 @@
-import { getTransactions, getAccount } from '../cacheHandler'
+import { getAccount } from '../cacheHandler'
 
+let transactionCache = {}
+
+/**
+ * for unit testing purposes
+ */
+export function __clearCache() {
```
akeem: Generally we should not add method solely for testing purposes. Another approach here would be possibly passing the transactionCache to web3ServiceHub

cellog: great idea!

cellog:
PTAL @akeem, I like the changes much better after your suggestion

PR #4200, opened 2019-07-15
add versioning and auto-invalidation if there is no match

### LINKED ISSUE #4199: paywall cache needs versioning, so if the format changes, it is invalidated
When the cache format changes, we need to invalidate it, or bad things can happen.

cellog:
# Description

This adds versioning to the localStorage cache, and auto-invalidation if the version does not match. In this way, we will never get a bug due to data format mis-match. This PR does not address the case of someone intentionally manipulating localStorage.

Most changes are 1-line additions to unit tests to add the `clear` function to their `window.localStorage`.

# Issues

Fixes #4199 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/data-iframe/blockchainHandler/web3ServiceHub.test.js
+++ b/paywall/src/__tests__/data-iframe/blockchainHandler/web3ServiceHub.test.js
@@ -9,6 +9,7 @@ describe('web3ServiceHub', () => {
     fakeWindow = {
       storage: {},
       localStorage: {
+        clear: () => (fakeWindow.storage = {}),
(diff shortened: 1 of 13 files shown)


julien51:
This won't solve the underlying problem in a robust way.
We MUST assume that the cache can be compromised and have a recovery mechanism which will _always_ result in not using the cache should there be any failure.

julien51:
Let's think about versioning later.
For now, we need a system which ensures that if we do not have the "shape" of cache that we expect, things do not fail.
Can we remove the versioning logic from there?

cellog:
closing in favor of using in-memory cache until localStorage is ready

PR #4218, opened 2019-07-16
Create paywall-specific account components

cnasc:
# Description

This PR duplicates the existing components used for logging in and signing up in the subdirectory used for embedding in the paywall and puts them into `AccountContent`. This sets us up to modify them to match the Figma design for the checkout, which is somewhat different from what we use on full pages. There should not be any functional changes to these components, only styles.

Additionally, this PR updates that specific login component to more closely conform with the Figma design.

image

# Issues

Refs #4164 

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/components/interface/user-account/SignUp.test.tsx
+++ b/unlock-app/src/__tests__/components/interface/user-account/SignUp.test.tsx
@@ -74,7 +74,7 @@ describe('Sign Up Page', () => {
       expect(emailAddress).toEqual('julien@unlock-protocol.com')
     })
 
-    it('should yield the emailAddress from the location if there are several', () => {
+    it('should yield the first emailAddress from the location if there are several', () => {
       expect.assertions(1)
       const state = {
         router: {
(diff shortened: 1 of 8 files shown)


unlock-app/src/components/interface/user-account/SignUp.tsx
```diff
@@ -0,0 +1,233 @@
...
+      return <SignupSuccess />
+    }
+
+    const LogInLink = (callToAction: string) => (
+      <LinkButton onClick={this.handleClick}>{callToAction}</LinkButton>
+    )
+
+    if (!emailAddress) {
+      return (
+        <div>
```
akeem: Not a blocker, but we should probably make future consideration to extract the return values here so that we can edit their make up without adjusting these files

unlock-app/src/__tests__/components/interface/user-account/SignUp.test.tsx
```diff
@@ -0,0 +1,328 @@
...
+            },
+          },
+        },
+      }
+
+      const { emailAddress } = mapStateToProps(state)
+      expect(emailAddress).toEqual('julien@unlock-protocol.com')
+    })
+
+    it('should yield the emailAddress from the location if there are several', () => {
```
akeem: does this mean that we should take the first one?

PR #4234, opened 2019-07-17
Updated credit card field styles

benwerd:
# Description

This PR brings credit card fields in line with other field styles, despite being imported from Stripe. This required adding a Stripe element global class style.

# Issues

Refs #4164

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [ ] This PR only contains configuration changes (package.json, etc.)
  - [x] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [x] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [x] I have performed a self-review of my own code
- [x] If my code involves visual changes, I am adding applicable screenshots to this thread

Screen Shot 2019-07-17 at 1 46 40 PM
Screen Shot 2019-07-17 at 1 45 45 PM

PR #4240, opened 2019-07-17
Credit card field refactor on top of unlock_provider_base 

benwerd:
# Description

This supersedes #4234, and is based on `unlock_provider_base`.

Really state and country should be a drop-down. At the very least, country should be. I'll follow up on this.

# Issues

Refs #4164

# Checklist:

Screen Shot 2019-07-17 at 3 52 53 PM
Screen Shot 2019-07-17 at 3 52 12 PM

### DIFF
--- a/unlock-app/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
+++ b/unlock-app/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
@@ -181,13 +181,13 @@ Object {
       rel="stylesheet"
     />
     <section
-      class="sc-cJSrbW dLwzxs"
+      class="sc-ksYbfQ jHwjmi"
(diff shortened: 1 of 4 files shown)


cnasc:
Now that I think about it, this really is an awful lot of fields. We should look into which are really necessary. The address fields were recommended by Stripe, but maybe we can find some best practices and trim this down (in a future PR).

Make sure to hit the edit button at the top right of the PR and set the base branch to `unlock_provider_base` so we can merge this to the right spot and I'll approve (just want to make sure we don't accidentally hit master with this)

benwerd:
It’s a huge number of fields. I don’t think we actually *need* anything beyond CC number, expiry, and CVC, but let’s talk about it. Editing the PR.

PR #4251, opened 2019-07-19
fix missing user accounts login window with clear cache

### LINKED ISSUE #4181: With fresh cache, paywall does not properly detect the need for user accounts
To test:

checkout `unlock_provider_base`
clear the localStorage of all unlock-related values.
start the dev environment, with paywall, unlock-app, and locksmith running
go to `http://localhost:3001/static/adremover/integrationtesting-loggedout.html?locks=%5B%7B%22address%22%3A%220x8276A24C03B7ff9307c5bb9c0f31aa60d284375f%22%2C%22%7D%5D&paywall=http%3A%2F%2Flocalhost%3A3001`

You will see the "no account" window instead of the login window

cellog:
# Description

If locks loaded after the initial "show the account iframe" message, the message was ignored. This fixes that race condition.

# Issues

Fixes #4181 

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/web3Proxy.ts
+++ b/paywall/src/unlock.js/web3Proxy.ts
@@ -223,6 +223,9 @@ export default function web3Proxy(
           currentLocks = locks
           if (showIframeWhenReady) {
             showIframe(window, accountIframe)
+            // turn the flag off, we don't want to re-open the user accounts iframe
+            // every time the locks change
+            showIframeWhenReady = false
           }
         }
       }
--- a/paywall/src/unlock.js/web3Proxy.ts
+++ b/paywall/src/unlock.js/web3Proxy.ts
@@ -121,6 +121,7 @@ export default function web3Proxy(
   let currentLocks: Locks
   let canUseUserAccounts: boolean = !window.web3
   let hasNativeWeb3Wallet = !!window.web3
+  let showIframeWhenReady = false
 
   const useUnlockAccount = () => !hasNativeWeb3Wallet && canUseUserAccounts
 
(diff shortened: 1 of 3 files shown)


paywall/src/unlock.js/web3Proxy.ts
```diff
@@ -121,6 +121,7 @@ export default function web3Proxy(
   let currentLocks: Locks
   let canUseUserAccounts: boolean = !window.web3
   let hasNativeWeb3Wallet = !!window.web3
+  let showIframeWhenReady = false
```
cnasc: Want to make sure I understand correctly:

- `showIframeWhenReady` is only true when we don't have any locks
- when it is true (i.e., we had not previously received locks) and we receive locks, we show the account iframe

Are there any other cases here?

cellog: This is almost correct.

`showIframeReady` is *only* set to true when the `PostMessages.SHOW_ACCOUNTS_MODAL` message is sent before the main window receives locks. In this case, we know the user accounts iframe is ready to receive logins or new account creation requests, but we don't yet know if this paywall has erc20 locks on it or not.

Then, when we get the locks, if there are erc20 locks, it checks this flag, and if set to true, it shows the iframe. Come to think of it, it should then reset it. I'll add that line.

cellog: added

cnasc: Perfect, reshowing the account modal was what I had in mind when asking about other cases

PR #4259, opened 2019-07-20
treat locks as unconnected to networks in the cache

### LINKED ISSUE #4256: ensureWalletReady adds many "once" listeners unnecessarily
ensureWalletReady calls `walletService.once('ready'...)` even if it is already ready.

### LINKED ISSUE #4258: cache handling of locks is incorrect
Currently, caching of locks assumes that they change per network. But this assumption is based on the flawed assumption that locks are retrieved by the user wallet. They are not. web3Service retrieves them based on a static provider URL that is *always the same network*.

Thus, caching should treat locks exactly the same way as account/network. These values do not change when the network changes. Locks don't either, we just show an error if the user is on the wrong network

cellog:
# Description

As stated in the issue this resolves, the cache was written with an incorrect assumption: that lock contents would change if the user changed the network of their wallet. Locks are immutable in the sense that in each environment (dev, staging, prod), locks can only come from one source.

This changes locks to be an unkeyed cache item.

# Issues

Fixes #4258

# Checklist:

### DIFF
--- a/paywall/src/__tests__/data-iframe/cache/cache.test.ts
+++ b/paywall/src/__tests__/data-iframe/cache/cache.test.ts
@@ -9,6 +9,8 @@ import {
   setNetwork,
   getNetwork,
   clear,
+  setLocks,
+  getLocks,
 } from '../../../data-iframe/cache/cache'
 
 describe('cache utility', () => {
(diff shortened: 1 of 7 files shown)


cellog:
this is obsoleted by current master

PR #4263, opened 2019-07-20
wait for locks to be downloaded before deciding whether to use user accounts

### LINKED ISSUE #4262: User accounts: web3Proxy must wait for locks to be actually set
Currently, the 2 conditions that web3Proxy uses to decide when loading has finished are:

1. did we receive locks from the data iframe?
2. is the account iframe ready?

As it turns out, because the cache sends whatever it has, with an empty cache, the locks values are `{}`. This is not enough information to determine whether we have erc20 locks or not.

Instead, we need to wait until the locks are fully populated

cellog:
# Description

As the description of #4262 states, the first locks retrieved are  empty, and so this is not  enough information to be able to determine if we can use user accounts yet.

# Issues

Fixes #4262 

# Checklist:


cellog:
closing, as this is already in master!

PR #4265, opened 2019-07-20
Implement a simple localStorage cache around the mailbox

cellog:
# Description

This adds caching to the `Mailbox`. The principles are extremely simple:

- The cache consists of the entire blockchainData structure, so includes `locks`, `account`, `balance`, and `network`.
- The cache is only used while the `BlockchainHandler` is loading

Here is the state flow:

1. on startup, if localStorage is available, it checks to see if the current paywall has a cache
2. if so, it sets the default value of `Mailbox.blockchainData` to this value. Otherwise, it uses a blank default.
3. Throughout the process, whenever the blockchain sends an update, it saves it to the localStorage cache
4. if the user has the same paywall open in another tab, and purchases a key (say a different article), the BlockchainHandler will refresh the blockchain data, thereby propagating keys between tabs. The refresh *only* happens if the paywall is exactly the same (it ignores updates to other caches)

Thus, the cache is a one-way data flow, and is only used at startup to speed things up. Otherwise, it is a write-only cache.

Each paywall cache is differentiated by the locks in the paywall configuration. Thus, a paywall that uses locks `'0x123...'` and `'0x456...'` is a different cache from paywall that uses '`0x789...'`, and is also different from a paywall that only uses one of the `'0x123'` or `'0x456'` locks.

If there are any errors at all, whether reading or writing from the cache, it clears the entire localStorage cache.

This branch includes a bunch of semi-related changes that can all be split off into their own pre-PRs:

1. updating the type signature for `addEventListener` is most of the semi-related complexity
2. adding LocalStorageWindow to the `FakeWindowHelper.ts`

# Checklist:

### DIFF
--- a/paywall/src/__tests__/data-iframe/Mailbox/getBlockchainDataFromLocalStorageCache.test.ts
+++ b/paywall/src/__tests__/data-iframe/Mailbox/getBlockchainDataFromLocalStorageCache.test.ts
@@ -1,0 +1,239 @@
+import { IframePostOfficeWindow } from '../../../windowTypes'
+import {
+  ConstantsType,
+  FetchWindow,
+  SetTimeoutWindow,
+  WalletServiceType,
+  Web3ServiceType,
+  BlockchainData,
+} from '../../../data-iframe/blockchainHandler/blockChainTypes'
+import Mailbox from '../../../data-iframe/Mailbox'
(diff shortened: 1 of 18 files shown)


cellog:
closing as this is already in master!

PR #4272, opened 2019-07-23
Basic access control by account type

cnasc:
# Description

This PR introduces a simple rule-based filter on the app navigation in the header. The idea is that we have some parts of the app that are only for crypto users, some parts that are only for managed accounts, and some parts that are common to both. **This brings us a step closer to merging the feature branch to production.**

The header now determines what kind of user is on the site and then presents just the navigation buttons that are relevant to them. My understanding is that the dashboard (maybe also the transaction log?) should only be accessible to crypto users, and the account settings page should only be accessible to managed accounts, and everything else allows both.

This change also has the effect of not showing any navigation at all if there is no account or if the account is not valid (empty object, email but no public key, for example) because the app cannot meaningfully work in that state. There should probably be additional handling app-wide in that circumstance (which I think only occurs when you go directly to the dashboard but don't allow metamask to connect to the app), which we should do post-V1.

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/Header.js
+++ b/unlock-app/src/components/interface/Header.js
@@ -17,20 +17,63 @@ const navigationButtons = [
   Buttons.Telegram,
 ]
 
+const accountTypes = {
+  managed: 'account/managed',
+  crypto: 'account/crypto',
+}
+
+export const typeOfAccount = account => {
+  if (!account) {
+    // this case mostly seems to happen in component tests that include the
+    // header but haven't had an account set.
+    return 'account/undefined'
+  }
+  const { address, emailAddress } = account
+  if (address && emailAddress) {
+    return accountTypes.managed
+  } else if (address) {
+    return accountTypes.crypto
+  } else {
+    // This final catch-all state is invalid. No app navigation will display for
+    // a user who doesn't match the above conditions. A user may be in an
+    // intermediate state (not all account updates have hit redux), which may
+    // cause a brief flicker of the icons. The return value here should indicate
+    // what pieces are missing for debugging purposes.
+    return `account/address={${address}}--emailAddress={${emailAddress}}`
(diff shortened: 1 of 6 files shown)


unlock-app/src/components/interface/Header.js
```diff
@@ -17,18 +17,63 @@ const navigationButtons = [
...
+    return accountTypes.managed
+  } else if (address) {
+    return accountTypes.crypto
+  } else {
+    // This final catch-all state is invalid. No app navigation will display for
+    // a user who doesn't match the above conditions. A user may be in an
+    // intermediate state (not all account updates have hit redux), which may
+    // cause a brief flicker of the icons. The return value here should indicate
+    // what pieces are missing for debugging purposes.
+    return `account/address={${address}}--emailAddress={${emailAddress}}`
```
akeem: Not a blocker from shipping this, we may need to do some research for understand the scenarios that have users that end up here and chip away at removing them

unlock-app/src/components/interface/Header.js
```diff
@@ -17,18 +17,63 @@ const navigationButtons = [
...
+  },
+  {
+    Button: PageNavButtons.Settings,
+    page: '/settings',
+    allowedUsers: [accountTypes.managed],
+  },
+  {
+    Button: PageNavButtons.KeyChain,
+    page: '/keychain',
+    allowedUsers: [accountTypes.crypto, accountTypes.managed],
```
akeem: Really appreciate the explicit nature of allowedUsers here so we don't have to do crazy tracing to figure out how people ended up in given paths! 👍

unlock-app/src/components/interface/Header.js
cnasc: Agreed 100%

PR #4296, opened 2019-07-25
on key expiration, re-send the data to re-lock the paywall

### LINKED ISSUE #4295: data iframe does not update data when a key expires
Currently, if the user is on a page when a key expires, the data iframe does not emit new data with the key being locked, so the UI cannot respond to this event. This is different from expected behavior, and different from how the legacy paywall works

cellog:
# Description

This PR adds missing behavior when a key expires. Currently, the paywall remains unlocked until a refresh. This PR adds a simple setTimeout call that re-dispatches the blockchain data when the key is expired. Because the "locked" state is re-calculated every time the data is sent, this will trigger a state change.

# Issues

Fixes #4295 

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/blockchainHandler/BlockchainHandler.ts
+++ b/paywall/src/data-iframe/blockchainHandler/BlockchainHandler.ts
@@ -222,8 +222,10 @@ export default class BlockchainHandler {
     // a key with no expiry yet does not need an expiration trigger
     if (!expiry) return
     const now = new Date().getTime() / 1000
+    const timeToExpirationInSeconds = expiry - now
+    const timeToExpirationInMilliseconds = timeToExpirationInSeconds * 1000
 
-    if (expiry - now <= 0) {
+    if (timeToExpirationInSeconds <= 0) {
       // key is expired already
       // as this is called by the key.updated listener, it will
       // dispatch changes.  If that ever changes, we need to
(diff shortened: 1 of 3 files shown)


akeem:
@cellog Have you been able to manually test the old paywall and new paywall behavior with metamask and the unlock provider with this code in place? We will need to manually test changes here until we have sufficient testing in place to catch breaking changes

cellog:
This code does not affect the old paywall. I did not yet run it on the new one

cellog:
out

This screencast shows the data iframe sending data 1 second after key expiry, and the checkout resetting

paywall/src/data-iframe/blockchainHandler/BlockchainHandler.ts
```diff
@@ -258,6 +283,10 @@ export default class BlockchainHandler {
     this.web3Service.on('key.updated', (_: any, key: KeyResult) => {
       key.lock = normalizeLockAddress(key.lock)
       this.store.keys[key.lock] = key
+      this.sendDataWhenKeyExpires(key.expiration)
+      // note: if this next line is ever removed, it will need to be
+      // added inside sendDataWhenKeyExpires() to ensure expired
```
julien51: It looks like this is actually called inside of `sendDataWhenKeyExpires` already?

cellog: line 232 would need to have the `dispatchChangesToPostOffice`. The flow is:

1. if the key is already expired, the `setTimeout` line is not called, and the dispatch is handled by line `290`.
2. if the key is not expired, the `setTimeout` sets up the dispatch.

The note is just to ensure the dependency on the dispatch existing in `'key.updated'` is explicit, so if in a future refactor that line is removed, we don't lose the "expired" state.

PR #4302, opened 2019-07-26
remove the old data iframe

cellog:
# Description

This removes the obsolete files from `paywall/src/data-iframe`

Yes, I manually tested the paywall before sending the pull request. There is no user account-specific behavior in the data iframe to test, it works with both erc20 and ether lock purchases.

# Checklist:


cellog:
> I trust as you wrote that you fully tested this because I have no way to review this further :/

Yes! The full story is:

First, I deleted all of the .js files. Then, I was honestly terrified that this would break something, so before I submitted it, I ran the app in 4 different ways (demo with eth/erc20, the integration test logged in page and the logged out page to make sure user accounts still worked). I found 2 `.js` files that were still being used in that testing process, so added those back in. This version works!

PR #4306, opened 2019-07-26
Implement ERC-721 Enumerable extension

### LINKED ISSUE #2082: Add ERC721 enumerable extension

NickCuso:
# Description

As noted in #2082, OpenSea currently expects this... even though as you can see from the implementation below it does not add much value for our use case.

# Issues

Fixes #2082
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/test/Lock/erc721/enumerable.js
+++ b/smart-contracts/test/Lock/erc721/enumerable.js
@@ -1,0 +1,47 @@
+const deployLocks = require('../../helpers/deployLocks')
+const shouldFail = require('../../helpers/shouldFail')
+
+const unlockContract = artifacts.require('../Unlock.sol')
+const getProxy = require('../../helpers/proxy')
+
+let unlock, locks, lock
+
+contract('Lock / erc721 / approve', accounts => {
+  before(async () => {
+    unlock = await getProxy(unlockContract)
+    locks = await deployLocks(unlock, accounts[0])
+    lock = locks['FIRST']
+
+    // Buy test keys for each account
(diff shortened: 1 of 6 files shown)


smart-contracts/test/Lock/erc721/enumerable.js
```diff
@@ -0,0 +1,47 @@
...
+    const keyPrice = await lock.keyPrice()
+    for (let i = 0; i < 5; i++) {
+      await lock.purchase(accounts[i], web3.utils.padLeft(0, 40), {
+        value: keyPrice.toString(),
+        from: accounts[i],
+      })
+    }
+  })
+
+  it('tokenByIndex is a no-op', async () => {
```
nfurfaro: Can you tell me briefly what this "no-op" is all about?

NickCuso: Ah I just mean that it's expected to return the parameter it was given here, unmodified.

PR #4324, opened 2019-07-30
display errors on the checkout component

cellog:
# Description

This is a draft pull request because there are many questions the team needs to weigh in on before it can be considered on its merits alone.

Currently, the checkout component does not do anything when errors occur. This PR adds the facility to display the errors as they happen.

There are a couple of major caveats.

1) The design is some crap I whipped up, and needs the caring and expert influence of @smombartz 
2) The basic concept was also something I whipped up, and so needs the discerning opinions of @julien51 

Questions I have:

1. Should we show the error message? 99% of the time it will be "purchase failed" but there is a chance that an error might be more blockchain-y, such as when alchemy fails to retrieve locks for some reason. Some options I've considered:
 - Just show "Something went wrong"
 - Show "purchase failed" for purchase failures, and "Something went wrong" for everything else
2. Currently, because of the way useListenForPostMessage works, if the user generates multiple errors with the same message, it will only show once. Thus, if you click to purchase, get the error, dismiss it (and ignore the instruction to refresh), and click to purchase again and it fails. no error will be shown. Should it re-render every time it gets an error even if the message is the same?
3. I could also easily extract this added code out into a new `useError` hook, which would make it cleaner. Opinions on that idea @julien51?
4. This component will/could also eventually be used to display specific validation problems with the paywall configuration (like `window.unlockProtocolConfig paywall type "pyawall" is invalid` or `window.unlockProtocolConfig lock address "0x123" is invalid`). So one question is whether those errors should be displayed differently than key purchase failure? The key purchase failure errors are for end users, whereas configuration errors are for creators integrating locks on their page.

Examples of the error displayer in action, both in the new storybook story, and on the actual app (I killed ganache to make the the purchase fail):

app
storybook story

Existing storybook story to show I didn't break the existing use cases:

Screen Shot 2019-07-30 at 2 58 40 PM

# Checklist:

### DIFF
--- a/paywall/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
+++ b/paywall/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
@@ -12716,6 +12716,3438 @@ Object {
 }
 `;
 
+exports[`Storyshots Checkout page Checkout page, error received from data iframe 1`] = `
+Object {
+  "asFragment": [Function],
+  "baseElement": <body>
+    .c3 {
+  background-color: var(--lightgrey);
+  cursor: pointer;
+  border-radius: 50%;
+  height: 24px;
+  width: 24px;
+  display: inline-block;
+  padding: 0;
+  border: 0;
+  line-height: 24px;
+}
+
+.c3 > svg {
+  fill: var(--grey);
+  height: 24px;
+  width: 24px;
+}
+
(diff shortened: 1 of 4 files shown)


julien51:
I think this is interesting and something we need to consider.
Let's chat with @smombartz about the design because I am not convinced about the "modal in modal" thing.

PR #4341, opened 2019-08-02
not plugging the user account iframe unless explicitly configured

julien51:
# Description

We're making use of the new `unlockUserAccounts` configuration by only loading the accounts iframe when explicitly asked by the creator. 
This PR will make things snappier for creators who are not using our user accounts.

# Checklist:


paywall/src/components/content/NewDemoContent.tsx
```diff
@@ -24,8 +24,10 @@ export default function NewDemoContent() {
   const locks: string[] = url.searchParams.getAll('lock')
   const names: string[] = url.searchParams.getAll('name')
   const type = url.searchParams.get('type')
+  const unlockUserAccounts = url.searchParams.get('unlockUserAccounts')
```
cnasc: Will this come back as `undefined` when this param isn't in the URL? Might be worth explicitly handling that so we always have a boolean here.

paywall/src/unlock.js/setupPostOffices.ts
```diff
@@ -15,26 +15,39 @@ interface process {
 declare const process: process
 
 export function normalizeConfig(unlockConfig: any) {
+  const unlockUserAccounts =
+    unlockConfig.unlockUserAccounts === 'true' ||
+    unlockConfig.unlockUserAccounts === true ||
+    false
```
cnasc: Is the `|| false` case necessary?

paywall/src/components/content/NewDemoContent.tsx
julien51: It returns `null`!

paywall/src/unlock.js/setupPostOffices.ts
julien51: nope! good catch!

julien51:
Looks like integration tests are failing so I will have to run them locally... but that is a rabbit hole :/

paywall/src/unlock.js/setupPostOffices.ts
```diff
@@ -15,26 +15,38 @@ interface process {
...
+    unlockConfig.unlockUserAccounts === 'true' ||
+    unlockConfig.unlockUserAccounts === true
+
   if (
     !unlockConfig ||
     !unlockConfig.locks ||
     typeof unlockConfig.locks !== 'object'
   )
-    return unlockConfig
+    return {
```
cellog: I'm surprised typescript allowed this through! There is a chance that `unlockConfig` could be not an object, like `NaN` or `null` here, so this code will crash in that case. If the config is not valid, we shouldn't touch it, as it will fail validation later anyways.

paywall/src/unlock.js/startup.ts
```diff
@@ -28,23 +28,35 @@ export default function startup(window: UnlockWindow) {
...
+      window,
+      process.env.USER_IFRAME_URL + origin
+    )
+  } else {
+    userAccountsIframe = makeIframe(
+      window,
+      '' // We make an "empty" accounts iframe
+    )
+  }
+  addIframeToDocument(window, userAccountsIframe)
```
cellog: this PR breaks because it changed the order that the iframes are loaded. We should do all of the creation at the top, and call the 3 `addIframeToDocument` where they originally were in the order that is there. Maybe add a comment that the order added is very important?

Also, add a comment that if the order needs to be changed for any reason, then the file `tests/helpers/iframes.js` needs to be changed as well to reflect the new order.

paywall/src/unlock.js/startup.ts
```diff
@@ -28,23 +28,35 @@ export default function startup(window: UnlockWindow) {
...
-    window,
-    process.env.USER_IFRAME_URL + origin
-  )
-  addIframeToDocument(window, dataIframe)
-  addIframeToDocument(window, userAccountsIframe)
   addIframeToDocument(window, checkoutIframe)
 
+  // Loading the user account iframe (if applicable)
+  let userAccountsIframe
+  if (normalizedConfig.unlockUserAccounts) {
```
cellog:
```suggestion
  if (normalizedConfig && normalizedConfig.unlockUserAccounts) {
```

This case will affirmatively check for unlockUserAccounts and work even if the config is invalid

paywall/src/unlock.js/startup.ts
julien51: I am worried that this code is extremely fragile if the order of iframes actually matters :/

cellog: It is primarily because it doesn't seem to be possible to figure out which iframe is which in puppeteer (accessing `src` flips out because it's cross-domain), so the integration tests rely on the order being constant. One solution (which just occurred to me) would be to put some kind of data attribute on the iframes to differentiate them. I'll open an issue so I don't forget!

julien51:
Closing in favor of a different approach taken by @cellog to simplify and make things a lot more rebust.

PR #4342, opened 2019-08-02
skipping storybook when there are no changes

julien51:
# Description

When we introduced the `monorepo.sh` script which terminates jobs prematurely when no change applies to the job,
we had to disable chromatic changes because they would not report success until they actually ran.
In the meantime, Chromatic added a `--skip` option which reports success even though the job has not run.
We can now restore the "required" check for Chromatic!

# Checklist:

### DIFF
--- a/unlock-app/package-lock.json
+++ b/unlock-app/package-lock.json
@@ -3366,9 +3366,9 @@
       }
     },
     "acorn-walk": {
-      "version": "6.1.1",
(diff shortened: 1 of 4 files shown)


unlock-app/package-lock.json
```diff
@@ -3366,9 +3366,9 @@
       }
     },
     "acorn-walk": {
```
cnasc: Should there be a matching `package.json` change for these?

PR #4343, opened 2019-08-05
Add getState() to window.unlockProtocol

cnasc:
# Description

In response to an issue with using the paywall inside react applications, this PR adds a method that allows end-users to query the current state of the paywall. There are 3 states:

1. `'locked'`
2. `'unlocked'`
3. `undefined` -- this is the value early in the page load when we haven't yet determined the page status

This method and the variable it wraps are not used inside our app, they simply provide a way for users to know what's going on if they remount a react component after a change has already been dispatched.

While I've manually verified that the event listener works and updates the variable that `getState` points to, there is only a unit test for the initial state. Since our unit tests work entirely with mock windows, I think we'll need to have an integration test for `getState` to get total coverage.

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/setupUnlockProtocolVariable.ts
+++ b/paywall/src/unlock.js/setupUnlockProtocolVariable.ts
@@ -1,8 +1,10 @@
 import { showIframe, hideIframe } from './iframeManager'
 import {
+  EventTypes,
   IframeManagingWindow,
-  UnlockProtocolWindow,
   IframeType,
+  LockStatus,
+  UnlockProtocolWindow,
 } from '../windowTypes'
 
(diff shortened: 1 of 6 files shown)


paywall/src/unlock.js/setupUnlockProtocolVariable.ts
```diff
@@ -23,16 +32,28 @@ export default function setupUnlockProtocolVariable(
   const hideCheckoutModal = () => {
     hideIframe(window, CheckoutUIIframe)
   }
+  const getState = () => lockStatus
```
julien51: Could you add a note that lockStatus needs to be private and can only updated by the event listener?

PR #4359, opened 2019-08-06
new locksAreDeployed helper

cellog:
# Description

This creates and enables a new `locksAreDeployed` helper which is used in the integration tests to ensure that all locks are ready prior to deploying. Because the locks are deployed in parallel, I have not seen the "waiting for" text invoked, but it is useful for the future case where the contract addresses may change.

Questions to resolve:

1) it makes me uncomfortable to hard-code the addresses we are waiting for, it would be great if there were a way to pull that in from docker standup somehow, or to predict it based on the base contract address.
2) If #1 isn't possible, it would be great to get suggestions of good ways to document the dependency and how to update it

# Issues

Refs #3848 

# Checklist:

### DIFF
--- a/docker/development/deploy-unlock.js
+++ b/docker/development/deploy-unlock.js
@@ -59,7 +59,7 @@ serverIsUp(1000 /* every second */, 120 /* up to 2 minutes */)
         readOnlyProvider: providerURL,
         unlockAddress: newContractInstance.options.address,
         requiredConfirmations: 1,
-        blockTime: 3,
+        blockTime: 3000, // this is in milliseconds
       })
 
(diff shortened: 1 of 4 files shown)


julien51:
The question of the lock addresses is a good one and I am not sure I have the right answer.
Now, my understanding is that the script will "fail" (timeout) if any of the addresses changes. 
Can we just then add log messages indicating that we are waiting for locks at addresses X, Y and Z and that if the addresses changed the integration tests will fail?

cellog:
ok @julien51 this is updated. Should I split out the change to `deploy-unlock.js`?

julien51:
No need. Thanks!

Issue #4015, opened 2019-06-26
A DAO owned lock


julien51:
One thing that I strongly believe is that some locks will be used as bundles, where site X, Y and Z all put the same lock on their pages and if a user has a key to the lock they can read content across all of these sites.

Now, there are 2 ways to build these bundles: top down or bottom up. The first approach requires X, Y and Z to "agree" on how they share revenues. Not great. The second approach is a lot more interesting and should be "self serve". Here is how we could implement it:

1. A DAO is formed with the following rules: each "member" of the DAO has a number of tokens which represents their claim on the revenues from the lock.
2. A lock is created and transferred to the DAO so that the DAO owns it.
3. Someone puts the lock on their site. 
4. Once someone purchases a key to the bundle lock, the "provenance" of the purchase has to be known (user U purchases a key on site X).
5. The DAO will mint a new token and grant it to X.

The funds from the lock are "withdrawn" by the lock and any DAO member can withdraw their share of these. 

This model is completely "permission-less" : anyone can put the bundle lock on their site and, if they sell a key, they will be able to claim their share of the revenue of that lock. 

I think there are interesting things about this approach:

1. At infinity, every co-owner of the DAO will receive exactly their share of the revenue that the lock generates across the network. [The ownership of tokens on the DAO will eventually be completely aligned with the revenue "origin" of the bundle]

2. Early adopters are rewarded until they slowly get diluted. This makes sense because they are taking a greater risk than late adopter who are benefiting from network effects.

I have no idea if this is viable but I have a thousand ideas of stuff which could be added to that model (the price of the keys could be adjusted based on the new supply of tokens too, in a log way so that the "larger" the bundle, the more expensive it becomes, while still being a better deal). 

Now, coming back down to earth, in order to implement this, I think we would need to improve the locks in 2 ways:

1. Add the ability to have "listeners" : the lock owner could add contracts which are "listening" on the the lock to see methods triggered when new keys are purchased, like `onPurchase` for example.
2. Accept a new parameter on `purchaseKey` which should be used for the event representing the "location" from which the key was purchased 

Ideas? As @akeem says it would be interested to run simulations!

For context: here is a blog post about bundles http://cdixon.org/2012/07/08/how-bundling-benefits-sellers-and-buyers/

julien51:
cc @nfurfaro @NickCuso @benwerd

NickCuso:
[...]

PR #4398, opened 2019-08-09
OnKeySold hook

NickCuso:
# Description

Implements a hook for key sales which enables new features.
Anyone (person or contract) can register an implementer via ERC-1820.  The test simply logs this to confirm it was actually called.  We can follow up in the future with a more practical example of how this may be leveraged.

# Issues

Fixes #
Refs #4015

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinPurchase.sol
+++ b/smart-contracts/contracts/mixins/MixinPurchase.sol
@@ -5,6 +5,7 @@ import './MixinKeys.sol';
 import './MixinLockCore.sol';
 import 'openzeppelin-eth/contracts/math/SafeMath.sol';
 import './MixinFunds.sol';
+import './MixinEventHooks.sol';
 
 
 /**
@@ -17,7 +18,8 @@ contract MixinPurchase is
   MixinFunds,
   MixinDisableAndDestroy,
   MixinLockCore,
-  MixinKeys
+  MixinKeys,
+  MixinEventHooks
 {
   using SafeMath for uint;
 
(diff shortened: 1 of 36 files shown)


smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -80,7 +84,10 @@ contract MixinPurchase is
     );
 
     // We explicitly allow for greater amounts of ETH to allow 'donations'
+    // Security: after state changes to minimize risk of re-entrancy
+    uint pricePaid = _chargeAtLeast(netPrice);
+
     // Security: last line to minimize risk of re-entrancy
-    _chargeAtLeast(netPrice);
+    _onKeySold(_recipient, _referrer, pricePaid, _data);
```
julien51: Isn't there a re-entrancy risk here if the hook calls back `purchase`?

NickCuso: It is possible to reenter via the hook.  However that would be fine.

Reentrancy becomes problematic when some state changes for the transaction are not complete before the external call is made.  Here we were able to make it the last line of the purchase function which makes this clearly safe.  Any reentrancy would be on-par with simply calling `purchase` again in a future tx.

NickCuso:
> Do you mind checking if this does not go over the gas limit upgrade unlock?

Deploy: 6.1 mil or 91.9% of the limit 
(available from the gas report at the end of the smart contract tests in CI)

julien51:
O, I really think we need to prioritize upgrading Unlock *without* changing the ABI for PublicLock.

NickCuso:
...I thought we discussed this and were okay breaking ABI.

`purchase` is already different from the current PublicLock..

smart-contracts/contracts/mixins/MixinPurchase.sol
nfurfaro: Yup, agreed that this looks fine as it follows the "checks, effects, interactions" pattern.

Issue #4381, opened 2019-08-08
"Unresponsive script" in paywall


julien51:
When at least 2 tabs are open which both load the paywall application, there are cases where the tabs become unresponsive.
See screenshot below taken on a Mac laptop using Chrome while showing a demo to a prospective user.
This has also happened in Firefox. 

My suspicion is that the various tabs/iframes are entering in some "loop" while exchanging messages.

<img width="1324" alt="image" src="https://user-images.githubusercontent.com/17735/62707792-0299bd00-b9c0-11e9-9b78-78de07601c3e.png">

cellog:
I just tried opening 6 tabs, 1 with ouvre boite, 2 with the demo on 2 production locks I have, 1 with unlock-protocol.com, and 2 with the forbes pages. I did this in the console:

```js
window.addEventListener('message', a => console.log(a.data))
```

to watch the incoming messages. I also opened the task manager, and then let it sit for a while.

Then, I switched it to the background and came over here, and suddenly, wham:

<img width="1653" alt="Screen Shot 2019-08-10 at 10 16 06 PM" src="https://user-images.githubusercontent.com/98250/62828804-cd0df300-bbbc-11e9-8e89-98c03a2e3e4d.png">
<img width="1653" alt="Screen Shot 2019-08-10 at 10 18 05 PM" src="https://user-images.githubusercontent.com/98250/62828805-cda68980-bbbc-11e9-8930-923153a8c01c.png">

Still not sure why this is happening yet

cellog:
I think I have a lead on why this is happening

cnasc:
Reopening, just had this happen when I had the blue checkmark page (production) and the staging-with-ads page (staging) open

julien51:
Dang. @cnasc did you see this recently? i have not in a couple of weeks but please do re-open if you see it again!

PR #4409, opened 2019-08-11
disable localStorage cache, add TODOs

cellog:
# Description

This PR is a temporary fix for #4381.

It works by simply disabling the localStorage cache.

## Background of the bug's root cause

This bug is caused when 2 paywalls with the same locks are loaded. The bug occurs when this loop occurs:

paywall 1 retrieves blockchain data.
paywall 2 retrieves blockchain data.
paywall 1 stores the updated data in localStorage
paywall 2 receives a notification "hey, localStorage changed for your locks"
paywall 2 requests updated blockchain data download
paywall 2 retrieves blockchain data
paywall 2 stores the updated data in localStorage
paywall 1 receives a notification "hey, localStorage changed for your locks"
paywall 1 requests updated blockchain data download
paywall 1 stores the updated data in localStorage
...etc.

Under some conditions, the loop does not occur, but in the right combo, it spawns a race condition of the localStorage notifications looping and causing chain retrievals in multiple async execution threads.

## Long-term solution

Over the long term, the solution is to be granular. We should validate the incoming localStorage value, and then check to see if any keys have been purchased, and only retrieve new data if a key has become valid/confirming/pending/submitted.

I added a TODO in this PR

# Issues

Refs #4381

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/Mailbox.ts
+++ b/paywall/src/data-iframe/Mailbox.ts
@@ -427,7 +427,7 @@ export default class Mailbox {
    * the whole cache is cleared out of an abundance of caution.
    */
   invalidateLocalStorageCache() {
-    if (!this.localStorageAvailable) return
+    if (!this.localStorageAvailable || !this.useLocalStorageCache) return
     if (!this.configuration) return
     try {
       this.window.localStorage.removeItem(this.getCacheKey())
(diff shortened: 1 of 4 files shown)


paywall/src/data-iframe/Mailbox.ts
```diff
@@ -101,6 +101,8 @@ export default class Mailbox {
    * We can use this to determine whether the user has purchased a key in another tab
    */
   setupStorageListener() {
+    if (!this.useLocalStorageCache) return
+    // TODO: check to see if there are any changes to the keys and don't retrieve if not
```
julien51: Do you mind linking to the issue/PR here in the comment so we avoid reproducing this?

Issue #4385, opened 2019-08-08
refactor unlock-js to be more SOLID


cellog:
This is an epic to track the PRs for refactoring unlock-js

julien51:
Is this unlock-js or unlock.min.js ?

PR #4419, opened 2019-08-12
new Checkout UI Handler for the refactored unlock.min.js

cellog:
# Description

This component data exchange between the checkout UI and the data iframe. Key purchase is the only component that is not here, and it is handled in the upcoming `Wallet` component.

# Issues

Refs #4385

# Checklist:

### DIFF
--- a/paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
+++ b/paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
@@ -1,0 +1,193 @@
+import FakeWindow from '../test-helpers/fakeWindowHelpers'
+import CheckoutUIHandler from '../../unlock.js/CheckoutUIHandler'
+import { PaywallConfig, Locks } from '../../unlockTypes'
+import IframeHandler from '../../unlock.js/IframeHandler'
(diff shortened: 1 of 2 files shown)


paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+declare const process: {
+  env: {
+    PAYWALL_URL: 'http://paywall'
+    USER_IFRAME_URL: 'http://app/account'
+  }
+}
+
+describe('CheckoutUIHandler', () => {
+  process.env.PAYWALL_URL = 'http://paywall'
```
julien51: I have seen this pattern a couple times and I think this is not the right way of doing this and will come back and bite us. Let's add a note to never use (and especially overwrite) `process.env` directly in the application code.
We should map the values of `process.env` inside a config file and then import ther confg file where we need it.

cellog: sure. The data iframe does this, and it is a pretty trivial change with a big payoff. I'll do that as a follow-up PR and modify the ones that are already out there too

PR #4421, opened 2019-08-12
new Wallet handler for refactored unlock.min.js

cellog:
# Description

This class handles everything involving communication with a wallet.

1. determining whether to use the crypto wallet or proxy user accounts wallet
2. routing key purchase to the crypto wallet or to credit card purchase
3. MUCH clearer setup of user accounts. Every method is SOLID here.

note: it adds some helpers to the fakeWindowHelper to wait for messages posted to an iframe to appear, and the same event dispatching as the `MainWindowHandler` PR.

# Issues

Refs #4385 

# Checklist:


paywall/src/unlock.js/Wallet.ts
```diff
@@ -0,0 +1,315 @@
+import { Web3Window, web3MethodCall } from '../windowTypes'
+import { PostMessages } from '../messageTypes'
+import IframeHandler from './IframeHandler'
+import { PaywallConfig } from '../unlockTypes'
+import StartupConstants from './startupTypes'
+
+declare const process: {
```
julien51: Since you do that in a lot of places can we just add this to  a types file? either `unlockTypes` or `startupTypes`?

paywall/src/unlock.js/Wallet.ts
```diff
@@ -0,0 +1,315 @@
...
+    constants: StartupConstants
+  ) {
+    this.window = window
+    this.iframes = iframes
+    this.config = config
+    this.userAccountNetwork = constants.network
+
+    // do we have a web3 wallet?
+    this.hasWallet = !!(this.window.web3 && this.window.web3.currentProvider)
+    this.isMetamask = !!(
```
julien51: `this.hasWallet && window.web3.currentProvider.isMetamask` ?

paywall/src/unlock.js/Wallet.ts
```diff
@@ -0,0 +1,315 @@
...
+    })
+    this.iframes.accounts.on(PostMessages.UPDATE_NETWORK, network => {
+      this.userAccountNetwork = network
+    })
+
+    // when a purchase is in progress, tell the data iframe to retrieve the transaction
+    this.iframes.accounts.on(PostMessages.INITIATED_TRANSACTION, () => {
+      this.iframes.data.postMessage(
+        PostMessages.INITIATED_TRANSACTION,
+        undefined
```
julien51: Do you mind putting a comment about the `undedined`? like what should it be?

paywall/src/unlock.js/Wallet.ts
```diff
@@ -0,0 +1,315 @@
...
+      )
+    })
+
+    // then create the iframe and ready its post office
+    this.iframes.accounts.createIframe()
+  }
+
+  setupProxyWallet() {
+    if (this.useUserAccounts && !this.hasWallet) {
```
julien51: Isn't that already checked when calling `setupProxyWallet` above?

julien51: Maybe we can get rid of that function and move all of its logic inside of the function calling it because it seems to be checking for the same things?

paywall/src/unlock.js/Wallet.ts
```diff
@@ -0,0 +1,315 @@
...
+        })
+        return
+      }
+      try {
+        // first, enable the wallet if necessary
+        await this.enableCryptoWallet()
+      } catch (e) {
+        // user declined to enable the crypto wallet
+        // they still have a wallet, but we need to re-enable it to use it
+        this.hasWeb3 = true
```
julien51: Move that line above the `try/catch` so that you do not have to duplicate it?

paywall/src/unlock.js/Wallet.ts
```diff
@@ -0,0 +1,315 @@
...
+      }
+
+      // the payload is validated inside DataIframeMessageEmitter
+      const { method, params, id }: web3MethodCall = payload
+
+      // we use call to bind the web3 call to the current provider
+      send &&
+        send.call(
+          this.window.web3 && this.window.web3.currentProvider,
+          {
```
julien51: I think I completely lost track of why we have to use RPC logic in there. This seems fragile =/ Do you mind providing context?

paywall/src/unlock.js/Wallet.ts
```diff
@@ -0,0 +1,315 @@
...
+      this.window.web3 &&
+      this.window.web3.currentProvider &&
+      (this.window.web3.currentProvider.sendAsync ||
+        this.window.web3.currentProvider.send)
+
+    this.iframes.data.on(PostMessages.WEB3, payload => {
+      // handler for the actual web3 calls
+      if (!this.hasWeb3) {
+        // error - no crypto wallet
+        this.iframes.data.postMessage(PostMessages.WEB3_RESULT, {
```
julien51: I think I completely lost track of why we have to use RPC logic in there. This seems fragile =/ Do you mind providing context?

paywall/src/unlock.js/Wallet.ts
cellog: same story: I fixed it earlier, and forgot to push my commit.

paywall/src/unlock.js/Wallet.ts
cellog: ah, I remember why I did it this way - typescript is not smart enough to recognize that `this.hasWallet` implies the existence of both `web3` and `web3.currentProvider`

I need to make a window type that has web3 defined to solve some of these little issues

paywall/src/unlock.js/Wallet.ts
cellog: this is a limitation of the typing. The postMessage does not actually need the parameter, but the current version of types does not allow the 2nd parameter to be missing. I have a branch I made a month ago that fixes this so that we don't have to pass the `undefined`, but I considered it lower priority than actually making it work. Once all of this stuff is merged, I planned to submit that branch as a PR, so the code is more natural.

paywall/src/unlock.js/Wallet.ts
cellog: ok, I moved setupProxyWallet inside `init()`

paywall/src/unlock.js/Wallet.ts
cellog: good call. done.

paywall/src/unlock.js/Wallet.ts
cellog: this is the main window side of `Web3ProxyProvider`. Another way of imagining it is that this replaces `src/unlock.js/web3Proxy.ts`. Unlike `web3Proxy.ts`, which intermingled user accounts and crypto wallet handling, this code separates the handling of proxying to the crpyto wallet from proxying to the user accounts iframe.

paywall/src/unlockTypes.ts
```diff
@@ -76,6 +76,7 @@ export type PaywallAppKind = 'adblock' | 'paywall'
 export interface PaywallConfig {
   icon?: string
   type?: PaywallAppKind
+  unlockUserAccounts?: true | 'true' | false
```
julien51: and 'false', too ;)

paywall/src/unlock.js/Wallet.ts
julien51: Sorry for the confusuin here. I was hoping you would add  a comment indicated what that value should be.
`undefined /* thething */` ;)
Thanks!

PR #4424, opened 2019-08-12
adding the persistentCheckout configuration for the checkout modal

julien51:
# Description

By default it is set to "false" which means that no behavior is changed.
If set to true, then the close button will not be visible.

# Checklist:


cellog:
I think we should have a larger discussion about the paywall config before we start adding new options. We might (for example) want to make a sub-section for "checkout" and for "accounts" rather than a bunch of top-level config options. Could we talk about that tomorrow perhaps?

julien51:
Let's do this one please. It is important that we move forward. We should indeed have a conversation but it can wait for things to be released ;)

julien51:
Note: this will remain an "undocumented" feature so no risk of having anyone be upset that we change that API...

cellog:
ok thanks, I'm convinced

paywall/src/utils/validators.js
```diff
@@ -119,6 +120,16 @@ export const isValidPaywallConfig = config => {
     return false
   }
 
+  if (
+    !(
+      typeof config.persistentCheckout === 'boolean' ||
```
cellog: could we name it `showCheckoutWhenLocked` instead so it describes what behavior it will elicit?

paywall/src/__tests__/utils/validators.test.js
```diff
@@ -666,6 +666,85 @@ describe('Form field validators', () => {
           ).toBe(false)
         })
       })
+
+      describe('persistentCheckout', () => {
+        it("should accept or 'true' as value", () => {
```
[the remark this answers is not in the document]
julien51: yes, I am worried about someone passing 'false'. The config is a JSON object which could be "generated" from a form (see donations...) and in that case, the type could be messed up.

I think it is important that we stay liberal on the input side... rather than strict ;)

paywall/src/utils/validators.js
julien51: hum. I have no idea what `showCheckoutWhenLocked` means? 
It is about not letting the user close the modal when set to `true`.

cellog: ok. so what happens when a key is purchased? can the user close the modal then?

julien51: Exactly like in the "paywall" type!

cellog: `persistentCheckout` implies that it is always there. I'm suggesting a name that captures this behavior: The checkout can't be closed when the content is locked. How about `whenLockedCheckoutPersists`

Issue #4378, opened 2019-08-08
Accounts are loading on paywall app when they should not


julien51:
I was doing a demo to one of our prospective customers who is using Unlock on their website. My demo was done with Chrome on Android (and no active wallet). 
I was expecting to see the modal asking me to download a crypto wallet (Coinbase Wallet, Metamask or Opera), but instead the user accounts modal loaded.

This should not happen as this publisher is using Eth locks. See screenshot.

![image](https://user-images.githubusercontent.com/17735/62707187-f2cda900-b9be-11e9-92f2-008002c21da4.png)

cnasc:
Looks like this was resolved in #4432, verified by checking the DOM and I only see 2 iframes (no account iframe) when the config does not have the  `unlockUserAccounts` property.

Issue #4250, opened 2019-07-19
Improve handling of account iframe hiding


cnasc:
Right now we provide a close button on the account iframe to dismiss the iframe, in the case that a user does not want to complete a purchase. This is fine on a manually triggered checkout (such as paying to remove ads), but it's a problem on hard paywalls. We'll need a way to re-show the iframe after that, or to signal that the close button should not appear. (Part of the task here is going to be to specify exactly what should happen)

Not a blocker for V1, but will need to be sorted out sooner rather than later.

cellog:
Probably the simplest way is to send the configuration to the accounts iframe also, then you can check the paywall type, and if it's paywall, disable the close button

PR #4432, opened 2019-08-13
hook up the refactored unlock.min.js

### LINKED ISSUE #4385: refactor unlock-js to be more SOLID

cellog:
# Description

This is the final piece of the puzzle in the new unlock.min.js

Once committed, it fixes several outstanding bugs.

The file is straightforward, it just instantiates the `IframeHandler`, and uses it to instantiate the pieces (`Wallet`, `MainWindowHandler`, `CheckoutUIHandler`) that make up the unlock.min.js script.

The test takes a birds-eye view. The goal is simply to test whether each of the components are created, and so it tests a single expected behavior of that class. The expectation is that the tests for the separate pieces will verify that each piece implements all of its functionality (and they do, currently). It's a sort of proof-of-correctness-by-induction.

A specific example: when the `CheckoutUIHandler` is instantiated, and its `init()` method is called, when we receive a `READY` message from the checkout iframe, the `CheckoutUIHandler` responds by sending the paywall config with a `CONFIG` message. Thus, to test the correct instantiation of the `CheckoutUIHandler`, all we need to do is test to see if the `CONFIG` message is sent in response to `READY`.

With this pattern, it becomes possible to refactor and the test will still pass. More importantly, it completely eliminates the hassle and also the danger of using jest mocks, which will continue to work even if the file being mocked completely changes.

In any case, since this is a new idea, if it doesn't feel right, the test can be rewritten to use mocks.

screenshots showing key purchases working for ETH lock, ERC20 lock with metamask, and a purchase in safari using user accounts, and showing that user accounts don't pop up without being explicit (again in safari, no wallet present).

NOTE: to do the purchase with user accounts, I had to transfer some DEV tokens to the default account from my metamask account, otherwise it is just the dev standup

out
out
out
out
out

# Issues

Fixes #4385 #4250 #4378

# Checklist:


paywall/src/unlock.js/index.ts
```diff
@@ -1,16 +1,58 @@
...
+declare const process: {
+  env: {
+    UNLOCK_ENV: 'dev' | 'test' | 'staging' | 'prod'
+    DEBUG: any
+    PAYWALL_URL: string
+    USER_IFRAME_URL: string
+  }
+}
+
+const constants: { [key: string]: StartupConstants } = {
```
julien51: I am a bit worried by this, but I guess we can move it to its own file and only use the ones we need there?

paywall/src/unlock.js/startup.ts
```diff
@@ -1,38 +1,81 @@
...
+import CheckoutUIHandler from './CheckoutUIHandler'
+import StartupConstants from './startupTypes'
 
-export default function startup(window: UnlockWindow) {
-  // Get the config
-  const normalizedConfig = normalizeConfig(window.unlockProtocolConfig)
+/**
+ * convert all of the lock addresses to lower-case so they are normalized across the app
+ */
+export function normalizeConfig(unlockConfig: any) {
```
julien51: I think I have seen that code somewhere else? If that's the case do you mind adding a comment `TODO: DRY-up with ...`?

paywall/src/unlock.js/index.ts
cellog: these are the shared constants that are needed, and the idea was to construct them in `index.ts`, similarly to how the data iframe does things. I don't think it matters much where they live, so a separate file would work fine

paywall/src/unlock.js/startup.ts
```diff
@@ -1,38 +1,81 @@
...
+ */
+export default function startup(
+  window: UnlockWindowNoProtocolYet,
+  constants: StartupConstants
+) {
+  // normalize all of the lock addresses
+  const config = normalizeConfig(window.unlockProtocolConfig)
+  // this next line ensures that the minimally valid configuration is passed to Wallet
+  // TODO: provide some kind of developer mode which lazy-loads more extensive validation
+  if (!config) {
```
julien51: 👍

paywall/src/unlock.js/startup.ts
cellog: it's used on line 42. Where you had seen it before was in `setupPostOffices.ts`, which is going to be deleted in the PR following this one as obsolete, so it's a straight move-for-refactor

Issue #4383, opened 2019-08-08
Show a "check your wallet overlay" in paywall


julien51:
When purchasing a key or waiting for the user to enable, we should show a "check your crypto wallet" overlay to the UI as Metamask sometimes only shows a very small icon change rather than open a full pop up.

cellog:
dupe of #3093?

julien51:
Overlaps for sure but this is specifically for the paywall, while that other issue is for the "enable".

cnasc:
Looking into this right now for purchases. It's easy enough to launch the overlay when the purchase button is clicked. I think we have to handle 2 cases for dismissing the overlay:

1. new transaction - the presence of a new transaction implies that we don't need the overlay anymore.
2. `Error: "purchase failed"` - this will happen when the user rejects the transaction in MetaMask, so we should also dismiss the modal in that case.

We can parse error postmessages for the second case. In the first case, I think we need to introduce a new postmessage to tap into the `transaction.new` event from `WalletService`.

cellog:
Any time a submitted transaction is received, dismiss the modal. No need for a new postmessage

PR #4455, opened 2019-08-19
Show wallet check message when purchasing on paywall

cnasc:
# Description

This PR is a partial fix to #4383 

image

It behaves as follows:

- When the purchase button is clicked, the checkout UI swaps over to the wallet check overlay ("Please check your browser wallet...") to make sure the user authorizes the transaction
- The overlay swaps back to the checkout UI in the following cases:
  - User clicks the dismiss button
  - User rejects the transaction in their browser wallet
  - User approves the transaction in their browser wallet

Now including a test, with more to come

# Issues

Refs #4383 

# Checklist:

### DIFF
--- a/paywall/src/components/content/CheckoutContent.tsx
+++ b/paywall/src/components/content/CheckoutContent.tsx
@@ -219,7 +219,7 @@ export default function CheckoutContent() {
     </CheckoutWrapper>
   )
 
-  if (showWalletCheckOverlay) {
+  if (showWalletCheckOverlay && !purchasingLocks.length) {
     return (
       <Greyout>
         <MessageBox>
--- a/paywall/src/components/content/CheckoutContent.tsx
+++ b/paywall/src/components/content/CheckoutContent.tsx
@@ -1,5 +1,6 @@
 import React, { Fragment, useState, useEffect, useCallback } from 'react'
 import Head from 'next/head'
+import styled from 'styled-components'
 
 import { pageTitle, ETHEREUM_NETWORKS_NAMES } from '../../constants'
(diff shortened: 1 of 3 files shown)


paywall/src/components/content/CheckoutContent.tsx
```diff
@@ -197,9 +219,49 @@ export default function CheckoutContent() {
     </CheckoutWrapper>
   )
 
+  if (showWalletCheckOverlay && !purchasingLocks.length) {
```
julien51: Would you mind clarifying the check for `purchasingLocks.length` ?

paywall/src/components/content/CheckoutContent.tsx
```diff
@@ -84,6 +92,20 @@ export default function CheckoutContent() {
     defaultValue: false,
     getValue: (val: any) => !!val,
   })
+
+  // This listener is used only for the side effect of closing the overlay when a purchase is rejected.
+  useListenForPostMessage({
+    type: POST_MESSAGE_ERROR,
+    defaultValue: undefined,
+    getValue: (payload: any) => {
+      if (payload === 'purchase failed') {
```
julien51: As discussed, we may not even need to check the payload and assume that, if an error happened, it must be because the user did something on their wallet?

cnasc:
@julien51 updated the comment and removed the unneeded condition PTAL

Currently I think all testing of the overlay is done in integration tests, so I think we should follow up this PR with one there if possible (we don't have metamask in integration tests, so we'll have to see how we can make this work)

julien51:
Would you be able to write tests and a quick story for this one please?

PR #4471, opened 2019-08-21
Actually remove the walletCheck overlay when a purchase happens

cnasc:
# Description

This PR updates CheckoutContent to set the state variable that controls whether the wallet overlay is visible to fix a case where the wallet overlay would pop up again after a purchase finished confirming, which would happen if you kept the checkout window open and watch all the confirmations.

I'm unsure what the best way to test this is, without a large effort in mocking the various components and simulating a key purchase.

# Checklist:


cnasc:
Updated code to remove the piece in the `useEffect` hook, it was unnecessary

PR #4472, opened 2019-08-21
Adds the ability to persist Lock metadata

akeem:
# Description

In effort to support Lock creators providing their own default Lock token metadata, we are providing an end point to create/update existing metadata. 

A follow up code change will follow to change the reading portion to read from the stored data as well. 

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/models/index.ts
```diff
@@ -7,6 +7,7 @@ import { Lock } from './lock'
 import { Transaction } from './transaction'
 import { AuthorizedLock } from './authorizedLock'
 import { EventLink } from './eventLink'
+import { LockMetadata } from './lockmetadata'
```
julien51: Let's keep naming consistent please ;)

```suggestion
import { LockMetadata } from './lockMetadata'
```

locksmith/__tests__/controllers/metadataController.test.ts
```diff
@@ -42,4 +79,32 @@ describe('Metadata Controller', () => {
       )
     })
   })
+
+  describe('updateDefaults', () => {
```
julien51: Can we add tests to make sure that the signature is checked and that updating fails if the signature does not match?

akeem:
@julien51 would you mind giving a final pass?

Issue #4377, opened 2019-08-08
isValidKey() returns false if owner is null


cellog:
in `paywall/src/utils/validators.js`, `isValidKey()` returns false if the owner is null, but this is a valid situation for a logged out user

julien51:
Hum. Not sure about that? Would you mind explaining the rational?

cellog:
when we are logged out, the key is a fake key, and the owner is `null` because we have no user. This is still a valid lock/key combo

julien51:
Oh, I think I am confused by what you meant. Sorry. I thought you meant the key is valid... but it is for the `validator` not to decide whether the key is expired or not ;)

PR #4474, opened 2019-08-21
Fix toggling between locked/unlocked states

cnasc:
# Description

There were several cases where the paywall would signal the wrong state, and then immediately correct itself. This caused annoying flickers between the locked and unlocked states of applications listening to the `unlockProtocol` events.

If the paywall hadn't yet received all necessary data from the chain, it would signal that it was locked. The symptom here was that it would always start off in the locked state, and then if you had a key it would move into the unlocked state. This was fixed by preventing the locked and unlocked messages from being sent until the paywall has a non-empty `locks` object.

When purchasing a key, when the purchase transitions from "Submitted" (purchase sent but not mined) to "Mined", the key would briefly be considered invalid because its expiration was 0. The correct expiration is pulled in shortly afterward. This case was fixed by special-casing keys with an `expiration` of 0 that have transactions associated with them.

# Checklist:


paywall/src/unlock.js/MainWindowHandler.ts
```diff
@@ -61,16 +61,20 @@ export default class MainWindowHandler {
     // dispatching "unlockProtocol" on the main window
     // and
     this.iframes.data.on(PostMessages.LOCKED, () => {
-      this.dispatchEvent('locked')
-      this.setCachedLockedState(true)
-      // Update the user-facing status with locked/unlocked updates
-      this.lockStatus = 'locked'
+      if (this.lockStatus !== 'locked') {
```
akeem: How would you feel about extracting this method and the other one below? It isnt a block from this shipping but should help us to start wrangling this

paywall/src/data-iframe/blockchainHandler/keyStatus.js
```diff
@@ -15,13 +15,20 @@ export function isValidKey(key) {
...
   if (!key.transactions || !key.transactions.length) {
     return valid ? 'valid' : 'none'
   }
   // if we reach this line, key.transactions has at least 1 transaction in it
   // the key transactions must be generated by linkTransactionsToKeys in order
   // to be properly sorted
+
+  // For a brief period, a transaction that moves from "submitted" to
+  // "confirming" will have an expiration of 0. We establish a special
+  // case for that here to avoid a bried flicker of the "locked"
```
cnasc:
```suggestion
  // case for that here to avoid a brief flicker of the "locked"
```

paywall/src/data-iframe/blockchainHandler/keyStatus.js
```diff
@@ -15,13 +15,20 @@ export function isValidKey(key) {
...
+  // paywall state.
+  valid = valid || key.expiration === 0
```
julien51: Is this related to https://github.com/unlock-protocol/unlock/issues/4377 ?

julien51: This is strange, because I believe when the user does not have a key unlockjs will actually return an expiration of 0.

paywall/src/__tests__/data-iframe/blockchainHandler/keyStatus.test.js
```diff
@@ -124,6 +124,17 @@ describe('key status manipulation', () => {
         )
       ).toBe('expired')
     })
+
+    it('should recognize a key with expiration of 0 but with transactions as valid', () => {
```
julien51: I guess that is optimistic unlocking? If we have a pending transaction we assume the key is valid?

paywall/src/data-iframe/Mailbox.ts
```diff
@@ -314,9 +318,11 @@ export default class Mailbox {
     this.postMessage(PostMessages.UPDATE_NETWORK, dataToSend.network)
     this.postMessage(PostMessages.UPDATE_LOCKS, dataToSend.locks)
     const unlockedLocks = this.getUnlockedLockAddresses()
+
     if (unlockedLocks.length) {
       this.postMessage(PostMessages.UNLOCKED, unlockedLocks)
-    } else {
+    } else if (Object.keys(dataToSend.locks).length) {
+      // Only send LOCKED if we actually have locks
```
julien51: But is that enough? Could there be a case when we have a lock but have not retrieved the corresponding key yet?

paywall/src/data-iframe/blockchainHandler/keyStatus.js
cnasc: No, this is a different set of key validation. The one from that issue is located here: https://github.com/unlock-protocol/unlock/blob/949bddfb6fed83064c74d3af16a5396e267a3aaa/paywall/src/utils/validators.js#L165

I agree that it's strange, and in fact that is the behavior of `web3Service`. What happens here is that we have a transaction associated with the key (which is not something that comes from `web3Service`). We haven't actually gotten the information from `web3Service` yet in this case, but we also don't want to treat a key as expired.

paywall/src/__tests__/data-iframe/blockchainHandler/keyStatus.test.js
cnasc: That may be the case

paywall/src/data-iframe/Mailbox.ts
cnasc: That case is certainly possible. I think as it currently stands, we don't have a way to know whether we have actually gotten all the keys. This change fixes the case that I was able to reproduce, and I think anything more thorough will require bigger changes to the design.

PR #4482, opened 2019-08-21
changed the prompt as the wallet can be checked also to be enabled.

julien51:
# Description

All in the title!

# Checklist:

### DIFF
--- a/paywall/src/components/content/CheckoutContent.tsx
+++ b/paywall/src/components/content/CheckoutContent.tsx
@@ -230,7 +230,7 @@ export default function CheckoutContent() {
     return (
       <Greyout>
         <MessageBox>
-          <p>Please check your browser wallet to complete the transaction.</p>
+          <p>Please check your browser wallet.</p>
           <Dismiss onClick={() => setShowWalletCheckOverlay(false)}>
             Dismiss
           </Dismiss>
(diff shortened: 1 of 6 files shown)


unlock-app/src/components/interface/modal-templates/WalletCheck.tsx
```diff
@@ -5,7 +5,7 @@ import { Dispatch } from '../../../unlockTypes' // eslint-disable-line
 
 const WalletCheck = ({ dispatch }: { dispatch: Dispatch }) => (
   <MessageBox>
-    <p>Please check your browser wallet to complete the transaction.</p>
+    <p>Please check your browser wallet.</p>
```
cnasc: Do you think it would be worth making this component take a property including the reason to check the wallet? That would let us have specific messages for transactions, signatures, and enabling access.

paywall/src/components/content/CheckoutContent.tsx
```diff
@@ -230,7 +230,7 @@ export default function CheckoutContent() {
     return (
       <Greyout>
         <MessageBox>
-          <p>Please check your browser wallet to complete the transaction.</p>
+          <p>Please check your browser wallet.</p>
```
cnasc: This particular use of the message only ever happens during a transaction, though I would be in favor of pulling this out and using the other component (which I forgot we moved into the paywall) with a parameter as mentioned in my other comment.

unlock-app/src/components/interface/modal-templates/WalletCheck.tsx
julien51: Maybe for a later PR when the need arises?

PR #4485, opened 2019-08-22
Remove legacy demo page from paywall

### LINKED ISSUE #4436: Deprecate the `demo` in favod or `newdemo` inside of paywall
Our paywall currently has 2 demo pages. We need to replace `demo` with `newdemo`.

cnasc:
# Description

This PR strips out the legacy demo page from the paywall. It also replaces references to the demo page in strings and comments.

There does seem to be some code, especially related to parsing the window location, that can also be removed since it was designed to specifically support the legacy demo. I will open an issue pointing that out.

We also may want to change the story for the paywall, since it is based on the same static demo component as the demo page.

# Issues

Fixes #4436

# Checklist:


paywall/next.config.js
```diff
@@ -95,7 +95,6 @@ module.exports = withTypescript(
 
       return {
         '/': { page: '/home' },
-        '/demo': { page: '/demo' },
```
julien51: Let's make sure a later PR renames `newdemo` in `demo` ;)

paywall/src/stories/interface/StaticDemo.stories.js
```diff
@@ -31,7 +31,7 @@ const store = createUnlockStore({
   },
   router: {
     location: {
-      pathname: '/demo/' + myLock.address,
+      pathname: '/newdemo?lock=' + myLock.address,
```
julien51: Ideally I'd love to use the same kind of pattern that we use for the donation page here with the config encoded as a query string? or maybe just turn the json into a valid query string?

PR #4488, opened 2019-08-22
using a hook to retrieve Eth price

### LINKED ISSUE #4330: implement ether price polling as a hook
The redux middleware for polling for price should be re-implemented as a hook for the new checkout UI. This will remove our last redux dependency.

julien51:
# Description

This is the last remaining thing which uses the redux store... which means we can get rid of it after that
We are not testing that this is handling updates when the hook useEffect triggers...
# Issues

Fixes #4330

# Checklist:


julien51:
So, snapshots are failing because they render with `---` instead of the price. It makes sense since the price retrieving is async. I'll fix them and merge.

PR #4492, opened 2019-08-22
keyStore experiment

cnasc:
# Description

This draft PR is an experiment around a way of ensuring we don't emit the `locked` or `unlocked` messages from the paywall until we have all the necessary information.

https://github.com/unlock-protocol/unlock/blob/df9b8bb745d78aa1785b6a24cceb86ccf77fe5a4/paywall/src/data-iframe/blockchainHandler/BlockchainHandler.ts#L283

Currently we maintain the list of keys in a bare object that the BlockchainHandler operates on. This PR introduces a `KeyStore` object that will be the source of truth on key validity. The `KeyStore` is instantiated with a list of the lock addresses included in the paywall config, and will only emit an event with the keys when it has received a key for each lock.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/blockchainHandler/KeyStore.ts
+++ b/paywall/src/data-iframe/blockchainHandler/KeyStore.ts
@@ -1,0 +1,38 @@
+import { EventEmitter } from 'events'
+import { KeyResult } from './blockChainTypes'
+
+/**
+ * KeyStore holds the keys corresponding to the locks on the
+ * page. When it has been filled up (i.e., `addKey()` has been called
+ * `numLocks` times), it emits an event containing all the currently
+ * valid keys.
+ */
+export default class KeyStore extends EventEmitter {
+  keys: { [lockAddress: string]: KeyResult | null }
+
+  constructor(lockAddresses: string[]) {
+    super()
+    this.keys = {}
+    lockAddresses.forEach(lockAddress => {
+      this.keys[lockAddress] = null
+    })
+  }
+
+  hasAllKeys(): boolean {
+    return Object.values(this.keys).every(value => value !== null)
+  }
+
+  addKey(key: KeyResult) {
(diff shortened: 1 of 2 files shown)


julien51:
That's an interesting approach but I wonder how if we could not do better.
My main worry here is that we have to fetch all of the locks first, and then the keys, when in practice, we should fetch the keys first. The lock info is only useful when there is no key, at least for now?

cnasc:
> That's an interesting approach but I wonder how if we could not do better.
> My main worry here is that we have to fetch all of the locks first, and then the keys, when in practice, we should fetch the keys first. The lock info is only useful when there is no key, at least for now?

We don't have to fetch locks in this case. We already have the lock addresses from the config file so this happens independently of fetching the locks.

paywall/src/data-iframe/blockchainHandler/KeyStore.ts
```diff
@@ -0,0 +1,38 @@
...
+ * `numLocks` times), it emits an event containing all the currently
+ * valid keys.
+ */
+export default class KeyStore extends EventEmitter {
+  keys: { [lockAddress: string]: KeyResult | null }
+
+  constructor(lockAddresses: string[]) {
+    super()
+    this.keys = {}
+    lockAddresses.forEach(lockAddress => {
```
julien51: Oh that's rigth! strings! Sorry I overlooked that!

julien51:
You are right! it looks like it could work!

paywall/src/data-iframe/blockchainHandler/KeyStore.ts
cnasc: No worries at all!

cnasc:
Closing in favor of a different approach

PR #4500, opened 2019-08-23
Refactoring metadata components for future extension

akeem:
# Description

Mostly a restructuring to support future changes and increase readability

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/operations/metadataOperations.ts
+++ b/locksmith/src/operations/metadataOperations.ts
@@ -1,0 +1,55 @@
+import { KeyMetadata } from '../models/keyMetadata'
+import { LockMetadata } from '../models/lockMetadata'
+import Metadata from '../../config/metadata'
+
+export const updateKeyMetadata = async (data: any) => {
+  try {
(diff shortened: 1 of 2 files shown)


locksmith/src/operations/metadataOperations.ts
```diff
@@ -0,0 +1,55 @@
...
+  try {
+    return !!(await KeyMetadata.upsert(data, { returning: true }))
+  } catch (e) {
+    return false
+  }
+}
+
+export const updateDefaultLockMetadata = async (data: any) => {
+  try {
+    return !!(await LockMetadata.upsert(data, { returning: true }))
```
cnasc: what does `upsert` return? For example, are there cases where it doesn't throw, but does return something other than `true`?

Just asking for my own understanding, everything looks good to me

akeem: Good question, the outline of the call is here.  https://sequelize.org/master/class/lib/model.js~Model.html#static-method-update

I should update these methods to be more robust to avoid future digging around issues

PR #4502, opened 2019-08-23
data-iframe now emits keys and transactions

cnasc:
# Description

This PR extends the data-iframe to emit keys and transactions in addition to the other postmessages. This does not alter any existing functionality, it only adds new capabilities.

# Issues

Refs #4492 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/data-iframe/Mailbox/sanitizeBlockchainData.test.ts
+++ b/paywall/src/__tests__/data-iframe/Mailbox/sanitizeBlockchainData.test.ts
@@ -16,10 +16,9 @@ import FakeWindow from '../../test-helpers/fakeWindowHelpers'
 import {
   getWalletService,
   getWeb3Service,
-  lockAddresses,
   addresses,
+  blockchainDataLocked,
 } from '../../test-helpers/setupBlockchainHelpers'
-import { TransactionType, TransactionStatus } from '../../../unlockTypes'
 
 let mockWalletService: WalletServiceType
 let mockWeb3Service: Web3ServiceType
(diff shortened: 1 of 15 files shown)


paywall/src/__tests__/data-iframe/Mailbox/sanitizeBlockchainData.test.ts
```diff
@@ -40,60 +39,15 @@ jest.mock('@unlock-protocol/unlock-js', () => {
   }
 })
 
-describe('Mailbox - getDataToSend', () => {
+describe('Mailbox - sanitizeBlockchainData', () => {
```
cnasc: This test was named incorrectly, which made it difficult to find which file the failing test was in

paywall/src/data-iframe/Mailbox.ts
```diff
@@ -359,11 +364,20 @@ export default class Mailbox {
...
     const chainData = data as BlockchainData
     const keys = Object.keys(chainData)
-    if (keys.length !== Object.keys(this.defaultBlockchainData).length)
+    if (keys.length !== Object.keys(this.defaultBlockchainData).length) {
       return nukeAndReturn()
+    }
     if (
       keys.filter(
-        key => !['locks', 'account', 'balance', 'network'].includes(key)
+        key =>
```
cnasc: Finding out that this was an issue was very difficult. I'm going to file an issue about the silent failures in the paywall (of which there are at least 2, but probably more)

paywall/src/data-iframe/Mailbox.ts
```diff
@@ -311,12 +313,15 @@ export default class Mailbox {
    * This is called by the BlockchainHandler when there is an update to chain data
    */
   emitChanges(newData: BlockchainData) {
+    const { keys, transactions } = newData
```
julien51: Why not using the `getDataToSend` function below?

paywall/src/data-iframe/Mailbox.ts
```diff
@@ -359,11 +364,20 @@ export default class Mailbox {
     if (typeof data !== 'object') return nukeAndReturn()
     const chainData = data as BlockchainData
     const keys = Object.keys(chainData)
-    if (keys.length !== Object.keys(this.defaultBlockchainData).length)
+    if (keys.length !== Object.keys(this.defaultBlockchainData).length) {
```
julien51: Thanks for this!

paywall/src/data-iframe/Mailbox.ts
cnasc: As I understand it, there's a chance that we can have new keys before we have new locks. I think if we used `getDataToSend` we might hold off on sending the new keys until we have gotten the locks.

I think we need to look at the architecture here, since `getDataToSend` seems to be oriented entirely around the caching.

Issue #4544, opened 2019-08-29
Add spinner on all buttons with lag


julien51:
Many of our buttons on the UI will perform things which have latency (signing requests, encrypting things, making requests to the server). Since these may take an "uncomfortable" amount of time and users are making "money" things we should show an indication of something "slowly" happening by adding a spinner on these buttons.

cnasc:
List of places where we need a spinner (or at least some indication that something is happening):
- [ ] sign up
- [ ] log in
- [ ] user account key purchase

PR #4561, opened 2019-08-30
Show a loading spinner on user account key purchase

cnasc:
# Description

Before clicking
image

After clicking
image

# Issues

Refs #4544 

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/user-account/KeyPurchaseConfirmation.tsx
+++ b/unlock-app/src/components/interface/user-account/KeyPurchaseConfirmation.tsx
@@ -11,63 +11,93 @@ import {
   SubmitButton,
   LockInfo,
   DisabledButton,
+  LoadingButton,
 } from './styles'
 import { Fees } from '../../../actions/keyPurchase'
 
-interface KeyPurchaseConfirmationProps {
+interface Props {
   address: string
   emailAddress: string
   lock?: Lock
-  cards: stripe.Card[]
+  card: string
   priceBreakdown: { [name: string]: string }
   signPurchaseData: (d: PurchaseData) => any
 }
 
-// TODO: get credit card in state, pass in here for use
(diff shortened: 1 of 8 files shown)


unlock-app/src/stories/content/AccountContent.stories.js
```diff
@@ -40,7 +40,12 @@ const loggedInWithCards = {
       keyPrice: '0.2',
       expirationDuration: 12345678,
     },
-    price: 1215,
+    fees: {
```
cnasc: Just had to update this story so it actually shows a price

unlock-app/src/components/interface/user-account/KeyPurchaseConfirmation.tsx
```diff
@@ -11,63 +11,93 @@ import {
...
+  LoadingButton,
 } from './styles'
 import { Fees } from '../../../actions/keyPurchase'
 
-interface KeyPurchaseConfirmationProps {
+interface Props {
   address: string
   emailAddress: string
   lock?: Lock
-  cards: stripe.Card[]
```
julien51: Any reason this goes from multiple cards to a single card?

cnasc: Previously this component just pulled the first card (if present) out of the array and rendered it. Now `mapStateToProps` does that. The actual output is the same (though some stories need to be updated, since the stories use the unconnected component) but it makes the component easier to read.

cnasc:
> The card number is not visible on the stories anymore apparently?

I'll update the stories to pass the correct value. I'll also have to move things around a bit so that the button doesn't change position on the password change form.

cnasc:
@julien51 made updates to fix the storyshots problems, and rectified the change to the button margin on `ChangePassword`. PTAL

PR #4668, opened 2019-09-03
fixing linting

julien51:
# Description

Package upgrades introduces new rules which broke master (not sure why tests did not catch that on the upgrade PR). This PR fixes that. We will rebase all other PRs.

# Checklist:

### DIFF
--- a/paywall/src/hooks/browser/useLocalStorage.js
+++ b/paywall/src/hooks/browser/useLocalStorage.js
@@ -16,7 +16,7 @@ export default function useLocalStorage(key) {
   useEffect(() => {
     if (!available) return // do nothing if localStorage can't be used
     window.localStorage.setItem(key, value)
-  }, [value]) // and not on every update to the component containing the hook // this effect only runs when the value changes
+  }, [available, key, value, window.localStorage]) // and not on every update to the component containing the hook // this effect only runs when the value changes
 
   return [value, setValue]
 }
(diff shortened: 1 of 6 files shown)


paywall/src/hooks/browser/useLocalStorage.js
```diff
@@ -16,7 +16,7 @@ export default function useLocalStorage(key) {
   useEffect(() => {
     if (!available) return // do nothing if localStorage can't be used
     window.localStorage.setItem(key, value)
-  }, [value]) // and not on every update to the component containing the hook // this effect only runs when the value changes
+  }, [available, key, value, window.localStorage]) // and not on every update to the component containing the hook // this effect only runs when the value changes
```
cnasc: I'm concerned that changing these might change the behavior of the hooks. What eslint rule was triggered by these?

julien51: hum. I was thinking about this too but then tests passed.  I think these values are not used ayway.
The rule is `react-hooks/exhaustive-deps`.

julien51: > /Users/julien/repos/unlock/paywall/src/hooks/browser/useLocalStorage.js
>  19:6  warning  React Hook useEffect has missing dependencies: 'available', 'key', and 'window.localStorage'. Either include them or remove the dependency array  react-hooks/exhaustive-deps

cnasc: Works for me, I guess if any of those changes it will actually cause a different effect to happen in any event.

PR #4672, opened 2019-09-03
adding a default export to messageTypes to fix eslint

julien51:
# Description

Follow on of https://github.com/unlock-protocol/unlock/pull/4668
There was also breakage in unlock-app.

# Checklist:

### DIFF
--- a/scripts/monorepo.sh
+++ b/scripts/monorepo.sh
@@ -7,8 +7,10 @@ BRANCH=$3
 # This script is highly specific to circleci and will "succeed" early when the
 # current git changes are not relevant to the folder being tested (and not on master)
 
+SHARED_CONFIG_FILES="package.json .eslintrc.js .prettierrc"
+
 if [ ! "$BRANCH" = "master" ]; then
-  HAS_CHANGES=$(git show --name-only $COMMIT -- package.json $SERVICE)
+  HAS_CHANGES=$(git show --name-only $COMMIT -- $SHARED_CONFIG_FILES $SERVICE)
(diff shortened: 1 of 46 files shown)


cnasc:
I think the recent lint upgrades broke a bunch of things.

julien51:
dang. I a re-installing things locally to look at that.

cnasc:
```
[breaking] enable `react/jsx-props-no-spreading`
```

https://github.com/unlock-protocol/unlock/pull/4652

Issue introduced here

julien51:
yup. I'll update this PR with fixes.

julien51:
So, I just pushed a large update which adds changes to a lot of files. There is no change in logic, just stylistic changes.

scripts/monorepo.sh
```diff
@@ -7,8 +7,10 @@ BRANCH=$3
 # This script is highly specific to circleci and will "succeed" early when the
 # current git changes are not relevant to the folder being tested (and not on master)
 
+SHARED_CONFIG_FILES=(package.json .eslintrc.js .prettierrc)
```
julien51: This is really the only important change here: it makes sure that if we change any of these 3 files, tests (including linting) will be run again in the subdirectory.

This should reduce the likeliness of this mess.

julien51:
There is something fishy. I would expect all tests to run because I changed the eslint config. Let me do more tests locally.

julien51:
confirmed. I was using an array var in the bash script when it should be a string. pushing a fix.

PR #4684, opened 2019-09-04
Design Folder

smombartz:
Created folder with branding and illustration assets


smombartz:
@julien51 removed blog folder

PR #4693, opened 2019-09-05
Remove createUnlockStore from the Paywall

cnasc:
# Description

@julien51 has been hard at work removing the last remaining bits of Redux from the Paywall app. Right now PR #4578 is blocked by CI failures, because `createUnlockStore` refers to the reducers (and is also no longer necessary).

This PR removes `createUnlockStore` and any place that referenced it in tests, which should provide a clear path to merge #4578.

In working on this, I noticed that we may have 2 tests for the same thing:
- `paywall/src/__tests__/components/lock/Lock.test.js`
- `paywall/src/__tests__/components/paywall/Lock.test.js`

It's worth looking at these in the future and seeing if one can be deleted.


paywall/src/__tests__/components/content/CheckoutContent.test.tsx
```diff
@@ -16,7 +14,9 @@ jest.mock('../../../hooks/utils/useConfig', () => {
 jest.mock('../../../hooks/useBlockchainData', () => {
   return () => ({
     account: {
-      balance: '200',
+      balance: {
```
cnasc: This was a proptype violation that I fixed while I was in here.

PR #4694, opened 2019-09-05
Provides service exposing key ownership

akeem:
# Description

In support of the key chain and providing address holders with the ability to obtain a list of the locks that they own keys to (keys the own). An endpoint is being added to expose this information.

The initial implementation is scrappy but should provide enough functionality to shape the feature and to be updated then our thresholds are met (increase in scalability, added time, etc)

the endpoint via GET of  `/users/<ethereum address>/keys`

Loosely this is how it all works, with the understanding that this will be simplified greatly when we cut over to the graph.

We leverage logs of transactions made by the Unlock Contract, with this information we should have a list of the locks deployed.

With deployed locks in tow, we leverage Alchemy's extended API  to provide us with token (key) ownership

The branch is knowingly light on testing in some spots with the understand that this will be swapped out for a much less complicated future implementation

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/controllers/userController.ts
+++ b/locksmith/src/controllers/userController.ts
@@ -1,7 +1,10 @@
 import { Request, Response } from 'express-serve-static-core' // eslint-disable-line no-unused-vars, import/no-unresolved
 import { DecoyUser } from '../utils/decoyUser'
+import * as OwnedKeys from '../utils/ownedKeys'
 
 import UserOperations = require('../operations/userOperations')
+
+const env = process.env.NODE_ENV || 'development'
 
 namespace UserController {
   export const createUser = async (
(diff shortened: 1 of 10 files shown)


locksmith/src/controllers/userController.ts
```diff
@@ -125,6 +128,17 @@ namespace UserController {
     let result = await UserOperations.getCards(emailAddress)
     return res.json(result)
   }
+
+  export const keys = async (req: Request, res: Response) => {
+    if (env == 'development') {
```
cnasc: Why do we not support this on dev?

akeem: At the moment, the feature is supported by external APIs (with support on mainnet(1) & rinkeby(4).

Initially I did have an implementation that would work locally but having two implementations and having to support two with the local implementation being particularly slow didnt seem worth it. 

This is what pushed me toward the exploration of a way to speed all of this up initially. Should be replaced and unified when we sort out the graph stuff

cnasc: Right, that makes sense

PR #4696, opened 2019-09-05
Remove Redux and Routers from paywall

cnasc:
# Description

We don't need these packages anymore, so let's ditch them.


paywall/src/stories/content/CheckoutContent.stories.js
```diff
@@ -120,151 +120,160 @@ storiesOf('Checkout page', module)
...
-        })
-        setTimeout(() => {
-          postedMessage({
-            ...messageTemplate,
-            data: {
-              type: POST_MESSAGE_UPDATE_ACCOUNT,
-              payload: lockAddress1,
-            },
-          })
+    const Wrapper = () => {
```
cnasc: The Git diff on this file is getting computed really weirdly. The only change is that each of these stories is put into a wrapper function. This solves an error caused by using the `useEffect` hook outside a component, which only surfaced after removing redux.

PR #4721, opened 2019-09-11
Provide an ERC20 balance representation for managed accounts

### LINKED ISSUE #4720: Managed accounts always show "Insufficient Balance" on locks
**Describe the bug**
After #4534, where we started checking whether users had a sufficient amount of ERC20 tokens to purchase keys for an ERC20 lock, managed user accounts (which always have a balance of `0` for every token) started to show the "Insufficient Balance" message.

Previously it worked because we did not check for balance on ERC20 locks, we deferred that warning to the wallet.

**To Reproduce**
1. Run local dev
2. Visit the demo page with an ERC20 lock set up and user accounts enabled
3. Log in and go to the checkout
4. See something like the screenshot below

**Expected behavior**
Since user accounts never actually have a balance, we should "fake it" for approved ERC20 locks.

**Screenshots**
image

cnasc:
# Description

This PR injects a default balance for ERC20 tokens when we use a managed user account. The balance is arbitrary, but assumed to be high enough to cover likely approved locks for the time being. This change allows the checkout to work with user accounts without needing to be aware that it is.

image

Fixes #4720 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
+++ b/paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
@@ -1,5 +1,8 @@
 import FakeWindow from '../test-helpers/fakeWindowHelpers'
-import CheckoutUIHandler from '../../unlock.js/CheckoutUIHandler'
+import CheckoutUIHandler, {
+  injectDefaultBalance,
+  defaultBalance,
+} from '../../unlock.js/CheckoutUIHandler'
 import { PaywallConfig, Locks, Transactions } from '../../unlockTypes'
 import { KeyResults } from '../../data-iframe/blockchainHandler/blockChainTypes'
 import IframeHandler from '../../unlock.js/IframeHandler'
@@ -74,7 +77,7 @@ describe('CheckoutUIHandler', () => {
     function makeReadyCheckout() {
       fakeWindow = new FakeWindow()
       const handler = makeCheckoutUIHandler(fakeWindow)
-      handler.init()
+      handler.init(false)
 
       fakeWindow.receivePostMessageFromIframe(
         PostMessages.READY,
@@ -184,7 +187,7 @@ describe('CheckoutUIHandler', () => {
         expect.assertions(1)
 
         const handler = makeCheckoutUIHandler(fakeWindow)
-        handler.init()
+        handler.init(false)
 
         fakeWindow.receivePostMessageFromIframe(
           type,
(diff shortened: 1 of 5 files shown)


paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
```diff
@@ -159,7 +162,13 @@ describe('CheckoutUIHandler', () => {
 
     const data: RelayedMessageTestType = [
       ['UPDATE_ACCOUNT', PostMessages.UPDATE_ACCOUNT, fakeAccount],
-      ['UPDATE_ACCOUNT_BALANCE', PostMessages.UPDATE_ACCOUNT_BALANCE, '123'],
+      [
+        'UPDATE_ACCOUNT_BALANCE',
+        PostMessages.UPDATE_ACCOUNT_BALANCE,
+        {
+          eth: '123',
```
cnasc: This test was still in the old balance format. Fixed it while I was in there, but it's not critical since this test just verifies that it passes through without touching the data.

paywall/src/unlock.js/CheckoutUIHandler.ts
```diff
@@ -1,6 +1,24 @@
 import IframeHandler from './IframeHandler'
 import { PostMessages } from '../messageTypes'
-import { PaywallConfig } from '../unlockTypes'
+import { PaywallConfig, Balance } from '../unlockTypes'
+
+// Arbitrarily defined number of tokens to assume a managed user
+// account holds for any given stablecoin
+export const defaultBalance = '35'
```
cnasc: I mentioned in discussion with @julien51 that because we are unlikely to approve any especially expensive locks, we can start by just assuming something like 35DAI.

paywall/src/unlock.js/CheckoutUIHandler.ts
```diff
@@ -1,6 +1,24 @@
...
+
+export const injectDefaultBalance = (oldBalance: Balance): Balance => {
+  const newBalance: Balance = {}
+  const tokens = Object.keys(oldBalance)
+  tokens.forEach(token => {
+    if (token.startsWith('0x')) {
```
cnasc: My understanding is that balances of ERC20 tokens are all stored with the currency contract address as the key. Is that assumption useful?

paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
```diff
@@ -74,7 +77,7 @@ describe('CheckoutUIHandler', () => {
     function makeReadyCheckout() {
       fakeWindow = new FakeWindow()
       const handler = makeCheckoutUIHandler(fakeWindow)
-      handler.init()
+      handler.init(false)
```
julien51: nit: do you mind adding a comment to indicate what `false` is for, like `handler.init(false /* usingManagedAccount */)`

paywall/src/unlock.js/CheckoutUIHandler.ts
julien51: Can we make this a constant? This would make re-using easier... just in case!

paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
cnasc: As an alternative, what about passing it in as an object? That would enforce that it's always clear at the caller what that value is

paywall/src/unlock.js/startup.ts
```diff
@@ -76,6 +76,6 @@ export default function startup(
   // go!
   mainWindow.init()
   wallet.init()
-  checkoutIframeHandler.init()
+  checkoutIframeHandler.init(wallet.useUserAccounts)
```
julien51: One thing that I want to clarify: is it true _if_ the account is actually provided by us, or is it true if the use of unlock accounts is authorized?

cnasc: In `Wallet.ts`, the decision goes as follows:

```typescript
this.useUserAccounts =
      !this.hasWallet &&
      (config.unlockUserAccounts === true ||
        config.unlockUserAccounts === 'true')
```

So if there is no browser wallet (such as metamask) _and_ the config allows for user accounts, it's true. For any other combination it is false.

PR #4744, opened 2019-09-17
Don't render anything stripe related until we have a stripe object

cnasc:
# Description

We noticed earlier today that the settings page was broken. Some investigation made it clear that the issue was related to `react-stripe-elements`. I was able to fix this issue by not rendering the credit card form until we absolutely already had a Stripe object instantiated. This is contrary to the Stripe docs which indicate that it is OK for it to be null before the object is ready. 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/content/SettingsContent.tsx
+++ b/unlock-app/src/components/content/SettingsContent.tsx
@@ -46,6 +46,10 @@ export class SettingsContent extends React.Component<
     this.interval = setInterval(this.getStripe, 500)
   }
 
+  componentWillUnmount() {
+    this.removeInterval()
+  }
+
   getStripe = () => {
     const {
       config: { stripeApiKey },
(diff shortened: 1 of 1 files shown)


unlock-app/src/components/content/SettingsContent.tsx
```diff
@@ -54,15 +58,20 @@ export class SettingsContent extends React.Component<
...
         stripe: window.Stripe(stripeApiKey),
       })
-      if (this.interval) {
-        clearInterval(this.interval)
-      }
+      this.removeInterval()
+    }
+  }
+
+  removeInterval() {
```
cnasc: This was just a bit of cleanup to make sure we get rid of the interval if needed.

unlock-app/src/components/content/SettingsContent.tsx
```diff
@@ -73,7 +82,7 @@ export class SettingsContent extends React.Component<
         <AccountInfo />
         <ChangePassword />
         {cards.length > 0 && <PaymentMethods cards={cards} />}
-        {!cards.length && <PaymentDetails stripe={stripe} />}
+        {stripe && !cards.length && <PaymentDetails stripe={stripe} />}
```
cnasc: This is the real meat of the change. Something about the transition from null to Stripe makes `react-stripe-elements` go wild and break the app.

PR #4777, opened 2019-09-20
now showing a link when there is no href

### LINKED ISSUE #4776: Ticket page should not show website link if there is none
This event does not have a website: https://tickets.unlock-protocol.com/event/0xaA23E8ceBd3817eeEC735Cd9aaF557cCB8e8407C but we still show a link (which goes nowhere!).

We should not show the link if there is none ;)

julien51:
# Description

now showing a link when there is no href
# Issues

Fixes #4776

# Checklist:

### DIFF
--- a/tickets/src/components/interface/EventLinks.js
+++ b/tickets/src/components/interface/EventLinks.js
@@ -46,13 +46,15 @@ export const EventLinks = ({ event }) => {
 
   const externalLinks = eventLinks.map(
     ({ href, text, icon = '/static/images/illustrations/link.svg' }) => {
-      return (
-        <Link key={href} icon={icon}>
-          <a target="_blank" rel="noopener noreferrer" href={href}>
-            {text}
-          </a>
-        </Link>
-      )
+      if (href) {
+        return (
+          <Link key={href} icon={icon}>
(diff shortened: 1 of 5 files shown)


julien51:
image

tickets/src/components/interface/EventLinks.js
```diff
@@ -5,7 +5,7 @@ import { googleCalendarLinkBuilder } from '../../utils/links'
 
 export const EventLinks = ({ event }) => {
   if (typeof window === 'undefined') {
-    return <></>
+    return <></> // WHAT IS THIS???
```
cnasc: https://reactjs.org/docs/fragments.html#short-syntax

julien51: Ho, TIL!

cnasc: In this case it's just being used to return "empty" JSX because that function has a dependency on `window` that may not be there (IIRC I added it because I got SSR errors a few times). I think in the future we'll be able to remove it, just at the moment that component is a verbatim copy/paste from the megafunction that is `EventContent`

PR #4783, opened 2019-09-21
Uniswap token value to track GDP in ETH

### LINKED ISSUE #3711: Use a price oracle to determine GDP in Unlock.sol
Today the GDP calculation sums ETH and ERC20 tokens as if they were the same units.  That means the GDP value we have ATM is worthless.

Fixing this is required in order to properly mint Unlock Discount Tokens.

Consider https://github.com/marbleprotocol/polaris or other solution to determine the value of an ERC20 token, and use that when updating the GDP.

If the token is not tracked by the oracle, I think we can assume 0 and carry on.

**Additional context**

NickCuso:
# Description

Allows us to register support for certain ERC-20 tokens.  If the token is supported, we use the registered Uniswap exchange to convert the value paid into the ETH equivalent.

If the token is not supported, the value is assumed to be 0.
If the lock is priced in ETH, no conversion is applied.

!! This breaks v0 Locks !!
Any Lock created with v0 will no longer support purchasing new keys.  Other functions should be fine.

We could workaround this by using assembly, but as discussed is Slack previously it should be okay to retire v0 at this point.

# Issues

Fixes #3711
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/Unlock.sol
+++ b/smart-contracts/contracts/Unlock.sol
@@ -30,6 +30,7 @@ import '@openzeppelin/contracts-ethereum-package/contracts/ownership/Ownable.sol
 import '@openzeppelin/upgrades/contracts/Initializable.sol';
 import './PublicLock.sol';
 import './interfaces/IUnlock.sol';
+import './interfaces/IUniswap.sol';
 import './mixins/CloneFactory.sol';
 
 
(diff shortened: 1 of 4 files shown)


smart-contracts/contracts/Unlock.sol
```diff
@@ -158,9 +163,27 @@ contract Unlock is
...
+        }
+      }
+      else {
+        // If priced in ETH (or value is 0), no conversion is required
+        valueInETH = _value;
+      }
+
+      grossNetworkProduct += valueInETH;
+      locks[msg.sender].totalSales += valueInETH;
+    }
```
nfurfaro: This looks ok.

PR #4787, opened 2019-09-21
Support free trials

### LINKED ISSUE #4762: Risk free trial
In some cases, a creator might want to offer a "trial" period during which someone who previously purchased a key can receive a full refund if they burn their key.

This trial period should default to 0 seconds.

NickCuso:
# Description

Adds a `freeTrialLength` to the PublicLock.  This defaults to 0.  When set to a non-zero value by the lock owner, this is the number of seconds where `cancelAndRefund` will provide a full refund.

The `getCancelAndRefundValueFor` function will reflect the trial as well.

Note that the time is based on when the transaction is mined, not when it's issued.  So it's possible that a user attempts to cancel during the free trial but it does not actually mine until after the trial has passed (resulting in a partial refund instead).

# Issues

Fixes #4762
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinRefunds.sol
+++ b/smart-contracts/contracts/mixins/MixinRefunds.sol
@@ -34,8 +34,7 @@ contract MixinRefunds is
   );
 
   event RefundPenaltyChanged(
-    uint oldRefundPenaltyNumerator,
-    uint oldRefundPenaltyDenominator,
+    uint freeTrialLength,
     uint refundPenaltyNumerator,
     uint refundPenaltyDenominator
   );
@@ -102,6 +101,7 @@ contract MixinRefunds is
    * Allow the owner to change the refund penalty.
    */
   function updateRefundPenalty(
+    uint _freeTrialLength,
     uint _refundPenaltyNumerator,
     uint _refundPenaltyDenominator
   )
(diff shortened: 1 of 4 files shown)


smart-contracts/contracts/mixins/MixinRefunds.sol
```diff
@@ -190,18 +193,23 @@ contract MixinRefunds is
...
     }
-    uint penalty = keyPrice.mul(refundPenaltyNumerator) / refundPenaltyDenominator;
-    if (refund > penalty) {
-      // Math: safeSub is not required since the if confirms this won't underflow
-      refund -= penalty;
-    } else {
-      refund = 0;
+
+    // Apply the penalty if this is not a free trial
+    if(freeTrialLength == 0 || timeRemaining + freeTrialLength < expirationDuration)
```
julien51: simplify with `timeRemaining + freeTrialLength < expirationDuration` only?

NickCuso: That changes a rare scenario a bit.
If I buy a key from a lock that does not offer trials.  Then I buy a second key (either normally or via a marketplace).  Now my timeRemaining is > the default expirationDuration... however the lock owner would still want the base penalty to apply on refund. (i suspect..)

PR #4788, opened 2019-09-22
Add a tip to ERC-20 lock purchases

### LINKED ISSUE #4518: Allow specifying price for ERC-20 purchases
Today you can name your price (>= keyPrice) for ETH locks but ERC-20 locks always charge the exact keyPrice.  We should allow tips.

Additionally specifying the price to pay on purchase guards the user from unexpected price changes.

Specify the value to charge in the purchase API.

**Additional context**

NickCuso:
# Description

Allows users to specify the value to pay when purchasing a key.  That value may be `>= keyPrice` in order to send a tip.  If set to 0, it's ignored and the current keyPrice is used.  

The front-end / JS implementation should always specify a value when using ERC-20 tokens.

# Issues

Fixes #4518
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinPurchase.sol
+++ b/smart-contracts/contracts/mixins/MixinPurchase.sol
@@ -27,11 +27,13 @@ contract MixinPurchase is
   * @dev Purchase function
   * @param _recipient address of the recipient of the purchased key
   * @param _referrer address of the user making the referral
+  * @param _value the price to pay for this purchase >=keyPrice
   * @param _data arbitrary data populated by the front-end which initiated the sale
   */
   function purchase(
     address _recipient,
     address _referrer,
+    uint256 _value,
     bytes calldata _data
   ) external payable
     onlyIfAlive
@@ -86,6 +88,7 @@ contract MixinPurchase is
     // We explicitly allow for greater amounts of ETH to allow 'donations'
     // Security: after state changes to minimize risk of re-entrancy
     uint pricePaid = _chargeAtLeast(inMemoryKeyPrice);
+    require(_value >= inMemoryKeyPrice, 'INSUFFICIENT_VALUE');
 
     // Security: last line to minimize risk of re-entrancy
     _onKeySold(_recipient, _referrer, pricePaid, _data);
--- a/smart-contracts/contracts/mixins/MixinPurchase.sol
+++ b/smart-contracts/contracts/mixins/MixinPurchase.sol
@@ -27,7 +27,7 @@ contract MixinPurchase is
   * @dev Purchase function
   * @param _recipient address of the recipient of the purchased key
   * @param _referrer address of the user making the referral
-  * @param _value the price to pay for this purchase >=keyPrice
+  * @param _value the price to pay for this purchase >=keyPrice (or 0 to ignore this check)
   * @param _data arbitrary data populated by the front-end which initiated the sale
   */
   function purchase(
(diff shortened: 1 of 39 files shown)


smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -25,11 +25,13 @@ contract MixinPurchase is
 
   /**
   * @dev Purchase function
+  * @param _value the price to pay for this purchase >=keyPrice (or 0 to ignore this check)
```
nfurfaro: Shouldn't this be `>keyPrice`? 
if _value is `=keyPrice`, this is the same as setting `_value` to 0 isn't it?

NickCuso: Setting value to keyPrice exactly doubles as a security feature.  That way if the lock owner increases the price while my transaction is pending I can't be charged more than I expected (only applicable to ERC-20 when more than keyPrice is approved for spending).

NickCuso: In fact I'll add this to the comments

nfurfaro: Ok, makes sense. I just don't get the comment:
> (or 0 to ignore this check)
If `_value` is set to 0 the purchase will fail won't it?

NickCuso: if _value is 0, you'll pay the current keyPrice

nfurfaro: got it, thanks

NickCuso: hmm actually I should change this.  0 should mean free...

nfurfaro: Just read that in the PR description!

NickCuso: that's going to require a lot of test updates! =p

nfurfaro: Hmm, if `_value` is set to `0`, you'll pay the current `keyPrice`, which would have to be `0` to indicate a free key right?

nfurfaro: I would think `0` only means free if it's referring to the current `keyPrice`

NickCuso: i'm going to remove _value == 0 means use the current keyPrice.  always specify the value.

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -83,7 +88,11 @@ contract MixinPurchase is
       toKey.tokenId
     );
 
-    // We explicitly allow for greater amounts of ETH to allow 'donations'
+    // We explicitly allow for greater amounts of ETH or tokens to allow 'donations'
+    if(_value > 0) {
```
nfurfaro: So from your last comments, `_value` will always be greater than 0 (with the exception of purchasing a free key) if we force the user to always specify the exact `_value`, right?

smart-contracts/test/Lock/erc20.js
```diff
@@ -147,7 +147,9 @@ contract('Lock / erc20', accounts => {
       await token.approve(lock.address, -1)
```
nfurfaro: CAn you remind me what the `-1` is in the approve call?

NickCuso: it's shorthand for MAX_UINT

smart-contracts/contracts/mixins/MixinPurchase.sol
NickCuso: Yea - give me a bit to make those updates.  Fixing the tests will be a bit tedious.

smart-contracts/test/Lock/erc20.js
nfurfaro: ahh, of couse,thx

nfurfaro:
Looking good, but will let you make changes and will take one last look.

NickCuso:
Okay should be good to go now.
The value is ignored if the lock is priced in ETH, we use the msg.value instead.  If it's priced in ERC-20 than the amount to pay, >= the current keyPrice - any applicable discount must be specified.

nfurfaro:
@NickCuso regarding this:
> any applicable discount must be specified

can you tell me more about where the discount would be specified?

NickCuso:
That's a future feature via the UnlockDiscountToken

PR #4790, opened 2019-09-22
Deploy PublicLock using a minimal proxy

### LINKED ISSUE #4708: Deploy PublicLock with a Minimal Proxy Contract
Deploying Unlock.sol is very expensive, soon we will hit the max per transaction and be unable to upgrade.

Deploy the PublicLock from Unlock using https://eips.ethereum.org/EIPS/eip-1167
This will remove the PublicLock bytecode from the Unlock.sol deployment, significantly reducing that cost.

..

**Additional context**
This is actually the only supported method for deploying contracts from Vyper.

NickCuso:
# Description

Deploys a proxy on `createLock` instead of the full contract each time.  This is a significant gas savings for `createLock` and for deploying `Unlock.sol`

Nothing is free though:

```
Before:
purchase 23128  ·     243420  ·     164059 
After:
purchase 23128  ·     244273  ·     164586
```
Which is an additional 527 gas for the worst case column, or a 0.32% increase.

!! It's important to note that we must add this to the upgrade process / checklist.  Call `configUnlock` with the address of the latest PublicLock template. 

# Issues

Fixes #4708
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/CloneFactory.sol
+++ b/smart-contracts/contracts/mixins/CloneFactory.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.5.10;
+pragma solidity 0.5.11;
 
 // From https://github.com/optionality/clone-factory/blob/master/contracts/CloneFactory.sol
 // Updated to support Solidity 5
@@ -6,6 +6,7 @@ pragma solidity 0.5.10;
 contract CloneFactory {
   function createClone(address target) internal returns (address result) {
     bytes20 targetBytes = bytes20(target);
+    // solium-disable-next-line
     assembly {
       let clone := mload(0x40)
       mstore(clone, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
--- a/smart-contracts/contracts/mixins/CloneFactory.sol
+++ b/smart-contracts/contracts/mixins/CloneFactory.sol
@@ -1,0 +1,17 @@
+pragma solidity 0.5.10;
+
+// From https://github.com/optionality/clone-factory/blob/master/contracts/CloneFactory.sol
+// Updated to support Solidity 5
+
+contract CloneFactory {
+  function createClone(address target) internal returns (address result) {
+    bytes20 targetBytes = bytes20(target);
+    assembly {
+      let clone := mload(0x40)
+      mstore(clone, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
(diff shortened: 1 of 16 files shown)


smart-contracts/test/Lock/grantKeys.js
```diff
@@ -1,5 +1,6 @@
 const Web3Utils = require('web3-utils')
 
+const truffleAssert = require('truffle-assertions')
```
nfurfaro: is this new as of this PR as well?

NickCuso: https://github.com/unlock-protocol/unlock/pull/4784

smart-contracts/contracts/mixins/CloneFactory.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity 0.5.11;
+
+// From https://github.com/optionality/clone-factory/blob/master/contracts/CloneFactory.sol
+// Updated to support Solidity 5
+
+contract CloneFactory {
+  function createClone(address target) internal returns (address result) {
```
nfurfaro: I see there's a `isClone` function available if we need it as well.

NickCuso: Yup.  I left it out though to save gas.  Knowing it's a clone doesn't help us much, we still need to confirm that Unlock.sol created it (or check the code itself).

smart-contracts/test/helpers/deployLocks.js
```diff
@@ -38,7 +38,7 @@ module.exports = function deployLocks(
       } else {
         const tx = await createCall
         // THIS API IS LIKELY TO BREAK BECAUSE IT ASSUMES SO MUCH
-        const evt = tx.logs[1]
+        const evt = tx.logs.find(v => v.event === 'NewLock')
```
nfurfaro: I'm not familiar with this syntax...it looks like a way to filter logs by event name?

NickCuso: Yea, it's JS shorthand to write a loop looking for a certain element.  for each element (`v` here) run the function until one returns true

nfurfaro: cool, thx

smart-contracts/contracts/mixins/CloneFactory.sol
nfurfaro: copy that.

PR #4794, opened 2019-09-22
Gas savings: remove getter function

NickCuso:
# Description

Saves 38 on purchase but adds 5,216 to deploy and createLock.  Still seems worthwhile since it simplifies the code a bit and purchase is the more important use case to optimize for.

# Issues

Fixes #
Refs #

# Checklist:


julien51:
The addition is just because the solidity compiler now adds this getter by itself?

NickCuso:
That's a good question.  I don't think it's the getter cause it's typically cheaper to use the generated one I've found.  I'll test a bit

PR #4806, opened 2019-09-22
Codechecks config for gas reports

### LINKED ISSUE #3621: Contract tests should confirm gas limits
Replace https://github.com/unlock-protocol/unlock/blob/master/smart-contracts/test/helpers/walletServiceMock.js with https://github.com/unlock-protocol/unlock/blob/master/unlock-js/src/constants.js and add gas limit tests for anything missing ATM such as deploy Unlock.

NickCuso:
# Description

@julien51 in order to enable this you need to go to codechecks.io
 - enable the project (there's a team selector in the top right in case it defaults to your personal account)
 - copy the secret
 - paste it as an environment variable in CircleCI called `CC_SECRET`

This will publish the gas report into the github status checks.  Note that the check always passes, this is just surfacing the information so that it is easy for us to consume.

e.g.
image
which links to a report like this https://github.com/HardlyDifficult/eth-gas-costs/pull/28/checks?check_run_id=231737999

# Issues

Fixes #3621
Refs #

# Checklist:


NickCuso:
I moved codechecks out of the package.json and install it on demand in the CI workflow.  

When it's in the package.json (as a dependency, not a devDependency) we get an error that the client is not installed.  I suspect this is related to the --production flag we use on npm ci.

This approach appears to be working.  CI ends with `Unexpected token function` which is the expected behavior until the secret environment variable is added.

julien51:
Oh! sorry I missde this!

.circleci/config.yml
```diff
@@ -98,6 +98,9 @@ jobs:
       - run:
           name: Smart Contract Tests
           command: scripts/tests.sh smart-contracts
+      - run: 
+          name: Codechecks
+          command: cd smart-contracts && npm i @codechecks/client && npx codechecks
```
julien51: I am not a huge fan of `npx` which will install the package?
Could we instead add `@codechecks/client` as a dev dependency and a corresponding script `codechecks` ( and put `npm run codechecks` in here)?

NickCuso: hmm I tried something like that in a previous commit, it didn't work.. i think it's due to the `--production` flag.  Can we try turning that off?

npx will not install the package since it's scoped by name.  that's why i have the `npm i` as well.

NickCuso: hmm nm maybe it did work, i can revert and try that again

NickCuso:
Will do, is the token added to CI?  It would be nice to look for a test pass vs the right fail message ;)

NickCuso:
Hmm yea this fails.  So I'm actually not sure how to call this?  Do we need to add something to the .sh script itself instead?

NickCuso:
okay I think we are good to go now.  just need that token added so we can test it out.

NickCuso:
@julien51 could you add the codechecks token to CI so we could test this out?

julien51:
Ok, I just did the required steps. I am not sure that `CC_SECRET` is the right variable name though. All of our tests are run inside of docker images and not all env vars are exposed to these containers. 

Let me know if things do not work as expected.

NickCuso:
Oh.  Well I guess I can re-run CI but it sounds like this won't work.  How do we expose a variable?

NickCuso:
@julien51 looks like you were right.  How do we expose a variable?

julien51:
See `https://github.com/unlock-protocol/unlock/blob/master/scripts/tests.sh`

julien51:
Also make sure you rebase too? I don't know when/if vars are set?

NickCuso:
Rebased.  If I'm reading this correctly we just need the environment variable renamed to `SMART_CONTRACTS_CC_SECRET`

julien51:
@NickCuso Any idea what is going on?
I believe the var is now set correctly...
```Error occured:  Could not find running CI.```

**Comment by NickCuso** [comment]
Hmm I don't know.  It seems to be related to Docker.  I'll dig around a bit to see if I can find what it's looking for..

**Comment by NickCuso** [comment]
@julien51 It's expecting the Circle environment variables to be available as well.  See https://github.com/codechecks/monorepo/blob/master/packages/client/src/ci-providers/Circle.ts#L8

Looks like it expects all the following variables:
 - CIRCLECI
 - CIRCLE_PULL_REQUEST
 - CIRCLE_SHA1
 - CIRCLE_PULL_REQUESTS
 - CIRCLE_PR_USERNAME 
 - CIRCLE_PR_REPONAME
 - CIRCLE_BUILD_URL

**Comment by NickCuso** [comment]
Adding CIRCLE* in test.sh made some progress.

Next error: https://github.com/codechecks/monorepo/blob/73f9a5cc1540b43a579bc81864369a9dd4ccc6d2/packages/client/src/runner.ts#L40
```
const gitRoot = findRootGitRepository(process.cwd());
  if (!gitRoot) {
    throw crash("Couldn't find git project root!");
  }
```

Implemented here: https://github.com/codechecks/monorepo/blob/fbccf0aa6102c6564863d37f614c294b6646e4f4/packages/client/src/utils/git.ts#L7
```
export function findRootGitRepository(path: string): string | undefined {
  const gitDirPath = join(path, ".git");
  const parentDir = join(path, "..");

  if (path === "/" || !path) {
    return undefined;
  }

  if (existsSync(gitDirPath)) {
    return path;
  } else {
    return findRootGitRepository(parentDir);
  }
}
```

Hmm I'm not sure why this fails..

**Comment by NickCuso** [comment]
We did it! \o/

image

**Comment by NickCuso** [comment]
@julien51 this is ready for review, finally.  I added changes to smart-contracts.dockerfile and test.sh.  Let me know if there are any concerns.

File: docker/smart-contracts.dockerfile [hunk 2]
```diff
@@ -2,6 +2,8 @@ FROM unlock-core
 
 # Dependencies for smart-contracts
 RUN mkdir /home/unlock/smart-contracts
+# Copy the .git directory for codechecks integration
+COPY --chown=node .git/* /home/unlock/.git/
```
julien51: Hum. I am fine copying this but you should know that it will make all builds significangtly slower because this folder will change on every commit and hence the docker build process will be much longer.

If you can skip this, then I'd recommend skipping. Otherwise you should move that as "late" as possible in the dockerfile in order to leverage the cache until the build process does this.

scripts/tests.sh
```diff
@@ -11,6 +11,8 @@ COMMAND="npm run ci"
 # UNLOCK_APP_X will be passed to the container for tests in unlock_app as X.
 UPCASE_SERVICE="${SERVICE^^}"
 ENV_VARS_PREFIX="${UPCASE_SERVICE//-/_}_"
-ENV_VARS=`env | grep $ENV_VARS_PREFIX | awk '{print "-e ",$1}' ORS=' ' | sed -e "s/$ENV_VARS_PREFIX//g"`
+# Also pass all env variables which begin with "CIRCLE" (which should represent CI config variables)
+# See also https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
+ENV_VARS=`env | grep -e $ENV_VARS_PREFIX -e "CIRCLE" | awk '{print "-e ",$1}' ORS=' ' | sed -e "s/$ENV_VARS_PREFIX//g"`
```
julien51: 👍

docker/smart-contracts.dockerfile
NickCuso: hmm okay.  I actually think all they need is for the directory to exist.  maybe we can `touch` instead - I'll give that a shot.

PR #4817, opened 2019-09-23
Create NPM module for paywall main window script

cnasc:
# Description

In response to a request from the MyCrypto team, this PR introduces a module that can be `npm install`-ed so that end users don't have to pull scripts from our server.

This is a quick and dirty first pass, please do offer any critique that comes to mind.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/module.ts
+++ b/paywall/src/unlock.js/module.ts
@@ -1,0 +1,68 @@
+import startup from './startup'
+import '../static/iframe.css'
+import { UnlockWindowNoProtocolYet } from '../windowTypes'
+import StartupConstants from './startupTypes'
+
+declare const process: {
+  env: {
+    UNLOCK_ENV: 'dev' | 'test' | 'staging' | 'prod'
+    DEBUG: any
+    PAYWALL_URL: string
+    USER_IFRAME_URL: string
+  }
+}
+
+const constants: { [key: string]: StartupConstants } = {
+  dev: {
+    network: 1984,
+    debug: process.env.DEBUG,
+    paywallUrl: process.env.PAYWALL_URL || 'http://localhost:3001',
+    accountsUrl: process.env.USER_IFRAME_URL || 'http://localhost:3000/account',
+  },
+  test: {
+    network: 1984,
+    debug: process.env.DEBUG,
+    paywallUrl: process.env.PAYWALL_URL || 'http://localhost:3001',
+    accountsUrl: process.env.USER_IFRAME_URL || 'http://localhost:3000/account',
+  },
+  staging: {
+    network: 4,
+    debug: process.env.DEBUG,
+    paywallUrl: process.env.PAYWALL_URL,
+    accountsUrl: process.env.USER_IFRAME_URL,
+  },
+  prod: {
+    network: 1,
+    debug: process.env.DEBUG,
+    paywallUrl: process.env.PAYWALL_URL,
+    accountsUrl: process.env.USER_IFRAME_URL,
+  },
+}
+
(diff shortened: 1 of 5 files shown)


paywall/src/unlock.js/module.ts
```diff
@@ -0,0 +1,68 @@
...
+    debug: process.env.DEBUG,
+    paywallUrl: process.env.PAYWALL_URL,
+    accountsUrl: process.env.USER_IFRAME_URL,
+  },
+}
+
+// TODO: make this more sophisticated? Would be nice to be able to pass config
+// as a param instead of relying on a window variable. For now this is just a
+// slight modification of `index.ts`
+export function init() {
```
cnasc: This is simple enough, it just wraps what the script would ordinarily do automatically with a function that gets exported.

paywall/src/unlock.js/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "unlock-checkout-builder",
```
cnasc: Suggestions on name appreciated!

julien51: That's not quite the checkout only that it builds, right

cnasc:
Right now we fail on ESLint because the introduction of a new `package.json` scopes the imports to that `package.json`. For the most part that seems to be OK, but the post message event emitters refer to a type from the parent `package.json`. We might want to hoist that file up to make this work without complicating our dep tree.

paywall/src/unlock.js/package.json
cnasc: Yeah, not quite sure what the best name for this would be unless we just want to call it `unlock-paywall`

paywall/src/unlock.js/module.ts
```diff
@@ -0,0 +1,68 @@
...
+declare const process: {
+  env: {
+    UNLOCK_ENV: 'dev' | 'test' | 'staging' | 'prod'
+    DEBUG: any
+    PAYWALL_URL: string
+    USER_IFRAME_URL: string
+  }
+}
+
+const constants: { [key: string]: StartupConstants } = {
```
julien51: Is there any way we can re-use what we already have ... somewhere?

cnasc: We could hoist the common pieces of `index.ts` and `module.ts` to another file and then import in both locations

paywall/src/unlock.js/module.ts
```diff
@@ -0,0 +1,68 @@
...
+  prod: {
+    network: 1,
+    debug: process.env.DEBUG,
+    paywallUrl: process.env.PAYWALL_URL,
+    accountsUrl: process.env.USER_IFRAME_URL,
+  },
+}
+
+// TODO: make this more sophisticated? Would be nice to be able to pass config
+// as a param instead of relying on a window variable. For now this is just a
```
julien51: YES!

cnasc:
> This looks good. Ideally there would be some re-using of some sort which you could do?

Agreed. Since this approach looks good, I'll do a little preliminary organization work in one PR and then introduce this module in a follow-up

paywall/src/unlock.js/package.json
julien51: I am actually fine with `unlock-paywall`, but what about `unlock-web` ?

cnasc: `unlock-web` sounds good to me!

cnasc:
Closing in favor of #4822 which has a cleaner approach

PR #4827, opened 2019-09-23
Removed Chromatic

julien51:
# Description

Next I'll explore replacements.

# Checklist:


docker/unlock-app.dockerfile
```diff
@@ -16,7 +16,7 @@ WORKDIR /home/unlock/unlock-app
 COPY --chown=node unlock-app/ /home/unlock/unlock-app/.
 RUN npm run build
 
-# Copy the .git stuff which is required by chromatic.
+# Copy the .git stuff
```
cnasc: Do we still actually need these parts if they were originally from chromatic?

julien51: I was wondering but I think it's actually good practice to have them in there so we can easily see what commits is being run when debugging things in ci.

PR #4885, opened 2019-09-27
Expose blockchain data on the paywall

cnasc:
# Description

This PR makes the main window handler track the state of the blockchain handler and adds a method to `window.unlockProtocol` that allows end users to get the relevant data.

image

The state that it provides mirrors the internal state of the blockchain handler

# Checklist:


julien51:
Ok, let's not publicly document that just yet though because I want to make sure we have a reflexion about the schema we use.

cnasc:
I’ll write some tests soon to ensure that we always pass through the post message without modification. Imo we should do any transformations at the blockchain handler level and the main window should mirror that internal state.

PR #4886, opened 2019-09-27
plugging email recovery

julien51:
# Description

~~This is the giant pull request for the password recovery mechanism.~~
~~I will split that in pieces but wanted you to look a this and tell me if you see any obvious flaws.~~

This PR is now fully ready. The last step mostly plugs things and adds the corresponding handlers to `storageMiddleware`.

# Checklist:

### DIFF
--- a/unlock-app/src/middlewares/storageMiddleware.js
+++ b/unlock-app/src/middlewares/storageMiddleware.js
@@ -1,6 +1,9 @@
 /* eslint promise/prefer-await-to-then: 0 */
-
-import { createAccountAndPasswordEncryptKey } from '@unlock-protocol/unlock-js'
+import queryString from 'query-string'
+import {
+  createAccountAndPasswordEncryptKey,
+  reEncryptPrivateKey,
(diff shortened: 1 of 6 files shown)


unlock-app/src/middlewares/storageMiddleware.js
```diff
@@ -158,6 +176,38 @@ const storageMiddleware = config => {
       dispatch(setError(Storage.Warning('Could not initiate key purchase.')))
     })
 
+    const { router } = getState()
+    if (router && router.location && router.location.pathname === '/recover/') {
```
cnasc: Would it be better to trigger this from the recover page with an action instead of parsing the router here? That might make it more flexible for unanticipated future use, but I'm ok with it either way.

julien51: That's always something that I am not sure about: how to dispatch actions without user input. We could use some kind of lifecycle method (like after rendering) but that fells gross.

PR #4903, opened 2019-09-30
the `/event` route maps to the `/newevent` page

cnasc:
# Description

Just putting this in the wings for when we want to cut over

### DIFF
--- a/tickets/next.config.js
+++ b/tickets/next.config.js
@@ -81,7 +81,7 @@ module.exports = withTypescript(
       return {
         '/': { page: '/home' },
         '/create': { page: '/create' },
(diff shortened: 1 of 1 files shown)


julien51:
Awesome! Let's do some thorough testing on some pages like the NFT meetup or the BOSS tickets or the ones for coinfund to make sure they're strictly identical!

julien51:
Would it be worth adding a redirect from `/newevent` to `/event` too for EthWaterloo?

julien51:
I guess that's probably for another PR!

cnasc:
Yeah I think we can normalize the naming in the future

PR #4916, opened 2019-10-01
Update the event email template

cnasc:
# Description

In response to some feedback from the EthWaterloo team, this PR will contain updates to our event email template.

image

Remaining TODOs:
- [x] Dynamic subject line based on event name
- [x] Custom text specific to the EthWaterloo event regarding ejection and refunds
- [x] Switching the order of QR and link instructions

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/wedlocks/src/templates/confirmEvent.js
+++ b/wedlocks/src/templates/confirmEvent.js
@@ -1,6 +1,6 @@
 const ethWaterlooMessage = `
 
-After you check in to ETHWaterloo, we'll refund your stake in DAI by the end of the weekend. You'll be able to access those funds once you eject your Unlock account into a crypto wallet. We'll provide more details on that process soon! Note that this event ticket is non-transferable.`
+If you purchased a ticket using an Unlock user account, after you check in to ETHWaterloo we'll refund your stake in DAI by the end of the weekend. You'll be able to access those funds once you eject your Unlock account into a crypto wallet. We'll provide more details on that process soon! Note that this event ticket is non-transferable.`
 
 export default {
   subject: params => `You've got your ticket for ${params.eventName}!`,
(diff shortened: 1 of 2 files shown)


wedlocks/src/templates/confirmEvent.js
```diff
@@ -1,9 +1,17 @@
...
-  subject: () => 'You are attending an event!',
+  subject: params => `You've got your ticket for ${params.eventName}!`,
   text: params =>
     `Hello,
 
-This is just a reminder that you are attending the event ${params.eventName} on ${params.eventDate}!
+This is just a reminder that you are attending the event ${
+      params.eventName
+    } on ${params.eventDate}!${
+      params.eventName === 'ETHWaterloo' ? ethWaterlooMessage : ''
```
cnasc: I'm dispatching on `eventName` here, but it might be wiser to do it by matching the lock address in the `ticketLink` field

akeem: Who currently owns the Lock? Do we have a risk of it changing? The lock Address should at least be consistent though

akeem: How do you feel about doing a downcase comparison here to avoid any funny stuff?

cnasc: We could do that -- any thoughts on going by the lock address in the url instead of event name?

cnasc: Julien created the lock, not sure if he transferred ownership to ETHWaterloo yet. In any event, the address should remain the same always.

akeem: sounds good, we'll probably need to double/triple test the method that does the extraction

PR #4923, opened 2019-10-02
add `blockchainData()` method to `window.unlockProtocol`

cnasc:
# Description

This PR adds a `blockchainData()` method to `window.unlockProtocol` which provides a mirror of the internal state of the blockchain handler. This allows consumers to get relevant information about the state of the chain.

The only contract provided here is that it mirrors the blockchain handler, and those internal details could change in the future (e.g., if we start using the graph and stop munging together the different kinds of data).

The immediate need is for this PR to expose the public key of the account using the paywall.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/MainWindowHandler.ts
+++ b/paywall/src/unlock.js/MainWindowHandler.ts
@@ -71,6 +71,13 @@ export default class MainWindowHandler {
     this.iframes.data.on(PostMessages.UNLOCKED, () => {
       this.toggleLockState(PostMessages.UNLOCKED)
     })
+    this.iframes.data.on(PostMessages.ERROR, e => {
+      if (e === 'no ethereum wallet is available') {
+        this.toggleLockState(PostMessages.LOCKED)
+      }
+    })
+
+    // When the data iframe sends updates, store them in the mirror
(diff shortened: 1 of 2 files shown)


paywall/src/unlock.js/MainWindowHandler.ts
```diff
@@ -65,6 +77,26 @@ export default class MainWindowHandler {
...
     })
 
+    // When the data iframe sends updates, store them in the mirror
+    this.iframes.data.on(PostMessages.UPDATE_LOCKS, locks => {
+      this.blockchainData.locks = locks
+    })
+    this.iframes.data.on(PostMessages.UPDATE_ACCOUNT, address => {
+      this.blockchainData.account = address
+    })
+    this.iframes.data.on(PostMessages.UPDATE_ACCOUNT_BALANCE, balance => {
```
akeem: Do we care about anything outside of address currently/for this use case?

cnasc: The immediate need is just for the address, but all the data is being emitted at present anyway. I can see a concern that it might commit us to the current structure if we show it all. Would you prefer if we only put out an object with the account address in it for now?

PR #4933, opened 2019-10-03
Paywall tells accounts iframe if page is locked/unlocked

cnasc:
# Description

In order to resolve an issue with the paywall prompting user account users to purchase a key they already own, this PR has the paywall tell the accounts iframe whether the page is locked or not. In a follow-on PR, the accounts iframe will use that information to determine whether it should show the purchase prompt.

# Issues

Fixes #
Refs #4934 

# Checklist:


cnasc:
> Makes sense, this is starting to grow all of the legs 🕸

Yeah, I'm also a bit concerned about how much each thing needs to know about each other thing, which is an indication that this architecture isn't ideal.

PR #4936, opened 2019-10-03
Accounts iframe buffers received postmessages

cnasc:
# Description

This PR makes sure that the message sent in #4933 will be received by the accounts iframe by buffering messages to it that happen before the iframe exists and sending them after creation, instead of just ignoring them.

This way, if the `LOCKED | UNLOCKED` message is sent before the account iframe is listening (likely), it won't be missed.

# Issues

Refs #4933 

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/PostMessageEmitters/AccountsIframeMessageEmitter.ts
+++ b/paywall/src/unlock.js/PostMessageEmitters/AccountsIframeMessageEmitter.ts
@@ -22,6 +22,14 @@ import {
   UserAccountsIframeEvents,
 } from '../../EventEmitterTypes'
 
+// This just keeps the typechecker happy. It may not be possible to
+// represent an array of heterogeneous generic types in TS, but this
+// gets the job done.
+interface Message {
+  type: any
+  payload: any
+}
+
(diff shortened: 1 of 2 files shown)


paywall/src/unlock.js/PostMessageEmitters/AccountsIframeMessageEmitter.ts
```diff
@@ -88,6 +98,7 @@ export default class AccountsIframeMessageEmitter extends FancyEmitter {
     this._postMessage = postMessage
     this._addHandler = addHandler
     this.setupListeners()
+    this.sendBufferedMessages()
```
akeem: This is more of a question around my lack of general understanding, but is `setupListeners` blocking until all listeners are "setup"?

cnasc: Yes, `setupListeners` synchronously adds event handlers.

The terminology is a bit confusing here:
- `postMessage` is how you send something to the iframe
- `addHandler` (called by `setupListeners`) is how you receive something from the iframe

PR #4945, opened 2019-10-04
Upgrade locksmith to unlockjs0317

akeem:
# Description

With a change introduced in unlockjs 0.3.17 the fixtures around pricing had gotten out of sync causing the build to fail. 

Updating the package and fixtures appropriately so that we dont end up with massive drift between our internal software

# Issues

Fixes #
Refs #

# Checklist:


akeem:
@cnasc It needs to be done manually and inspected, ideally it would occur closer to the version that changes behavior. Basically a method in the library added additional network calls, the mocking library screamed that new calls were being made which is the right thing to do.  I think we can probably make a giant switch that will make recording easier but we shouldnt let it drive itself because things might actually be broken

PR #4951, opened 2019-10-05
Fix determination of expensiveness

### LINKED ISSUE #4950: DAI lock not showing "Insufficient funds" even though I do not have them?
When using a crypto wallet (MetaMask) with [an account](0x173E1535B113b8dcaFB44DA068e757F142a056f6) which has no Ether and no DAI, we're only showing Insufficient funds for the Ether lock (See https://www.ouvre-boite.com/members/)

image

I believe this is a consequence of user accounts where we want to let the user click on a lock to let them purchase with their CC?

cnasc:
# Description

There was a bug in the function we use for determining whether a key is too expensive for a user. If there was no entry for a token's balance in the account, it would get parsed as `NaN`. And since `NaN` is _not smaller than_ anything (or bigger than anything, for that matter), the balance appeared to the app to be sufficient for a key purchase.

~~This PR fixes that, but in doing so peels back another layer and reveals a different bug. Balance updates aren't correctly fetched when a user logs in to MetaMask while on the page. So now the checkout shows insufficient funds when there is no entry for token balance, but the entry doesn't get added. I will have some more code shortly to address that, and we shouldn't merge this until we have that.~~

This PR also makes sure that when we reset the internal state of the blockchain handler, we also request balances for the currency contract address of each ERC20 lock on the page. In my testing on localhost, this resolved the issue.

# Issues

Fixes #4950 

# Checklist:

### DIFF
--- a/paywall/src/utils/locks.js
+++ b/paywall/src/utils/locks.js
@@ -63,8 +63,10 @@ export function isTooExpensiveForUserByCurrency(lock, account, currencyKey) {
   }
 
   if (keyPrice <= balance) {
-    // This doesn't take gas into account. If keyPrice and balance are exactly
-    // equal, it's probable that no transaction can be carried out.
+    // This doesn't take gas into account. If a user has exactly as much eth as
+    // the price of a key, the wallet will warn that it cannot do the
+    // transaction without a little extra for gas. For an ERC20-priced lock, it
+    // makes sense to be able to spend all of your token on one thing.
     return LOCK_IS_AFFORDABLE
   }
 
--- a/paywall/src/utils/locks.js
+++ b/paywall/src/utils/locks.js
@@ -16,31 +16,76 @@ export function currencySymbolForLock(lock, config) {
 }
 
 /**
+ * Returns true if the keys for `lock` are priced in ERC20, false otherwise.
+ * @param {*} lock
+ */
+export const isERC20Lock = lock => {
+  return !!lock.currencyContractAddress
+}
+
+/**
+ * Returns true if the keys for `lock` are priced in Eth, false otherwise.
+ * @param {*} lock
+ */
+export const isEthLock = lock => {
+  return !lock.currencyContractAddress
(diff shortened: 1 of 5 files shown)


paywall/src/utils/locks.js
```diff
@@ -15,32 +15,77 @@ export function currencySymbolForLock(lock, config) {
...
+  const balance = parseFloat(account.balance[currencyKey])
+  if (!balance) {
+    return LOCK_IS_TOO_EXPENSIVE
+  }
+
+  const keyPrice = parseFloat(lock.keyPrice)
+  if (!keyPrice) {
+    // All locks should have a keyPrice, but if for some reason one doesn't,
+    // we're in some kind of broken state and we'd better prevent purchasing.
+    return LOCK_IS_TOO_EXPENSIVE
```
julien51: 👍

paywall/src/utils/locks.js
```diff
@@ -15,32 +15,77 @@ export function currencySymbolForLock(lock, config) {
...
+  }
+
+  if (keyPrice <= balance) {
+    // This doesn't take gas into account. If keyPrice and balance are exactly
```
julien51: So I think this is only true if this is an Ether lock. For ERC20 locks, we are probably fine with just the right amount?

julien51: Let's to a strict equal for now and the crypto wallet will actually show an error on Ether locks when the user trying to purchase indicating that the balance is not high enough.

cnasc: If we do a strict `<` that would stop someone with exactly 20DAI from buying a key to a 20DAI lock, which feels like it might not be the right thing. With the current logic the wallet will still catch the balance being wrong once gas is accounted for.

julien51: Oh sorry that was exactly what I wanted, I misread your code. We should not block for the strict equality!

PR #4962, opened 2019-10-08
Switch from using fractions to basisPoints

NickCuso:
# Description

Instead of using a numerator / denominator for values, we can simplify and define the units in basis points.  For the impacted use cases this should be plenty of precision, simplifies (for both users and the code itself) and would be a bit cheaper on gas.

Impacted transferFee and refundPenalty 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinLockCore.sol
+++ b/smart-contracts/contracts/mixins/MixinLockCore.sol
@@ -51,6 +51,9 @@ contract MixinLockCore is
   // The account which will receive funds on withdrawal
   address public beneficiary;
 
+  // The denominator component for values specified in basis points.
+  uint public constant BASIS_POINTS_DEN = 10000;
+
   // Ensure that the Lock has not sold all of its keys.
   modifier notSoldOut() {
     require(maxNumberOfKeys > totalSupply, 'LOCK_SOLD_OUT');
--- a/smart-contracts/contracts/mixins/MixinRefunds.sol
+++ b/smart-contracts/contracts/mixins/MixinRefunds.sol
@@ -19,9 +19,8 @@ contract MixinRefunds is
   using SafeMath for uint;
 
   // CancelAndRefund will return funds based on time remaining minus this penalty.
-  // This is calculated as `proRatedRefund * refundPenaltyNumerator / refundPenaltyDenominator`.
-  uint public refundPenaltyNumerator;
-  uint public refundPenaltyDenominator;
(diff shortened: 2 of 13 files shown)


smart-contracts/contracts/mixins/MixinLockCore.sol
```diff
@@ -51,6 +51,9 @@ contract MixinLockCore is
   // The account which will receive funds on withdrawal
   address public beneficiary;
 
+  // The denominator component for values specified in basis points.
```
nfurfaro: I'm unclear why a denominator is needed if we're switching to basis points to get away from fractions?

smart-contracts/contracts/mixins/MixinRefunds.sol
```diff
@@ -37,14 +36,13 @@ contract MixinRefunds is
...
-    uint refundPenaltyDenominator
+    uint refundPenaltyBasisPoints
   );
 
   function initialize() public
   {
-    refundPenaltyNumerator = 1;
-    refundPenaltyDenominator = 10;
+    // default to 10%
+    refundPenaltyBasisPoints = 1000;
```
nfurfaro: So 1000 basis points is the same as 0.1, correct?

nfurfaro:
Why are basis points cheaper on gas?

smart-contracts/contracts/mixins/MixinRefunds.sol
NickCuso: yup

smart-contracts/contracts/mixins/MixinLockCore.sol
NickCuso: we want to support values < 1.  so it's still fractions under the hood, I'm just hard-coding the denominator to simplify the api

nfurfaro: Ahh, I think I get it. This value seems to get used as the denominator in a few calculations below, so it's just a naming thing.

NickCuso:
Gas savings are for the our deployment and lock owner, not users.  It's less data when deploying the template and less data to update config.  But the gas savings are not significant.

I thought to do this when our team had questions about the numerator/denominator thing.  I'm hoping basis points is more natural / intuitive for people.

NickCuso:
Oh, it is cheaper for users too!

That is because a constant is cheaper to access than a variable.

Once we get codechecks added the diff will be easier to see.

PR #4990, opened 2019-10-10
Assert your identity from the /identity page

cnasc:
# Description

This PR is the other half of #4987. It is probably best to hold off on review until they are more complete.

This PR introduces a page, `/identity` which allows someone to generate a QR code asserting that they own a given address. This will be useful, for example when checking people into events.

At the moment it's quite basic, but it prompts for signature when they click the button and then renders the appropriate QR code. There are some TODOs around here:
- [x] prompt for login when there is no address
- [ ] slick up the styling a bit

image

# Issues

Fixes #
Refs #

# Checklist:


unlock-app/src/middlewares/walletMiddleware.js
```diff
@@ -218,6 +219,21 @@ const walletMiddleware = config => {
               action.price
             )
           })
+        } else if (action.type === SIGN_DATA) {
+          const { data } = action
+          walletService.signDataPersonal(
+            '', // account address -- unused in walletService
```
akeem: What would this be used for in the other cases? This is not on you but kind of weird

PR #4994, opened 2019-10-11
Update website to work with Node 10

NickCuso:
# Description

Before this PR, attempting to run `npm i` would fail if you have Node 10 and not Node 8 specifically.  The package change allows `npm i` to execute.

Then `npm run dev` fails with a callback error.  Switching to use `writeFileSync` corrects that issue.

Now it works for me, even on Windows \o/

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-protocol.com/package.json
+++ b/unlock-protocol.com/package.json
@@ -46,10 +46,6 @@
     "@svgr/cli": "4.3.2",
     "nodemon": "1.19.2"
   },
-  "engines": {
-    "node": "8.16.1"
-  },
-  "engineStrict": true,
   "scripts": {
     "predev": "npm run build-blog",
     "dev": "nodemon src/server.js",
--- a/unlock-protocol.com/package.json
+++ b/unlock-protocol.com/package.json
@@ -46,6 +46,10 @@
     "@svgr/cli": "4.3.2",
     "nodemon": "1.19.2"
   },
+  "engines": {
+    "node": "10.16.3"
+  },
+  "engineStrict": true,
   "scripts": {
     "predev": "npm run build-blog",
     "dev": "nodemon src/server.js",
--- a/docker/unlock-core.dockerfile
+++ b/docker/unlock-core.dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.16.1-alpine
+FROM node:10.16.3-alpine
 LABEL Unlock <ops@unlock-protocol.com>
 
 # Adding bash which is missing from the based alpine image
--- a/locksmith/Dockerfile
+++ b/locksmith/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.16.1-alpine
+FROM node:10.16.3-alpine
 LABEL Locksmith <ops@unlock-protocol.com>
 LABEL maintainer="ops@unlock-protocol.com"
 
--- a/paywall/package.json
+++ b/paywall/package.json
@@ -97,7 +97,7 @@
     "not op_mini all"
   ],
   "engines": {
-    "node": "8.16.1"
+    "node": "10.16.3"
   },
   "engineStrict": true
 }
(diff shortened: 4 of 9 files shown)


unlock-protocol.com/package.json
```diff
@@ -46,10 +46,6 @@
     "@svgr/cli": "4.3.2",
     "nodemon": "1.19.2"
   },
-  "engines": {
-    "node": "8.16.1"
```
julien51: I think we need to keep a hardcoded version in there so we have some "locked" version to make sure things run well!

NickCuso: hmm, can we lock it to the current LTS version then?  10.16.3

cnasc:
I support this, and I agree that we should put the latest LTS in the `engines`.

As an aside, I've been running with v10 for the last few days and haven't noticed any problems, though I haven't run all the apps with it. We should update all the `package.json`s at the same time.

cnasc:
(and if you run into any issues updating the other apps, I'm more than happy to help remediate)

NickCuso:
Updated packages but have not tested the other apps yet. 

I've tried unlock-protocol.com and smart-contracts so far.

NickCuso:
`unlock-core.dockerfile` starts with `FROM node:8.16.1-alpine`, I assume we should update this as well but unsure where the `-alpine` came from..

cnasc:
> `unlock-core.dockerfile` starts with `FROM node:8.16.1-alpine`, I assume we should update this as well but unsure where the `-alpine` came from..

https://github.com/nodejs/docker-node/blob/a9c583095d4cf08bbd68f570a1f9a99780820351/10/alpine/Dockerfile

I think you just have to update it with the relevant Alpine Linux based node dockerfile

NickCuso:
Ah that's a linux flavor.  Okay updated.

Little worried about CI's fail fast here.  Anyway to force all tests to run?

cnasc:
> Ah that's a linux flavor. Okay updated.
> 
> Little worried about CI's quick-pass here. Anyway to force all tests to run?

Looks like it's running ok, what do you mean?

NickCuso:
```
#!/bin/bash -eo pipefail
scripts/monorepo.sh unlock-js $CIRCLE_SHA1 $CIRCLE_BRANCH
No change in unlock-js, skipping job.
```

That's not much reassurance ;)

This is from `monorepo.sh` but I'm not much of a scripter.  Does not look like there is any workaround, but we should add one.  Is it possible to say something like `if title contains FORCE`?

NickCuso:
Hmm alternatively I could add a `.forceBuild` file to each directory which contains a number.  Anytime we want to ensure tests are run for a certain app we increment the number.

cnasc:
> ```
> #!/bin/bash -eo pipefail
> scripts/monorepo.sh unlock-js $CIRCLE_SHA1 $CIRCLE_BRANCH
> No change in unlock-js, skipping job.
> ```
> 
> That's not much reassurance ;)
> 
> This is from `monorepo.sh` but I'm not much of a scripter. Does not look like there is any workaround, but we should add one. Is it possible to say something like `if title contains FORCE`?

🤔 Hm, unlock-js doesn't have any files changed in this PR, but it looks like tests were also skipped for projects that do

NickCuso:
That part is cause it's based on the last specific commit, it should probably change to compare to master instead.

My concern is that this rule considers the change to docker/unlock-core.dockerfile a no-op.

cnasc:
> My concern is that this rule considers the change to docker/unlock-core.dockerfile a no-op.

I see what you mean. Let me take a look at the script

cnasc:
It looks like we don't have direct access to the PR title there, and I'm not familiar enough with the surrounding CI context to know how to easily do it.

One option we have is to merge, since CI always runs all tests on master builds. If we see any failures we can back them out quickly.

Issue #4949, opened 2019-10-05
User Account flow purchase improvements


julien51:
The current user account flow purchase is the following: (as tested [on this site](https://www.ouvre-boite.com/members/)):

## Visit locked paged
<img width="1206" alt="image" src="https://user-images.githubusercontent.com/17735/66251399-b7671500-e78a-11e9-9c01-dd157d1ee0c0.png">

## Click on "Become a member"
<img width="1193" alt="image" src="https://user-images.githubusercontent.com/17735/66251407-cbab1200-e78a-11e9-8338-f7127d435170.png">

On that page, we show both options, which *may* be an issue.  We should show "Balance Insufficient" for the Eth option because user accounts are *just* for DAI lock.

## Click on the **Ether** option 
We rightfully show the login page 
<img width="1187" alt="image" src="https://user-images.githubusercontent.com/17735/66251454-3fe5b580-e78b-11e9-8c11-5ee27d36063c.png">

One the log in is successful, even when the user has enter their payment info, we *briefly* show the credit card form (probably because the app does not know that the credit card has already been added. We should avoid that and only show it if the user does not have a CC.

## After that we show the purchase confirmation 
<img width="655" alt="image" src="https://user-images.githubusercontent.com/17735/66251507-c1d5de80-e78b-11e9-922a-8bf9050b1c22.png">
As you can see on the screenshot here, since I picked the Ether lock, we show a purcahse for it, which will fail.

<img width="637" alt="image" src="https://user-images.githubusercontent.com/17735/66251518-d87c3580-e78b-11e9-940b-7bf428c40da3.png">

I suggest that, after the user has been logged in, rather than moving on with the purchase we send the user _back_ to the "pick your lock" screen, with a notice at the top like "You're now logged in!"

This way, we solve multiple problems:
1. we will show "insufficient funds" on the ether lock
2. the user can *only* pick the DAI lock (we could maybe even show insufficient funds for any lock that locksmith did not approve, but we would probably need an endpoint for that? cc @akeem )
3. we will not show the CC form, because by the time the user clicks again on the right lock we will likely have loaded that info

Note: this flow can be "tested" by closing the modal after signing in and re-opening it by clicking again on the "Become a member" button.

julien51:
What do you think @cnasc ?

It might be a little bit "weird" to redirect the user back to the lock selecting phase. If that's the case, then i suggest showing "Log in" instead of "Purchase" under the locks until the user is logged in, so that the process becomes clearer.

[...]

PR #5014, opened 2019-10-14
Make sure to zero out eth balance for logged out users

cnasc:
# Description

In #4949, we noticed that the [null ethereum account](https://etherscan.io/address/0x0000000000000000000000000000000000000000) that we use as the default for users who are not logged in actually has an enormous eth balance, which allowed people to initiate key purchases on eth locks that would fail immediately. This PR makes sure we intercept that balance and zero it out so that the eth lock is always disabled for user accounts.

This PR makes an assumption that eth should always be zeroed and all other tokens should receive the default stablecoin balance constant, under the assumption that locks can only be created with stablecoins that we allow purchasing on (presently, DAI). If that is not correct, we should ensure that we zero everything except for approved tokens.

# Checklist:

### DIFF
--- a/paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
+++ b/paywall/src/__tests__/unlock.js/CheckoutUIHandler.test.ts
@@ -7,6 +7,7 @@ import { KeyResults } from '../../data-iframe/blockchainHandler/blockChainTypes'
 import IframeHandler from '../../unlock.js/IframeHandler'
 import { PostMessages, ExtractPayload } from '../../messageTypes'
 import { DEFAULT_STABLECOIN_BALANCE } from '../../constants'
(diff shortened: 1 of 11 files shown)


cnasc:
@akeem @julien51 PTAL, I have updated this to zero out every balance except for the expected token.

Token addresses used:
- dev, test: `0x591AD9066603f5499d12fF4bC207e2f577448c46` (taken from my `.env.dev.local`)
- staging: [0x8f2e097E79B1c51Be9cBA42658862f0192C3E487](https://rinkeby.etherscan.io/address/0x8f2e097e79b1c51be9cba42658862f0192c3e487)
- prod: [0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359](0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359)

PR #5022, opened 2019-10-15
rather than using the commit, use the diff to master

### LINKED ISSUE #5007: monorepo.sh: CI should build if diff from master (vs previous commit)
The current configuration for `monorepo.sh` will skip a CI pass if there were no changes in the commit.  The trouble is CI also automatically cancels builds when another commit is added to a PR.  This can lead to completely missing a test pass that you'd expect was included.  It's easy to miss, looking at the CI build times is the easiest way to know this might have happened.

Instead of checking if the commit has changes, check if the diff vs master has changes.

**Additional context**

julien51:
# Description

this should let us detect changes when there are multiple commits in a branch.

Fixes #5007

# Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - checkout
       - run:
-          command: scripts/monorepo.sh unlockprotocol $CIRCLE_SHA1 $CIRCLE_BRANCH
+          command: scripts/monorepo.sh unlock-protocol.com $CIRCLE_BRANCH
       - run:
           name: Build unlock-protocol-com Image
           command: scripts/build-image.sh unlock-protocol-com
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     steps:
       - checkout
       - run:
-          command: scripts/monorepo.sh unlockprotocol $CIRCLE_SHA1 $CIRCLE_BRANCH
+          command: scripts/monorepo.sh unlock-protocol.com $CIRCLE_BRANCH
```
julien51: `unlockprotocol` was never matching anything!

cnasc: Good catch!

cnasc:
> This will be slightly less performant because if the currently pushed branch is not rebased against master, t may run test which would not be required. Maybe a later improvement would be to diff against the earliest diverging commit?

I'd wager that most of our PRs have a small diff against master, though as a best-practices sort of thing we should definitely periodically rebase the long-running PRs to keep them roughly up to date.

julien51:
Dammed, killinhg the job made tests fails. I'll rereun!

PR #5029, opened 2019-10-15
Add a modal for QR codes

cnasc:
# Description

Split from #5024, this PR adds a modal to hold QR codes. You pass the action with the string representing the value that the QR code evaluates to.

Mobile
image

Desktop
image

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/actions/fullScreenModals.test.ts
+++ b/unlock-app/src/__tests__/actions/fullScreenModals.test.ts
@@ -3,10 +3,8 @@ import {
   DISMISS_MODAL,
   waitForWallet,
   dismissWalletCheck,
-  promptForPassword,
-  dismissPasswordPrompt,
-  promptForResetPassword,
-  dismissResetPasswordPrompt,
+  displayQR,
+  dismissQR,
 } from '../../actions/fullScreenModals'
 import { KindOfModal } from '../../unlockTypes'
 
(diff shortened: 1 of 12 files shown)


julien51:
Let's ask @smombartz to look at the design later...

unlock-app/src/__tests__/actions/fullScreenModals.test.ts
```diff
@@ -31,43 +29,24 @@ describe('FullScreenModals actions', () => {
     expect(dismissWalletCheck()).toEqual(expectedAction)
   })
 
-  it('should create an action to prompt for password', () => {
```
julien51: Did you mean to replace these tests?

unlock-app/src/actions/fullScreenModals.ts
```diff
@@ -5,9 +5,10 @@ export const DISMISS_MODAL = 'fullScreenModal/DISMISS'
 
 // General events, in terms of which other kinds of modal interaction should be
 // defined.
-export const launchModal = (kindOfModal: KindOfModal) => ({
+export const launchModal = (kindOfModal: KindOfModal, data?: any) => ({
```
julien51: The previous modals were gettin their data from actions. do you think we could keep that approach for consistency?

unlock-app/src/__tests__/actions/fullScreenModals.test.ts
cnasc: Yes, those are dead code

unlock-app/src/actions/fullScreenModals.ts
cnasc: The other ones always rendered the same thing no matter what, and communicated the bundles through the `dispatch` prop passed in to the modal template (i.e., no callback, just putting the credential payload on the redux chain). This one is a bit different since the template itself needs the value for the QR code.

I agree that my approach here is a little unclean. It's because the original modal infrastructure I implemented here was to have many kinds of static modals. Since right now in practice we have only 2 kinds, it might be better to have distinct actions instead of one parent action with different kinds of modals attached.

cnasc:
> I think this is good but it is also a bit too dense to feel confident. Please make sure you do manual testing for edge cases if you can think of any.

I'd rather not force it through if you're not comfortable on the approach here. What in particular raises your concern?

PR #5030, opened 2019-10-15
Update reducer to track multiple signatures at the same time

cnasc:
# Description

This PR updates the signature handling code in `unlock-app` to be able to store signatures for multiple things at a time. Split off from #5024.

# Checklist:


cnasc:
> I am a bit worried about the `id` (and collisions) but let's not go too far down that rabbit hole!

In this case the `id` is the lock address associated with the key. I _think_ that in the event someone managed to have 2 separate keys to the same lock it would still be fine because from the verifier's perspective they just have _a_ valid key to the lock, not any particular one. It's something we can look into in the future though.

PR #5039, opened 2019-10-17
Frontend component for account ejection

cnasc:
# Description

This draft PR is the component for ejection on the frontend.

**Status**: nearly complete. Just need to send the ejection notice to locksmith when the button is clicked.

image

image

# Checklist:


julien51:
Great! 
So I think the first message should show a bullet list of "what is going to happen".

```
Exporting your account will eject it from the Unlock platform. The following things will happen:
1. A JSON file will be downloaded to your computer. This file includes your private key, encrypted with your Unlock password. You should then import it into a wallet like MetaMask.
2. Your Unlock account will be deleted permanently, including all personal information.
3. After that, you will not be able to log back in with Unlock and the recovery link we sent by email will not work anymore either. 
```

Then, inside the "Danger zone", we should write a much shorter (in red)

```
Clicking the eject button below is final. There is no way back after that.  Do not delete the JSON file until you have safely imported it in another crypto wallet.
```

Issue #5055, opened 2019-10-18
Integrate ENS resolver


julien51:
When showing an address on the front end (paywall or dashboard), we could use the ENS to resolve the domain (if any) which would be a significant UI improvement over the `0x...` address.

If we do this the ENS folks will likely include us on their website which should bring some visibility.

PR #5065, opened 2019-10-21
adding ethers as it is required to use ens resolution

julien51:
# Description

This weeks integration will be adding support for ENS and 3box (maybe).
these libraries need to make their own calls to providers... which means we need to have Etherjs in the project directly (rather than wrapped inside of unlockjs)

See #5055

# Checklist:


julien51:
It also makes me think we can actually move that provider logic of off unlock-js at some point to make it behave like other libraries?

cnasc:
> It also makes me think we can actually move that provider logic of off unlock-js at some point to make it behave like other libraries?

Definitely possible, @akeem and I had been talking about doing that

Issue #5026, opened 2019-10-15
Duplicate tokenID assigned


NickCuso:
**Describe the bug**
We have a duplicate tokenID https://etherscan.io/tx/0x6d1788fe468c1e548bf68db1f04317d05d3ec86ad4abb0679aec94264bf3dcfc vs https://etherscan.io/tx/0xec1ae3c322380918e274f6a4ca3baafb14a481f939dee8539bf7309505f4b5fb

**To Reproduce**
Unclear ATM

**Expected behavior**
They should be unique IDs.

**Screenshots**
![image](https://user-images.githubusercontent.com/14855515/66863514-01e55000-ef48-11e9-8e72-ff290a52a436.png)

**Additional context**

NickCuso:
https://etherscan.io/token/0xb0114bbdce17e0af91b2be32916a1e236cf6034f?a=0xdb92c096bc5efa8adb48f05cd601dddb75228203

I owned an expired key when I called `purchaseFor` again, shortly after "pat" called `purchaseFor`.

NickCuso:
hmm so did pat's account: https://etherscan.io/token/0xb0114bbdce17e0af91b2be32916a1e236cf6034f?a=0xfacafb5baefcd4cf95a72ca88bcefc4ba12a0aa7

nfurfaro:
Hmm. It looks like "Pat" (0xfacafb5baefcd4cf95a72ca88bcefc4ba12a0aa7) purchased a key about 4 hrs after you Nick...but yes, it seems you both already had an expired key.

nfurfaro:
So here's one thing I found that needs to be looked at. in `MixinKeys.sol`(for V4 as well as the current code) `_assignNewTokenId` looks like this:
```
/**
   * Assigns the key a new tokenId (from totalSupply) if it does not already have
   * one assigned.
   */
  function _assignNewTokenId(
    Key storage _key
  ) internal
  {
    if (_key.tokenId == 0) {
      // This is a brand new owner, else an owner of an expired key buying an extension.
      // We increment the tokenId counter
      totalSupply++;
      // we assign the incremented `totalSupply` as the tokenId for the new key
      _key.tokenId = totalSupply;
    }
  }
```
I believe the comments is flawed...`an owner of an expired key buying an extension` will have a tokenId which is not `0`, and therefore just needs an extension of `expirationDuration`, not an incremented `tokenId`.
I will add this to a TODO list here for reference.

PR #5069, opened 2019-10-21
WIP: Duplicate tokenIDs Bug

nfurfaro:
# Description

I wrote this to try to reproduce a bug we found where duplicate tokenID's seemed to be getting created. I did manage to repro the bug, as well as to confirm that it is no longer an issue for post-v4 Locks.
I still don't know what the root cause of the issue is, but here's a description of what is happening.
So far, the only scenario that results in duplicate (but not really) tokenIds seems to be:
(let's say the tokenID is `46`)
- 2 key owners, each with an expired key
- both of them repurchase another key, one after the other. 
- the Transfer event emitted for both shows the tokenId as being `46` (this is the duplicate tokenID we're seeing)
- neither person has their actual tokenID modified. That is, calling `getTokenIdFor()` either one will return the tokenId of the first key they purchased. This is by design. 

So, the transfer event is emitting the wrong tokenID! 
In the current version of Lock code, this doensn't happen, and the actual tokenID matches the emitted event arg `_tokenID`

# Issues

Fixes #
Refs #5026 

# Checklist:

### DIFF
--- a/smart-contracts/test/Unlock/upgrades/v4ToLatest.js
+++ b/smart-contracts/test/Unlock/upgrades/v4ToLatest.js
@@ -154,7 +154,7 @@ contract('Unlock / upgrades', accounts => {
       })
 
       it('re-purchasing 2 expired keys from V4 should not duplicate tokenIDs', async () => {
-        let keyOwners = [keyOwner3, keyOwner4]
+        let keyOwners = [keyOwner3, keyOwner4, accounts[5], accounts[6]]
         const purchases = keyOwners.map(account => {
           return lockV4.methods.purchaseFor(account).send({
             value: keyPrice,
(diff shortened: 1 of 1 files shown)


NickCuso:
Good findings.

Here's the bug: https://github.com/unlock-protocol/unlock/blob/b7220ca3c898642e493035c89681fba3bce5b8d9/smart-contracts/contracts/mixins/MixinPurchase.sol#L115

NickCuso:
And here's the fix https://github.com/unlock-protocol/unlock/blob/master/smart-contracts/contracts/mixins/MixinPurchase.sol#L89

nfurfaro:
closing this as it is not needed anymore. It confirmed the bug is no longer an issue.

Issue #5060, opened 2019-10-21
Add ability to send a signed key (ticket) code via email to keychain


julien51:
The keychain now provides the ability for a key owner to sign their own key and shows a QR code.
We should let the user send this QR code to themselves by email to provide parity with the tickets feature.

cnasc:
I'm going to create a stub verification page just so the wedlocks email has something to point to, if that's alright

PR #5075, opened 2019-10-21
Add QR modal based on InlineModal

cnasc:
# Description

This PR adds a QR modal to be used on the keychain in the very near future. It includes an email form that will enable us to send an email containing the QR code to users. 

image

image

image

# Issues

Refs #5060 

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/components/interface/keyChain/QRModal.test.tsx
+++ b/unlock-app/src/__tests__/components/interface/keyChain/QRModal.test.tsx
@@ -1,0 +1,15 @@
+import React from 'react'
+import * as rtl from 'react-testing-library'
+import QRModal from '../../../../components/interface/keyChain/QRModal'
+
+describe('QRModal', () => {
+  it('should render a QR code', () => {
+    expect.assertions(1)
+    const dismiss = jest.fn()
+    const { container } = rtl.render(
+      <QRModal active dismiss={dismiss} value="just some test data" />
+    )
+
+    expect(container.querySelector('canvas')).not.toBeNull()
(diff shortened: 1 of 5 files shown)


cnasc:
Momentarily blocked by the failure on `canvas.getContext` in the tests. Fix inbound in separate PR.

cnasc:
Alright, needs to be rebased on #5079, but this is ready for review

unlock-app/src/__tests__/components/interface/keyChain/QRModal.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  })
+
+  it('should have a disabled submit button when input is empty', () => {
+    expect.assertions(1)
+    const { getByText } = render()
+
+    const submit = getByText('Send Email')
+    rtl.fireEvent.click(submit)
+
+    expect(sendEmail).not.toHaveBeenCalled()
```
akeem: In this test is there a way to see if the button has a disabled state? From first glance it looks like the test does ensure that it will not fire off an email but I guess it may still be enabled?

cnasc: In the code, there are 2 buttons and we choose which to render based on the validity of the input field. One button is hooked up to the submit function, and one is not. So this test demonstrates (in conjunction with the other test that shows the function _does_ get called when the input is filled) which of the buttons renders when.

There is a more direct way to identify the buttons, by giving each a `data-testid` that we can query against. That's less idiomatic though.

PR #5087, opened 2019-10-22
Adds evaluation method to middleware

akeem:
# Description

As part of our need to allow lock owners the ability to view additional token metadata, we are providing a non-error returning path to the middleware. The general gist is that signature concerns will stay here and users/consumers will now have two paths.

1. Reject when appropriate
2. All requesters further downstream for conditional behavior.

**The tests have been shifted to typescript and will be consolidated in an additional pass**

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/middlewares/signatureValidationMiddleware.ts
+++ b/locksmith/src/middlewares/signatureValidationMiddleware.ts
@@ -96,6 +96,30 @@ namespace SignatureValidationMiddleware {
       }
     }
   }
+
+  export const evaluateSignature = (
+    configuration: SignatureValidationConfiguration
+  ): any => {
+    return (req: any, _res: Response, next: any) => {
+      var signature = extractToken(req)
+
+      if (signature === null) {
+        next()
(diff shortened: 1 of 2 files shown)


locksmith/src/middlewares/signatureValidationMiddleware.ts
```diff
@@ -96,6 +96,30 @@ namespace SignatureValidationMiddleware {
       }
     }
   }
+
+  export const evaluateSignature = (
```
cnasc: Since this returns a function that evaluates a signature, rather than evaluating the signature itself, what do you think about renaming this to something like `makeSignatureEvaluator` or something similar? LG otherwise

akeem: sounds good, will do

cnasc: I'll approve pending that change

PR #5092, opened 2019-10-22
When presented with a signature proving lock ownership expose protected information

akeem:
# Description

In expanding on previous work, we are opening the previously shielded information for this that pass the new criteria.

The next step in supporting the feature will be to ensure that protected information is encrypted.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/controllers/metadataController.ts
+++ b/locksmith/src/controllers/metadataController.ts
@@ -1,4 +1,4 @@
-import { Request, Response } from 'express-serve-static-core' // eslint-disable-line no-unused-vars, import/no-unresolved
+import { Response } from 'express-serve-static-core' // eslint-disable-line no-unused-vars, import/no-unresolved
 import Normalizer from '../utils/normalizer'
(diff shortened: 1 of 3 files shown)


locksmith/src/controllers/metadataController.ts
```diff
@@ -10,11 +10,11 @@ const config = require('../../config/config')[env]
 const metadataOperations = require('../operations/metadataOperations')
 
 namespace MetadataController {
-  export const data = async (req: Request, res: Response): Promise<any> => {
+  export const data = async (req: any, res: Response): Promise<any> => {
```
cnasc: Why did this type signature change?

akeem: Previously we were using the standard Request type here, in the new filter that this sits in front of this i've add a `signee` if there is one.

It might be worthwhile to make a type for this in the future but i'd rather let it soak a bit to see if anything changes after everything is in place

PR #5093, opened 2019-10-22
Make tests pass without automine enabled

### LINKED ISSUE #2466: Run smart contract tests without automine
Our smart contract test suite currently assumes that the nodes are running with "auto-mining" on (each transaction is immediately mined when the node receives it). 
We should remove that assumption and ensure that the smart contract test suite actually runs on any network (not just ganache).

nfurfaro:
# Description
- I adjusted some of the time-based parameters for the failing tests. Most of the issues came from the fact that without automine, we need to wait after certain transactions for the block to be mined before testing to see if certain values have been set on the contract, etc...
As different networks will have different blocktimes, I tried to set worst-case values so they work with mainnet (and so should work with any other network too). 
- I refactored out the `sleep()` function that Nick added into its own module and made more extensive use of it to wait for mining. 
- I tested the entire smart-contract test suite with a blocktime of 3 seconds, and tested individual files with a blocktime of 14 seconds successfully.

# Issues

Fixes #2466
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/test/Lock/freeTrial.js
+++ b/smart-contracts/test/Lock/freeTrial.js
@@ -35,7 +35,7 @@ contract('Lock / freeTrial', accounts => {
     let initialLockBalance
 
     beforeEach(async () => {
-      await lock.updateRefundPenalty(5, 2000)
+      await lock.updateRefundPenalty(20, 2000)
       initialLockBalance = new BigNumber(
         await web3.eth.getBalance(lock.address)
       )
@@ -58,7 +58,7 @@ contract('Lock / freeTrial', accounts => {
 
     describe('should cancel and provide a partial refund after the trial expires', () => {
       beforeEach(async () => {
-        await sleep(6000)
+        await sleep(21000)
         await lock.cancelAndRefund({
           from: keyOwners[0],
         })
--- a/smart-contracts/test/Lock/freeTrial.js
+++ b/smart-contracts/test/Lock/freeTrial.js
@@ -2,7 +2,7 @@ const Units = require('ethereumjs-units')
 const BigNumber = require('bignumber.js')
 
 const deployLocks = require('../helpers/deployLocks')
-const sleep = require('../helpers/sleep')
+const { sleep, sleepTime } = require('../helpers/sleep')
 
 const unlockContract = artifacts.require('Unlock.sol')
 const getProxy = require('../helpers/proxy')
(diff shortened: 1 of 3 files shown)


nfurfaro:
I have hardcoded sleep values here, but we could potentially check to see which network the tests are being run on and set values accordingly. Also, even with hardcoded values, we may be able to go lower than what I have (mostly 20 sec). I wanted to be a bit generous to be on the safe side, but mainnet blocktime is only around 14 seconds I believe.

julien51:
Let's extract that 2000 wait value in a single place so we can change it all over in a single change?

NickCuso:
hmm I don't understand why only two tests are impacted?  what's unique about those?

nfurfaro:
This is a good point, and I don't have a good answer yet. I need to look at this some more.

julien51:
Can you actually change the config to disable automine on this PR to make sure it does the trick?

nfurfaro:
>Can you actually change the config 

Yes, i can and will. Note that a block-time of 3 seconds on my machine took close to an hour to run the test suite.

julien51:
@nfurfaro I think this is good then! @NickCuso what do you think?

NickCuso:
I'm not sure I understand the changes here.  But if it works that's fine, we can try and maybe simplify later.

How about we set CI to use a 1-second block time for tests going forward?

nfurfaro:
@julien51 RE:
> Can you actually change the config to disable automine on this PR to make sure it does the trick?

I didn't do this yet...It looks like CI is already using a block time of 3 seconds? ( which is what I was testing with locally).

NickCuso:
I don't believe CI is using 3 seconds for the smart contract tests.  We would have noticed much longer test times, right?

nfurfaro:
You're right, I went digging through the config and was looking at docker/development/standup.sh, but yes, we would definitely notice a blocktime of 3 secs, its painfully slow. Even 1 sec would be slow, but might be worth it.

NickCuso:
I think it's this line https://github.com/unlock-protocol/unlock/blob/0cc647b92f6a2d23fd0e6e75486a41c9911cfbf0/docker/docker-compose.ci.yml#L8

nfurfaro:
Update: Not ready to merge this, as there's now 1 newly-introduced  test failing.

nfurfaro:
Closing, replaced by #5472

PR #5098, opened 2019-10-23
not catching errors anymore when sending transactions

julien51:
# Description

The way we catch errors is really bad because if anything happens "downstream" then the error was caught by these blocks completely obfuscating what really happened.
Also I believe the application should be the one acting on these errors.

# Checklist:


cnasc:
> Also I believe the application should be the one acting on these errors.

Looks good overall, and I agree that we shouldn't hide this information from the consuming application

PR #5148, opened 2019-10-29
Add gasLimit and gasPrice overrides to external refund

cnasc:
# Description

While testing before the external refund deploy, we noticed that web3 wallets would error when automatically calculating gas prices and limits for the refund transaction. This PR fixes that by specifying concrete values (based on @akeem's successful manual transaction [here](https://rinkeby.etherscan.io/tx/0xb12aba624674a47734b30c5e0b81ceafb9049470341f5916a8920c6e1c1242c1))

Additionally, because there is some lag when running on a real network, this PR makes the UI respond faster by disabling the button right away instead of waiting for the transaction promise to resolve.

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/components/interface/RefundButton.test.tsx
+++ b/unlock-app/src/__tests__/components/interface/RefundButton.test.tsx
@@ -42,7 +42,10 @@ describe('RefundButton', () => {
 
     rtl.fireEvent.click(refundButton)
 
-    expect(refundFunction).toHaveBeenCalledWith('0xdeadbeef')
+    expect(refundFunction).toHaveBeenCalledWith('0xdeadbeef', {
+      gasLimit: 600000,
+      gasPrice: 20000000000,
+    })
 
     // Button text changes after click
     await findByText('Refund Initiated')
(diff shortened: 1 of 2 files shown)


NickCuso:
You are currently sending almost 10x what it needs.  That should be fine.

Alternatively you could use https://web3js.readthedocs.io/en/v1.2.2/web3-eth-contract.html#methods-mymethod-estimategas
That attempts to return the exact gas required.  The value may be incorrect if another transaction is mined before yours and it changed some significant state.  Or it may be incorrect due to a simple bug (they found a gas estimation bug just a couple months ago).

cnasc:
> You are currently sending almost 10x what it needs. That should be fine.
> 
> Alternatively you could use https://web3js.readthedocs.io/en/v1.2.2/web3-eth-contract.html#methods-mymethod-estimategas
> That attempts to return the exact gas required. The value may be incorrect if another transaction is mined before yours and it changed some significant state. Or it may be incorrect due to a simple bug (they found a gas estimation bug just a couple months ago).

Thanks @NickCuso! We don't have web3js in `unlock-app` anymore, so I'm comfortable just leaving it with a very high limit.

NickCuso:
That API is offered by the node - `ethers` should have a way to hit that too.  But yea, a high value seems fine here.

PR #5155, opened 2019-10-30
WIP: adding timeMachine function

nfurfaro:
# Description
The idea here is to add a simple function capable of increasing or decreasing the `expirationTimestamp` for a given key by a given amount. This could be used for taking a time-based fee from a key, as well as in partial transfers to reduce the original keys time, and increase the new keys time.(this last use case is maybe not needed, and this could be simplified to only reduce time from a key).
- I've made it public for ease of testing, but it should be made internal.
- access control should be handled by the function which calls `timeMachine`
- tests are not complete, I just want to get another opinion on this approach.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -202,7 +202,7 @@ contract MixinTransfer is
     address _owner,
     uint256 _deltaT,
     bool _addTime
-  ) public
+  ) internal
     hasValidKey(_owner)
   {
     Key storage key = keyByOwner[_owner];
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -189,6 +189,32 @@ contract MixinTransfer is
   }
 
   /**
+  * @notice Modify the expirationTimestamp of a key
+  * by a given amount.
+  * @param _owner The owner of the key to modify
+  * @param _deltaT The amount of time in seconds by which
+  * to modify the keys expirationTimestamp
+  * @param _addTime Choose whether to increase or decrease
+  * expirationTimestamp (false == decrease, true == increase)
+  * @dev Throws if owner does not have a valid key.
+  */
+  function timeMachine(
+    address _owner,
+    uint256 _deltaT,
+    bool _addTime
+  ) internal
+    hasValidKey(_owner)
+  {
+    Key storage key = keyByOwner[_owner];
+    uint formerTimestamp = key.expirationTimestamp;
+    if(_addTime) {
+      key.expirationTimestamp = formerTimestamp.add(_deltaT);
+    } else {
+      key.expirationTimestamp = formerTimestamp.sub(_deltaT);
+    }
+  }
+
+  /**
    * @dev Internal function to invoke `onERC721Received` on a target address
    * The call is not executed if the target address is not a contract
    * @param from address representing the previous owner of the given token ID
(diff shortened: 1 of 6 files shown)


NickCuso:
I see what you are trying to do, but I do not like the idea of exposing this publicly especially without an onlyOwner modifier.

Use a mock instead.  See https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/mocks for examples of the pattern

nfurfaro:
@NickCuso Thanks for the tip on using mocks. I fully agree about not making this public, as per the description:
> I've made it public for ease of testing, but it should be made internal.

Will look to complete this the right way tonight.

nfurfaro:
@NickCuso When you have a bit of time, I'd appreciate a second set of eyes on this. Note that I haven't got things working when `function _timeMachine` is internal yet (all my calls to the function in `timeMachine.js` revert when I try to use a mock, so I still need to solve this).

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -96,7 +99,7 @@ contract MixinTransfer is
       _tokenId
     );
 
-    _chargeAtLeast(fee);
+    // _chargeAtLeast(fee);
```
nfurfaro: It doesn't seem like this would be needed anymore?

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -180,14 +183,48 @@ contract MixinTransfer is
     uint fee;
     if(timeRemaining >= expirationDuration) {
       // Max the potential impact of this fee for keys with long durations remaining
-      fee = keyPrice;
+      fee = timeRemaining;
     } else {
       // Math: using safeMul in case keyPrice or timeRemaining is very large
-      fee = keyPrice.mul(timeRemaining) / expirationDuration;
+      fee = timeRemaining.mul(timeRemaining) / expirationDuration;
```
nfurfaro: I'd like your thoughts on this stuff here.

smart-contracts/test/helpers/deployMocks.js
```diff
@@ -0,0 +1,38 @@
+// const PublicLock = artifacts.require('./PublicLock.sol')
```
nfurfaro: This works for deploying the mock contract, but is ugly and verbose.

smart-contracts/test/Lock/transferFee.js
```diff
@@ -36,82 +36,82 @@ contract('Lock / transferFee', accounts => {
...
-        it('transfer fee was received by the contract', async () => {
-          const lockBalance = new BigNumber(
-            await web3.eth.getBalance(lock.address)
-          )
-          assert.equal(
-            lockBalance.toFixed(),
-            lockInitialBalance.plus(estimatedTransferFee).toFixed()
-          )
-        })
+        // it.skip('transfer fee was paid by the keyOwner', async () => {
```
nfurfaro: `timeMachine.js` tests that the fee is paid correctly already.

smart-contracts/test/Lock/transferFee.js
```diff
@@ -36,82 +36,82 @@ contract('Lock / transferFee', accounts => {
...
+        //   assert.equal(
+        //     keyOwnerBalance.toFixed(),
+        //     keyOwnerInitialBalance
+        //       .minus(transferGasCost)
+        //       .minus(estimatedTransferFee)
+        //       .toFixed()
+        //   )
+        // })
+
+        // it.skip('transfer fee was received by the contract', async () => {
```
nfurfaro: No longer relevant.

nfurfaro:
closing in favor of a few smaller PR's inbound.

PR #5158, opened 2019-10-30
Migrate unlock-app to @testing-library/react

cnasc:
# Description

For some time we've been hanging on to an old version of `react-testing-library`. I've decided to start moving us along by moving each consuming app to the new package and making the necessary updates to handle breaking changes.

This PR looks huge, but most of it was just changing the imports.

# Checklist:


unlock-app/jest.setup.ts
```diff
@@ -1,7 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import 'jest-dom/extend-expect'
-import 'react-testing-library/cleanup-after-each'
```
cnasc: This import is deprecated under the new library. Also changed to TS here so that we could get type inference for jest-dom

Issue #5161, opened 2019-10-30
Upgrade all react subrepos to use @testing-library/react


cnasc:
We've been a few releases behind. We should incrementally update each subrepo until we've purged the last remnants of `react-testing-library`.

- [x] unlock-app (done in #5158)
- [x] paywall (done in #5164)
- [x] tickets (done in #5162)
- [x] unlock-protocol.com (done in #5163)

PR #5162, opened 2019-10-30
Update tickets to use @testing-library/react

cnasc:
# Description

Much like #5158, this PR makes tickets use the new `@testing-library` packages.

# Issues

Refs #5161 

# Checklist:


tickets/src/__tests__/components/developer/DeveloperOverlay.test.js
```diff
@@ -40,17 +40,17 @@ describe('DeveloperOverlay', () => {
...
     )
 
-    expect(component.queryByText('HTTP')).not.toBeNull()
+    component.getByText('HTTP')
     expect(component.queryByText('Metamask')).not.toBeNull()
   })
 
   it('sets selected provider from prop', () => {
-    expect.assertions(1)
+    expect.assertions(0)
```
julien51: That's a bit strange? I think we should have at leaast one expectation?

cnasc: `getBy<query name>('a string')` fails the test if it doesn't find the element, so it works about the same as `expect(queryBy<query name>('a string')).not.toBeNull()` but doesn't contain an `expect`.

julien51: I understand but it's strange to have this `expect.assertions(0)`
Let's unlblock for now though.

PR #5165, opened 2019-10-31
Remove default gas price for external refund

cnasc:
# Description

This PR removes the default gas price in our interaction with the refund contract so that we can refund people cheaply. We'll have to be careful testing this, the error the defaults fixed only showed up in staging before. The hope is that simply setting the gas limit without the price will still prevent the error.

# Checklist:


julien51:
LGTM!

PR #5174, opened 2019-10-31
Apply renovate updates to paywall

cnasc:
# Description

This PR applies the updates from #5011 to the paywall. It looks like just a single hook test was failing, and deleting a line of code makes the test pass. That doesn't seem like it should be the solution, which reduces my confidence in the existing hooks tests further. We can discuss in detail. The error was `finishFetch is not a function`, and it was actually undefined at this point in the test.

# Checklist:


julien51:
>  reduces my confidence in the existing hooks tests further. 

:( as we refactor paywall stuff let's make sure we use a proper testing tool for these hooks.

PR #5179, opened 2019-10-31
Adds locksmith as part of the development docker-compose setup

akeem:
# Description

This is a piecemeal update, allowing passersby the ability to use the dev setup more or less out of the box. Some additional work will come after ensuring that we are publishing the latest version of the docker image to docker hub to avoid the need to build the image as well.

There are some additional components around managed purchaser that are being explicitly set here as we don't have a mock stripe server setup but that will be a fast follow.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/docker/docker-compose.development.yml
+++ b/docker/docker-compose.development.yml
@@ -50,4 +50,35 @@ services:
     environment:
       POSTGRES_USER: graph-node
       POSTGRES_PASSWORD: let-me-in
-      POSTGRES_DB: graph-node
+      POSTGRES_DB: graph-node    
+  locksmith-migration:
+    build: 
+      context: ../locksmith/
+      dockerfile: Dockerfile
+    command: npm run db:migrate
+    environment: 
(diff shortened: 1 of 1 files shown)


docker/docker-compose.development.yml
```diff
@@ -50,4 +50,35 @@ services:
     environment:
       POSTGRES_USER: graph-node
       POSTGRES_PASSWORD: let-me-in
-      POSTGRES_DB: graph-node
+      POSTGRES_DB: graph-node    
+  locksmith-migration:
```
julien51: Do we need this one too?

akeem: the short answer is yes we need it. In production we have a two step process, migrate then start the application. We do this instead of chaining commands together to ensure that we don't end up with the application unable to server requests due to someone weird going on with a migration.

In the instance above a container will be spun up for migration and will be terminated aftewards

PR #5182, opened 2019-10-31
Update React apps to Next 9

cnasc:
# Description

This PR updates all the React apps to use Next 9. Most of my changes were as found here: https://github.com/zeit/next.js/blob/canary/UPGRADING.md

Some tests failed. In general they were simple fixes. More tests failed in the paywall, but all of the failing tests were for dead code so I just deleted the code and the tests.

# Issues

Fixes #4814 

# Checklist:


cnasc:
Hmm, these all run fine locally. I'll pick this up later and see what's up.

cnasc:
The problem is webpack, but not sure how to resolve

unlock-app/src/pages/_error.js
```diff
@@ -1,34 +1,18 @@
...
 import withConfig from '../utils/withConfig'
 
 class Error extends React.Component {
-  static getInitialProps({ res, err, req }) {
+  static getInitialProps({ res, err }) {
     const statusCode = res ? res.statusCode : err ? err.statusCode : null
 
-    // redirect to home now if not found
+    // why did we previously automatically redirect to home?
+    // at any rate, we cannot use Router here because Router doesn't exist server-side
```
cnasc: We should discuss this

cnasc: Maybe we can just adjust that function to be like the example here: https://stackoverflow.com/a/56928111

julien51: I actually don't think it matters. We can just show an error if not found.

julien51: (note that netlify's logic will do that _before_ it hits our app anyway)

cnasc:
🤷‍♂️ another one that runs fine locally. I don't think updating next is worth spending more time figuring out why the integration tests fail.

cnasc:
I think the integration test failures are caused by a weird redirect I've started seeing when I visit `http://locahost:3000`. It immediately tries to redirect me to the static site (`http://locahost:3002`), which may cause some issues when running the integration tests.

cnasc:
Integration tests fail on the dashboard, it looks like the account never populates. I was able to make this run successfully by manually running the test, not sure what makes it fail when running via the script.

cnasc:
Pinning this for now, I'll incrementally update app-by-app so we only have to deal with `unlock-app` in the final PR

PR #5196, opened 2019-11-01
Enable the use of Yarn for package management

cnasc:
# Description

This PR introduces the ability to use Yarn for package management.

In the root Unlock directory, simply run `yarn` and it will install deps for all apps. (might be worth doing `yarn nuke` first).

On a first install, it takes about 8 minutes to download and install everything. Once the cache is full, subsequent installs take less than 2 minutes.

We'll need to talk about using yarn with our CI and docker images and whatnot in the future, right now this is only present for local dev.

# Checklist:


package.json
```diff
@@ -3,6 +3,24 @@
   "version": "1.0.0",
   "description": "Unlock is a protocol for decentralized access control",
   "main": "index.js",
+  "private": true,
```
cnasc: Repo has to be marked private to enable use of workspaces.

package.json
```diff
@@ -3,6 +3,24 @@
...
+    "rover",
+    "smart-contracts",
+    "tests",
+    "tickets",
+    "unlock-app",
+    "unlock-js",
+    "unlock-protocol.com",
+    "wedlocks"
+  ],
+  "resolutions": {
```
cnasc: See: https://yarnpkg.com/lang/en/docs/selective-version-resolutions/

Basically, this makes everything use the same version of React no matter which it actually specifies.

unlock-protocol.com/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "unlock-app",
+  "name": "unlock-protocol.com",
```
cnasc: Duplicate name made yarn angry

package.json
julien51: Does it have any other consequnce?

unlock-protocol.com/package.json
julien51: 😠

package.json
cnasc: I _believe_ it would only impact publishing this parent module on npm but I will double check.

cnasc: Confirmed, this would only prevent publishing the root-level unlock module. Does not interfere with subrepos.

cnasc:
Just did another push -- it looks like snapshots are different between the yarn and npm environments (just css class names). If that is the case and my recent push didn't fix things, we'll need to update CI to use yarn to move this forward.

cnasc:
This is pretty much dead in the water, things run fine locally but then resolution of the storybook modules fail in the docker image. Not worth burning more time on.

julien51:
snapshots!

cnasc:
> snapshots!

IIRC we have the issue here where snapshots generated with Yarn will be different from the ones done with NPM. Should be able to fix it so that we just get deps with Yarn for now and run the scripts with NPM still

cnasc:
Closing in favor of #5613

PR #5215, opened 2019-11-05
Adding timeMachine function

### LINKED ISSUE #5110: Transfer fee to be paid in time
ATM we charge a transferFee which is defined in terms of the current keyPrice.  

We'd like to switch to reduce the total time remaining on transfer instead.

**Additional context**
With this we can switch some APIs back to non-payable, maybe even start using the ZOS ERC-721 files going forward.

nfurfaro:
# Description
This adds a function `timeMachine` which can be used to add or subtract time from a given key.
It will be used to implement partial transfers, as well as to pay the transfer fee in time.

# Issues

Fixes #5110 
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -216,7 +216,7 @@ contract MixinTransfer is
         key.expirationTimestamp = block.timestamp.add(_deltaT);
       }
     } else {
-      key.expirationTimestamp = formerTimestamp - _deltaT;
+      key.expirationTimestamp = formerTimestamp.sub(_deltaT);
     }
   }
 
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -191,7 +191,7 @@ contract MixinTransfer is
   /**
   * @notice Modify the expirationTimestamp of a key
   * by a given amount.
-  * @param _owner The owner of the key to modify
+  * @param _tokenId The ID of the key to modify.
   * @param _deltaT The amount of time in seconds by which
   * to modify the keys expirationTimestamp
   * @param _addTime Choose whether to increase or decrease
@@ -199,14 +199,16 @@ contract MixinTransfer is
   * @dev Throws if owner does not have a valid key.
   */
   function _timeMachine(
-    address _owner,
+    uint _tokenId,
     uint256 _deltaT,
     bool _addTime
   ) internal
   {
-    Key storage key = keyByOwner[_owner];
+    address tokenOwner = ownerOf[_tokenId];
+    require(tokenOwner != address(0), 'NON_EXISTENT_KEY');
+    Key storage key = keyByOwner[tokenOwner];
     uint formerTimestamp = key.expirationTimestamp;
-    bool validKey = getHasValidKey(_owner);
+    bool validKey = getHasValidKey(tokenOwner);
     if(_addTime) {
       if(validKey) {
         key.expirationTimestamp = formerTimestamp.add(_deltaT);
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -214,11 +214,7 @@ contract MixinTransfer is
         key.expirationTimestamp = block.timestamp.add(_deltaT);
       }
     } else {
-      if(formerTimestamp.sub(_deltaT) <= block.timestamp) {
-        key.expirationTimestamp = block.timestamp;
-      } else {
-        key.expirationTimestamp = formerTimestamp - _deltaT;
-      }
+      key.expirationTimestamp = formerTimestamp - _deltaT;
     }
   }
 
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -203,12 +203,16 @@ contract MixinTransfer is
     uint256 _deltaT,
     bool _addTime
   ) internal
-    hasValidKey(_owner)
   {
     Key storage key = keyByOwner[_owner];
     uint formerTimestamp = key.expirationTimestamp;
+    bool validKey = getHasValidKey(_owner);
     if(_addTime) {
-      key.expirationTimestamp = formerTimestamp.add(_deltaT);
+      if(validKey) {
+        key.expirationTimestamp = formerTimestamp.add(_deltaT);
+      } else {
+        key.expirationTimestamp = block.timestamp.add(_deltaT);
+      }
     } else {
       if(formerTimestamp.sub(_deltaT) <= block.timestamp) {
         key.expirationTimestamp = block.timestamp;
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -208,11 +208,7 @@ contract MixinTransfer is
     Key storage key = keyByOwner[_owner];
     uint formerTimestamp = key.expirationTimestamp;
     if(_addTime) {
-      if(formerTimestamp.add(_deltaT) > block.timestamp.add(expirationDuration)) {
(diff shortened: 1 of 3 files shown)


smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -188,6 +188,40 @@ contract MixinTransfer is
...
+    uint256 _deltaT,
+    bool _addTime
+  ) internal
+    hasValidKey(_owner)
+  {
+    Key storage key = keyByOwner[_owner];
+    uint formerTimestamp = key.expirationTimestamp;
+    if(_addTime) {
+      if(formerTimestamp.add(_deltaT) > block.timestamp.add(expirationDuration)) {
+        key.expirationTimestamp = block.timestamp.add(expirationDuration);
```
NickCuso: Hmm do we really want to cap transfers?  It seems if I own the time I should be able to transfer it all to a different account.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -188,6 +188,40 @@ contract MixinTransfer is
...
+      } else {
+        key.expirationTimestamp = formerTimestamp.add(_deltaT);
+      }
+    } else {
+      if(formerTimestamp.sub(_deltaT) <= block.timestamp) {
+        key.expirationTimestamp = block.timestamp;
```
NickCuso: This may be okay as either way the key is expired, but is this what we are looking for?

For the partial transfer scenario, if this is < block.timestamp then they may be transferring too much.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -188,6 +188,40 @@ contract MixinTransfer is
...
+  * @param _addTime Choose whether to increase or decrease
+  * expirationTimestamp (false == decrease, true == increase)
+  * @dev Throws if owner does not have a valid key.
+  */
+  function _timeMachine(
+    address _owner,
+    uint256 _deltaT,
+    bool _addTime
+  ) internal
+    hasValidKey(_owner)
```
NickCuso: What if the destination has an expired key?
For partial transfer we would want to be able to add time to that account.  But not based on their current expiration - it should use block.timestamp + deltaT instead.

smart-contracts/contracts/mocks/TimeMachineMock.sol
```diff
@@ -0,0 +1,26 @@
+pragma solidity ^0.5.0;
+
+import '../PublicLock.sol';
+
+contract TimeMachineMock is
+  PublicLock
+{
+  /**
+  * @notice Modify the expirationTimestamp of a key
```
NickCuso: Minor: but maybe let's not duplicate comments too much.  They are likely to get out of sync as we make changes.  Since this is just a test artifact I'd say either drop the comment or only comment on the mock specific side of things here.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: Yes, you are right. I will modify this to allow adding an arbitrary amount of time.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: Agreed. My thinking was that the `partialTransfer()` function (not yet introduced) will need to make the appropriate checks to ensure that a key owner can't transfer more than they have. 
This timeMachine function doesn't/shouldn't have to anticipate all use cases, only that it does what it is supposed to do, add or subtract correctly. I currently see at least 2 different uses for it, paying the transfer fee in time, and partial transfers. They may well be more down the road and each will have different requirements. Does this seem reasonable?

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: Yes, agreed. I could use the `getHasValidKey` modifier here to check, and if the key is expired then use block.timestamp + deltaT instead.

smart-contracts/contracts/mixins/MixinTransfer.sol
NickCuso: Yes it does not need to consider everything.  I'm asking about `_addTime = true`

Use case: I want to add x minutes to account y.  It seems that could be useful for various.  It seems we would want that to work instead of fail because account y does not already own a valid key.  But I can understand wanting to keep new key creation logic separate.

NickCuso: lol sorry confusing issues.  this is addTime = false.

NickCuso: Okay so for this one - it seems like a silent fail.  Why would we ever call this with a value which is too large?

smart-contracts/contracts/mocks/TimeMachineMock.sol
nfurfaro: done.

smart-contracts/contracts/mixins/MixinTransfer.sol
NickCuso: Another way to look at it is why is this helpful?  Isn't deleting these lines valid as well?  If so, let's simplify right?

nfurfaro: Gotcha. Just as for adding time, when deducting time, the check on how much time can be deducted could/should be made in the calling function, not here. I think this is what you are saying. To simplify, this can just be:
```
else {
      key.expirationTimestamp = formerTimestamp - _deltaT;
    }
```

nfurfaro:
Ok I've made changes and pushing now

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -188,6 +188,36 @@ contract MixinTransfer is
...
+    uint256 _deltaT,
+    bool _addTime
+  ) internal
+  {
+    Key storage key = keyByOwner[_owner];
+    uint formerTimestamp = key.expirationTimestamp;
+    bool validKey = getHasValidKey(_owner);
+    if(_addTime) {
+      if(validKey) {
+        key.expirationTimestamp = formerTimestamp.add(_deltaT);
```
NickCuso: What if the user has never owned a key previously?  This would not add them to the owners array, for e.g..  Nor issue a transfer event.  

We could maybe add the modifier `ownsOrHasOwnedKey` and then higher level branch to key creation when needed.  Or add key creation logic here.

NickCuso: Hmm just a thought: maybe this function could accept a `Key storage _key` instead of the _owner, clarifying that the caller is responsible for ensuring the key exists.

nfurfaro: Yep, This last suggestion seems like a good fit. I was thinking that the caller should be responsible for ensuring the key exists, but this would make that explicit.

nfurfaro: Hmm. What about pasing `uint _tokenId` as a proxy for the actual key? This is what we do elsewhere rather than passing actual keys around, and may prove to be simpler (especially for testing) while still making it clear that the key needs to already exist.

NickCuso: Seems fine

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -188,6 +188,38 @@ contract MixinTransfer is
...
+    uint formerTimestamp = key.expirationTimestamp;
+    bool validKey = getHasValidKey(tokenOwner);
+    if(_addTime) {
+      if(validKey) {
+        key.expirationTimestamp = formerTimestamp.add(_deltaT);
+      } else {
+        key.expirationTimestamp = block.timestamp.add(_deltaT);
+      }
+    } else {
+      key.expirationTimestamp = formerTimestamp - _deltaT;
```
NickCuso: Should we maybe use `.sub` here?

nfurfaro: absolutely. Ive been staring at this thing too long ;)

PR #5231, opened 2019-11-06
Revert "Codechecks config for gas reports"

NickCuso:
Reverts unlock-protocol/unlock#4806

It seems we do need to copy in the actual .git directory.


julien51:
Dang. Just make sure you add it last then.

PR #5232, opened 2019-11-06
Start requesting metadata from locksmith

cnasc:
# Description

This PR is an intermediate commit on the path to getting additional key metadata from locksmith. Pushing this now to avoid a truly massive PR later. The only user-visible change right now is that the `Members` page initiates signature requests for each lock being queried against, though the `Members` page is not actually "released" so I think this is ok.

# Checklist:


unlock-app/src/middlewares/storageMiddleware.js
```diff
@@ -297,6 +309,13 @@ const storageMiddleware = config => {
           storageService.getKeyPrice(lock.address)
         }
 
+        if (action.type === SIGN_METADATA_RESPONSE) {
+          const { signature, keyIds, lockAddress } = action
+          keyIds.forEach(id => {
```
julien51: Do you mind adding a quick TODO that at some point we should able to retrieve all the keyys at once?

unlock-app/src/components/content/MembersContent.tsx
```diff
@@ -38,18 +45,41 @@ export const MembersContent = ({ account, network, lockAddresses }: Props) => {
...
+  signMetadataRequest,
+  accountAddress,
+}: GraphWrapperProps) => {
   const { loading, error, data } = useQuery(keyHolderQuery(), {
     variables: { addresses: lockAddresses },
   })
 
+  useEffect(() => {
+    // Dispatch request for key metadata here, only when data changes
+    if (data) {
```
julien51: I think this works. I also thought we might be able to query locksmith from here (instead of thru Redux) provided that we got the signature thru Redux (but don't change your approach for this for now)

cnasc:
We have a fix for getting the protected data fields from `locksmith`, but this should not be merged until we look at it and clean it up.

unlock-app/src/components/content/MembersContent.tsx
cnasc: My only concern about that is the signatures are transient, so putting them in the store adds complexity because we'd also need to be able to tell when they are too stale. I think generating them at the point of use works a little simpler, although it does require the actions to carry additional data so that everything is in place for storageMiddleware.

cnasc:
For the moment going to start splitting out parts of this so we can make progress on it faster

cnasc:
Now rebased and somewhat more compact. Will begin work on integrating with @akeem's locksmith update

unlock-app/src/services/storageService.js
```diff
@@ -371,4 +373,37 @@ export class StorageService extends EventEmitter {
...
+   * @param {*} data
+   */
+  async getMetadataFor(lockAddress, keyId, signature, data) {
+    const opts = {
+      headers: this.genAuthorizationHeader(signature),
+    }
+    const stringData = encodeURIComponent(JSON.stringify(data))
+    try {
+      const result = await axios.get(
+        `${this.host}/api/key/${lockAddress}/${keyId}?data=${stringData}&signature=${signature}`,
```
akeem: there may be a way of building the query string using whatever axios provides, this is not a blocker but may make this slightly cleaner

unlock-app/src/components/content/MembersContent.tsx
```diff
@@ -38,18 +45,41 @@ export const MembersContent = ({ account, network, lockAddresses }: Props) => {
...
   )
 }
 
-const GraphWrapper = ({ lockAddresses }: { lockAddresses: string[] }) => {
+interface GraphWrapperProps {
+  lockAddresses: string[]
+  signMetadataRequest: typeof signMetadataRequest
+  accountAddress: string
+}
+const GraphWrapper = ({
```
akeem: Not exactly sure what this should be called but we may want to consider a more meaningful name than GraphWrapper we may not be able to remember what it is supposed to do in 3 weeks

cnasc: Yeah, the general gist is just to avoid bloating one function with too many concerns. I'll think of something better to call this.

PR #5233, opened 2019-11-06
Updating getTransferFee function

nfurfaro:
# Description
This modifies the getTransferFee function to return a time-based fee in seconds. 
- It also adds a new second param to the function to enable calculating fees for both transfers and partial transfers.
- finally, the lock interface has been updated to reflect this, and most of the tests in `transferFee.js` have been re-enabled.

# Issues

Fixes #
Refs # 5110

# Checklist:

### DIFF
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -293,7 +293,8 @@ interface IPublicLock {
    * @dev Throws if _owner does not have a valid key
    * @param _owner The owner of the key check the transfer fee for.
    * @param _time The amount of time to calculate the fee for.
-   * @return The transfer fee in basis-points(bps).
+   * "0" means use all the time remaining on the key for the calculation.
+   * @return The transfer fee in seconds.
    */
   function getTransferFee(
     address _owner,
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -292,10 +292,12 @@ interface IPublicLock {
    * overtime.
    * @dev Throws if _owner does not have a valid key
    * @param _owner The owner of the key check the transfer fee for.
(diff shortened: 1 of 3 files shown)


smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -292,10 +292,12 @@ interface IPublicLock {
    * overtime.
    * @dev Throws if _owner does not have a valid key
    * @param _owner The owner of the key check the transfer fee for.
+   * @param _time The amount of time to calculate the fee for.
    * @return The transfer fee in basis-points(bps).
```
NickCuso: Hmm the return units is not bps (?)

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -157,26 +157,29 @@ contract MixinTransfer is
    * transfer the key to another account.  This is pro-rated so the fee goes down
    * overtime.
    * @param _owner The owner of the key check the transfer fee for.
+   * @param _time The amount of time to calculate the fee for.
+   * "0" means use all the time remaining on the key for the calculation.
```
NickCuso: Might want the comment in the interface to match

nfurfaro: yes! on it.

smart-contracts/contracts/interfaces/IPublicLock.sol
nfurfaro: I'm not sure returning bps makes sense now? The basis points give us a simple way to set the fee percentage, but when we want to calculate the fee, it seems like returning seconds would be the easiest to work with. That said, I don't feel strongly one way or the other.

nfurfaro: I'm going to proceed with modifying the "@return" comment to reflect the return unit in seconds.

PR #5277, opened 2019-11-13
Remove unneeded modal infrastructure

### LINKED ISSUE #5036: Improve FullScreenModal infrastructure
#5029 revealed that the current infrastructure for modals in `unlock-app` doesn't really fit with our current needs. We should refactor this, or perhaps distinguish between modals that need to be globally available and those that are location-specific.

cnasc:
# Description

This PR reverts some changes that were made to the modal infrastructure. Since we ended up doing QR codes on the `keychain` as "inline modals," once again the only modal that needs to be launched from Redux instead of being component-local is the wallet check.

While doing this, I made sure we retained the cleanup I originally did in #5029 and also removed some no-longer-necessary `eslint-disable`s.

# Issues

Fixes #5036 

# Checklist:


unlock-app/src/unlockTypes.ts
```diff
@@ -3,29 +3,22 @@
 
 // A bug in eslint causes it to think that this exported enum is "unused". So
 // disable eslint for that declaration until they fix it. TODO: follow up on this.
-/* eslint-disable no-unused-vars */
```
cnasc: While I was in here I took the opportunity to remove these `eslint-disable`s, which are no longer required.

PR #5288, opened 2019-11-14
Update keyHoldersByLock query and associated functions

cnasc:
# Description

The `keyHoldersByLock` query used for batch key metadata querying always returned an empty array. This PR fixes that.

# Checklist:

### DIFF
--- a/locksmith/src/graphql/datasource/keyholdersByLock.ts
+++ b/locksmith/src/graphql/datasource/keyholdersByLock.ts
@@ -1,5 +1,6 @@
 import { gql } from 'apollo-server-express'
 import { UnlockGraphQLDataSource } from './unlockGraphQLDataSource'
+import Normalizer from '../../utils/normalizer'
 
 // eslint-disable-next-line import/prefer-default-export
 export class KeyHoldersByLock extends UnlockGraphQLDataSource {
@@ -7,7 +8,7 @@ export class KeyHoldersByLock extends UnlockGraphQLDataSource {
   genKeyHolderQuery(address: string) {
     return gql`
     {
-      locks(where: {address: ${address}} }) {
+      locks(where: {address: "${address}" }) {
       keys {
         owner {
           address
@@ -18,7 +19,7 @@ export class KeyHoldersByLock extends UnlockGraphQLDataSource {
   `
   }
 
-  async getKeyHolders(address: string) {
+  async getKeyHolders(address: string): Promise<any[]> {
     try {
       const response = await this.query(this.genKeyHolderQuery(address))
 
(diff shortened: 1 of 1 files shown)


locksmith/src/graphql/datasource/keyholdersByLock.ts
```diff
@@ -1,13 +1,14 @@
 import { gql } from 'apollo-server-express'
```
julien51: Let's also use that PR to fix the filename please `keyHoldersByLock`.
Thanks!

locksmith/src/graphql/datasource/keyholdersByLock.ts
```diff
@@ -32,12 +33,12 @@ export class KeyHoldersByLock extends UnlockGraphQLDataSource {
...
 
     try {
-      if (queryResults.data.locks === 0) {
+      if (queryResults.length === 0) {
         return []
       }
-      return queryResults.data.locks
-        .map((lock: any) => lock.keys)[0]
-        .map((key: any) => key.owner.address)
+      return queryResults[0].keys.map((key: any) =>
```
julien51: Actually, I think we can fix that query in a better way

julien51:
```
export const locksMemberQuery = locks => `{
    locks(where: {
      address_in: ${JSON.stringify(locks)}
    }) {
     keys(first: 1000) {
      owner {
        address
      }
      keyId
      expiration
    }
    name
    address
  }
}
```

That's what the guys from The Graph recommended I use for Member wall and it works great.

locksmith/src/graphql/datasource/keyholdersByLock.ts
julien51: I am confused by the `[0]` here?

cnasc: It retrieves the first and only element of the arrray

julien51: that I think I understand ;p
What I am confused about is what this element is? is that "per lock"?

julien51:
This way you can get all member son several locks at once?

locksmith/src/graphql/datasource/keyholdersByLock.ts
cnasc: The graph always returns a collection. In this case, because we are trying to get the keyholders for a particular lock, we can have a collection with either 0 or 1 elements (no lock found with that address/an error occurred and we will treat that as not having found a lock, or exactly one lock found with that address).

So when we want to operate on that lock, we have to pull it out of the array. Otherwise mapping over a collection of one lock so that we can map over its keys would make the return result a nested array, which would complicate usage at call sites.

We can make changes to make this support multiple locks, but for that to be beneficial we would have to make changes elsewhere in the stack as well which I think is not strictly necessary at the moment.

julien51: ok!

locksmith/src/graphql/datasource/keyholdersByLock.ts
julien51: Oh no!!! this did not happen :(

PR #5290, opened 2019-11-14
Implementing shareKey (partial transfers)

### LINKED ISSUE #5110: Transfer fee to be paid in time

nfurfaro:
# Description
It's all in the title.

Possible changes:
- [ ] add/switch to `safeShareKey` following example of `safeTransferFrom`

# Issues

Fixes #5110 
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -450,6 +450,48 @@ interface IPublicLock {
   /// @param _tokenId The NFT to transfer
   function transferFrom(address _from, address _to, uint _tokenId) external;
 
+  /**
+  * @notice Allows the key owner to share their key (parent key) by
+  * transferring a portion of the remaining time to a new key (child key).
+  * @dev Throws if _from does not have a valid key.
+  * @dev Throws if _from does not have enough remaining time on
+  * parent key to both pay transfer fee and share time with a child key.
+  * @dev Throws if `_to` is the zero address
+  * @param _from The owner of the parent key
+  * @param _to The recipient of the shared key
+  * @param _tokenId the key to share
+  * @param _timeShared The amount of time shared
+  * @dev Emit Transfer event
+  */
+  function shareKey(
+    address _from,
+    address _to,
+    uint _tokenId,
+    uint _timeShared
+  ) external;
+
+  /**
+  * @notice Allows the key owner to share their key (parent key) by
+  * transferring a portion of the remaining time to a new key (child key).
+  * @dev Throws if _from does not have a valid key.
+  * @dev Throws if _from does not have enough remaining time
+  * on parent key to both pay transfer fee and share time with a child key.
+  * @dev Throws if _timeShared > maxSharableTime.
+  * @dev Throws if `_to` is the zero address. When transfer is complete, this function
+  * checks if `_to` is a smart contract (code size > 0). If so, it calls
+  * `onERC721Received` on `_to` and throws if the return value is not
+  * `bytes4(keccak256('onERC721Received(address,address,uint,bytes)'))`.
+  * @param _from The owner of the parent key
+  * @param _to The recipient of the shared key
+  * @param _timeShared The amount of time shared
+  * @dev Emit Transfer event
+  */
+  // function safeShareKey(
+  //   address _from,
+  //   address _to,
+  //   uint _timeShared
+  // ) external;
+
   /// @notice Set or reaffirm the approved address for an NFT
   /// @dev The zero address indicates there is no approved address.
   /// @dev Throws unless `msg.sender` is the current NFT owner, or an authorized
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -453,9 +453,7 @@ interface IPublicLock {
   /**
   * @notice Allows the key owner to share their key (parent key) by
   * transferring a portion of the remaining time to a new key (child key).
-  * @dev Throws if _from does not have a valid key.
-  * @dev Throws if _from does not have enough remaining time on
-  * parent key to both pay transfer fee and share time with a child key.
+  * @dev Throws if key is not valid.
   * @dev Throws if `_to` is the zero address
   * @param _to The recipient of the shared key
   * @param _tokenId the key to share
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -457,14 +457,12 @@ interface IPublicLock {
   * @dev Throws if _from does not have enough remaining time on
   * parent key to both pay transfer fee and share time with a child key.
   * @dev Throws if `_to` is the zero address
-  * @param _from The owner of the parent key
   * @param _to The recipient of the shared key
   * @param _tokenId the key to share
   * @param _timeShared The amount of time shared
   * @dev Emit Transfer event
   */
   function shareKey(
-    address _from,
     address _to,
     uint _tokenId,
     uint _timeShared
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -66,20 +66,34 @@ contract MixinTransfer is
 
     Key storage fromKey = keyByOwner[_from];
     Key storage toKey = keyByOwner[_to];
-    uint IdFrom = fromKey.tokenId;
     uint IdTo = toKey.tokenId;
-    // deduct time from parent key, including transfer fee
-    _timeMachine(IdFrom, _timeShared, false);
+    uint time;
+    // get the remaining time for the origin key
+    uint timeRemaining = fromKey.expirationTimestamp - block.timestamp;
+    // get the transfer fee
+    uint fee = getTransferFee(_from, _timeShared);
+    uint totalTime = _timeShared + fee;
+
+    // ensure that we don't try to share too much
+    if(totalTime < timeRemaining) {
+      time = totalTime;
+      // deduct time from parent key, including transfer fee
+      _timeMachine(_tokenId, time, false);
+    } else {
+      time = timeRemaining - fee;
+      fromKey.expirationTimestamp = block.timestamp; // Effectively expiring the key
+      emit ExpireKey(_tokenId);
+    }
 
     if (toKey.tokenId == 0) {
       _assignNewTokenId(toKey);
-      _recordOwner(_to, toKey.tokenId);
+      _recordOwner(_to, IdTo);
     }
     // add time to new key
-    _timeMachine(IdTo, _timeShared, true);
+    _timeMachine(IdTo, time, true);
     // trigger event
       emit Transfer(
-        address(0), // This is a creation.
+        address(0), // This is a creation or time-sharing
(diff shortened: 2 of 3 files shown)


smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -44,6 +44,65 @@ contract MixinTransfer is
...
+      address(0), // This is a creation or time-sharing
+      _to,
+      IdTo
+    );
+  }
+
+  /**
+   * This is payable because at some point we want to allow the LOCK to capture a fee on 2ndary
+   * market transactions...
+   */
```
nfurfaro: oops. not sure why this is back but I will remove it.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -44,6 +44,65 @@ contract MixinTransfer is
...
+      fromKey.expirationTimestamp = block.timestamp; // Effectively expiring the key
+      emit ExpireKey(_tokenId);
+    }
+
+    if (toKey.tokenId == 0) {
+      _assignNewTokenId(toKey);
+      _recordOwner(_to, IdTo);
+    }
+    // add time to new key
+    _timeMachine(IdTo, time, true);
```
NickCuso: Hmm this is transferring `_timeShared + fee` to the other account?  For this maybe we should add a test that confirms the total time remaining across both accounts is <= the original time remaining - fee

nfurfaro: Good catch. I've reworked this to handle this better, but I'll add a test case as you suggest as well.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -44,6 +44,62 @@ contract MixinTransfer is
...
+  /**
+  * @notice Allows the key owner to share their key (parent key) by
+  * transferring a portion of the remaining time to a new key (child key).
+  * @param _from The owner of the parent key
+  * @param _to The recipient of the shared key
+  * @param _tokenId the key to share
+  * @param _timeShared The amount of time shared
+  */
+  function shareKey(
+    address _from,
```
NickCuso: Why accept _from?

By using onlyKeyOwnerOrApproved you are confirming the msg.sender has access to that tokenId, but then in the code below you pull the key indexed by _from.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -44,6 +44,62 @@ contract MixinTransfer is
...
+  ) public
+    onlyIfAlive
+    hasValidKey(_from)
+    onlyKeyOwnerOrApproved(_tokenId)
+  {
+    require(_to != address(0), 'INVALID_ADDRESS');
+
+    Key storage fromKey = keyByOwner[_from];
+    Key storage toKey = keyByOwner[_to];
+    uint IdTo = toKey.tokenId;
```
NickCuso: minor: style convention, not sure we ever start a variable with a capital letter..
where's ESlint on this?!  heh I still wonder if that's actually running at all sometimes.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -44,6 +44,62 @@ contract MixinTransfer is
...
+    uint time;
+    // get the remaining time for the origin key
+    uint timeRemaining = fromKey.expirationTimestamp - block.timestamp;
+    // get the transfer fee
+    uint fee = getTransferFee(_from, _timeShared);
+
+    // ensure that we don't try to share too much
+    if(_timeShared.add(fee) < timeRemaining) {
+      time = _timeShared;
```
NickCuso: This results in overcrediting the other account still, right?  If I try to transfer 1 year to a friend but only have 1 day remaining, it seems I'm charged a day but they still got they full year (?)

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -44,6 +44,62 @@ contract MixinTransfer is
...
+
+    if (toKey.tokenId == 0) {
+      _assignNewTokenId(toKey);
+      _recordOwner(_to, IdTo);
+    }
+    // add time to new key
+    _timeMachine(IdTo, time, true);
+    // trigger event
+    emit Transfer(
+      address(0), // This is a creation or time-sharing
```
NickCuso: Hmm is from address(0) best here?  maybe we should use the from account?

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: Well, first, we want to make sure that `_from` has a valid key before trying to share it, and second, with from we can look up the key in one step, vs using `_tokenId` to look up the owner, then looking up the key.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: oops, my bad. fixed.

smart-contracts/contracts/mixins/MixinTransfer.sol
NickCuso: if _from as approved to spend _tokenId (which is owned by someone else), this function proceeds and then _fromKey.tokenId != _tokenId

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: Yeah, could do. I guess it depends on the recipient. If they don't have a key, a new key is being created for them. If they already have a key, they just get a top-up. 
I really don't like either option...but equally fine with using either.

NickCuso: I think doing it once or twice makes sense.  If it's a new key, emit from 0 as the spec requires that on creation.  Then always emit from -> to as well at the end.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: So, if the time I want to share + fee < the timeRemaining on my key, then `time = _timeShared`, but if the time I want to share + fee >= the timeRemaining on my key, then we expire my key and share the time remaining -fee

Make sense?

nfurfaro: Basically, the value of `_time` depends on whether I'm trying to share less than I have or more. In your example, "they" only got a day minus the fee.

NickCuso: yup my bad

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: That works for me.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: I was thinking that even if this function is called by someone else who is approved, `_from` is still the address of the key owner, not msg.sender.

nfurfaro: We could remove this param, and just use `_tokenId` to look up the owner, and then get the key and check its validity using that...

nfurfaro: But we can't assume the `_from` address is the message.sender. In fact, as per the interface, `_from` is "The owner of the parent key"

nfurfaro: We can change this if you like, but it's the exact same pattern we use in `transferFrom`

nfurfaro: resolved in slack channel.

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -450,6 +450,46 @@ interface IPublicLock {
   /// @param _tokenId The NFT to transfer
   function transferFrom(address _from, address _to, uint _tokenId) external;
 
+  /**
+  * @notice Allows the key owner to share their key (parent key) by
+  * transferring a portion of the remaining time to a new key (child key).
+  * @dev Throws if _from does not have a valid key.
+  * @dev Throws if _from does not have enough remaining time on
```
NickCuso: comments about _from no longer applicable

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -44,6 +44,65 @@ contract MixinTransfer is
...
+    uint time;
+    // get the remaining time for the origin key
+    uint timeRemaining = fromKey.expirationTimestamp - block.timestamp;
+    // get the transfer fee
+    uint fee = getTransferFee(keyOwner, _timeShared);
+
+    // ensure that we don't try to share too much
+    if(_timeShared.add(fee) < timeRemaining) {
+      time = _timeShared;
+      uint timePlusFee = _timeShared.add(fee);
```
NickCuso: Minor: but we already safeMath added these values.  For gas, we could cache the result (or use + here)

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -44,6 +44,65 @@ contract MixinTransfer is
...
+      // deduct time from parent key, including transfer fee
+      _timeMachine(_tokenId, timePlusFee, false);
+    } else {
+      fee = getTransferFee(keyOwner, timeRemaining);
```
NickCuso: Hmm this seems okay but looking at it again, maybe we could simplify by using a standard transfer here.  At this point it should be on-par with the logic there.  We could maybe do that later..

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -44,6 +44,65 @@ contract MixinTransfer is
...
+      time = timeRemaining.sub(fee);
```
NickCuso: For gas savings, we could probably remove the .sub here.  We know that the fee is <= 100% making this safe.

nfurfaro:
Thx for the feedback. I'll address your concerns and merge tonight.

smart-contracts/contracts/interfaces/IPublicLock.sol
nfurfaro: yup. fixed.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: Yeah, I had a similar thought about this. 
The one difference here is that shareKey allows the original owner to retain their original tokenId (for an expired key), but maybe that doesn't matter. Maybe something to look at after this release.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: yep. done.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: done.

PR #5296, opened 2019-11-15
changed getTransactionsHashesSentBy into getRecentTransactionsHashesSentBy

julien51:
# Description

In order to improve performance on the dashboard, we do not bother retrieving old transactions (> 1 day)
which should have all be mined at this point...

# Checklist:

### DIFF
--- a/unlock-app/src/services/storageService.js
+++ b/unlock-app/src/services/storageService.js
@@ -75,15 +75,16 @@ export class StorageService extends EventEmitter {
   }
 
   /**
-   * Gets all the transactions sent by a given address.
+   * Gets all the transactions sent by a given address, in the last 24 hours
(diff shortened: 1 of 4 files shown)


unlock-app/src/services/storageService.js
```diff
@@ -80,10 +80,11 @@ export class StorageService extends EventEmitter {
    * TODO: consider a more robust url building
    * @param {*} senderAddress
    */
-  async getTransactionsHashesSentBy(senderAddress) {
+  async getRecentTransactionsHashesSentBy(senderAddress) {
```
cnasc: May be worth including the default time interval and reason for change in the JSDoc above to indicate why this is different from the old one

PR #5321, opened 2019-11-18
adding method to deploy template lock contract

julien51:
# Description

This is the first of all PRs to add support for v12.

# Checklist:


unlock-js/scripts/compressAbi.js
```diff
@@ -94,6 +94,8 @@ function formatDeployableBytecode() {
 const bytecode = {
 ${Object.keys(output).map(version => {
   return `  ${version}: {
+    PublicLock:
```
julien51: For some reason, this bit was missing previously!

unlock-js/src/walletService.js
```diff
@@ -149,6 +152,32 @@ export default class WalletService extends UnlockService {
...
+  /**
+   * Deploys a new template for locks
+   * It's a regular lock, but it will never work (purchase function fails)
+   * It just used as template whose address is fed into configUnlock to deploy
+   * locks through a proxy (keeping gas prices much lower)
+   * @param {*} version
+   * @param {*} callback
+   */
+  async deployTemplate(version, callback) {
+    const factory = new ethers.ContractFactory(
```
nfurfaro: @julien51 I wonder if this is where the problem is. From the `ethers` docs, :
> new ethers . ContractFactory ( abi , bytecode [ , signer ] )
Creates a factory for deployment of the Contract with bytecode, and the constructor defined in the abi. The signer will be used to send any deployment transaction.

Our contract(PublicLock.sol) doesn't have a constructor defined. Maybe you could try this with `ethers . ContractFactory . fromSolidity ( compilerOutput [ , signer ] )` instead?

https://docs.ethers.io/ethers.js/html/api-contract.html#deploying-a-contract

julien51: hum. I am not sure. What's strange is that the issue is transient. Sometime, the test just deploys fine, some other times, it fails.

julien51: Also, we actually do not have the `.sol` files at our disposal here inside of `unlock-js`.

nfurfaro: RIght. I was thinking we could use the artifacts(publicLock.json) here, but not sure if those are available?
However, the fact that its transient makes me doubt that this is the issue

julien51:
ok, that was an issue with constructors! All fixed now as we onkly need this for v1.2/5!

PTAL

PR #5326, opened 2019-11-19
Begins updating Locksmith with more idiomatic TS

akeem:
# Description

Locksmith's beginnings as a JS application has lead to a few places where we
could update the code to benefit from some of TS functionality.

This is a bit of a piecemeal effort trending in a better direction.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/controllers/metadataController.ts
+++ b/locksmith/src/controllers/metadataController.ts
@@ -2,16 +2,15 @@
 /* eslint-disable no-use-before-define */
 import { Response } from 'express-serve-static-core' // eslint-disable-line no-unused-vars, import/no-unresolved
 import Normalizer from '../utils/normalizer'
-import { SignedRequest } from '../types' // eslint-disable-line no-unused-vars, import/no-unresolved
 import LockData from '../utils/lockData'
 import { expiredSignature } from '../utils/signature'
 import { addMetadata } from '../operations/userMetadataOperations'
 import { KeyHoldersByLock } from '../graphql/datasource/keyholdersByLock'
+import * as lockOperations from '../operations/lockOperations'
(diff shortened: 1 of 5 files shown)


locksmith/src/controllers/metadataController.ts
```diff
@@ -2,16 +2,15 @@
 /* eslint-disable no-use-before-define */
 import { Response } from 'express-serve-static-core' // eslint-disable-line no-unused-vars, import/no-unresolved
 import Normalizer from '../utils/normalizer'
-import { SignedRequest } from '../types' // eslint-disable-line no-unused-vars, import/no-unresolved
```
cnasc: Any reason to remove this typedef?

locksmith/src/routes/metadata.ts
```diff
@@ -1,8 +1,8 @@
 import express from 'express'
 import signatureValidationMiddleware from '../middlewares/signatureValidationMiddleware'
+import MetaDataController from '../controllers/metadataController'
```
cnasc:
```suggestion
import MetadataController from '../controllers/metadataController'
```

Not critical, just a naming suggestion to have it match the namespace definition and be more like the file name.

locksmith/src/controllers/metadataController.ts
akeem: Cutting over importing the controller in the router revealed a type mismatch that previously wasn't raised. While the types in this controller are as expected, the path mapping code in the router expects an object that contains a `Request` making this incompatible

locksmith/src/routes/metadata.ts
akeem: sounds good, i'll make this change and adjust its usage below

Issue #5128, opened 2019-10-26
Lock images consistency!


julien51:
Our dashboard currently "generates" images based on the addresses. This is work that @cnasc did when he applied for Unlock ;)

When somebody purchases a key, they view a *different* image in their wallets : the one from the metadata. I think we should bring consistency to this!

Here's my thinking:
1. Move the image generating to locksmith so that the "Default" lock image is different for each lock
2. Load the image from the metadata inside of the dashboard and keychain!
3. Let the lock owner easily change the image from the dashboard (thru an image upload)

Later I believe we should always "overlay" the default lock icon in the lower right corner of the images uploaded.

Note: Chris rightfully asked about size requirements: let's always generate a square (largest possible) and crop anything that is not a square.

julien51:
> 1. Move the image generating to locksmith so that the "Default" lock image is different for each lock 
✅

> 2. Load the image from the metadata inside of the dashboard and keychain! 
Done for keychain, missing for the dashboard.

> 3. Let the lock owner easily change the image from the dashboard (thru an image upload)

Still not done

julien51:
Done!

PR #5329, opened 2019-11-19
adding method to return svg icon

julien51:
# Description

In order to be consistent in our icons for NFT and locks, we need to have the icons be "served" by locksmith.
This PR does add the function to locksmith to serve the icon based on the lock address.

# Issues

Refs #5128

# Checklist:

### DIFF
--- a/locksmith/__tests__/utils/lockIcon.test.ts
+++ b/locksmith/__tests__/utils/lockIcon.test.ts
@@ -1,0 +1,11 @@
+import { lockIcon } from '../../src/utils/lockIcon'
+
+describe('lockIcon', () => {
+  it('should generate the right svg for different locks', () => {
+    expect.assertions(2)
+    const lockIcon1 = lockIcon('0xAaAdEED4c0B861cB36f4cE006a9C90BA2E43fdc2')
+    const lockIcon2 = lockIcon('0x95de5F777A3e283bFf0c47374998E10D8A2183C7')
+    expect(lockIcon1).toMatch(/^<svg/)
+    expect(lockIcon1).not.toEqual(lockIcon2)
+  })
+})
--- a/locksmith/package-lock.json
+++ b/locksmith/package-lock.json
@@ -4550,14 +4550,6 @@
       "resolved": "https://registry.npmjs.org/ethereum-protocol/-/ethereum-protocol-1.0.1.tgz",
       "integrity": "sha512-3KLX1mHuEsBW0dKG+c6EOJS1NBNqdCICvZW9sInmZTt5aY0oxmHVggYRE0lJu1tcnMD1K+AKHdLi6U43Awm1Vg=="
     },
-    "ethereumjs-abi": {
-      "version": "git+https://github.com/ethereumjs/ethereumjs-abi.git#1cfbb13862f90f0b391d8a699544d5fe4dfb8c7b",
-      "from": "git+https://github.com/ethereumjs/ethereumjs-abi.git",
-      "requires": {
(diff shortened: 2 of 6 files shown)


julien51:
This misses tests and I will split the dependency addition, but I wanted to get a feedback on the approach?

locksmith/src/apollo-datasource-graphql.d.ts
```diff
@@ -1 +1,2 @@
 declare module 'apollo-datasource-graphql'
+declare module 'color-scheme'
```
cnasc: I think this one can be removed since it's in the global types file

locksmith/src/operations/lockOperations.ts
```diff
@@ -74,3 +75,139 @@ export async function getKeyHolderMetadata(
     },
   })
 }
+
+export async function getIcon(address: string) {
```
cnasc: Does this need to be async?

cnasc:
This seems mostly reasonable to me. The one major concern I have is putting the SVG code in `lockOperations`, since it doesn't directly relate to the DB. Probably in a separate util file?

julien51:
Yeah, I think that's right. I was hesitant about this, because it feels "different" from the other stuff (not just about the database). I'll move that to a utils file and add some tests.

locksmith/src/operations/lockOperations.ts
akeem: Lets move this and  associated code into its own library, it can still be called from the Lock Controller.  We have been using the "operations" code to separate persistence code from the controller but this does quite fit that paradigm

locksmith/src/operations/lockOperations.ts
```diff
@@ -74,3 +75,139 @@ export async function getKeyHolderMetadata(
...
+  let colors = ['#8c8c8c', '#e8e8e8', '#c3c3c3']
+  const scheme = new ColorScheme()
+  const mainColor = address.substring(2, 8).toUpperCase()
+  scheme
+    .from_hex(mainColor)
+    .scheme('triade')
+    .variation('pastel')
+  colors = scheme.colors().map((c: string) => `#${c}`)
+
+  const originalIcon = [
```
akeem: There are quite a bit of constants (eg: stripedIcon,chomIcon,bitIcon) it may be useful to provide some comments on why they are meaningful   being leveraged in this code, it

julien51:
ok, all refactored, PTAL!

julien51:
puzzled because tests do pass locally.

locksmith/__tests__/utils/lockIcon.test.ts
```diff
@@ -0,0 +1,11 @@
+import { lockIcon } from '../../src/utils/lockIcon'
+
+describe('lockIcon', () => {
+  it('should generate the right svg for different locks', () => {
+    expect.assertions(2)
+    const lockIcon1 = lockIcon('0xAaAdEED4c0B861cB36f4cE006a9C90BA2E43fdc2')
+    const lockIcon2 = lockIcon('0x95de5F777A3e283bFf0c47374998E10D8A2183C7')
+    expect(lockIcon1).toMatch(/^<svg/)
```
akeem: Not a blocker but we may want to see if we can add a bit better criteria to match up on here. I think an empty svg may end up sliding though this test.

julien51: Yeah, that's a fair point. I am testing that it's different for 2 lock addresses though so that should probably at least make sure that this code does something...

PR #5336, opened 2019-11-19
using the lock icon as default for the NFT icon

julien51:
# Description

We're now using the image rendered by locksmith a the default nft image.

I think I'll have to ask you where/how env variables are set for locksmith...

# Checklist:

### DIFF
--- a/locksmith/src/operations/metadataOperations.ts
+++ b/locksmith/src/operations/metadataOperations.ts
@@ -31,7 +31,8 @@ export const updateDefaultLockMetadata = async (data: any) => {
 export const generateKeyMetadata = async (
   address: string,
   keyId: string,
-  isLockOwner: boolean
+  isLockOwner: boolean,
+  host: string
 ) => {
   let onChainKeyMetadata = await fetchChainData(address, keyId)
   if (Object.keys(onChainKeyMetadata).length == 0) {
@@ -45,7 +46,7 @@ export const generateKeyMetadata = async (
     : {}
 
   let keyCentricData = await getKeyCentricData(address, keyId)
-  let baseTokenData = await getBaseTokenData(address)
+  let baseTokenData = await getBaseTokenData(address, host)
   return Object.assign(
     baseTokenData,
     keyCentricData,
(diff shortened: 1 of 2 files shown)


locksmith/src/operations/metadataOperations.ts
```diff
@@ -112,7 +112,7 @@ const defaultMappings = (address: string) => {
   let defaultResponse = {
     name: 'Unlock Key',
     description: 'A Key to an Unlock lock.',
-    image: `${baseURIFragement}/unlock-default-key-image.png`,
+    image: `${config.locksmithUri}/lock/${address}/icon`,
```
akeem: I think we might be able to do this a bit differently, since the SVG code provides static code regardless of where is coming from and we are returning this relative to a request being made lets see if we can construct this based on information received at a higher level around the host the request is being made to.

This should help to eliminate the need for a constant referring back to the instance serving this request

locksmith/src/operations/metadataOperations.ts
```diff
@@ -61,7 +61,7 @@ const getBaseTokenData = async (address: string) => {
   })
 
   let assetLocation = Asset.tokenMetadataDefaultImage({
-    base: baseURIFragement,
+    base: config.locksmithUri,
```
akeem: I think this is still required, the use case this supports is around  a user providing a default image for their  token  (lock) being host in the organization provided storage

locksmith/src/utils/forage.ts
```diff
@@ -4,7 +4,7 @@ class Forage {
   }
 
   tokenMetadataDefaultImage(data: any) {
-    return `${this.normalizedAddress(data.address)}/metadata/default_image`
+    return `lock/${this.normalizedAddress(data.address)}/icon`
```
akeem: I think we can revert this and the associated test as this code is in support of the lock owner provided default image

locksmith/src/operations/metadataOperations.ts
julien51: Tell me more? I am not sure I understand?

locksmith/src/utils/forage.ts
julien51: ohhh, I was confused about that. This makes sense. I'll revert that bit.

PR #5339, opened 2019-11-20
Move setupWeb3ProxyWallet to postMessageHub

cnasc:
# Description

Another piece of the main window paywall postmessaging consolidation. Pretty much the same story as the other ones.

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/Wallet.ts
+++ b/paywall/src/unlock.js/Wallet.ts
@@ -1,19 +1,13 @@
-import {
-  Web3Window,
-  web3MethodCall,
-  CryptoWalletWindow,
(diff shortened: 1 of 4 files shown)


paywall/src/unlock.js/Wallet.ts
```diff
@@ -92,10 +86,18 @@ export default class Wallet {
     return this.userAccountNetwork
   }
 
+  getHasWallet = () => {
```
cnasc: I intend for these accessors to be temporary. Once you strip the postmessages out of this class, all the real work it does happens in the constructor which means this class can probably be replaced entirely.

PR #5344, opened 2019-11-20
Data iframe handles postMessages more fluidly

cnasc:
# Description

This PR is another piece of the postMessage refactoring and serves as a template for future work.

Currently, all the various iframes have an EventEmitter abstraction layer standing between them and the messages they want to receive. They all also have to register which messages they're interested in. As a result people don't know where to add code for new messages (I know @julien51 and I have both been bitten by this).

Any code changes to postMessages should only happen at 2 points: where the message is produced, and where the message is consumed. This PR moves us toward that, simplifying a lot of the interaction on the data iframe. Once we've done this for all the iframe event emitters, there will be a lot of dead code we can ditch.

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/PostMessageEmitters/DataIframeMessageEmitter.ts
+++ b/paywall/src/unlock.js/PostMessageEmitters/DataIframeMessageEmitter.ts
@@ -2,8 +2,7 @@ import { EventEmitter } from 'events'
 import { PostMessages } from '../../messageTypes'
 import {
   PostMessageResponder,
-  mainWindowPostOffice,
-  PostMessageListener,
+  emitPostMessagesFrom,
 } from '../../utils/postOffice'
 import {
   IframeType,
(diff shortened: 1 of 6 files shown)


paywall/src/unlock.js/PostMessageEmitters/DataIframeMessageEmitter.ts
```diff
@@ -53,15 +42,33 @@ export default class DataIframeMessageEmitter extends FancyEmitter {
...
-      'main window',
-      'Data iframe'
+      window,
+      this.emitWrapper
     )
     this.postMessage = postMessage
-    this.addHandler = addHandler
+  }
+
+  emitWrapper = (type: PostMessages, payload?: any) => {
```
cnasc: This is a bit of a hack to match the previous behavior of the registered listeners -- in the future I think we should do as little manipulation of messages as possible. If we don't think it's useful to receive an update with an empty object of locks, the data iframe should never emit one. Likewise for web3 calls.

See: [the robustness principle](https://en.wikipedia.org/wiki/Robustness_principle)

paywall/src/utils/postOffice.ts
```diff
@@ -129,6 +138,82 @@ export function setupPostOffice<T extends MessageTypes = MessageTypes>(
   }
 }
 
+export function postMessageIsSafe(
```
cnasc: Extracted from the head of `mainWindowPostOffice` -- makes sense to have this as a separate function, and now when debug mode is on you'll be able to see why a message gets rejected.

PR #5346, opened 2019-11-21
updated graph service and query when retrieving user locks

julien51:
# Description

All in the title!

# Checklist:

### DIFF
--- a/unlock-app/src/queries/locksByOwner.js
+++ b/unlock-app/src/queries/locksByOwner.js
@@ -7,18 +7,9 @@ export default function locksByOwnerQuery() {
         id
         address
         name
-        tokenAddress
-        price
         expirationDuration
         totalSupply
         maxNumberOfKeys
-        keys {
-          id
-          lock
-          keyId
-          owner
-          expiration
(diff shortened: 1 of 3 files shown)


unlock-app/src/queries/locksByOwner.js
```diff
@@ -7,18 +7,9 @@ export default function locksByOwnerQuery() {
         id
         address
         name
-        tokenAddress
-        price
         expirationDuration
         totalSupply
         maxNumberOfKeys
-        keys {
```
julien51: Not needed here, so let's not load them...

unlock-app/src/queries/locksByOwner.js
```diff
@@ -7,18 +7,9 @@ export default function locksByOwnerQuery() {
         id
         address
         name
-        tokenAddress
```
julien51: This is not normalized so we cannot show it on the UI for now. (locks are loaded from alchemy anyway...) We just use The Graph has a "hint" until we have all the data we need in there.

PR #5362, opened 2019-11-21
Consolidate unconditional post messages

cnasc:
# Description

This PR is part of the Great PostMessage Refactor of 2019. It consolidates all of the postMessage listeners that should always be active -- they don't depend on some condition (such as "using a user account") to avoid stepping on each other.

This PR will break until more refactoring happens, and it's a bit huger than I'd like. A lot of it is interdependent, so I'll have to see where it can be split up. Want to get the draft up to invite comment.

# Checklist:


cnasc:
I will most likely extract the `postmessageHub` changes into their own PR, then restructure a few other things (not in this PR) before merging the shift to the unified postmessage handling.

PR #5379, opened 2019-11-27
Downgrade Truffle and unlock-abi-1-2 [fix master]

NickCuso:
This gets our tests to run again.

It's not yet clear why tests do not run with the new truffle version.  However they had a bug with the very same impact just a few versions back.


NickCuso:
v5 upgrade test also fails as the v5 package bumped as well

PR #5380, opened 2019-11-27
Pass the correct flag to checkoutHandlerInit [fixes master regression]

### LINKED ISSUE #5378: Injected balance is wrong for managed key purchases
**Describe the bug**
Pasted from #5377:
> image

> This is on the Ouvre-Boite staging page with no metamask (i.e., trying to use a managed account). The Eth lock should be disabled, and the WEE lock should be active. I think there's something wrong somehow in the injected balance.

This issue is only in staging at the moment, prod is unaffected. I believe this is related to the paywall refactoring.

**To Reproduce**
1. On staging, go to a page that supports user accounts
2. Launch the checkout modal
3. Note that ERC20 locks show as having insufficient balance and ETH locks show as having correct balance

**Expected behavior**
ERC20 locks should have an injected balance that allows a user to do a managed key purchase with them. ETH locks should be disabled.

cnasc:
# Description

In the course of refactoring the paywall, I forgot to remove the Wallet-local variable used to determine whether we used user accounts. It defaulted to false, and was never updated. Therefore, when it was passed to the checkout handler init, it would never allow the default balance injection to occur, which led to the issue we discovered yesterday.

This PR cleans that up. _We should do a manual production push after this is merged_

# Issues

Fixes #5378 

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/CheckoutUIHandler.ts
+++ b/paywall/src/unlock.js/CheckoutUIHandler.ts
@@ -56,7 +56,6 @@ export default class CheckoutUIHandler {
       config: this.config,
       dataIframe: this.iframes.data,
       checkoutIframe: this.iframes.checkout,
-      injectDefaultBalance,
     })
   }
 }
--- a/paywall/src/unlock.js/Wallet.ts
+++ b/paywall/src/unlock.js/Wallet.ts
@@ -27,7 +27,6 @@ export default class Wallet {
   private readonly window: Web3Window
   private readonly config: PaywallConfig
   private hasWeb3: boolean = false
-  useUserAccounts: boolean = false
 
   private userAccountAddress: string | null = null
   private userAccountNetwork: unlockNetworks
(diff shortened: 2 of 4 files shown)


paywall/src/unlock.js/CheckoutUIHandler.ts
```diff
@@ -56,7 +56,6 @@ export default class CheckoutUIHandler {
       config: this.config,
       dataIframe: this.iframes.data,
       checkoutIframe: this.iframes.checkout,
-      injectDefaultBalance,
```
cnasc: In this case I removed the class-local version of injectDefaultBalance from being used in the init, now that we've extracted it to a util. A next step will be to delete the code duplication.

paywall/src/unlock.js/Wallet.ts
```diff
@@ -27,7 +27,6 @@ export default class Wallet {
   private readonly window: Web3Window
   private readonly config: PaywallConfig
   private hasWeb3: boolean = false
-  useUserAccounts: boolean = false
```
cnasc: That was the unused variable at the root cause of this issue.

julien51:
Let's schedule a push on Monday, this should be 4 days in?

PR #5383, opened 2019-11-28
Extracting the erc721 interface

nfurfaro:
# Description
This PR helps pave the way for a related one, switching to using the standard openzeppelin `IERC.sol` interface. Here, I've extracted all the erc721-specific stuff out of `IPublicLock.sol`(which is now abstract) and imported the IECR721 & IECR721Enumerable interfaces
Once this is merged, the next one can simply switch the soure of those imports to the openzeppelin package.

- Note: the multiple comments directives regarding indentation in `IPublicLock.sol` are a simple (hopefully tempory) workaround to a Solium bug described here: https://github.com/duaraghav8/Ethlint/issues/268
Basically, once the functions are made abstract (rather than being pure interfaces) Solium complains about indentation that was previously considered fine.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -11,9 +11,9 @@ import './IERC721Enumerable.sol';
 
 contract IPublicLock is IERC721Enumerable, IERC721 {
 
-/// See indentationissue description here:
-/// https://github.com/duaraghav8/Ethlint/issues/268
-/// solium-disable indentation
+// See indentationissue description here:
+// https://github.com/duaraghav8/Ethlint/issues/268
+// solium-disable indentation
 
   ///===================================================================
   /// Events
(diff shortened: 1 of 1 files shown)


smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -304,7 +290,7 @@ interface IPublicLock {
   function getTransferFee(
     address _owner,
     uint _time
-  ) external view returns (uint);
+  ) external view returns (uint); // solium-disable-line indentation
```
NickCuso: hmm instead of adding this to each impacted line, could we just do it once for the entire file?  https://ethlint.readthedocs.io/en/latest/user-guide.html#configuring-with-comments

And maybe also include a comment pointing to the bug

nfurfaro: I could. Although this is repetitive, I didn't want to disable the linting for the whole file if I didn't have to. I don't mind switching it though.
I will add a link to the bug.

PR #5394, opened 2019-12-02
Adds ability to limit size of result set of queried keys

akeem:
# Description

As part of making the local graphql endpoint consistent with the service previously used, we are implementing the 'first' predicate to be passed to the service being utilized below

# Issues

Fixes #
Refs #

# Checklist:


akeem:
> What happens in this case if no `first` parameter is passed?
> 
> LG

In the case of first not being passed, we will pass to  graphprotocol's endpoint without a limit and we will get their limit back. I think its something like 999

PR #5395, opened 2019-12-02
adding erc1820

julien51:
# Description

This is required by v12 for testing.

# Checklist:

### DIFF
--- a/unlock-js/package-lock.json
+++ b/unlock-js/package-lock.json
@@ -2238,6 +2238,16 @@
       "integrity": "sha512-2scffjvioEmNz0OyDSLGWDfKCVwaKc6l9Pm9kOIREU13ClXZvHpg/nRL5xyjSSSLhOnXqft2HpsAzNEEA8cFFg==",
       "dev": true
     },
+    "accepts": {
+      "version": "1.3.7",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.7.tgz",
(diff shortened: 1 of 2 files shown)


NickCuso:
erc1820 does expect a web3 instance, which i think would make the -js package heavier.  to avoid maybe we should port the code to ethers.js.  it's not much code: https://github.com/0xjac/ERC1820/blob/master/js/deployment.js

julien51:
Dang. what is a `web3` instance exactly? I was indeed hoping to use ethers.js

julien51:
Maybe we can use the `generateDeployTx` function.

NickCuso:
it expects this package https://www.npmjs.com/package/web3

we could use generateDeployTx and then just port the deploy section, but that still adds a couple dependencies.  i'm not sure if that would be significant.  however ethers.js should have the same capabilities already, so it would be redundant it seems

julien51:
I think I got this. `generateDeployTx` yields the stuff we need for `sendTransaction` https://docs.ethers.io/ethers.js/html/api-providers.html#contract-execution

NickCuso:
We could also consider using a proxy to forward web3 apis to ethers.  Something like:
const web3Proxy = {
eth: {
getCode: (address) => {
return provider.getCode(address); // https://docs.ethers.io/ethers.js/html/api-providers.html?highlight=getcode
} 
}
}

PR #5397, opened 2019-12-03
Validators for specification of metadata fields in paywall config

cnasc:
# Description

This PR is the first step toward integrating metadata collection with the paywall. This adds validators for metadata fields, which will be used in the near future to validate the fields supplied in the metadata config.

The idea is that a property called something like `metadataInputs` will contain an array of objects like
```javascript
{
  name: 'field name', // any string
  type: 'date', // any valid html input type
  required: false, // a boolean
}
```

When collecting metadata, the paywall will then render those fields in the order specified, with the appropriate attributes applied to the input (`type` and `required`) and the name used as a label for the input.

# Checklist:

### DIFF
--- a/paywall/src/utils/validators.js
+++ b/paywall/src/utils/validators.js
@@ -387,4 +387,22 @@ export const isValidMetadataField = field => {
 
   return true
 }
+
+export const isValidMetadataArray = fields => {
+  if (!Array.isArray(fields)) {
+    console.error('Paywall config metadata property is not an array.')
+    return false
+  }
+
+  // TODO: disallow multiple fields with the same name?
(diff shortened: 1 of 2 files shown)


paywall/src/utils/validators.js
```diff
@@ -344,3 +343,66 @@ export const isValidBalance = balance => {
     )
   }, true)
 }
+
+const allowedInputTypes = ['text', 'date', 'color', 'email', 'url']
```
cnasc: Would appreciate comment on which fields we should include. I believe these are the only fields that don't require extra data (`select`s would need a series of options, for example)

PR #5399, opened 2019-12-03
Break out parts of config validation for easier reading

cnasc:
# Description

In advance of adding metadata validation to the config validator, this PR splits up the mega-method that does the validation into smaller units. This is just a first pass, but it stops the validator from being enormous and doing 12 different things which is good enough for now.

No behavioral change, just some shuffling.

# Checklist:

### DIFF
--- a/paywall/src/utils/validators.js
+++ b/paywall/src/utils/validators.js
@@ -83,6 +83,25 @@ export const isValidCTA = callToAction => {
   return true
 }
 
+export const isValidConfigLock = (lock, configLocks) => {
+  if (!isAccount(lock)) return false
+  const thisLock = configLocks[lock]
+  if (!thisLock || typeof thisLock !== 'object') return false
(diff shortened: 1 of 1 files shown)


paywall/src/utils/validators.js
```diff
@@ -27,6 +29,94 @@ export const isAccountOrNull = val => {
...
+
+export const isValidConfigLocks = configLocks => {
+  if (typeof configLocks !== 'object') {
+    console.error(`The paywall configs's "locks" field is not an object.`)
+    return false
+  }
+  const locks = Object.keys(configLocks)
+  if (!locks.length) return false
+  if (
+    locks.filter(lock => {
```
akeem: I know this is more of a reshift, but this is a bit wild. How would you feel about seeing if we can extract this whole thing into its own method?

cnasc: Reasonable. I'll do that now.

PR #5401, opened 2019-12-03
Metadata inputs are validated as part of config validation

cnasc:
# Description

Finalizing the metadata validation work, this PR activates it in the paywall config validation and adds some tests and type definitions.

# Checklist:

### DIFF
--- a/paywall/src/utils/validators.js
+++ b/paywall/src/utils/validators.js
@@ -212,6 +212,12 @@ export const isValidPaywallConfig = config => {
     return false
   }
 
+  if (config.metadataInputs) {
+    if (!isValidMetadataArray(config.metadataInputs)) {
+      return false
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


paywall/src/utils/validators.js
```diff
@@ -212,6 +212,12 @@ export const isValidPaywallConfig = config => {
     return false
   }
 
+  if (config.metadataInputs) {
+    if (!isValidMetadataArray(config.metadataInputs)) {
```
akeem: could we return `isValidMetadataArray(config.metadataInputs)` here ?

cnasc: We can't, because we can't return `true` unless every condition passes. If we hypothetically added other validations later that come after this one, that could be a subtle bug because the later additions would never get checked.

PR #5417, opened 2019-12-05
First pass at leaner data-iframe

cnasc:
# Description

Per my discussion with @julien51 this morning, I decided to take a swipe at a leaner `BlockchainHandler` to use with our more flexible iteration on the paywall. It's similar to the existing one, but with an emphasis on avoiding unnecessary work.

I think it's also valuable to have a split between the reader and the writer (should be same iframe, separate objects), so right now I'm conceptualizing this as being the read-only portion. Think of it like some glue attaching Web3Service and the pending transactions in Locksmith.

This _isn't_ feature complete yet, and it's also not hooked up to anything. Just want to get some feedback on the approach.

Here are some things this object will explicitly **not** do:

- create "default" keys
- purchase keys (handled elsewhere)
- handle key expiration beyond simply saying the timestamp at which a key expires
- munge together locks, keys, and transactions
- be "resettable" (if something like an account or network changes and invalidates data, throw the object away and make a new one. This case is rare enough in reality that it isn't worth complicating the code for)

In terms of use, when this is hooked up there should be an `index.ts` file that constructs a `Web3Service` and sets up something like a [postmate](https://github.com/dollarshaveclub/postmate) `Child`. When the `Child` receives the necessary information (lock addresses from paywall config and account address), it build a `BlockchainHandler` and get things started. This line of communication will also allow us to share the data we fetch.

# Checklist:


cnasc:
Assuming this approach is palatable, next steps are to implement all the relevant Web3Service listeners and then also set up a bare MainWindowHandler with postmate so that we can actually have it running and grow it from there as a v2 option for the script.

cnasc:
This is ready for a review. It will probably solidify more as we integrate it with messaging, but for now I think it's in a good place.

cnasc:
This can be split if needed, breaking it into the locksmith util and the object itself

PR #5418, opened 2019-12-05
Showing the key metadata on the key verification page

julien51:
# Description

This uses a hybrid approach between redux and hooks but that's actually faily straightforward.
I am still confused about the right way to write tests for this.
Any suggestion?

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useGetMetadataFor.js
+++ b/unlock-app/src/hooks/useGetMetadataFor.js
@@ -1,0 +1,23 @@
+import { useMemo } from 'react'
+import { useSelector, useDispatch } from 'react-redux'
+import { signBulkMetadataRequest } from '../actions/keyMetadata'
+
+/**
+ * This hook retrieves metadata for a token
+ * @param {*} address
+ */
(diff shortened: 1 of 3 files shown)


julien51:
image

unlock-app/src/hooks/useGetMetadataFor.js
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from 'react'
+import { useSelector, useDispatch } from 'react-redux'
+import { signBulkMetadataRequest } from '../actions/keyMetadata'
+
+/**
+ * This hook retrieves metadata for a token
+ * @param {*} address
+ */
+export const useGetMetadataFor = (lockAddress, keyOwner) => {
+  const { account, metadata } = useSelector(state => state)
```
cnasc: I _think_ to test this you'll want to mock `useSelector` so that it returns various values for `metadata` and then check that the hook returns what you expect in those cases.

Additionally, you can check that if `account` is present, `dispatch` gets called with the appropriate message.

Issue #5428, opened 2019-12-06
Hook up new BlockchainHandler in parallel on unlock.js script in paywall


cnasc:
With #5417 and #5427 we have enough pieces to trial the new data iframe in parallel with the existing one.

To do this, we'll need a few different things:

- [x] Get the new BlockchainHandler to build and be served in such a way that we can load an iframe into it (#5433)
- [x] define a postmate `Child` for the BlockchainHandler (#5433)
- [ ] define a postmate `Parent` for the main window (in progress at #5436)

Once we have these in, we can start handling some pieces of the app's needs with the new handler. Unfortunately there may be downstream changes required (checkout iframe currently relies on some of the old behavior of the data iframe).

I think once these pieces are in place it should be easy to quickly iterate on this junction and it will give us a template to follow in hooking up the other iframes and moving to our new desired architecture.

PR #5433, opened 2019-12-09
New data iframe has a Postmate model and can be built

cnasc:
# Description

This PR does the following:

- adds `index.2.0.ts` to the `data-frame` directory, giving us an entry point to build the 2.0 version
  - that index file sets up a postmate model so the data iframe can talk to the main window
  - some ancillary shuffling of code to make things cleaner happened as a result
- adds a webpack config to build the 2.0 version
- adds an html file so we can open an iframe to the 2.0 version
- adds a build command to `package.json`

# Issues

Refs #5428 

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/index.2.0.ts
+++ b/paywall/src/data-iframe/index.2.0.ts
@@ -22,7 +22,7 @@ const web3Service = new Web3Service({
 let blockchainReader = new BlockchainDataStorable()
 
 const initializeReader = (lockAddresses: string[], accountAddress: string) => {
-  // since we may destroy a previous BlockchainHandler with this
+  // since we may destroy a previous BlockchainReader with this
   // operation, we should also make sure not to retain any event
   // listeners that involve the old object
   web3Service.removeAllListeners()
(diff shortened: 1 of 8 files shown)


akeem:
👏  on actually bumping the version number of this project

paywall/src/data-iframe/index.2.0.ts
```diff
@@ -0,0 +1,45 @@
...
+  unlockAddress,
+  blockTime,
+  requiredConfirmations,
+})
+
+// Start with a null object, it will be replaced when the authenticate
+// method in the model is called.
+let blockchainHandler = new BlockchainDataStorable()
+
+const authenticate = (lockAddresses: string[], accountAddress: string) => {
```
akeem: the naming of this method feels a bit weird or appears a bit weird given the parameters being passed. Would we need to pass lock addresses here?

cnasc: Maybe `init` is a better choice -- those are the 2 pieces of data required to instantiate the handler:
- the list of lock addresses from the paywall config
- the account address for the user

cnasc: In fact, since there is a separation between the read and write operations, it's probably even better to disambiguate exactly what we're starting here.

cnasc:
Improvements made to naming, ptal @akeem

PR #5435, opened 2019-12-09
Adding wedlocks to the developer setup

akeem:
# Description

As part of trending towards a saner dev/testing setup we are adding wedlocks to the development compose file.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/docker/docker-compose.development.yml
+++ b/docker/docker-compose.development.yml
@@ -91,3 +91,14 @@ services:
     command: ['node', './deploy-subgraph.js']
     depends_on:
       - graph-node
+  wedlocks:
+    image: unlockprotocol/wedlocks
+    command: npm run dev
+    environment: 
+      SMTP_HOST: 'smtp_host'
+      SMTP_USERNAME: 'smtp_username'
+      SMTP_PASSWORD: 'smtp_password'
+      BASE64_WEDLOCKS_PRIVATE_KEY: 'pk'
+
+    ports:
+      - '1337:1337'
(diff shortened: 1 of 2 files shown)


docker/docker-compose.development.yml
```diff
@@ -91,3 +91,14 @@ services:
...
     depends_on:
       - graph-node
+  wedlocks:
+    image: unlockprotocol/wedlocks
+    command: npm run dev
+    environment: 
+      SMTP_HOST: 'smtp_host'
+      SMTP_USERNAME: 'smtp_username'
+      SMTP_PASSWORD: 'smtp_password'
+      BASE64_WEDLOCKS_PRIVATE_KEY: 'pk'
```
cnasc: This may break some functionality in dev -- I think it will impact the query strings in the generated URLs. @julien51 can confirm

akeem: Interesting, would we be able to generate the relevant data utilized here explicitly for the local development use case?

julien51: Sorry for the delay here. Not sure it would actually break? Worth a try...

julien51: I think you can generate a private key like this (see https://github.com/unlock-protocol/unlock/blob/master/wedlocks/src/__tests__/encrypter.test.js#L11)
```
    forge.rsa.generateKeyPair({ bits: 2048, workers: 2 }, function(
      err,
      keypair
    ) {
      privateKey = forge.pki.privateKeyToPem(keypair.privateKey)
      publicKey = forge.pki.publicKeyToPem(keypair.publicKey)
      done()
    })
```

PR #5436, opened 2019-12-09
A swipe at connecting the parallel data iframe

### LINKED ISSUE #5428: Hook up new BlockchainHandler in parallel on unlock.js script in paywall

cnasc:
# Description

This PR connects the v2.0 data iframe in parallel with the existing one (no changed functionality)

# Issues

Fixes #5428 

# Checklist:

### DIFF
--- a/paywall/data-iframe.2.0.webpack.config.js
+++ b/paywall/data-iframe.2.0.webpack.config.js
@@ -6,7 +6,7 @@ const configVariables = require('./environment')
 module.exports = () => {
   return {
     cache: false,
-    mode: 'production',
+    mode: 'development',
     devtool: 'source-map',
     entry: path.resolve(__dirname, 'src', 'data-iframe', 'index.2.0.ts'),
     output: {
--- a/paywall/src/data-iframe/index.2.0.ts
+++ b/paywall/src/data-iframe/index.2.0.ts
@@ -2,14 +2,49 @@ import Postmate from 'postmate'
 import { Web3Service } from '@unlock-protocol/unlock-js'
 import { BlockchainReader } from './BlockchainReader'
 import { BlockchainDataStorable } from './BlockchainDataStorable'
-import configure from '../config'
+
+declare var __ENVIRONMENT_VARIABLES__: any
+
+// TODO FIND A WAY TO MOVE THIS IN A BETTER PLACE
+// NOTE: NEVER PUT ENV VARS HERE
+const constants: { [key: string]: any } = {
+  dev: {
+    unlockAddress: '0x885EF47c3439ADE0CB9b33a4D3c534C99964Db93',
+    blockTime: 3000,
+    requiredConfirmations: 6,
+    defaultNetwork: 1984,
+  },
+  test: {
+    unlockAddress: '0x885EF47c3439ADE0CB9b33a4D3c534C99964Db93',
+    blockTime: 3000,
+    requiredConfirmations: 6,
+    defaultNetwork: 1984,
+  },
+  staging: {
+    unlockAddress: '0xD8C88BE5e8EB88E38E6ff5cE186d764676012B0b',
+    blockTime: 8000,
+    requiredConfirmations: 12,
+    defaultNetwork: 4,
+  },
+  prod: {
+    unlockAddress: '0x3d5409CcE1d45233dE1D4eBDEe74b8E004abDD13',
+    blockTime: 8000,
+    requiredConfirmations: 12,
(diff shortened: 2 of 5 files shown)


paywall/data-iframe.2.0.webpack.config.js
```diff
@@ -6,7 +6,7 @@ const configVariables = require('./environment')
 module.exports = () => {
   return {
     cache: false,
-    mode: 'production',
+    mode: 'development',
```
cnasc: I had to switch this temporarily because the browser complained it couldn't find the related source maps. I'm sure there's a configuration setting I have to switch somewhere.

paywall/src/data-iframe/index.2.0.ts
```diff
@@ -2,14 +2,49 @@ import Postmate from 'postmate'
...
+  ...__ENVIRONMENT_VARIABLES__,
+  ...constants[__ENVIRONMENT_VARIABLES__.unlockEnv],
+}
 
 const {
   readOnlyProvider,
   unlockAddress,
   blockTime,
   requiredConfirmations,
-} = configure()
```
cnasc: `configure()` has a dependency on `next`, which we don't have access to here. Luckily all the required info is in the environment variables we bundle.

paywall/src/data-iframe/index.2.0.ts
```diff
@@ -2,14 +2,49 @@ import Postmate from 'postmate'
 import { Web3Service } from '@unlock-protocol/unlock-js'
 import { BlockchainReader } from './BlockchainReader'
 import { BlockchainDataStorable } from './BlockchainDataStorable'
-import configure from '../config'
+
+declare var __ENVIRONMENT_VARIABLES__: any
+
+// TODO FIND A WAY TO MOVE THIS IN A BETTER PLACE
+// NOTE: NEVER PUT ENV VARS HERE
+const constants: { [key: string]: any } = {
```
cnasc: This is an artifact from the previous implementation. I think we can sort this out. It _might_ be worth splitting out the various iframes into different subrepos (or at least into one `paywall-iframes` subrepo), which would allow us to keep the configuration specific to them in one place.

paywall/src/data-iframe/index.2.0.ts
```diff
@@ -21,10 +56,18 @@ const web3Service = new Web3Service({
 // method in the model is called.
 let blockchainReader = new BlockchainDataStorable()
 
-const initializeReader = (lockAddresses: string[], accountAddress: string) => {
+interface ReaderArgs {
+  lockAddresses: string[]
+  accountAddress: string
+}
+const initializeReader = ({ lockAddresses, accountAddress }: ReaderArgs) => {
```
cnasc: Had an issue with calling `initializeReader` from the other end with multiple arguments, so now we pass an object to it.

paywall/src/unlock.js/startup.ts
```diff
@@ -65,6 +66,16 @@ export function startup(
...
   iframes.init(config)
 
+  const handshake = new Postmate({
+    url: constants.paywallUrl + '/static/data-iframe.2.0.html',
+    classListArray: ['unlock', 'start'],
+  })
+
+  handshake.then(child => {
+    console.log('Data iframe handshake completed!')
+    ;(window as any).dataIframe = child
```
cnasc: This is just here for testing purposes. We have to figure out exactly where this step belongs. However, once this promise resolves, you can go into the console and run:

```javascript
dataIframe.call('initializeReader', {
  lockAddresses: [ /* a list of lock addresses */ ], 
  accountAddress: "", /* your address */
})
```

And the reader will initialize and start querying for relevant data. Then calls to the model will start to bear fruit:

```javascript
dataIframe.get('locks').then(console.log)

dataIframe.get('keys').then(console.log)
```

PR #5467, opened 2019-12-13
Support for v12

julien51:
# Description

This PR introduces support for v12.
I am now confident that this is the last version implemented in this manner.
We need to refactor all unit tests to not include nock as these tests combine both our code un ethers tests.
We should mock anything ethers related, not the underlying network calls.

# Checklist:


unlock-js/src/constants.js
```diff
@@ -1,9 +1,11 @@
 import { ethers } from 'ethers'
 
+// TODO: gas prices should be custom per version!
```
cnasc: Are these prices or limits?

unlock-js/src/v12/createLock.js
```diff
@@ -42,14 +42,20 @@ export default async function(lock, callback) {
   let currencyContractAddress = lock.currencyContractAddress || ZERO
 
   const lockName = lock.name || 'New Lock'
+
+  // Building a salt from the lock name will prevent creators from creating 2 locks with the same name.
```
cnasc: Just curious what the rationale is behind this

unlock-js/src/constants.js
julien51: you are right, they are not prices, but amounts of gas to be spent. I'll fix the comment

unlock-js/src/v12/createLock.js
julien51: The salt is used to compute the lock address, along with the user's nonce.
This will prevent someone from creating 2 locks at the same time with the same name. (like pressing "create lock" twice...)

PR #5488, opened 2019-12-16
Wrap WalletService for data-iframe

cnasc:
# Description

This PR replaces #5457. It's a somewhat simpler approach. There is a little more messaging involved, but it keeps the code itself cleaner and handles the complexity of interaction between web3service and walletservice (which are meant to be separate, but what happens in walletservice can impact web3service)

In short, the wallet wrapper is completely stateless. For the moment, it just passes messages up to the main window. In the event of an account or network change, the main window will handle it by reinitializing the blockchain reader. The main window will handle a new transaction by passing it back down as a transaction update (not yet implemented).

The data iframe will need a bit more work, but I'm hopeful that we can soon use it for real, at least for the read-only pieces.

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/walletWrapper.ts
+++ b/paywall/src/data-iframe/walletWrapper.ts
@@ -1,0 +1,68 @@
+import { WalletService } from '@unlock-protocol/unlock-js'
+import {
+  WalletServiceType,
+  TransactionDefaults,
+} from './blockchainHandler/blockChainTypes'
+import { normalizeLockAddress } from '../utils/normalizeAddresses'
+import { POLLING_INTERVAL } from '../constants'
+
+// There are slight differences between what WalletService emits and
+// what our other code expects, this piece massages the data so that
+// they match.
+export const formatTransaction = (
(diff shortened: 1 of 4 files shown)


paywall/src/data-iframe/walletWrapper.ts
```diff
@@ -0,0 +1,74 @@
...
+    to: normalizedTo,
+    input,
+    type,
+    status,
+    blockNumber: Number.MAX_SAFE_INTEGER,
+  }
+}
+
+// hideous
+type NewTransactionArgs = [string, string, string, string, string, string]
```
akeem: Not a deal breaker considering where we started from with this, but we should probably revisit this at some point

cnasc: Definitely agree. Until we have a good picture of the downstream consumer expectations I’m trying to avoid messing with the formatting too much (once we start using this iframe for real we’ll be able to make larger changes with confidence thanks to the integration tests)

PR #5501, opened 2019-12-17
Invalidate config when we don't have access to web3

cnasc:
# Description

This is a turbo-hack, but now we force the config to be invalid whenever user accounts are not allowed and there is no injected wallet. (For example, this will be the case when a typical Forbes visitor with no wallet visits a page with the paywall on it).

We should thoroughly manually test this in staging.

(didn't have time to rebase my prior PR so there are a handful of extra lines in here)

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/startup.ts
+++ b/paywall/src/unlock.js/startup.ts
@@ -40,7 +40,24 @@ export function startup(
   constants: StartupConstants
 ) {
   // normalize all of the lock addresses
-  const config = normalizeConfig(window.unlockProtocolConfig)
+  let config = normalizeConfig(window.unlockProtocolConfig)
+
+  // There is no reason to do anything if window.web3 does not exist
+  // and the config does not allow for user accounts. As a quick hack,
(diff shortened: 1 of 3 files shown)


paywall/src/unlock.js/startup.ts
```diff
@@ -41,7 +40,24 @@ export function startup(
...
+  // that we don't make any requests for lock data.
+  const userAccountsAllowed = !!config.unlockUserAccounts
+  const web3Present = !!window.web3
+  if (!web3Present && !userAccountsAllowed) {
+    config = {
+      ...config,
+      // This violates the expected value for locks on the paywall,
+      // which will force the checkout into the "no wallet" state
+      // without ever querying for any locks.
+      locks: new XMLHttpRequest(),
```
julien51: wierd. can we just set it to be `{}`?

cnasc: I figured that would be guaranteed to be invalid, but I can change it

PR #5515, opened 2019-12-18
Cache account address when account is updated in walletservice

cnasc:
# Description

This PR caches an account address in localStorage whenever we get an account address update from localstorage. This will allow us to defer loading the paywall unless we already have a stored account.

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/MainWindowHandler.ts
+++ b/paywall/src/unlock.js/MainWindowHandler.ts
@@ -56,6 +56,7 @@ export default class MainWindowHandler {
       showAccountIframe: this.showAccountIframe,
       hideCheckoutIframe: this.hideCheckoutIframe,
       blockchainData: this.blockchainData,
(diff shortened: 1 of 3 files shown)


paywall/src/unlock.js/MainWindowHandler.ts
```diff
@@ -106,11 +107,10 @@ export default class MainWindowHandler {
...
     try {
-      // this is a fast cache. The value will only be used
-      // to prevent a flash of ads on startup. If a cheeky
-      // user attempts to prevent display of ads by setting
-      // the localStorage cache, it will only work for a
-      // few milliseconds
+      // this is a fast cache. The value will only be used to prevent
+      // a flash of ads on startup. If a user attempts to prevent
+      // display of ads by setting the localStorage cache, it will
+      // only work for a few milliseconds
```
akeem: We should extract this to a module with a reasonable name so that we can remove this comment and the use case is clear

PR #5517, opened 2019-12-18
Defer startup for users who don't have cached address

cnasc:
# Description

In response to a request from Forbes, we won't try to connect to Metamask (or other wallets) until the user has interacted with the button. In the case of users for whom we have a cached address, we see that as indication they have previously chosen to connect, and go ahead and initialize the app.

This performed nicely in dev, but we should thoroughly run down the use cases in staging before doing a push.

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Agreed, thanks a lot @cnasc ! I'll merge so we can test on staging extensively!

Issue #5518, opened 2019-12-19
Not prompting for MetaMask


julien51:
Our paywall application prompts for MetaMask the moment it is loaded. This is a sub-par experience because people may not necessarily have an understanding of why they are being prompted.

On the other end, if we wait for users to actually perform an action to get their identity (by prompting the wallet), we end up in a situation where we cannot assess whether the users have already unlocked the content... which, if they did, would be another bad experience.

julien51:
@cnasc implemented a clever solution which consists of 2 steps:

1. caching (localstorage) the address of the user if it is known

2. upon loading
* if no identity is known/cached: 
  * emit the "locked" event
  * skip initialization of the paywall application, until the user performs an action
* If an identity is known, resume normal loading, as the user has already "connected" their wallet, so the wallet should not prompt.

julien51:
It's now time to test this on staging using the followin scenarios:
1. no crypto wallet (Safari)
   * Nothing should happen, until the user clicks on the "unlock" button.
   * page state should be "locked" 
   * The "no wallet" message should be shown.
2. crypto wallet, user has *never* approved the site
   * Metamask should not prompt the user when the page loads
   * page state should be "locked"
   * When the user clicks on the "unlock" button:
      * MetaMask should prompt to connect
      * The checkout modal should open
      * Page state should be "locked"
3. crypto wallet, user *already* approved the site, but did not purchase
   * Metamask should not prompt the user when the page loads
   * Page state should be "locked"
   * When the user clicks on the "unlock" button:
      * The checkout modal should open
3. crypto wallet, user *already* approved the site, and previously purchased a key
   * Metamask should not prompt the user when the page loads
   * Page state should be "unlocked"
      * The checkout modal should open

julien51:
In test case 1. I can confirm that indeed, there is an issue where the first time the clicks on the button, nothing happens. The second time, it does load the modal. Everything else is happening has expected. (more below)

Case 2. ✅
Case 3. ✅
Case 4. ✅

julien51:
My understanding of the issue is 1. is that the `PostMessages.UPDATE_ACCOUNT` [event may not be triggered when there is no web3](https://github.com/unlock-protocol/unlock/pull/5517/files#diff-c3b2363f95c0188b18fdd376bda9349dR118), because indeed, there is no account...

Now, I wonder if it is strictly required to have this event having been triggered to open the checkout modal? Maybe @cnasc could clarify here.

[...]

PR #5519, opened 2019-12-19
Don't defer startup for users with no wallet

cnasc:
# Description

This PR checks whether we have an injected wallet, and only does the deferred paywall startup if there is one. If not, we can just let the startup run because it should immediately end up in the expected state of showing the no wallet message.

# Issues
Refs #5518 

# Checklist:

### DIFF
--- a/paywall/src/unlock.js/startup.ts
+++ b/paywall/src/unlock.js/startup.ts
@@ -1,3 +1,4 @@
+import { PaywallConfig } from 'src/unlockTypes'
 import { UnlockWindowNoProtocolYet } from '../windowTypes'
 import IframeHandler from './IframeHandler'
 import Wallet from './Wallet'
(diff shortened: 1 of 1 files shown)


paywall/src/unlock.js/startup.ts
```diff
@@ -141,34 +141,41 @@ export default function startupWhenReady(
...
-          startup(
-            (window as unknown) as UnlockWindowNoProtocolYet,
-            startupConstants,
-            true
-          )
-        },
+  const web3Present = !!(window as any).web3
+
+  // Only try to do the deferred setup when we have a wallet. If we
+  // don't, we can do setup rright away, since it will just result in
```
julien51:
```suggestion
  // don't, we can do setup right away, since it will just result in
```

For once, I am the one telling you about a typo 🏅

PR #5530, opened 2019-12-20
web3Service now accepts an optional network number

### LINKED ISSUE #5526: Reduce the number of net version calls being made in production
After some inspection, it looks like we are making a ton of 'net version' calls that do not need to be made. 

We have isolated a probable case of the issue;  the instance of provider being used does not provide the network requiring an additional call to be made. 

An adjusting passing the network should ideally resolved the issue. 

https://docs.ethers.io/ethers.js/html/api-providers.html

julien51:
# Description

This is now ready for review!

# Issues
Fixes #5526

# Checklist:

### DIFF
--- a/unlock-js/src/web3Service.js
+++ b/unlock-js/src/web3Service.js
@@ -20,10 +20,11 @@ export default class Web3Service extends UnlockService {
     unlockAddress,
     blockTime,
     requiredConfirmations,
+    network,
   }) {
     super({ unlockAddress })
 
-    this.setup(readOnlyProvider)
+    this.setup(readOnlyProvider, network)
     this.blockTime = blockTime
     this.requiredConfirmations = requiredConfirmations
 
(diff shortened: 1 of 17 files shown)


unlock-js/src/web3Service.js
```diff
@@ -120,10 +121,14 @@ export default class Web3Service extends UnlockService {
    * Temporary function that allows us to use ethers functionality
    * without interfering with web3. This will be moved to the constructor when
    * we remove web3
+   * TODO: ^ assess this?
```
cnasc: +1

unlock-js/src/web3Service.js
```diff
@@ -120,10 +121,14 @@ export default class Web3Service extends UnlockService {
...
    */
-  setup(readOnlyProvider) {
+  setup(readOnlyProvider, network = 1) {
```
cnasc: I'm curious if this default is the correct choice. I think it might be better to leave it entirely undefined, and have the fallback behavior be for it to get the network ID on its own, like it does now.

julien51: That's a very good point. I'll leave as undefined.

PR #5532, opened 2019-12-20
Force paywall to initialize on the paywall-with-crypto-wallet test

cnasc:
# Description

This PR re-enables one of the skipped paywall integration tests. In order to make the test run successfully, in the beforeAll we load the page once, set a cached account address in localStorage, and then reload the page so we have a clean start for the remaining tests.

I think we can use this pattern for all of the currently skipped tests, but I'll put them up individually so we can gradually bring the tests back in.

# Issues

Fixes #
Refs #

# Checklist:


tests/test/paywall-with-crypto-wallet.test.js
```diff
@@ -55,6 +55,10 @@ describe.skip('The Unlock Ad Remover Paywall (logged in user)', () => {
       )}&provider=${encodeURIComponent(url.readOnlyProvider())}&logindelay=0`
     )
     await page.goto(testUrl, { waitUntil: 'networkidle2' })
+    await page.evaluate(() => {
+      localStorage.setItem('__unlockProtocol.accountAddress', '0xwhatever')
```
cnasc: Using an obviously invalid account address here so that if we ever
come to depend on the account address stored there, the failure will
be loud and clear, prompting us to handle it correctly for that case.

PR #5563, opened 2020-01-02
Purging setEncryptedPrivateKey from unlock-app

cnasc:
# Description

There was a TODO in the `privateKeyReducer` file indicating that it isn't used anymore:
> // TODO: this is not used: cleanup!

As a result, I've removed the reducer, its tests, and the action that would trigger it. However, the account ejection component relies on the presence of this value in the state to provide the encrypted key download. This is a lot of code just to handle that one use-case, so I think it would make more sense to use a hook to download the JSON blob.

# Checklist:


cnasc:
In the end, we'll still need to provide the email address, so we'd still have some Redux involved. Better to remove the TODO and update the file and module for clarity.

PR #5591, opened 2020-01-04
adding generateLockAddress to the inputHandler in order to determine the upcoming lock address when it is not known

julien51:
# Description

I previously forgot that the inputsHandlers also called generateLockAddress...
This will let us "guess" the address of a new lock after a page refresh!

# Checklist:

### DIFF
--- a/unlock-js/src/web3Service.js
+++ b/unlock-js/src/web3Service.js
@@ -77,13 +77,15 @@ export default class Web3Service extends UnlockService {
     }
 
     this.inputsHandlers = {
-      createLock: async (transactionHash, contractAddress, params) => {
+      createLock: async (transactionHash, contractAddress, sender, params) => {
         // The annoying part here is that we do not have the lock address...
         // Since it is not an argument to the function.
         // We will 'guess' it again using generateLockAddress
         // knowing that this may create a race condition another lock creation pending transaction
         // exists.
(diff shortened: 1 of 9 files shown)


unlock-js/src/web3Service.js
```diff
@@ -77,13 +77,15 @@ export default class Web3Service extends UnlockService {
     }
 
     this.inputsHandlers = {
-      createLock: async (transactionHash, contractAddress, params) => {
+      createLock: async (transactionHash, contractAddress, sender, params) => {
         // The annoying part here is that we do not have the lock address...
         // Since it is not an argument to the function.
         // We will 'guess' it again using generateLockAddress
         // knowing that this may create a race condition another lock creation pending transaction
```
NickCuso: With the change to `create2` it's no longer a "guess".  If there's a conflict (e.g. same user reusing a salt) the createLock transaction will fail.

NickCuso:
Looks like some tests need to be updated to pass the new `sender` param.  Or maybe would it make sense to add that field to the `params` object?

julien51:
HA, yes, I just pushed the tests ;) thanks!

unlock-js/src/web3Service.js
julien51: Good point! fixed the comment,

PR #5613, opened 2020-01-06
Starting to enable yarn

cnasc:
# Description

The basic form of this PR involves converting each subrepo to yarn individually.

1. Pull deps from root repo into each subproject
2. Replace usage of `npm` with equivalent `yarn` usage in `package.json`
3. Generate `yarn.lock`
4. Update relevant dockerfile to use yarn to get deps

It's a lot of changed files, but almost without exception they all follow that pattern above. One key difference is that the `@types` definition for Next is out of sync with Next itself, so when we pull in `setConfig` in `jest.setup.ts` in some repos, we get an error. I'm not sure why we only see this breakage when using with `yarn`, but changing the setup file to '.js' resolves it and tests run correctly. That issue will be resolved when we upgrade to Next 9, because it is bundled with first-party type definitions.

When this goes live, `yarn` will have to be installed globally to work with our project. The overall workflow remains the same for the time being, a future PR will bring in the Yarn monorepo workflow features. Additionally, you will have to use `yarn test`, for example to run tests. All usages of `npm` to run a script will have to be replaced with `yarn` as in https://yarnpkg.com/en/docs/cli/

This PR deprecates `link-parent-bin`. As a result, we'll want to audit deps for each subrepo to make sure we only pull in things we need (which we should really do anyway, because there has been a lot of `package.json` copypasta).

# Checklist:


scripts/postinstall.sh
```diff
@@ -6,13 +6,13 @@
 
 if [ "$SKIP_SERVICES" != "true" ]; then
 
-  SERVICES=( paywall smart-contracts unlock-app locksmith tests tickets newsletter wedlocks unlock-js unlock-protocol.com docker/development )
-
+  SERVICES=( smart-contracts tests tickets wedlocks unlock-js unlock-protocol.com docker/development )
+  ROOT_DIR=$(pwd)
```
cnasc: There was a bug before where in the last iteration of this loop we would be in `docker/development` and then `cd ..` leaving us in `docker/`, so the later `rm` commands wouldn't do anything. Always `cd`ing to a full path fixes that.

cnasc:
@akeem @julien51 this PR is now ready to go. Let me know if you have any questions about the process

PR #5619, opened 2020-01-07
Adding the ability to update token address

### LINKED ISSUE #5486: We should allow lock owner to change the token that the lock is priced in.
Use case #1.) A lock has been created and is priced in SAI. The Lock owner now wants to switch this to be priced in DAI.

We should provide a clear path to make this happen, as well as define how refunds/cancelations are handled. This mostly comes down to deciding which token a key purchaser should receive. Ideally if they paid in SAI they should be refunded in SAI, but this presents some challenges:
- there is currently no way to select which token to refund. Whatever tokenAddress is set in the lock contract is what will be used for a refund. We could work around this by adding a `_tokenAddress` param to our refund functions, _BUT_ if the lock pricing is changed to DAI, there's no guarantee that there would be enough SAI left in the contract to handle refunds... as the owner may have withdrawn it. 
For that matter, if I bought a key with SAI, and then the owner switched the lock to be priced in DAI, and then I want a refund immediately, there may be 0 DAI in the contract to grant me my refund.

We could make it so that in order to update the tokenAddress, a few things need to happen. In the SAI/DAI example above:
1.) the lock could retain enough of a reserve of SAI to handle refunds for the owners who paid with SAI
2.) the lock owner has to send the lock contract enough DAI to replace the remaing SAI in the contract, which is transferred to the lock owner.

I'm thinking 1.)would be nice, but maybe too much work for too little value, while 2.) is probably a necessity. Thoughts?
cc @julien51 @NickCuso

nfurfaro:
# Description
This adds a param to the `updateKeyPrice` function (now called `updateKeyPricing), as the pricing is really a function of 2 things; the tokenaddress and the price in said tokens. In most cases, updating 1 will require updating the other. It is still possible to update only one of these params by (for example) passing the existing keyPrice as the new key price. This is because passing 0 as the new keyprice is a valid input and is used to makes keys free, just as passing 0 as the token address is a valid input when switching the pricing to use ETH.

# Issues

Fixes #5486
Refs #

# Checklist:


smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -45,6 +45,11 @@ contract IPublicLock is IERC721Enumerable {
     uint keyPrice
   );
 
+  event TokenAddressChanged(
```
NickCuso: This is redundant with PricingChanged

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -116,9 +121,11 @@ contract IPublicLock is IERC721Enumerable {
    * A function which lets the owner of the lock to change the price for future purchases.
    * @dev Throws if called by other than owner
    * @dev Throws if lock has been disabled
+   * @dev Throws if _tokenAddress is not a valid token
    * @param _keyPrice The new price to set for keys
+   * @param _tokenAddress The address of the erc20 token to use for pricing the keys
```
NickCuso: ..or address(0) to use ETH

smart-contracts/contracts/interfaces/IPublicLock.sol
nfurfaro: Yes, this is an artifact of the first approach I took.

smart-contracts/contracts/interfaces/IPublicLock.sol
nfurfaro: yep. done.

PR #5636, opened 2020-01-09
refactored function which checks if locks are deployed

julien51:
# Description

This should actually die when the expected locks have not been deployed...
A much bigger refactor of this is needed. We'll get down to this soon.

# Checklist:

### DIFF
--- a/docker/wedlocks.dockerfile
+++ b/docker/wedlocks.dockerfile
@@ -2,7 +2,6 @@ FROM unlock-core
 
 # Dependencies for wedlocks
 RUN mkdir /home/unlock/wedlocks
-COPY --chown=node wedlocks/package-lock.json /home/unlock/wedlocks/.
 COPY --chown=node wedlocks/package.json /home/unlock/wedlocks/.
 WORKDIR /home/unlock/wedlocks
 RUN yarn --production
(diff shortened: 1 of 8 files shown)


tests/helpers/locksAreDeployed.js
```diff
@@ -17,6 +17,15 @@ function resolveWhenDeployed(address, delay, maxAttempts) {
   return new Promise((resolve, reject) => {
     function retrieveCode() {
       attempts += 1
+
+      const retryOrReject = () => {
```
akeem: It may be worthwhile for us to tighten up this method in the future, I could see future us tracing this to figure out how it is supposed to work

docker/wedlocks.dockerfile
```diff
@@ -2,7 +2,6 @@ FROM unlock-core
 
 # Dependencies for wedlocks
 RUN mkdir /home/unlock/wedlocks
-COPY --chown=node wedlocks/package-lock.json /home/unlock/wedlocks/.
```
cnasc: Is this based on an older version of master? The one that went in yesterday should have changed all of these to `yarn.lock`

julien51: hum. I think it's rebased :/

PR #5628, opened 2020-01-08
Replace `npm run` with `yarn`

cnasc:
# Description

Found a number of places in the code where we still used `npm` instead of `yarn` for script running. Additionally, some documentation still had the old usage.

This PR swaps in `yarn` for the sake of consistency.

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
  - [x] This PR only contains configuration changes (package.json, etc.)
  - [ ] This PR only contains code changes (if configuration changes are required, do a separate PR first, then re-base)
- [ ] My code follows the style guidelines of this project, including naming conventions
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If my code adds or changes components, I have written corresponding stories with Storybook
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #5637, opened 2020-01-09
Some more npm cleanup!

julien51:
# Description

Some more npm cleanup!

# Issues

Refs #5628

# Checklist:


docker/wedlocks.dockerfile
```diff
@@ -2,7 +2,6 @@ FROM unlock-core
 
 # Dependencies for wedlocks
 RUN mkdir /home/unlock/wedlocks
-COPY --chown=node wedlocks/package-lock.json /home/unlock/wedlocks/.
```
cnasc: Same comment as in the other PR, I think this branch is missing the `yarn.lock` files

cnasc: Even if these are ones that slipped through the last PR, we should replace `package-lock.json` with the associated `yarn.lock`, because without it we may not always get the same set of deps on each build.

cnasc:
Additionally, `rover` wasn't included in the first pass at `yarn` because it wasn't one of the subrepos managed by `postinstall.sh`, so we shouldn't delete its `package-lock.json` until it has yarn in place.

julien51:
Good points. I'll hold on this one.

PR #5645, opened 2020-01-09
Smart-contracts uses Eslint 6

cnasc:
# Description

Updating the smart contracts to use Eslint 6. All of the code changes in here were done automatically by `eslint --fix`. For the ones it could not do automatically, I added them as exceptions in the `.eslintrc.js`

# Checklist:

### DIFF
--- a/smart-contracts/.eslintrc.js
+++ b/smart-contracts/.eslintrc.js
@@ -1,3 +1,19 @@
+const rulesToIgnore = [
+  'no-underscore-dangle',
+  'no-param-reassign',
+  'no-use-before-define',
+  'no-plusplus',
+  'no-await-in-loop',
+  'radix',
+  'prefer-destructuring',
+  'no-shadow',
+  'no-loop-func',
+  'eqeqeq',
+  'no-useless-concat',
+  'prefer-const',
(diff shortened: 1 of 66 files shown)


cnasc:
Looks like the automatically-applied fixes break the tests by assuming later JS features than are available in this subrepo. Not sure if we do any transpiling here, but we'll either need to change it or introduce it if we want to use the same lint config as everything else.

NickCuso:
Hmm not sure, but maybe this is because truffle uses babel 6 instead of babel 7 (?)

cnasc:
None of these are critical since we only use JS for tests here; we lived without them before this update. Not updating ESLint for the time being is a valid path.

NickCuso:
Hmm we could also look for which lint rule is breaking our tests and then turn those rules off in config.  That could allow us to keep a consistent style across repos.

cnasc:
The one that shows the failure is this one: https://eslint.org/docs/rules/prefer-object-spread

I'll try to peel it back further and see if I can make it pass, not sure if any others will fail after this is resolved.

cnasc:
Closing in favor of #5656

PR #5657, opened 2020-01-10
Trimming root package.json since we don't use link-parent-bin now

cnasc:
# Description

This PR removes some deps from the root `package.json`; most of these were in place because we were copying them down to the child repos. Since we no longer use `link-parent-bin`, we no longer need them at this level. AFAIK the minimum deps at this level are `husky` and `lint-staged`; we can also probably remove `husky` and `lint-staged` in the child repos, but I'll save that for another PR.

This probably reduces the size of the core docker image, and should reduce its build time a little.

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,17 +26,13 @@
     "eslint-plugin-react": "7.17.0",
     "eslint-plugin-react-hooks": "2.3.0",
     "eslint-plugin-standard": "4.0.1",
-    "jest": "24.9.0",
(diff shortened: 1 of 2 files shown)


cnasc:
Ah, misunderstood the way the eslint config worked. In that case, we should probably keep those at this level and then remove them from the child repos.

cnasc:
Going to have to re-evaluate this

PR #5659, opened 2020-01-10
Reorganize dependencies at root level

cnasc:
# Description

Shuffling some deps around, removing unneeded ones and putting others in dev dependencies to minimize our image size.

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,7 @@
     "@typescript-eslint/eslint-plugin": "2.14.0",
     "@typescript-eslint/parser": "2.14.0",
     "babel-eslint": "10.0.3",
-    "babel-jest": "24.9.0",
-    "eslint": "5.16.0",
+    "eslint": "6.8.0",
     "eslint-config-airbnb": "18.0.1",
     "eslint-config-prettier": "6.9.0",
     "eslint-config-standard": "14.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,8 +10,7 @@
     "@typescript-eslint/eslint-plugin": "2.14.0",
     "@typescript-eslint/parser": "2.14.0",
     "babel-eslint": "10.0.3",
-    "babel-jest": "24.9.0",
-    "eslint": "5.16.0",
+    "eslint": "6.8.0",
```
cnasc: ESLint does need to be installed at this level, the config can't resolve things it needs without it. However, moving this to a higher version doesn't break subrepos that haven't been updated so it looks like the final source of truth is the version of ESLint that initiates the run.

package.json
```diff
@@ -26,18 +25,15 @@
...
-    "ts-jest": "24.2.0",
     "typescript": "3.7.4"
   },
   "engines": {
     "node": "^10"
   },
   "devDependencies": {
-    "husky": "3.1.0"
+    "husky": "3.1.0",
+    "lint-staged": "8.2.1"
```
cnasc: `lint-staged` is only ever used when committing, which doesn't happen in CI

PR #5669, opened 2020-01-10
Reduces the size of the unlock-core image

akeem:
# Description

As part speeding things up, this is a step toward reducing the unlock-core image.  The main step taking place here is that we are removing the apk installed packages after installing the base node dependencies

# Issues

Fixes #
Refs #

# Checklist:


akeem:
A bit more research needs to be done here, I think i'm encountering some things when using this images as a base for some of the other images locally

julien51:
ohhh interesting. How much "lighter" is it?

akeem:
It is about 200mb or so lighter. Some of the images that inherit from this one require python so a bit of further redistribution needs to occur

PR #5684, opened 2020-01-13
Add keyId to keychain

cnasc:
# Description

This PR adds the `keyId` field from the graph onto the keychain so that users can see which token they have

image

# Checklist:


unlock-app/package.json
```diff
@@ -36,6 +36,7 @@
     "babel-loader": "8.0.6",
     "babel-plugin-require-context-hook": "1.0.0",
     "babel-plugin-styled-components": "1.10.6",
+    "babel-preset-react-app": "^9.1.0",
```
cnasc: Storybook failed to start without this. Not certain why, but it makes sense to add it here.

PR #5691, opened 2020-01-14
Update the Ganache version in Docker

NickCuso:
# Description

Switching ganache back to a stable build.

BTW is `ganache.dockerfile` still applicable to anything?  It had a rather old version so would not be working with current code..

# Issues

Fixes #
Refs #

# Checklist:


akeem:
@NickCuso `ganache.dockerfile` is currently being used in the development environment via `docker-compose.override.yml`. It is responsible for deploying the unlock contract and some initial locks. It appears to be working as expected, where are some of the places that it would exhibit unexpected behavior with the previous version?

NickCuso:
Oh my bad, I was working on a feature that includes using chainid but we set that down a few weeks ago.  I was thinking it was already included.

So the next contract update should require this docker change.

PR #5692, opened 2020-01-14
Moved unlockProvider back into unlock-app

julien51:
# Description

This required surprising little changes beyond paths
I had to install a package (`nock`) for tests (maybe a follow up PR to see if we can remove that dependency...) and did a few cosmetic changes/renames.

# Checklist:


julien51:
strangely, tests pass when I do `npm run test src/__tests__/services/unlockProvider.test.js` so they must be conflicting with something else.

unlock-app/src/__tests__/services/unlockProvider.test.js
```diff
@@ -1,40 +1,196 @@
...
+
+        expect(
+          sigUtil.recoverPersonalSignature({
+            data: messageHash,
+            sig,
+          })
+        ).toEqual(publicKey.toLowerCase())
+      })
+
+      it('should be compatible with walletService', async done => {
```
cnasc: Weird that this test failed

julien51: Probably some network call. trying to see why.

julien51: What's weirder is that it passes locally.

julien51: I found it!

cnasc: What was the problem?

julien51: We are making a network call to get the network id! 

But apparently there is smething else.

PR #5696, opened 2020-01-14
Add lock owner to keyholder query

cnasc:
# Description

For the verification page, we need to be able to see if the user on the page is also the owner of the lock in question. Adding this property to the query will allow us to determine that.

# Checklist:


julien51:
What Akeem said! LG otherwise

Issue #5701, opened 2020-01-14
Port metadata storage and retrieval functions to unlock-js


cnasc:
Currently we have [implementations of these operations](https://github.com/unlock-protocol/unlock/blob/master/newsletter/utils/token.js) in the Newsletter. We can move these into `unlock-js` and make them a bit more generic.

- [x] less manual work, take advantage of existing signature methods in `unlock-js`
- [x] make them more generic, allow storing and querying arbitrary values
- [x] support storage of metadata for keyholder
- [x] support storage of metadata for lock owner

This will enable us in the immediate future to mark tickets as checked in, and should make it easier to work with metadata for other uses in the future. One interesting feature I discussed with @akeem earlier is making an internal test page where we can stick metadata on a token, which can be a bit easier than using the scripts we have in `locksmith`

julien51:
This is going to have a large impact in several parts of the app, which is exciting:
* the [paywall will be able to "collect" data before a user purchases a key](https://github.com/unlock-protocol/unlock/issues/5693)
* the keychain will be able to show the data that a token has stored (as well as maybe edit it!)

cnasc:
This is done!

PR #5702, opened 2020-01-15
Adding typed data for metadata uploads to unlock-js

cnasc:
# Description

This PR adds the necessary typed data for metadata uploads to unlock-js. We'll use this to implement the metadata upload methods in a future PR. Based on the existing scripts and sample data in locksmith.

# Issues

Refs #5701 

# Checklist:


cnasc:
>     * the lock metadata which acts as default for the key metadata
The template for this particular metadata isn't included in this PR, but if we brought it in would likely go in the `lockOwnerMetadata.js` file (or whatever we end up calling it)

>     * the key metadata which is not not specific to the owner of the key (like its icon?... etc)
This is included in this PR, and is the one in `lockOwnerMetadata.js`. My understanding is that this metadata is only set by the lock owner and is the kind of metadata we would use to store whether or not a given ticket has been checked in (correct me if I'm wrong on that)

>     * the key owner metadata which is data which is related to the owner of the key (email address in the context of a ticket for example.)
This is the template included in 'keyHolderMetadata.js`. My understanding is that this metadata is only set by the key holder.

The general scheme I had for naming these was based on who sets the data, not what the data pertains to. I can make that more clear in filenames, like `lockOwnerControlledMetadata` for example. Alternately, I can move things around so that they're grouped by what they operate on (keys vs locks), but I think we'll want to indicate somehow who controls which set of metadata so the code is more understandable.

cnasc:
@julien51 updated and ready for a fresh look

PR #5730, opened 2020-01-17
[draft] add setKeyMetadata to unlock-js

cnasc:
# Description

This is a draft PR, going up just so I can annotate some things inline as I go along

# Issues

Refs #5701

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/integration/walletServiceIntegration.test.js
+++ b/unlock-js/src/__tests__/integration/walletServiceIntegration.test.js
@@ -5,16 +5,19 @@ import { waitForContractDeployed } from '../helpers/waitForContractDeployed'
 import 'cross-fetch/polyfill'
 
 let host
+let locksmithHost
 const port = 8545
 const locksmithPort = 8080
 if (process.env.CI) {
   host = 'ganache-integration'
+  locksmithHost = 'locksmith'
 } else {
   host = '127.0.0.1'
+  locksmithHost = '127.0.0.1'
 }
 
 const provider = `http://${host}:${port}`
-const locksmithBaseUrl = `http://${host}:${locksmithPort}`
+const locksmithBaseUrl = `http://${locksmithHost}:${locksmithPort}`
 
 // This test suite will do the following:
 // For each version of the Unlock contract
--- a/unlock-js/src/__tests__/integration/walletServiceIntegration.test.js
+++ b/unlock-js/src/__tests__/integration/walletServiceIntegration.test.js
@@ -32,7 +32,7 @@ let accounts
 
 // Tests
 describe('Wallet Service Integration', () => {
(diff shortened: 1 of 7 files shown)


unlock-js/src/walletService.js
```diff
@@ -88,17 +89,18 @@ export default class WalletService extends UnlockService {
 
     if (!accounts.length) {
       // We do not have an account, can't do anything until we have one.
-      return (this.ready = false)
+      this.ready = false
```
cnasc: This is a lint change, only picked up in the pre-commit hook for whatever reason. It doesn't like assigning in a return statement (and I agree)

unlock-js/src/walletService.js
```diff
@@ -88,17 +89,18 @@ export default class WalletService extends UnlockService {
...
-    let address = accounts[0]
+    const address = accounts[0]
 
     this.emit('account.changed', address)
     if (this.provider.emailAddress) {
       this.emit('account.updated', { emailAddress: this.provider.emailAddress })
     }
     this.emit('ready')
-    return Promise.resolve(address)
+    return address
```
cnasc: `async` functions automatically `Promise.resolve` their returns

unlock-js/src/walletService.js
```diff
@@ -245,7 +248,7 @@ export default class WalletService extends UnlockService {
       callback(null, transaction.hash)
     }
 
-    return await this.provider.waitForTransaction(transaction.hash)
+    return this.provider.waitForTransaction(transaction.hash)
```
cnasc: see: https://eslint.org/docs/rules/no-return-await

unlock-js/src/walletService.js
```diff
@@ -289,20 +292,23 @@ export default class WalletService extends UnlockService {
...
-
-      callback(null, Buffer.from(result).toString('base64'))
+      const result = await this.unformattedSignTypedData(account, data)
+      return callback(null, Buffer.from(result).toString('base64'))
     } catch (err) {
       return callback(err, null)
     }
   }
 
+  async unformattedSignTypedData(account, data) {
```
cnasc: This is where the real changes begin. The `signData` function base64s its return value (I think this was originally used for tickets or something). I split it in 2:

- `unformattedSignTypedData`, which just does the signature so we can use it more generically
- `signData`, which uses the former function but returns it in base64 to maintain the expected interface.

unlock-js/src/__tests__/integration/walletServiceIntegration.test.js
```diff
@@ -29,7 +32,7 @@ let accounts
 
 // Tests
 describe('Wallet Service Integration', () => {
-  const versions = ['v0', 'v01', 'v02', 'v10', 'v11', 'v12', 'v13']
+  const versions = [/* 'v0', 'v01', 'v02', 'v10', 'v11', 'v12', */ 'v13']
```
cnasc: This will be cleaned up once I have passing tests

PR #5735, opened 2020-01-17
Update eslint in unlock-js to v6

cnasc:
# Description

Like the updates from the early part of this year, this PR updates eslint in `unlock-js`, removes unneeded lint deps, applies some automated fixes, and ignores everything we couldn't automate.

It looks like the cause of some of the issues with the pre-commit hook were because husky runs at the root of the monorepo, which has a newer version of eslint than `unlock-js` does. This brings them in sync with each other and should completely alleviate that pain point.

# Checklist:


cnasc:
(this will allow me to commit and push my tests on the key metadata PR!)

PR #5751, opened 2020-01-17
Add library to enrich token information with email address when appropriate

akeem:
# Description

Add library to enrich token information with email address when appropriate

# Issues

Fixes #
Refs #

# Checklist:


nudge/src/emailList.ts
```diff
@@ -0,0 +1,21 @@
+import { generateKeyMetadata } from './metadataOperations'
```
julien51: Can we keep the filename consistent with the function name here?

PR #5754, opened 2020-01-20
adding call to initialize when deploying template

julien51:
# Description

This is part of the post mortem from last week.
Having this inside of the library means we are less likely to "forget"!

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/v13/initializeTemplate.test.js
+++ b/unlock-js/src/__tests__/v13/initializeTemplate.test.js
@@ -1,0 +1,66 @@
+import WalletService from '../../walletService'
+import v13 from '../../v13'
+import { ZERO } from '../../constants'
+
+let walletService
+
+const transaction = {
+  hash: '0xtransactionHash',
+}
+
+// Mock the lockContract (can we use the abi here?)
+const lockContract = {
+  initialize: jest.fn(() => Promise.resolve(transaction)),
+}
+const templateAddress = '0xtemplate'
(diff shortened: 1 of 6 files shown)


unlock-js/src/__tests__/v13/initializeTemplate.test.js
```diff
@@ -0,0 +1,66 @@
+import WalletService from '../../walletService'
+import v13 from '../../v13'
+import { ZERO } from '../../constants'
+
+let walletService
+
+const transaction = {
```
julien51: I took an extra effort to NOT use nock in this as I really want to get rid of it inside of our unlock-js unit tests. 
I believe it is cleaner this way

unlock-js/src/v13/initializeTemplate.js
```diff
@@ -0,0 +1,30 @@
+import { ZERO } from '../constants'
+
+/**
+ * Initialize a lock (will be called on the template only)
+ * @param {PropTypes.contract} contract
+ */
+export default async function({ templateAddress }, callback) {
+  const signer = this.provider.getSigner()
+  const _owner = await signer.getAddress()
```
cnasc: Do these need to be underscored? It looks like there are no collisions inside this function definition

julien51: Probably not, I'll change them. It's just that these are the argument names on the smart contract and I used the same ;)

PR #5771, opened 2020-01-20
refactoring tests for v0/createLock.js

julien51:
# Description

This is the first of many pull requests which removes our crazy "nock" setup in unlockjs unit tests.
Basicaly rather than mock things at the network level, we mock things inside of the class itself.
Integratiion tests do cover the the fact that we test what ethers does so there is no need to test
it in unit tests.

I believe this will bring a lot of simplication/cleanup when that is done.
I am commenting the PR for more details

# Checklist:

### DIFF
--- a/unlock-js/src/v0/createLock.js
+++ b/unlock-js/src/v0/createLock.js
@@ -1,5 +1,5 @@
 import ethersUtils from '../utils'
-import { GAS_AMOUNTS, ETHERS_MAX_UINT } from '../constants'
+import { ETHERS_MAX_UINT } from '../constants'
 import TransactionTypes from '../transactionTypes'
 import { UNLIMITED_KEYS_COUNT } from '../../lib/constants'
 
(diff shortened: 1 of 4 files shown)


unlock-js/src/v0/createLock.js
```diff
@@ -14,15 +14,10 @@ export default async function(lock, callback) {
...
   }
-  const transactionPromise = unlockContract.functions[
-    'createLock(uint256,uint256,uint256)'
-  ](
+  const transactionPromise = unlockContract.createLock(
     lock.expirationDuration,
     ethersUtils.toWei(lock.keyPrice, 'ether'),
-    maxNumberOfKeys,
-    {
-      gasLimit: GAS_AMOUNTS.createLock,
```
julien51: ethers is actually smart enough to extract the expected gas price for a transaction, so there is no need to pass one.

unlock-js/src/v0/createLock.js
```diff
@@ -14,15 +14,10 @@ export default async function(lock, callback) {
   if (maxNumberOfKeys === UNLIMITED_KEYS_COUNT) {
     maxNumberOfKeys = ETHERS_MAX_UINT
   }
-  const transactionPromise = unlockContract.functions[
-    'createLock(uint256,uint256,uint256)'
-  ](
+  const transactionPromise = unlockContract.createLock(
```
julien51: This syntax is much easier to handle/deal with.

PR #5793, opened 2020-01-22
Removed price polling whih is generating re-rendering of our apps when the odds of a price changing significantly are not that important

julien51:
# Description

In an effort to make the app MUCH snappier I am removing this polling mechanism which is IMO pre-mature optimization.
The price we show is *just* indicative and the odds of it changing significantly in the time spent on the page are very slim anyway.

Eventually we should move this to a hook (if we can memoize the value)

# Checklist:


julien51:
Doh!

PR #5801, opened 2020-01-23
Downgrade svgr to 4.3.3

### LINKED ISSUE #5719: svg-2-components script doesn't produce usable output
At the moment, running the `svg-2-components` script produces broken output in several ways:

- Sizes, colors, and position of existing SVG react components get changed (icons off-center, etc)
- It adds a `titleId` prop that we don't pass, which breaks any TypeScript components using SVGs, because it interprets that prop as required

This will get in the way of us working with SVGs in the future (as @smombartz and I discovered this morning). We should fix the script and its associated template and config so that this won't be a problem.

cnasc:
# Description

v5 of svgr breaks our app. It's really not critical to upgrade, and too much work to do so, so we should stick with 4.3.3 until we have a _really_ compelling reason to switch.

# Issues

Fixes #5719 
Refs #

# Checklist:

### DIFF
--- a/unlock-app/package.json
+++ b/unlock-app/package.json
@@ -75,7 +75,7 @@
     "validator": "12.1.0"
   },
   "devDependencies": {
-    "@svgr/cli": "5.0.1",
+    "@svgr/cli": "4.3.3",
     "nodemon": "2.0.2",
     "npm-check": "5.9.0"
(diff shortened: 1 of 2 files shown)


cnasc:
CC @smombartz

PR #5803, opened 2020-01-23
CheckoutWrapper now holds the header and footer itself

### LINKED ISSUE #5787: CheckoutWrapper component should include header and footer
With the addition of metadata collection to the paywall, we need to augment the `<CheckoutWrapper>` component with the header (typically site icon included in paywall config) and footer ("Powered by Unlock"). This change is necessary because the checkout has grown to a size that makes it hard to manage without factoring into smaller pieces -- not having to worry about duplicating the header and footer everywhere will make it easier.

cnasc:
# Description

This PR finishes moving the header and footer parts of the checkout into the wrapper component. This will make it easier to keep them consistent in the future (in working on this I found that the size of the icon would vary between views) and also make it easier to add new views in the future.

No real visible changes:
image

# Issues

Fixes #5787 

# Checklist:

### DIFF
--- a/paywall/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
+++ b/paywall/src/__tests__/storybook/__snapshots__/storyshots.test.js.snap
@@ -1696,27 +1696,6 @@ Object {
   fill: white;
 }
 
-.c14 {
-  margin-top: 50px;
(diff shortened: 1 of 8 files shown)


julien51:
snapshots ;)

PR #5804, opened 2020-01-23
using hook to retrieve user locks

julien51:
# Description

This PR just adds the hook and its tests.
The next one wil replace the redux logic

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useLocks.js
+++ b/unlock-app/src/hooks/useLocks.js
@@ -1,0 +1,77 @@
+import { useState, useEffect } from 'react'
+import { Web3Service } from '@unlock-protocol/unlock-js'
+import { UNLIMITED_KEYS_COUNT } from '../constants'
+
+import GraphService from '../services/graphService'
+import configure from '../config'
+
+const config = configure()
+
+/**
+ * A hook which yields locks
+ */
+/**
+ * This hook yields the list of locks for the owner based on data from the graph and the chain
+ * @param {*} address
+ */
+export const useLocks = owner => {
+  // Let's retrieve the locks!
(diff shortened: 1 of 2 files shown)


unlock-app/src/hooks/useLocks.js
```diff
@@ -0,0 +1,77 @@
...
+import GraphService from '../services/graphService'
+import configure from '../config'
+
+const config = configure()
+
+/**
+ * A hook which yields locks
+ */
+/**
+ * This hook reverse resolves any Ethereum address using the Ethereum Name Service
```
cnasc: 🤔

unlock-app/src/hooks/useLocks.js
```diff
@@ -0,0 +1,77 @@
...
+    subgraphURI,
+    readOnlyProvider,
+    unlockAddress,
+    blockTime,
+    requiredConfirmations,
+    requiredNetworkId,
+  } = config
+  const graphService = new GraphService(subgraphURI)
+
+  const web3Service = new Web3Service({
```
cnasc: Not sure if it will matter in this case, but we might conceivably instantiate multiple `Web3Service`s if this hook (or others like it) gets wider use. If they do we might investigate making sure that Web3Service is a singleton that each hook can get a reference to.

unlock-app/src/hooks/useLocks.js
julien51: Oups! Let me fix that!

unlock-app/src/hooks/useLocks.js
julien51: Good point. I think no state is shared in Web3Service so we're probably good.

PR #5806, opened 2020-01-23
Adding nudge staging deployment configuration

akeem:
# Description

There are some caveats to be considered here. 

* The configuration does indeed leverage a good portion of Locksmith's environment variables. This is mostly to ensure that they are using the same database config. The second reason being the elimination of cut and paste errors at this stage of development.

* Due to an ELB requirement a random port has been exposed.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/nudge/scripts/deploy-elasticbeanstalk-docker.sh
+++ b/nudge/scripts/deploy-elasticbeanstalk-docker.sh
@@ -6,7 +6,6 @@ db_name=$5
 db_hostname=$6
 is_forked_pr=$8
 build_id=$9
-message="${10}"
 web3_provider_host=${13}
 graphql_base_url=${15}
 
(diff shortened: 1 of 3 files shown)


nudge/scripts/deploy-elasticbeanstalk-docker.sh
```diff
@@ -0,0 +1,50 @@
...
+db_name=$5
+db_hostname=$6
+is_forked_pr=$7
+build_id=$8
+web3_provider_host=${9}
+graphql_base_url=${10}
+
+function check_is_forked_pr()
+{
+    if [ is_forked_pr = "true" ]; then
```
cnasc: Can you make an issue for this? I think it bears investigating further, though it doesn't block this PR

PR #5807, opened 2020-01-23
lock retrieval moved to hook from redux

julien51:
# Description

This means that the locks are now not retrieved when the visitor is not on the dashboard page.

# Checklist:


julien51:
Ok, I found a flaw in that approach. I will have to also move the lock creation process to the hook.
Closing this PR in the meantime.

PR #5841, opened 2020-01-27
Nudge add lockname email

akeem:
# Description

The Nudge code has been updated to ensure that the Lock's name is retrieved so that it can be included in the templated information passed along to the email service

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/nudge/src/emailList.ts
+++ b/nudge/src/emailList.ts
@@ -1,29 +1,27 @@
 import { generateKeyMetadata } from './generateKeyMetadata'
 
 export async function extractEmails(results: any[]): Promise<Key[]> {
-  let keysData = results.map((result: any) =>
-    enrichKeyWithEmailAddress(result.lockAddress, result.keyId)
-  )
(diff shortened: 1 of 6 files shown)


nudge/src/emailList.ts
```diff
@@ -1,29 +1,27 @@
 import { generateKeyMetadata } from './generateKeyMetadata'
 
 export async function extractEmails(results: any[]): Promise<Key[]> {
-  let keysData = results.map((result: any) =>
-    enrichKeyWithEmailAddress(result.lockAddress, result.keyId)
-  )
+  let keysData = results.map((result: any) => enrichKeyWithEmailAddress(result))
```
cnasc:
```suggestion
  let keysData = results.map(enrichKeyWithEmailAddress)
```

Really minor suggestion, up to you if you want to use it

akeem: It looks like ts isn't a huge fan of this. I will probably need to tighten up the types but we can do the tightening before we move this from staging to prod!

Issue #5728, opened 2020-01-17
Change event used when renewing your purchase


NickCuso:
**Describe the bug**
If you purchase 2 copies of the same key, some screens on etherscan consider this 2 tokens: https://etherscan.io/token/0x91002310cec24d818f42a169dd0d27ecbfa8bcd5?a=0xdb92c096bc5efa8adb48f05cd601dddb75228203#inventory

**To Reproduce**
1. purchase a key
2. purchase it again

**Expected behavior**
Per EIP-721 we should emit Transfer(from: 0) when minting.  But our use case is a bit unique.

I think we could address this by creating a new event type.  Emit Transfer only for the first purchase (when the tokenId is assigned).  Then when a token is renewed we can emit a new `Renewal` event instead.

nfurfaro:
We could stop emitting Transfer for renewals and just emit TimestampChanged in the timeMachine function. We already fire this event, we would just need to refactor a bit to use the timeMachine function inside `purchase` when it's a renewal/top-up.

julien51:
Let's use `ExpirationChanged` so it is more explicit than `timestamp`...

nfurfaro:
Yeah, we could change the existing event name. It's probably not being used anywhere yet so I doubt it would break anything except a few tests.

nfurfaro:
I just noticed this as well.

There is another location in the code which could potentially cause some confusion with the transfer event, though I don’t think we’re seeing it show up yet. In the shareKey function, I believe the Transfer event  will sometimes emit twice under certain(likely common) conditions.

When sharing with a recipient who has no key, we emit `Transfer(address(0), _to, iDTo)` to signal the new key creation.
We then proceed to emit `emit Transfer( keyOwner, _to, iDTo` 
below this to signal the transfer of time from the existing key to the new one.
I think the second event should actually be removed, as the call to timeMachine already emits the `TimestampChanged` event once for each key under most cases. 

https://github.com/unlock-protocol/unlock/blob/3bf5e79ad77321ca4256c0aad25dd38797640cb8/smart-contracts/contracts/mixins/MixinTransfer.sol#L93

PR #5859, opened 2020-01-29
Moving the _timeMachine function to MixinKeys

nfurfaro:
# Description

This is preparation for refactoring the `purchase` function to use `_timeMachine`. This would get us to a point where we consistently reuse the same logic  in `shareKeys`, `transferFrom` and `purchase`. This should also pave the way for refactoring the events we emit around these flows, as well as where/when we emit them. 

# Issues

Fixes #
Refs #5728

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinKeys.sol
@@ -23,8 +23,10 @@ contract MixinKeys is
     uint expirationTimestamp;
   }
 
-  // Called when the Lock owner expires a user's Key
+  // Emitted when the Lock owner expires a user's Key
   event ExpireKey(uint indexed tokenId);
+
+  // Emitted when the expiration of a key is modified
   event ExpirationChanged(
     uint indexed _tokenId,
     uint _amount,
--- a/smart-contracts/contracts/mixins/MixinKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinKeys.sol
@@ -2,6 +2,7 @@ pragma solidity 0.5.14;
 
 import '@openzeppelin/contracts-ethereum-package/contracts/ownership/Ownable.sol';
 import './MixinLockCore.sol';
+import '@openzeppelin/contracts-ethereum-package/contracts/math/SafeMath.sol';
 
 
 /**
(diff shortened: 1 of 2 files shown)


smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -22,6 +25,11 @@ contract MixinKeys is
 
   // Called when the Lock owner expires a user's Key
   event ExpireKey(uint indexed tokenId);
+  event ExpirationChanged(
```
NickCuso: Other event needs to be removed.

And should we put it in the interface?  We've been inconsistent about that.
To avoid redundancy - maybe we ONLY put events in the interface.

nfurfaro: Actually this is something I've been thinking as well. I do think events should ONLY be defined in the interface. I will move this one over.

nfurfaro: > Other event needs to be removed.
are you referring to the `ExpireKey` event? That one has to stay... `ExpirationChanged` is just a renaming of `TimestampChanged`

NickCuso: https://github.com/unlock-protocol/unlock/blob/n44o-issue-5728-2/smart-contracts/contracts/mixins/MixinTransfer.sol#L33

nfurfaro: Ahh, yes. Thx. 
Regarding putting all the events in the interface, I guess we'll need to import & inherit the interface in pretty much all the mixins. Should be fine, as IPublicLock is abstract.

nfurfaro: @NickCuso I'm thinking that  moving all event definitions into the interface could be a separate PR. I'm pretty sure that when inheriting the interface in each Mixin, we're going to hit a bunch of circular dependency errors.
One possible (cleaner?) approach would be to split up our current abstract `IPublicLock` into "pure" interfaces (one for each mixin) each inherited by its respective mixin. `IPublicLock` could inherit each of these small interfaces as well, and then PublicLock still only has to inherit the single interface. I will create an issue around these ideas if you think its worth pursueing?

nfurfaro: New issue here: https://app.zenhub.com/workspaces/unlock-5bfd5c894b5806bc2bf85d03/issues/unlock-protocol/unlock/5866

NickCuso:
Could you remove the redundant event?
How is the test passing?

PR #5861, opened 2020-01-30
Add setUserMetadata wrapper for paywall

cnasc:
# Description

This PR contains a simple wrapper over the unlock-js method for setting user metadata so that we can conveniently access it in the paywall. Next step will be to render the form and hook it up with this function when metadata is present in the config.

# Checklist:

### DIFF
--- a/paywall/src/utils/setUserMetadata.ts
+++ b/paywall/src/utils/setUserMetadata.ts
@@ -1,0 +1,27 @@
+import { WalletService } from '@unlock-protocol/unlock-js'
+import { UserMetadata } from '../unlockTypes'
+import configure from '../config'
+
+export function setUserMetadata(
+  lockAddress: string,
+  userAddress: string,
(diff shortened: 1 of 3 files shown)


paywall/src/utils/setUserMetadata.ts
```diff
@@ -0,0 +1,27 @@
...
+  lockAddress: string,
+  userAddress: string,
+  metadata: UserMetadata,
+  callback: (error: any, value: any) => void
+) {
+  const { unlockAddress, locksmithUri } = configure()
+
+  // TODO: investigate having a singleton WalletService to avoid any
+  // instantiation overhead if we start to use it more frequently
+  const walletService = new WalletService({ unlockAddress })
```
julien51: I think this migt require to call `.connect` because that's how the provider object is set and the provider is then used by `walletService.setUserMetadata` to sign?

PR #5863, opened 2020-01-30
Include network id in Web3ServiceParams

cnasc:
# Description

The types exported from `unlock-js` are slightly wrong, because it doesn't recognize that `Web3Service` needs to take the network id in the constructor. This PR updates that. When it gets published we can revisit #5862 and replace some of the types defined there with the exported ones.

# Checklist:

### DIFF
--- a/unlock-js/index.d.ts
+++ b/unlock-js/index.d.ts
@@ -7,7 +7,7 @@ interface Web3ServiceParams {
   unlockAddress: string
   blockTime: number
   requiredConfirmations: number
-  network: 1 | 4 | 1984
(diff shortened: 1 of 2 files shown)


unlock-js/index.d.ts
```diff
@@ -7,6 +7,7 @@ interface Web3ServiceParams {
   unlockAddress: string
   blockTime: number
   requiredConfirmations: number
+  network: 1 | 4 | 1984
```
akeem: Would you mind adding the id for kovan here as well?

PR #5868, opened 2020-01-30
Shifting subproject specific Dockerfile a part of a process to clean setups

akeem:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/newsletter/Dockerfile
+++ b/newsletter/Dockerfile
@@ -1,0 +1,14 @@
+FROM unlock-core
+
+# Dependencies for newsletter
+RUN mkdir /home/unlock/newsletter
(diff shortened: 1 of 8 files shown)


akeem:
> LGTM
> 
> Is the end result of this process that the docker directory goes away, or just that it's simplified?

@cnasc  the docker directory will end up being simplified and so should any scripts that we have the currently operate on the files. 

The dockerfiles in the docker directory currently effectively require that the images be built from there. We will shift to a slightly easier to reason about model where the Dockerfile in the project can be built from their and the scripts currently in play are less knowledgeable about each project specific dockerfile

PR #5869, opened 2020-01-30
Bump unlock-js to 0.8.3

cnasc:
# Description

Not having up-to-date types in the paywall is getting annoying, so I'm cutting 0.8.3 to get the proper type definitions in place. I won't publish until this is merged.

# Checklist:

### DIFF
--- a/unlock-js/CHANGELOG.md
+++ b/unlock-js/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changes
 
-# Next Version
+# 0.8.3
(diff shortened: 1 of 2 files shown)


cnasc:
> Lets see if can find a way to reduce the associate pain here and make the releases such a lift

This one's a bit of a special case, it only happened because the Paywall had it's own hand-rolled types from before we exported types from unlock-js, and replacing those couldn't happen without this. Ideally in the future all type changes will happen in lock step with unlock-js feature releases, so we shouldn't have this dance.

PR #5872, opened 2020-01-31
[DRAFT] Add useSetUserMetadata hook to paywall

cnasc:
# Description

example for @julien51 

Example usage here would be something like
```javascript

const [ metadata, setMetadata ] = useState(undefined)
const { status } = useSetUserMetadata(metadata)

// ... other code happens
// when we want to set the user metadata, we do this
setMetadata(someMetadata)

// changing the state causes a render, which causes the hook
// to fire again and make the call to walletService
```

It may be possible to wrap the `useState` part inside the hook, and expose a setter as part of the exports

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Thanks for this @cnasc :) I think we can close now!

PR #5876, opened 2020-01-31
adding withWalletService and related context

julien51:
# Description

Rather than instantiating the walletService inside of the walletMiddleware we do it at the `_app` level and then pass it down to the middleware and the context so it can be consumed in several places in the app.

Pretty sure some tests will fail... let me see which ones from Circle ;p

# Checklist:


unlock-app/src/__tests__/middlewares/walletMiddleware.test.js
```diff
@@ -76,17 +76,6 @@ class MockWalletService extends EventEmitter {
 }
 
 let mockWalletService = new MockWalletService()
-
```
julien51: Removing this below feels good actually :)

PR #5886, opened 2020-02-01
allowing lock owner to check people in

julien51:
# Description

This adds the ability for lock owners to mark keys as checked in.

# Checklist:


julien51:
image

unlock-app/src/__tests__/components/interface/VerificationStatus.test.tsx
```diff
@@ -131,43 +110,4 @@ describe('VerificationStatus', () => {
 
     getByText('Key Invalid')
   })
-
-  it('should render a message to indicate that the key is valid when applicable', () => {
```
julien51: I had to remove this test because it was becoming really hard to maintain as it required mocking the behavior/hooks inside of the `ValidKey` making it less and less of a unit test.

In practice, this test broke even though nothing in the logic of `VerificationStatus` actually changed :/

PR #5889, opened 2020-02-03
Add messaging infrastructure for setting user metadata

cnasc:
# Description

I forgot just how labyrinthine this subsystem could be. This PR adds the necessary messages for the checkout iframe to request that metadata be set for a user and for the data iframe to indicate that it did so successfully. I'll annotate inline, because there are some pieces here that are a little different from the other messages.

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/Mailbox.ts
+++ b/paywall/src/data-iframe/Mailbox.ts
@@ -14,7 +14,7 @@ import {
   isAccountOrNull,
   isValidBalance,
 } from '../utils/validators'
-import { PaywallConfig, PurchaseKeyRequest } from '../unlockTypes'
+import { PaywallConfig, PurchaseKeyRequest, UserMetadata } from '../unlockTypes'
 import {
   IframePostOfficeWindow,
   ConsoleWindow,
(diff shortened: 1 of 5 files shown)


paywall/src/data-iframe/Mailbox.ts
```diff
@@ -316,6 +320,33 @@ export default class Mailbox {
...
+      this.emitError(
+        new Error('blockchain handler not instantiated, cannot set metadata')
+      )
+      return
+    }
+
+    try {
+      await this.handler.setUserMetadata(lockAddress, metadata)
+      // success, no error returned
+      this.postMessage(PostMessages.SET_USER_METADATA_SUCCESS, undefined)
```
cnasc: The success branch has its own message. Typically the BlockchainHandler doesn't return successes, because they are eventually reflected in the data it emits. This use is different, because it's purely a side-effectful method that doesn't change the chain data we return.

paywall/src/data-iframe/Mailbox.ts
```diff
@@ -316,6 +320,33 @@ export default class Mailbox {
...
+    } catch (error) {
+      this.emitError(error)
```
cnasc: There is already a generic `PostMessages.ERROR` that is used for failures, and I opted to continue using it here for consistency, so there is no specific metadata error postmessage.

PR #5900, opened 2020-02-04
adding useMembers hook

julien51:
# Description

This will drammatically simplify how we load data on the members page...

# Checklist:

### DIFF
--- a/unlock-app/src/queries/keyholdersByLock.js
+++ b/unlock-app/src/queries/keyholdersByLock.js
@@ -13,6 +13,7 @@ export default function keyholdersByLockQuery() {
         }
         name
         address
+        owner
       }
     }
   `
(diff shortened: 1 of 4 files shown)


unlock-app/src/queries/keyholdersByLock.js
```diff
@@ -13,6 +13,7 @@ export default function keyholdersByLockQuery() {
         }
         name
         address
+        owner
```
julien51: This is used to only ask the user to check their wallet if they are the lock owner.

unlock-app/src/services/storageService.js
```diff
@@ -418,6 +418,7 @@ export class StorageService extends EventEmitter {
       )
 
       this.emit(success.getBulkMetadataFor, lockAddress, result.data)
+      return result.data
```
julien51: Easier to use this as a promise than an event emitter...

cnasc:
LG, coverage just needs to be bumped

PR #5901, opened 2020-02-04
Show metadata form when purchasing a key

cnasc:
# Description

This PR introduces a metadata collection flow to key purchases.

In essence, when `metadataInputs` are specified in the paywall config the checkout modal will now prompt for the user to fill out the fields before the key purchase occurs.

There is probably more that can be done to refine this particular flow, but it will take bigger shifts in the paywall to get there.

# Checklist:

### DIFF
--- a/paywall/src/components/content/NewDemoContent.tsx
+++ b/paywall/src/components/content/NewDemoContent.tsx
@@ -62,6 +62,19 @@ export default function NewDemoContent() {
       }),
       {}
     ),
+    metadataInputs: [
+      {
+        name: 'First Name',
+        type: 'text',
+        required: true,
+        public: true,
+      },
+      {
+        name: 'Last Name',
+        type: 'text',
+        required: true,
+      },
+    ],
     callToAction: {
       default:
         'You have reached your limit of free articles. Please purchase access to continue reading',
--- a/paywall/src/components/content/NewDemoContent.tsx
+++ b/paywall/src/components/content/NewDemoContent.tsx
@@ -49,6 +49,8 @@ export default function NewDemoContent() {
   const names: string[] = url.searchParams.getAll('name')
   const unlockUserAccounts: boolean =
     url.searchParams.get('unlockUserAccounts') === 'true'
+  const useMetadataInputs: boolean =
+    url.searchParams.get('metadataInputs') === 'true'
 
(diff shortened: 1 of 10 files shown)


cnasc:
image

The current code works to a degree. It will show the metadata form when inputs are specified, metamask will pop up to do the signature, then the key purchase will initiate (result of successful run shown above).

However, the way we show the wallet check message clips the checkout component out of the tree, which interrupts all the listeners. If we show the wallet check when getting the signature, it breaks the flow. So before this is ready for prime time we'll have to figure out a better way to work with that.

cnasc:
As far as I can tell, the resolution here is going to be moving the wallet check message down a layer into the `<Checkout>` component.

cnasc:
> As far as I can tell, the resolution here is going to be moving the wallet check message down a layer into the `<Checkout>` component.

This did not resolve the issue

cnasc:
Right now there is almost no unit-test coverage of the checkout flow, everything is deferred to the integration tests. LMK if integration tests should be added as part of this PR or as a separate follow-up.

I'm happy to demo the paywall with metadata collection in localdev if that will ease review.

julien51:
I think it would be good to write integration tests but I worry this is going to end up quite intense for logic which we'll soon move to the unlock-app so let's write some when we get there.

paywall/src/components/content/NewDemoContent.tsx
```diff
@@ -49,6 +49,8 @@ export default function NewDemoContent() {
   const names: string[] = url.searchParams.getAll('name')
   const unlockUserAccounts: boolean =
     url.searchParams.get('unlockUserAccounts') === 'true'
+  const useMetadataInputs: boolean =
+    url.searchParams.get('metadataInputs') === 'true'
```
julien51: 👍

paywall/src/components/interface/MetadataForm.tsx
```diff
@@ -26,7 +37,10 @@ export const MetadataForm = ({ fields, onSubmit }: Props) => {
           </Label>
         ))}
       </FieldWrapper>
-      <ActionButton type="submit">Continue</ActionButton>
+      {submittedForm && (
```
julien51: 👍

PR #5871, opened 2020-01-30
Refactor Purchase to use timeMachine

nfurfaro:
# Description
This fixes the issues around events described in issue 5728, and simplifies the branching logic in `purchase` a fair bit.

# Issues

Fixes #5728 
Refs #

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [x] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #5904, opened 2020-02-04
Transfer event bugfix

### LINKED ISSUE #5728: Change event used when renewing your purchase

nfurfaro:
# Description
This pr replaces #5871 . It is a more targeted approach which only tackles the issue at hand.
I've added a couple extra tests for the new usage of the `ExpirationChanged` event in the `purchase` function (only for renewals/repurchases)
- still working on optimizing gas for this. The new event used _should_ be costing less. It has 3 params like `Transfer`, but only one of them is indexed...

# Issues

Fixes #5728 
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinPurchase.sol
+++ b/smart-contracts/contracts/mixins/MixinPurchase.sol
@@ -52,15 +52,24 @@ contract MixinPurchase is
       // Assign a new tokenId (if a new owner or previously transferred)
       _assignNewTokenId(toKey);
       _recordOwner(_recipient, toKey.tokenId);
+
+      // trigger event
+      emit Transfer(
+        address(0), // This is a creation.
+        _recipient,
+        toKey.tokenId
+      );
     }
 
     if (toKey.expirationTimestamp >= block.timestamp) {
       // This is an existing owner trying to extend their key
       toKey.expirationTimestamp = toKey.expirationTimestamp.add(expirationDuration);
+      emit ExpirationChanged(toKey.tokenId, expirationDuration, true);
     } else {
       // SafeAdd is not required here since expirationDuration is capped to a tiny value
       // (relative to the size of a uint)
       toKey.expirationTimestamp = block.timestamp + expirationDuration;
+      emit ExpirationChanged(toKey.tokenId, expirationDuration, true);
     }
 
     // Let's get the actual price for the key from the Unlock smart contract
@@ -81,13 +90,6 @@ contract MixinPurchase is
     }
 
     unlockProtocol.recordKeyPurchase(inMemoryKeyPrice, getHasValidKey(_referrer) ? _referrer : address(0));
-
-    // trigger event
-    emit Transfer(
-      address(0), // This is a creation.
-      _recipient,
-      toKey.tokenId
-    );
 
     // We explicitly allow for greater amounts of ETH or tokens to allow 'donations'
     if(tokenAddress != address(0)) {
--- a/smart-contracts/contracts/mixins/MixinPurchase.sol
+++ b/smart-contracts/contracts/mixins/MixinPurchase.sol
@@ -49,11 +49,14 @@ contract MixinPurchase is
 
     // Assign the key
     Key storage toKey = keyByOwner[_recipient];
+    uint newTimeStamp;
 
     if (toKey.tokenId == 0) {
       // Assign a new tokenId (if a new owner or previously transferred)
       _assignNewTokenId(toKey);
       _recordOwner(_recipient, toKey.tokenId);
+      newTimeStamp = block.timestamp + expirationDuration;
+      toKey.expirationTimestamp = newTimeStamp;
 
       // trigger event
       emit Transfer(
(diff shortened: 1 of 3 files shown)


smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -52,15 +52,24 @@ contract MixinPurchase is
...
 
     if (toKey.expirationTimestamp >= block.timestamp) {
       // This is an existing owner trying to extend their key
       toKey.expirationTimestamp = toKey.expirationTimestamp.add(expirationDuration);
+      emit ExpirationChanged(toKey.tokenId, expirationDuration, true);
     } else {
       // SafeAdd is not required here since expirationDuration is capped to a tiny value
       // (relative to the size of a uint)
       toKey.expirationTimestamp = block.timestamp + expirationDuration;
+      emit ExpirationChanged(toKey.tokenId, expirationDuration, true);
```
NickCuso: Hmm here we technically added more than just expirationDuration... 

I'm not sure this matters though :)  We might not need this level of detail - maybe in another PR we could change the event to `ExpirationChanged(tokenId, newExpiration)`.

Hmm another thought on that, using the user's address in the event is probably more useful than tokenId given that each user can only have one id.  That would allow front end implementations to listen for events filtered by the user's address.  So maybe `ExpirationChanged(address user, uint newExpiration)`

nfurfaro: >we technically added more than just expirationDuration

Hmm. Yeah, I should probably change this to be using `block.timestamp + expirationDuration` as the amount in the event. I was trying to reuse the existing event `ExpirationChanged` here, which I originally created for use in the `timeMachine` function. The purpose there was/is to log the amount of time added or removed to a given key. It seems like it would not achieve this if we logged the `newExpiration` there instead... Logging the owner's address instead of the tokenId could work. It makes me reconsider trying to reuse this event vs creating a new one specifically tailored to key renewals...

nfurfaro: It makes me reconsider trying to reuse this event vs creating a new one specifically tailored to key renewals, as the semantics are a bit different (logging the amount of time added or removed from a key, vs logging the updated expiration. I realize that one can be derived from the other, so I'm not feeling too strongly either way about reusing the existing event vs creating a new one.

NickCuso: A new one for this use case seems good.  Maybe `RenewKeyPurchase(address indexed owner, uint newExpiration)`

NickCuso: or maybe to be closer to the purchase event: `RenewKeyPurchase(address indexed owner, uint indexed tokenId)`.  ... or maybe both tokenId and newExpiration.

hmm event best practices need some work :)

nfurfaro: Ok. I think this is probably a better way forward too. I'll make the changes and push one more time. I don't want to be guessing about the meaning of values in events, or introduce more event-related bugs down the road.

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -64,12 +66,12 @@ contract MixinPurchase is
...
       // This is an existing owner trying to extend their key
       toKey.expirationTimestamp = toKey.expirationTimestamp.add(expirationDuration);
-      emit ExpirationChanged(toKey.tokenId, expirationDuration, true);
+      emit RenewKeyPurchase(_recipient, toKey.expirationTimestamp.add(expirationDuration));
     } else {
       // SafeAdd is not required here since expirationDuration is capped to a tiny value
       // (relative to the size of a uint)
       toKey.expirationTimestamp = block.timestamp + expirationDuration;
-      emit ExpirationChanged(toKey.tokenId, expirationDuration, true);
+      emit RenewKeyPurchase(_recipient, block.timestamp + expirationDuration);
```
NickCuso: Hmm I'm just noticing this will emit on fist purchase - right?

nfurfaro: Dang it. yes it will. but of course we don't want it to.

nfurfaro: The flow for a new (first) purchase needs to skip this entire `if/else` block

NickCuso: yea - we have 3 use cases here.  i think we can if elseif else it.  1) new purchase 2) extend a non-expired key 3) repurchase an expired key

nfurfaro: yup. copy that.

nfurfaro:
Ok, I fixed the branching, and modified / added some tests where needed. No rush , but would be great if you can give it one more look.

PR #5908, opened 2020-02-04
Fix usePaywallConfig loop

cnasc:
# Description

This is a partial fix for the weird issue we found on local dev today. Calling `usePaywallConfig` twice caused a loop that a) rendered the paywall nonfunctional on Firefox (and sometimes Chrome) and b) would grind my machine to a halt in relatively short order.

Removing the second call and making the component take the config as a param stops the bleeding, although it doesn't explain why we saw different behavior in different environments.

# Checklist:

### DIFF
--- a/paywall/src/components/content/CheckoutContent.tsx
+++ b/paywall/src/components/content/CheckoutContent.tsx
@@ -65,6 +65,10 @@ export const Wrapper: React.FunctionComponent<WrapperProps> = ({
  */
 export default function CheckoutContent() {
   const window = useWindow()
+
+  // This is the ONLY place where usePaywallConfig should be
+  // called. Calling it anywhere else will introduce loops that break the
+  // app.
   const paywallConfig = usePaywallConfig()
   const { account, network, locks, checkWallet } = useBlockchainData(
     window,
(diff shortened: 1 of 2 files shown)


paywall/src/components/content/CheckoutContent.tsx
```diff
@@ -38,8 +40,6 @@ export const Wrapper: React.FunctionComponent<WrapperProps> = ({
     })
   }, [postMessage])
 
-  const paywallConfig = usePaywallConfig()
```
julien51: Can you add a comment inside of `usePaywallConfig` saying that calling it twice WILL create issues?

julien51: so that we don't end up in the same situation soon...

julien51:
This also makes me wonder why this is not happening on staging. Any clue?

cnasc:
> This also makes me wonder why this is not happening on staging. Any clue?

Still a mystery to me

julien51:
ok, well if you confirm that;'s fixed locally, let's proceed because that's where we had the problem :/

paywall/src/hooks/usePaywallConfig.js
```diff
@@ -5,6 +5,14 @@ import { PostMessages } from '../messageTypes'
 import useListenForPostMessage from './browser/useListenForPostMessage'
 import usePostMessage from './browser/usePostMessage'
 
+/**
+ * NOTE: Do not use this hook anywhere other than CheckoutContent. It
+ * is not safely written, so having more than one instance of it will
+ * introduce major breakage.
+ *
+ * We should consider it deprecated at this point.
```
julien51: 🗑

PR #5913, opened 2020-02-05
Be able to convert form submission to metadata format

cnasc:
# Description

Required for #5901 

This PR introduces a small util that converts the output of the metadata form (a string:string dictionary) into the metadata format we need for user metadata. For maximum flexibility we can specify in the paywall config whether a field should be public. If not, it is assumed to be protected. 

# Checklist:

### DIFF
--- a/paywall/src/unlockTypes.ts
+++ b/paywall/src/unlockTypes.ts
@@ -78,6 +78,7 @@ export interface MetadataInput {
   name: string
   type: 'text' | 'date' | 'color' | 'email' | 'url'
   required: boolean
+  protected?: true // optional, all non-protected fields are treated as public
 }
 
 // This interface describes an individual paywall's config
--- a/paywall/src/unlockTypes.ts
+++ b/paywall/src/unlockTypes.ts
@@ -78,7 +78,7 @@ export interface MetadataInput {
   name: string
   type: 'text' | 'date' | 'color' | 'email' | 'url'
   required: boolean
-  protected?: true // optional, all non-protected fields are treated as public
+  public?: true // optional, all non-public fields are treated as protected
 }
 
 // This interface describes an individual paywall's config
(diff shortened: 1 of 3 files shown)


paywall/src/utils/userMetadata.ts
```diff
@@ -0,0 +1,30 @@
...
+  let result: { [key: string]: boolean } = {}
+  inputs.forEach(input => (result[input.name] = input.protected || false))
+  return result
+}
+
+export function formResultToMetadata(
+  formResult: { [key: string]: string },
+  inputs: MetadataInput[]
+): UserMetadata {
+  let result: UserMetadata = {
```
julien51: Let's make sure we have the same format as used on https://github.com/unlock-protocol/unlock/blob/master/newsletter/utils/token.js

paywall/src/unlockTypes.ts
```diff
@@ -78,6 +78,7 @@ export interface MetadataInput {
   name: string
   type: 'text' | 'date' | 'color' | 'email' | 'url'
   required: boolean
+  protected?: true // optional, all non-protected fields are treated as public
```
julien51: Since this is user centric data (from the owner) let;s play it safe and make it all protected for now.

paywall/src/utils/userMetadata.ts
```diff
@@ -0,0 +1,30 @@
...
+    publicData: {},
+    privateData: {},
```
julien51: `protectedData` ;)

paywall/src/utils/userMetadata.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  const protectedInputs = getProtectedInputs(inputs)
+  Object.keys(formResult).forEach(name => {
+    if (protectedInputs[name]) {
+      result.privateData![name] = formResult[name]
```
julien51:
```suggestion
      result.protectedData![name] = formResult[name]
```

cnasc:
@julien51 ready for another look

PR #5926, opened 2020-02-07
CheckoutContent reads paywall config from URL param

cnasc:
# Description

This PR makes the `CheckoutContent` component parse the paywall config from a URL param if it is present. No visible change here yet, the prop is just available to the component for future use.

One surprisingly huge part of this PR was moving the associated validators over from the paywall subrepo.

# Checklist:

### DIFF
--- a/unlock-app/jest.config.js
+++ b/unlock-app/jest.config.js
@@ -21,7 +21,7 @@ module.exports = {
       branches: 85.2,
       functions: 88.7,
       lines: 95.3,
-      statements: -165,
+      statements: -172,
     },
   },
 }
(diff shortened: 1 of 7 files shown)


unlock-app/jest.config.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
       branches: 85.2,
       functions: 88.7,
       lines: 95.3,
-      statements: -165,
+      statements: -172,
```
cnasc: The code I brought over from the paywall has a lot of uncovered `console.error` statements. Since these lines are non-critical and the new code all has 100% coverage, it makes sense to bump the allowed statements to move this along without creating busywork.

PR #5945, opened 2020-02-12
Add smart-contract-extensions repo

NickCuso:
# Description

This repo would contain contracts which interact with other contracts.  ATM it's just swap-and-call but we can add unlock-example-contracts and the KeyPurchaser.

I wasn't sure about the naming - we could change it if you have suggestions.

# Issues

Fixes #
Refs #

# Checklist:


julien51:
IIRC you need to add that image to the docker-compose cluster? I think @akeem can help...

akeem:
@NickCuso yeah you will need to add a stanza named `smart-contract-extensions` in this file https://github.com/unlock-protocol/unlock/blob/master/docker/docker-compose.ci.yml .

We have a few levels in the scripts that could/should probably be flatten, but the tests script runs the tests in the container via the service name defined in the compose file

PR #5946, opened 2020-02-12
Fix compile warnings

NickCuso:
# Description

Fix all the Solidity compile warnings.  This should not change our bytecode at all.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/Unlock.sol
+++ b/smart-contracts/contracts/Unlock.sol
@@ -159,8 +159,8 @@ contract Unlock is
    * TODO: actually implement this.
    */
   function computeAvailableDiscountFor(
-    address _purchaser, // solhint-disable-line no-unused-vars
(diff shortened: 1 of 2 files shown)


nfurfaro:
So you've switched from comment directives, to simply commenting out the parameter names. Why the `suppressCompileWarning++` ? Would we not see warnings reappear if this approach failed?

NickCuso:
The suppressCompileWarning variable thing is for the warning which says could be `pure` instead.

NickCuso:
I'm not sure if there's a better way to get rid of those.  But it's just a test artifact so this might be sufficient.

nfurfaro:
Ah, I see. Yeah, maybe this is sufficient then. Could we just mark these as `pure` to fix the warning?

NickCuso:
No, it's trying to match the erc20 abi.

PR #5959, opened 2020-02-13
Edit updateBeneficiary permissions to allow both LockManagers and the Beneficiary

### LINKED ISSUE #5952: Allow lockManager to update beneficiary
This is just re-implementing a feature we already had not long ago, based on a better understanding of the issue.

nfurfaro:
# Description
This is undoing a change made recently, while also migrating to using lockManager instead of owner. It also adds a new test file under the `test/Lock/permissions` directory

# Issues

Fixes #5952 
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinLockCore.sol
+++ b/smart-contracts/contracts/mixins/MixinLockCore.sol
@@ -164,8 +164,8 @@ contract MixinLockCore is
   function updateBeneficiary(
     address _beneficiary
   ) external
-    onlyLockManager
   {
+    require(msg.sender == beneficiary || isLockManager(msg.sender), 'ONLY_BENEFICIARY_OR_LOCKMANAGER');
     require(_beneficiary != address(0), 'INVALID_ADDRESS');
     beneficiary = _beneficiary;
   }
--- a/smart-contracts/contracts/mixins/MixinLockCore.sol
+++ b/smart-contracts/contracts/mixins/MixinLockCore.sol
@@ -164,8 +164,8 @@ contract MixinLockCore is
   function updateBeneficiary(
     address _beneficiary
   ) external
+    onlyLockManager
   {
-    require(msg.sender == beneficiary, 'ONLY_BENEFICIARY');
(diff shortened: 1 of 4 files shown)


smart-contracts/contracts/mixins/MixinLockCore.sol
```diff
@@ -164,8 +164,8 @@ contract MixinLockCore is
   function updateBeneficiary(
     address _beneficiary
   ) external
+    onlyLockManager
```
NickCuso: I thought we agreed on lockManager || beneficiary?

nfurfaro: I just caught that and renoved the request for review, but I was too late.

smart-contracts/contracts/mixins/MixinLockCore.sol
```diff
@@ -165,7 +165,7 @@ contract MixinLockCore is
     address _beneficiary
   ) external
   {
-    require(msg.sender == beneficiary, 'ONLY_BENEFICIARY');
+    require(msg.sender == beneficiary || isLockManager(msg.sender), 'ONLY_BENEFICIARY_OR_LOCKMANAGER');
```
NickCuso: Hmm should we maybe update the onlyOwnerOrBeneficiary modifier instead so that withdrawal gets the fix too?

NickCuso:
Could you also fix the title?  It can be a little confusing when we copy these into the release notes later.

smart-contracts/contracts/mixins/MixinLockCore.sol
nfurfaro: It makes sense, but I wonder if it's worth it to even have a modifier for this if it's only used twice?

NickCuso: heh yes! I question modifier best practices as well. It might be nice to do a quick test to understand the tradeoffs - but we can revisit that.

nfurfaro: sounds good

PR #5970, opened 2020-02-14
waits for loading to be false to fix flaky test

julien51:
# Description

This fixes master!

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/hooks/useLocks.test.js
+++ b/unlock-app/src/__tests__/hooks/useLocks.test.js
@@ -239,11 +239,10 @@ describe('useLocks', () => {
       })
     })
 
-    const { result, waitForNextUpdate } = renderHook(() =>
-      useLocks(ownerAddress)
-    )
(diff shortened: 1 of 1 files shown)


unlock-app/src/__tests__/hooks/useLocks.test.js
```diff
@@ -239,11 +239,10 @@ describe('useLocks', () => {
       })
     })
 
-    const { result, waitForNextUpdate } = renderHook(() =>
-      useLocks(ownerAddress)
-    )
-    await waitForNextUpdate() // First render when lock is loaded through The Graph
-    await waitForNextUpdate() // Second render when the lock is loaded through past transactions
+    const { result, wait } = renderHook(() => useLocks(ownerAddress))
+    await wait(() => {
```
julien51: instead of waiting for a number of "renders" which could actually differ (given that the the `useReducer` callback will only trigger a re-render if the data differs and the data may be different based on the timing of the data retrievals), we wait for loading to be done, which will always happen!

PR #5984, opened 2020-02-15
Add CodeRequiredHook

NickCuso:
# Description

Require the user know a code in order to purchase a key.

The test shows the use case where the lock is free if you know the code.  I'm thinking we can also offer a second lock, which has a non-zero keyPrice that people can purchase if they don't have the code.

e.g. blog about an honeypot example:

HoneyPot blog
A few paragraphs about the scam, links to a specific example, and some clues about what's going on.

Then behind the lock is a few paragraphs about with proof of manipulation and various ways to detect the scam yourself.

You can buy a key for $1 or it's free if you can find the tx that turned our contract that looks like a simple game into a honeypot scam.  Enter tx hash [0x..........]

--

Idea from #ideas slack channel

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contract-extensions/contracts/unlock-examples/hooks/CodeRequiredHook.sol
+++ b/smart-contract-extensions/contracts/unlock-examples/hooks/CodeRequiredHook.sol
@@ -53,7 +53,8 @@ contract CodeRequiredHook is WhitelistAdminRole
   }
 
   /**
-   * @notice This call allows the owner to make any call, such as calling withdrawal on the lock.
+   * @notice This call allows the owner to make any call, such as calling updateBeneficiary on the lock
+   * which would remove the code requirement.
    */
   function proxyCall(
     address _contract,
--- a/smart-contract-extensions/contracts/unlock-examples/hooks/CodeRequiredHook.sol
+++ b/smart-contract-extensions/contracts/unlock-examples/hooks/CodeRequiredHook.sol
@@ -1,0 +1,64 @@
+pragma solidity 0.5.16;
+
+import '@openzeppelin/contracts/cryptography/ECDSA.sol';
+import '@openzeppelin/contracts/access/roles/WhitelistAdminRole.sol';
+import '@openzeppelin/contracts/introspection/IERC1820Registry.sol';
+import 'hardlydifficult-ethereum-contracts/contracts/proxies/CallContract.sol';
+
+/**
+ * @notice Used with a Lock in order to require the user knows
+ * a code in order to buy.
+ * @dev Set this contract as the beneficiary of the lock.
+ */
+contract CodeRequiredHook is WhitelistAdminRole
+{
+  using CallContract for address;
+
+  IERC1820Registry public constant erc1820 = IERC1820Registry(0x1820a4B7618BdE71Dce8cdc73aAB6C95905faD24);
+
+  // `keccak256("IUnlockEventHooks_keySold")`
+  bytes32 public constant keySoldInterfaceId = 0x4d99da10ff5120f726d35edd8dbd417bbe55d90453b8432acd284e650ee2c6f0;
+
+  /**
+   * @notice The answer expressed as an address where the private key is
+   * keccak256(abi.encode(answer, lock.address))
+   */
+  address public answerAddress;
+
+  constructor(
+    address _answerAddress
+  ) public
+  {
+    answerAddress = _answerAddress;
+    erc1820.setInterfaceImplementer(address(this), keySoldInterfaceId, address(this));
+  }
+
+  /**
+   * @notice This is called by the Lock when a purchase is attempted.
+   * When this fails so does the purchase itself.
+   * @param _data contains the signature from the `answerAddress`.
+   */
+  function keySold(
+    address /* _from */,
+    address _to,
+    address /* _referrer */,
+    uint /* _pricePaid */,
+    bytes calldata _data
+  ) external view
+  {
+    // Confirm `_to` (the new keyOwner)
+    bytes32 secretHash = ECDSA.toEthSignedMessageHash(keccak256(abi.encode(_to)));
+    require(ECDSA.recover(secretHash, _data) == answerAddress, 'INCORRECT_CODE');
+  }
+
+  /**
+   * @notice This call allows the owner to make any call, such as calling withdrawal on the lock.
+   */
+  function proxyCall(
+    address _contract,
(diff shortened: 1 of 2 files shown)


NickCuso:
This contract works for a more standard use case as well: a simple redeem code.  One lock that's free (or discounted) if you have the code, and another that's paid if you don't.

NickCuso:
BTW we can also create a similar contract which offers discount codes.  e.g. enter `unlock-10` for 10% off.  We implement that by withdrawing funds from the lock and transferring them to the buyer.  So it's not the best solution - you pay full price and then get an instant refund.

julien51:
Very cool! @nfurfaro I'll trust your judgement on the solidity parts!

nfurfaro:
>Require the user know a code in order to purchase a key.

What sor t of "code" are we thinking here? a tx hash as per the Ideas slack channel, or something different?

smart-contract-extensions/contracts/unlock-examples/hooks/CodeRequiredHook.sol
```diff
@@ -0,0 +1,66 @@
...
+
+import '@openzeppelin/contracts/cryptography/ECDSA.sol';
+import '@openzeppelin/contracts/access/roles/WhitelistAdminRole.sol';
+import '@openzeppelin/contracts/introspection/IERC1820Registry.sol';
+import 'hardlydifficult-ethereum-contracts/contracts/proxies/CallContract.sol';
+
+/**
+ * @notice Used with a Lock in order to require the user knows
+ * a code in order to buy.
+ * @dev Set this contract as the beneficiary of the lock.
```
nfurfaro: So here we set the CodeRequiredHook contract as the beneficiary, but I'm not seeing how changing the beneficiary(as per comments for proxyCall) would remove the code requirement? I would have thought we'd need to update the call to this hook from the purchase function.

NickCuso:
The code can be anything - that's determined by the frontend implementation.  The test shows a string and that string could be any length - it could represent a redeem code or an answer to a question.  The frontend could lower().trim() ect to ensure things like casing do not cause it to fail.

The code could also be a number or whatever.

By feeding it through a hash, any type with effectively unlimited inputs is fine.

smart-contract-extensions/contracts/unlock-examples/hooks/CodeRequiredHook.sol
NickCuso: We check for a hook today by asking erc-1820 if the beneficiary has registered one.

NickCuso: We could either make this contract the beneficiary, or instead have the beneficiary call erc1820 to say this is the hook to use.

The downside with the latter is the beneficiary's other locks will call the hook as well.  So the hook needs to check if it's the lock of interest... which is what motivated our desire to remove 1820 completely.

nfurfaro: Ah, right. So while the new hook contract is the beneficiary, there is a registered hook (itself). Changing the beneficary means the new beneficiary won't have (by default) a registered hook, and the purchase will proceed as a normal purchase.

One question about this: If all purchase calls have to check the erc-1820 contract for registered hooks, what sort of gas-penalty do users pay for this in the cases where a hook is not used?

NickCuso: That's right.

It's a non-zero cost to use 1820 for sure, another motivation to change.

smart-contract-extensions/test/unlock-examples/hooks/codeRequiredHook.js
```diff
@@ -0,0 +1,110 @@
...
+  const keyBuyer = accounts[5]
+  let lock
+  let hookContract
+  // Never expose the answer or its PK directly in JS for any integrations
+  // Only the answer account's address can safely be exposed
+  let answerAccount
+
+  beforeEach(async () => {
+    // Ganache only setup (a no-op if erc1820 already exists)
+    await erc1820.deploy(web3)
```
nfurfaro: Maybe use the oz-testhelper here? 
https://docs.openzeppelin.com/test-helpers/0.5/api#erc1820

NickCuso: Cool - didn't know that was a thing.

But it's a different use case.  We only use interfaceIds in Solidity for this example... I never use them in JS here.

The line you tagged is about deploying the erc1820 contract to ganache

smart-contract-extensions/test/unlock-examples/hooks/codeRequiredHook.js
```diff
@@ -0,0 +1,110 @@
+const { constants, protocols } = require('hardlydifficult-ethereum-contracts')
+const erc1820 = require('erc1820')
+const { reverts } = require('truffle-assertions')
+
+const CodeRequiredHook = artifacts.require('CodeRequiredHook.sol')
+
+contract('CodeRequiredHook', accounts => {
+  const lockOwner = accounts[1]
```
nfurfaro: Minor, but could we maybe use lockCreator here moving forward? 
We'll be doing a fairly large migration away from using "owner" fairly soon. 
"lockCreato"r is pretty descriptive, and may help with the transition away from using Ownable, at least cognitively.

NickCuso:
Yes! This contract supports discount codes already - if we offer one lock with a code required that's discounted and another lock that's full price.

smart-contract-extensions/test/unlock-examples/hooks/codeRequiredHook.js
nfurfaro: oops, sorry. meant to link to this:
https://docs.openzeppelin.com/test-helpers/0.5/api#erc1820registry

NickCuso: OZ is great and all but I'm going to leave this as is for the moment.  It's the same feature, but the library I'm using is the official example from the EIP.

If/when we make this change we can do smart-contracts and smart-contract-extensions at the same time.

NickCuso: Oh! I just remembered this step is optional cause `createTestLock` will take care of it.  I'll delete to simplify the test.

PR #6013, opened 2020-02-18
blog post on portability

smombartz:
created. to be published on the 19th of Feb

image
image


unlock-protocol.com/blog/portability.md
```diff
@@ -0,0 +1,21 @@
...
+authorName: Sascha Mombartz
+publishDate: Feb 19, 2020
+description: Instead of ads following us around our access should.
+image: /static/images/blog/portability/hero.jpg
+---
+
+![Take it with you](/static/images/blog/portability/hero.jpg)
+
+
+You’ve experienced this: You‘ve bought something (or just searched for it) and now you have ads for that or a similar item following you around. What’s happening is you have two services – the shop you bought from and the ad delivery network – that are talking to you but not each other, and that leads to a not so great user experience.
```
cnasc:
```suggestion
You’ve experienced this: you‘ve bought something (or just searched for it) and now you have ads for that or a similar item following you around. What’s happening is you have two services—the shop you bought from and the ad delivery network—that are talking to you but not each other, and that leads to a not-so-great user experience. 
```

cnasc: I changed the hyphens after "two services" to be em dashes, but not sure if they display correctly here

unlock-protocol.com/blog/portability.md
```diff
@@ -0,0 +1,21 @@
...
+
+I can see a future where Spotify will check my ethereum wallet, see that I have the New York Times membership token and give me The Daily version for members without promotions. Maybe they will even create a version without ads and special content for long time or higher tier subscribers.
+
+In that future I could take advantage of many other benefits and services based on my existing purchases and memberships. Event discounts for NYTimes or Spotify members? I just need to sign a message from my wallet that checks that my membership is valid without having to sign up or share any additional information. Bought the audio book on Amazon and the publisher offers the ebook for free on their site? Easy to verify your purchase, no need to login or share your Amazon data. You’re a New York WeWork member and are in Berlin? They can easily verify your membership and give you free or discounted access without having to look you up in their system. The blockchain let’s different services talk to each other, cross checking your memberships and purchases with minimal data sharing, if you allow it. Your purchase and membership tokens function as universal keys.
```
cnasc: > I just need to sign a message from my wallet that checks that my membership is valid without having to sign up or share any additional information.


This would probably be done without requiring a signature (for example, we don't ask for a signature before unlocking a page)

unlock-protocol.com/blog/portability.md
smombartz: I hate em dashes and only use en dashes! but thats ok.

unlock-protocol.com/blog/portability.md
smombartz: you still need to connect your wallet with the site – i.e. give the site permission to read it. maybe i should have called it connect.

also can we have content discussion in google docs? much easier than here.

PR #6018, opened 2020-02-19
Be able to show a list of locks on the Checkout page

cnasc:
# Description

This PR combines a few recent pieces into a working list of locks on the checkout page. Comments inline.

image

For easy validation of these results on local dev, you can use the following URL:
```
http://localhost:3000/checkout?paywallConfig=%7B%22locks%22:%7B%220x6D9e231E65ABf68d8eA91f3D13C134BB007E4854%22:%7B%22name%22:%22ERC20%20Lock%22%7D,%220xEE9FE39966DF737eECa5920ABa975c283784Faf8%22:%7B%22name%22:%22Eth%20Lock%22%7D%7D,%22callToAction%22:%7B%22default%22:%22This%20is%20the%20default%20call%20to%20action%22,%22expired%22:%22Your%20key%20is%20past%20its%20due%20date%22,%22pending%22:%22Purchase%20pending...%22,%22confirmed%22:%22Confirmed%22,%22noWallet%22:%22Get%20a%20wallet%22%7D%7D
```

# Checklist:

### DIFF
--- a/unlock-app/src/components/content/CheckoutContent.tsx
+++ b/unlock-app/src/components/content/CheckoutContent.tsx
@@ -6,8 +6,7 @@ import BrowserOnly from '../helpers/BrowserOnly'
 import Layout from '../interface/Layout'
 import Account from '../interface/Account'
 import { pageTitle } from '../../constants'
-import { Lock } from '../interface/checkout/Lock'
-import Loading from '../interface/Loading'
+import { Lock, LoadingLock } from '../interface/checkout/Lock'
 import {
   Account as AccountType,
   Network,
@@ -55,7 +54,13 @@ export const CheckoutContent = ({
       {account && (
         <BrowserOnly>
           <Account network={network} account={account} />
-          {loading && <Loading />}
+          {loading && (
+            <div>
+              {lockAddresses.map(address => (
+                <LoadingLock key={address} />
+              ))}
+            </div>
+          )}
           {locks && (
             <div>
               {locks.map(lock => (
--- a/unlock-app/src/components/content/CheckoutContent.tsx
+++ b/unlock-app/src/components/content/CheckoutContent.tsx
@@ -6,14 +6,18 @@ import BrowserOnly from '../helpers/BrowserOnly'
 import Layout from '../interface/Layout'
 import Account from '../interface/Account'
 import { pageTitle } from '../../constants'
-import LogInSignUp from '../interface/LogInSignUp'
+import { Lock } from '../interface/checkout/Lock'
+import Loading from '../interface/Loading'
 import {
   Account as AccountType,
   Network,
   Router,
   PaywallConfig,
+  RawLock,
 } from '../../unlockTypes'
 import getConfigFromSearch from '../../utils/getConfigFromSearch'
+import { durationsAsTextFromSeconds } from '../../utils/durations'
+import { usePaywallLocks } from '../../hooks/usePaywallLocks'
 
 interface CheckoutContentProps {
   account: AccountType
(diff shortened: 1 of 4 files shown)


unlock-app/src/components/content/CheckoutContent.tsx
```diff
@@ -6,22 +6,47 @@ import BrowserOnly from '../helpers/BrowserOnly'
...
   account: AccountType
   network: Network
   config?: PaywallConfig
 }
 
-export const CheckoutContent = ({ account, network }: CheckoutContentProps) => {
+const defaultLockAddresses: string[] = []
+
+const lockKeysAvailable = (lock: RawLock) => {
+  if ((lock as any).unlimitedKeys) {
```
cnasc: We're going to want to firm up the various lock types we have. Worse, it seems that Web3Service doesn't consistently return the same set of properties for all locks, which makes it difficult to reconcile these so there are some `any`s in this code.

unlock-app/src/components/content/CheckoutContent.tsx
```diff
@@ -30,12 +55,25 @@ export const CheckoutContent = ({ account, network }: CheckoutContentProps) => {
...
+          {loading && <Loading />}
+          {locks && (
+            <div>
+              {locks.map(lock => (
+                <Lock
+                  name={lock.name}
+                  keysAvailable={lockKeysAvailable(lock)}
+                  key={lock.name}
+                  price={lock.keyPrice}
+                  symbol={(lock as any).currencySymbol || 'ETH'}
```
cnasc: Right now this code assumes that any lock that does not have a currency symbol in the payload is an ETH lock. That may revisiting in the future.

unlock-app/src/components/interface/checkout/Lock.tsx
```diff
@@ -123,8 +123,7 @@ export const TickerSymbol = styled.span`
   font-style: normal;
   font-weight: bold;
   font-size: 20px;
-  margin-left: 4px;
-  text-align: center;
+  text-align: right;
```
cnasc: Once I saw the 2 locks on top of each other, it looked nicer to have the currency symbols aligned

unlock-app/src/hooks/usePaywallLocks.ts
```diff
@@ -17,13 +17,13 @@ export const usePaywallLocks = (lockAddresses: string[]) => {
...
     const locks = await Promise.all(lockPromises)
-    setLocks(locks)
     setLoading(false)
+    setLocks(locks)
   }
 
   useEffect(() => {
     getLocks()
-  }, [lockAddresses])
+  }, [])
```
cnasc: This was running infinitely before. Since the config is statically in the URL and is not expected to change, it's ok to make this `useEffect` run once and only once.

unlock-app/src/hooks/usePaywallLocks.ts
```diff
@@ -17,13 +17,13 @@ export const usePaywallLocks = (lockAddresses: string[]) => {
     })
 
     const locks = await Promise.all(lockPromises)
-    setLocks(locks)
     setLoading(false)
+    setLocks(locks)
```
cnasc: We now stop loading before we set the locks, this way we don't show both the placeholder locks and the real locks at the same time.

unlock-app/src/components/content/CheckoutContent.tsx
julien51: Yeah, Web3Service is due for a large refactor :/

PR #6023, opened 2020-02-20
Adding the usePurchaseKey hook

cnasc:
# Description

This PR adds a hook wrapping the `purchaseKey` method of `WalletService`. Additionally, it integrates it into the `Lock` component we use on the checkout page, so it's now possible to buy a key from this page by clicking the lock.

# Checklist:

### DIFF
--- a/unlock-app/src/components/content/CheckoutContent.tsx
+++ b/unlock-app/src/components/content/CheckoutContent.tsx
@@ -12,10 +12,8 @@ import {
   Network,
   Router,
   PaywallConfig,
-  RawLock,
 } from '../../unlockTypes'
 import getConfigFromSearch from '../../utils/getConfigFromSearch'
-import { durationsAsTextFromSeconds } from '../../utils/durations'
 import { usePaywallLocks } from '../../hooks/usePaywallLocks'
 
 interface CheckoutContentProps {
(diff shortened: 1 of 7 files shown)


unlock-app/src/components/content/CheckoutContent.tsx
```diff
@@ -26,16 +24,6 @@ interface CheckoutContentProps {
 
 const defaultLockAddresses: string[] = []
 
-const lockKeysAvailable = (lock: RawLock) => {
```
cnasc: I changed this so we pass the whole lock object to the `Lock` component (so we don't bloat the props with the things needed for the hook), so all of this is now handled internally.

unlock-app/src/hooks/usePaywallLocks.ts
```diff
@@ -17,6 +17,8 @@ export const usePaywallLocks = (lockAddresses: string[]) => {
     })
 
     const locks = await Promise.all(lockPromises)
+    // getLock doesn't always return the lock address apparently
+    locks.forEach((lock, index) => (lock.address = lockAddresses[index]))
```
cnasc: Had to add this patch to `usePaywallLocks` because the locks I tried it on didn't include their addresses in the response.

unlock-app/src/hooks/usePurchaseKey.ts
```diff
@@ -0,0 +1,21 @@
...
+    setInitiatedPurchase(true)
+    walletService.purchaseKey({
+      lockAddress: lock.address,
+      owner: lock.owner!,
+      keyPrice: lock.keyPrice,
+      erc20Address: lock.currencyContractAddress,
+    })
+  }
+
+  return { purchaseKey, initiatedPurchase }
```
cnasc: We may want to track and return several different things here, like whether it completed the purchase or whether it errored in some way. Leaving them out for now until we have a concrete use for them.

PR #6024, opened 2020-02-20
tokenId bug fix

nfurfaro:
# Description
I found a bug in the way tokenId's were getting mapped to a key owner when `shareKey` was used to share a key with a new owner(someone not currently owning a key).
- Basically, the new key's tokenId was incremented correctly, but the new owner was set as the owner of the key with a tokenId of `0`. This was caused by incorrect handling of a cached value `iDTo` which wasn't refreshed after the call to `_assignNewTokenId()`

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -81,8 +81,9 @@ contract MixinTransfer is
       emit ExpireKey(_tokenId);
     }
 
-    if (toKey.tokenId == 0) {
+    if (iDTo == 0) {
       _assignNewTokenId(toKey);
+      iDTo = toKey.tokenId;
       _recordOwner(_to, iDTo);
       emit Transfer(
         address(0), // This is a creation or time-sharing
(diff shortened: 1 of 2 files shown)


smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -81,8 +81,9 @@ contract MixinTransfer is
       emit ExpireKey(_tokenId);
     }
 
-    if (toKey.tokenId == 0) {
```
NickCuso: hmm i'm not sure i understand what's going on here.  how is using iDto better than toKey.tokenId, on line 62 we set iDTo to the toKey.tokenId

nfurfaro: Fair point. This change is just on optimization, it saves 1 SLOAD by using the cached value. The fix is on line 86, where we update iDTo with the newly assigned tokenId.

NickCuso: ah i see

PR #6031, opened 2020-02-21
Checkout lock can show when balance is insufficient

cnasc:
# Description

This PR allows the new checkout UI to show when a lock is too expensive for a user per @smombartz's design. Part of doing this involved shuffling some pieces around to avoid errors with the hooks.

**Edit**
In response to @akeem's concerns, I've gone beyond the figma and made it a bit more clear that the user cannot interact with the lock when they don't have enough money.

_Normal_:
image

_Insufficient Funds_:
image

# Checklist:

### DIFF
--- a/unlock-app/src/components/content/CheckoutContent.tsx
+++ b/unlock-app/src/components/content/CheckoutContent.tsx
@@ -6,7 +6,7 @@ import BrowserOnly from '../helpers/BrowserOnly'
 import Layout from '../interface/Layout'
 import Account from '../interface/Account'
 import { pageTitle } from '../../constants'
-import { Lock, LoadingLock } from '../interface/checkout/Lock'
+import { Locks } from '../interface/checkout/Locks'
 import {
   Account as AccountType,
   Network,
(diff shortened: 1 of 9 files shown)


unlock-app/src/components/content/CheckoutContent.tsx
```diff
@@ -43,20 +40,10 @@ export const CheckoutContent = ({
...
-            </div>
-          )}
-          {locks && (
-            <div>
-              {locks.map(lock => (
-                <Lock lock={lock} key={lock.name} />
-              ))}
-            </div>
-          )}
+          <Locks
```
cnasc: Compartmentalizing this piece inside an area where we know we always have an account is important since the hooks rely on the account address to get balance.

unlock-app/src/hooks/usePaywallLocks.ts
```diff
@@ -17,8 +20,14 @@ export const usePaywallLocks = (lockAddresses: string[]) => {
     })
 
     const locks = await Promise.all(lockPromises)
-    // getLock doesn't always return the lock address apparently
-    locks.forEach((lock, index) => (lock.address = lockAddresses[index]))
+    locks.forEach((lock, index) => {
+      if (lock.currencyContractAddress) {
+        getTokenBalance(lock.currencyContractAddress)
```
cnasc: We fire off a request for token balance whenever we run into an ERC20 lock. This can result in duplicate requests, but there should never be many of these.

cnasc:
The _new_ new design implies an entirely different DOM, closing this as a result

PR #6043, opened 2020-02-24
Get token balance while fetching locks

cnasc:
# Description

Split off from #6031. When we get lock information, if a given lock is denominated in an ERC20 token, we query for and store the user's balance of that token.

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/usePaywallLocks.ts
+++ b/unlock-app/src/hooks/usePaywallLocks.ts
@@ -3,7 +3,10 @@ import { Web3Service } from '@unlock-protocol/unlock-js'
 import { Web3ServiceContext } from '../utils/withWeb3Service'
 import { RawLock } from '../unlockTypes'
 
-export const usePaywallLocks = (lockAddresses: string[]) => {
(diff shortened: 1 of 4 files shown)


unlock-app/src/hooks/usePaywallLocks.ts
```diff
@@ -17,8 +20,18 @@ export const usePaywallLocks = (lockAddresses: string[]) => {
...
 
     const locks = await Promise.all(lockPromises)
-    // getLock doesn't always return the lock address apparently
-    locks.forEach((lock, index) => (lock.address = lockAddresses[index]))
+
+    locks.forEach((lock, index) => {
+      if (lock.currencyContractAddress) {
+        // This is the easiest place to determine which tokens we need
+        // to query for the user's balance of. The values are tracked
+        // in the useGetTokenBalance hook.
```
cnasc: Unfortunately, this is a bit at-odds with the ideal of showing locks before we have an account. This is fine for now and we can figure out how to resolve it in the future.

PR #6048, opened 2020-02-25
Updates the code coverage path ignore patterns to exclude stories

akeem:
# Description

We noticed that these files where being considered in the code coverage calculation. As this metric is currently push blocking, it is reasonable that it correctly reflect the intended items under measurement.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/jest.config.js
+++ b/unlock-app/jest.config.js
@@ -14,6 +14,7 @@ module.exports = {
     '^.+\\.(js|jsx)$': 'babel-jest',
     '^.+\\.(ts|tsx)$': 'ts-jest',
   },
+  coveragePathIgnorePatterns: ['/node_modules/', 'src/stories/.*/*.stories.js'],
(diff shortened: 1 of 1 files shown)


julien51:
Thanks, would it be worth quickly adding all of the stories in other applications as well?

julien51:
Also, as part of this PR let's update the `coverageThreshold` once this has run once?

akeem:
> Also, as part of this PR let's update the `coverageThreshold` once this has run once?

I've updated the branch to reflect the thresholds as they stand now.  TBH it might end up blocking a bit of code but we'll see how that shakes out

akeem:
> Thanks, would it be worth quickly adding all of the stories in other applications as well?

I think this will serve as a good template for those looking to make similar changes in other projects

cnasc:
`Jest: Uncovered count for statements (170)exceeds global threshold (168)` close!

PR #6051, opened 2020-02-25
KeyPurchaser

### LINKED ISSUE #5939: KeyPurchaser
When devising about strategies to allow for recurring payments, we identified the ability to implement a `keyPurchaser` contract whose role is to let 3rd parties renew keys on behalf of users who have approved for renewals.

## Flow
1. A lock is deployed
2. A specific keyPurchaser contract is deployed
3. A user approves an ERC20 withdrawal from the keyPurchaser of K times the keyPrice.
4. The user sends a first `renewal` themselves thru the `keyPurchaser` contract. At that point, the contract keeps track of the `keyPrice`.
5. When/If the user has a (nearly) expired key to the lock, 
_any user_ can trigger a `renewal` on the keyPurchaser.  The `renewal` method will withdraw the ERC20 amount required for itself and will purchase a key for the user who has approved the key purchases.

## Constraints
The `Renewal` call should only be successful if
1. The ERC20 approved amount is high enough (the `purchase` call will fail otherwise)
2. The `keyPrice` has not changed by more than +Δ since approval. Let's default to 10%.
3. The user who has approved for renewals has no key or a nearly expired key.

NickCuso:
# Description

Adds the KeyPurchaser extension.
Comments should explain usage. Let me know if there are questions or suggestions

Docs https://docs.unlock-protocol.com/developers/smart-contracts-architecture/key-purchaser

# Issues

Fixes #5939
Refs #

# Checklist:

### DIFF
--- a/smart-contract-extensions/contracts/KeyPurchaser.sol
+++ b/smart-contract-extensions/contracts/KeyPurchaser.sol
@@ -217,7 +217,7 @@ contract KeyPurchaser is Initializable, LockRoles
     purchasePrice = purchasePrice.add(msgSenderReward);
 
     // It's okay if the lock changes tokenAddress as the ERC-20 approval is specifically
-    // the token the endUser wanted to spend
+    // for the token the endUser wanted to spend
     IERC20 token = IERC20(lock.tokenAddress());
     require(token.balanceOf(_recipient) >= purchasePrice, 'INSUFFICIENT_BALANCE');
     require(token.allowance(_recipient, address(this)) >= purchasePrice, 'INSUFFICIENT_ALLOWANCE');
--- a/smart-contract-extensions/contracts/KeyPurchaser.sol
+++ b/smart-contract-extensions/contracts/KeyPurchaser.sol
@@ -7,6 +7,7 @@ import '@openzeppelin/contracts/token/ERC20/IERC20.sol';
 import '@openzeppelin/contracts/token/ERC20/SafeERC20.sol';
 import '@openzeppelin/contracts/math/SafeMath.sol';
 import 'unlock-abi-7/IPublicLockV7.sol';
+import './mixins/LockRoles.sol';
 
 /**
  * @notice Purchase a key priced in any ERC-20 token - either once or as a regular subscription.
@@ -18,11 +19,16 @@ import 'unlock-abi-7/IPublicLockV7.sol';
  * Risk: if the user transfers or cancels the key, they would naturally expect that also cancels
  * the subscription but it does not. This should be handled by the frontend.
  */
-contract KeyPurchaser is Initializable, Stoppable
+contract KeyPurchaser is Initializable, LockRoles
 {
   using Address for address payable;
   using SafeERC20 for IERC20;
   using SafeMath for uint;
+
+  /**
+   * @notice Emitted when the stop is triggered by a stopper (`account`).
+   */
+  event Stopped(address account);
 
   // set on initialize and cannot change
 
@@ -71,6 +77,12 @@ contract KeyPurchaser is Initializable, Stoppable
    */
   bool internal hidden;
 
+  /**
+   * @notice Indicates if the contract has been disabled by a lock manager. Once stopped it
+   * may never start again.
+   */
+  bool public stopped;
+
   // store minimal history
 
   /**
@@ -86,7 +98,6 @@ contract KeyPurchaser is Initializable, Stoppable
    */
   function initialize(
     IPublicLockV7 _lock,
-    address _admin,
     uint _maxPurchasePrice,
     uint _renewWindow,
     uint _renewMinFrequency,
@@ -94,13 +105,30 @@ contract KeyPurchaser is Initializable, Stoppable
   ) public
     initializer()
   {
-    _initializeAdminRole(_admin);
     lock = _lock;
     maxPurchasePrice = _maxPurchasePrice;
     renewWindow = _renewWindow;
     renewMinFrequency = _renewMinFrequency;
     msgSenderReward = _msgSenderReward;
     approveSpending();
+  }
+
+  /**
+   * @dev Modifier to make a function callable only when the contract is not stopped.
+   */
+  modifier whenNotStopped()
+  {
+    require(!stopped, 'Stoppable: stopped');
+    _;
+  }
+
+  /**
+    * @notice Called by a stopper to stop, triggers stopped state.
+    */
+  function stop() public onlyLockManager(lock) whenNotStopped
+  {
+    stopped = true;
+    emit Stopped(msg.sender);
   }
 
   /**
@@ -126,7 +154,7 @@ contract KeyPurchaser is Initializable, Stoppable
     string memory _name,
     bool _hidden
   ) public
-    onlyAdmin()
+    onlyLockManager(lock)
   {
     name = _name;
     hidden = _hidden;
@@ -138,7 +166,7 @@ contract KeyPurchaser is Initializable, Stoppable
    */
   function shouldBeDisplayed() public view returns(bool)
   {
-    return !stopped() && !hidden;
+    return !stopped && !hidden;
   }
 
   /**
--- a/smart-contract-extensions/contracts/KeyPurchaser.sol
+++ b/smart-contract-extensions/contracts/KeyPurchaser.sol
@@ -1,0 +1,235 @@
+pragma solidity 0.5.17;
+
+import 'hardlydifficult-ethereum-contracts/contracts/lifecycle/Stoppable.sol';
+import '@openzeppelin/upgrades/contracts/Initializable.sol';
+import '@openzeppelin/contracts/utils/Address.sol';
+import '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+import '@openzeppelin/contracts/token/ERC20/SafeERC20.sol';
+import '@openzeppelin/contracts/math/SafeMath.sol';
+import 'unlock-abi-7/IPublicLockV7.sol';
+
+/**
+ * @notice Purchase a key priced in any ERC-20 token - either once or as a regular subscription.
+ * This allows the user to purchase or subscribe to a key with 1 tx (`approve`)
+ * or if the token supports it, with 1 signed message (`permit`).
+ *
+ * The user can remove the ERC-20 approval to cancel anytime.
+ *
+ * Risk: if the user transfers or cancels the key, they would naturally expect that also cancels
+ * the subscription but it does not. This should be handled by the frontend.
+ */
+contract KeyPurchaser is Initializable, Stoppable
+{
+  using Address for address payable;
+  using SafeERC20 for IERC20;
+  using SafeMath for uint;
+
+  // set on initialize and cannot change
+
+  /**
+   * @notice This is the lock for the content users are subscribing to.
+   */
+  IPublicLockV7 public lock;
+
+  /**
+   * @notice The most you will spend on a single key purchase.
+   * @dev This allows the lock owner to increase the key price overtime, up to a point, without
+   * breaking subscriptions.
+   */
+  uint public maxPurchasePrice;
+
+  /**
+   * @notice How close to the end of a subscription someone may renew the purchase.
+   * @dev To ensure that the subscription never expires, we allow this amount of time (maybe 1 day)
+   * before it expires for the renewal to happen.
+   */
+  uint public renewWindow;
+
+  /**
+   * @notice The earliest a renewal may be purchased after the previous purchase was done.
+   * @dev This typically would not apply, but helps to avoid potential abuse or confusion
+   * around cancel and transfer scenarios.
+   */
+  uint public renewMinFrequency;
+
+  /**
+   * @notice The amount of tokens rewarded from the end user to the msg.sender for enabling this feature.
+   * This is paid with each renewal of the key.
+   */
+  uint public msgSenderReward;
+
+  // admin can change these anytime
+
+  /**
+   * @notice Metadata for these subscription terms which may be displayed on the frontend.
+   */
+  string public name;
+
+  /**
+   * @notice Metadata for these subscription terms which can be used to supress the offering
+   * from the frontend.
+   */
+  bool internal hidden;
+
+  // store minimal history
+
+  /**
+   * @notice Tracks when the key was last purchased for a user's subscription.
+   * @dev This is used to enforce renewWindow and renewMinFrequency.
+   */
+  mapping(address => uint) public timestampOfLastPurchase;
+
+  /**
+   * @notice Called once to set terms that cannot change later on.
+   * @dev We are using initialize instead of a constructor so that this
+   * contract may be deployed with a minimal proxy.
+   */
+  function initialize(
+    IPublicLockV7 _lock,
+    address _admin,
+    uint _maxPurchasePrice,
+    uint _renewWindow,
+    uint _renewMinFrequency,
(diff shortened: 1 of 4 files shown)


smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,141 @@
...
+      temp = lock.keyExpirationTimestampFor(_user); // This reverts if getHasValidKey is false
+      require(temp <= now || temp - now <= renewWindow, 'OUTSIDE_RENEW_WINDOW');
+    }
+
+    temp = lock.keyPrice();
+    require(temp <= maxKeyPrice, 'PRICE_TOO_HIGH');
+
+    // It's okay if the lock changes tokenAddress as the ERC-20 approval is specifically
+    // the token the endUser wanted to spend
+    IERC20 token = IERC20(lock.tokenAddress());
```
NickCuso: Hmm maybe we should check if the keyPrice is 0, and then avoid the transfer / approval.

NickCuso: fixed

smart-contract-extensions/contracts/KeyPurchaserFactory.sol
```diff
@@ -0,0 +1,56 @@
...
+  constructor() public
+  {
+    keyPurchaserTemplate = address(new KeyPurchaser());
+  }
+
+  /**
+   * @notice Deploys a new KeyPurchaser for a lock and stores the address for reference.
+   */
+  function deployKeyPurchaser(
+    IPublicLock _lock,
```
nfurfaro: shouldn't this be `IPublicLockV6` as that's what you imported?

smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,162 @@
...
+import '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+import '@openzeppelin/contracts/token/ERC20/SafeERC20.sol';
+import 'unlock-abi-1-3/IPublicLockV6.sol';
+
+/**
+ * @notice Purchase a key priced in any ERC-20 token - either once or as a regular subscription.
+ * This allows the user to purchase or subscribe to a key with 1 tx (`approve`)
+ * or if the token supports it, with 1 signed message (`permit`).
+ *
+ * The user can remove approval to cancel anytime.
```
nfurfaro: Just to be clear, this is referring to `erc-20 approve()` right?

smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,162 @@
...
+ * Risk: if the user transfers or cancels the key, they would naturally expect that also cancels
+ * the subscription but it does not. This should be handled by the frontend.
+ */
+contract KeyPurchaser is Initializable, Stoppable
+{
+  using Address for address payable;
+  using SafeERC20 for IERC20;
+
+  // set on initialize and cannot change
+  IPublicLock public lock;
```
nfurfaro: Same as comment for factory contract.

smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * @notice Used by admins to update metadata which may be leveraged by the frontend.
+   */
+  function config(
+    string memory _name,
+    bool _disabled
+  ) public
+    onlyAdmin()
```
nfurfaro: I think that the Admin here is the lock owner right? Where does the lock owner get added to the admin role?

smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * @notice Called by anyone to purchase or renew a key on behalf of a user.
+   * The user must have ERC-20 spending approved and the purchase must meet the terms
+   * defined during initialization.
+   */
+  function purchaseFor(
+    address payable _user,
+    address _referrer,
```
nfurfaro: I wonder if _recipient(or _beneficiary) would be more clear than _user here?

smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,162 @@
...
+  uint public renewWindow;
+  uint public renewMinFrequency;
+  bool public isSubscription;
+
+  // admin can change these anytime
+  string public name;
+  bool internal disabled;
+
+  // store minimal history
+  mapping(address => uint) public timeOfLastPurchase;
```
nfurfaro: iI think this is a timestamp?

smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * @notice Called once to set terms that cannot change later on.
+   */
+  function initialize(
+    IPublicLock _lock,
+    uint _maxKeyPrice,
+    uint _renewWindow,
+    uint _renewMinFrequency,
```
nfurfaro: If you could add some @param comments above it will probably save me a bunch of questions further down in purchaseFor()

smart-contract-extensions/contracts/KeyPurchaserFactory.sol
NickCuso: The name of the contract != the name of the file. We might want to fix that and republish.  I'd vote removing V6 from the filename.

smart-contract-extensions/contracts/KeyPurchaser.sol
NickCuso: yup. I could improve the comment

smart-contract-extensions/contracts/KeyPurchaser.sol
NickCuso: Via Stoppable.

nfurfaro: Right, but when a lock creator deploys a KeyPurchaser contract, are they added to the role automatically?

NickCuso: yup

smart-contract-extensions/contracts/KeyPurchaserFactory.sol
nfurfaro: Haha, right you are. I'd second that vote!

smart-contract-extensions/contracts/KeyPurchaser.sol
NickCuso: hmm or does it use msg.sender? lol i got to check again - still waking up over here ;)

nfurfaro: yeah it uses message.sender, so the factory address

smart-contract-extensions/contracts/KeyPurchaserFactory.sol
NickCuso: hmm second thought.  an advantage to renaming the contract name in the interface file is an extension could then support multiple lock versions.  without that - truffle breaks on a dupe name.

smart-contract-extensions/contracts/KeyPurchaser.sol
NickCuso: k renamed

smart-contract-extensions/contracts/KeyPurchaser.sol
NickCuso: fixed

NickCuso:
Should this be limited to the lock creator? Currently the factory checks that but it's more a frontend thing than anything - ensure the frontend doesn't surface unreasonable options for users to sort through.

What if we switch to a create2 factory and calc the seed based on the configuration? That way users can actually use it before it exists. We would need a backend service to be informed that a new purchaser should be deployed and what configuration is expected. If the lock creator did not explicitly create one, we could have a default configuration - e.g. up to a 5% slippage from the current price.

smart-contract-extensions/contracts/KeyPurchaser.sol
NickCuso: updated

smart-contract-extensions/contracts/KeyPurchaser.sol
NickCuso: done

NickCuso:
KeyPurchaser and the factory have been refactored.  Aiming to simplify the terms for end users as well as add support for Create2 so technically someone could subscribe before the keyPurchaser even exists.

NickCuso:
Open question:

```
    // This require is not strictly necessary, but helps to maintain trust when surfacing these
    // options on the frontend.
    require(_lock.isLockManager(msg.sender), 'ONLY_LOCK_MANAGER');
```

Should we remove this requirement?

Pro: It would allow any broadcaster to process subscriptions even if the lock manager has not explicitly enabled them.
Con: Someone may create a keyPurchaser with absurd terms that may mislead users. e.g. a maxKeyPrice that's 1000x the current keyPrice.  Depending on how well terms are communicated on the frontend that may mislead or confuse users.

I'm leaning towards removing it.  What do you think?

NickCuso:
CC @julien51

smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,235 @@
...
+    IPublicLockV7 _lock,
+    address _admin,
+    uint _maxPurchasePrice,
+    uint _renewWindow,
+    uint _renewMinFrequency,
+    uint _msgSenderReward
+  ) public
+    initializer()
+  {
+    _initializeAdminRole(_admin);
```
NickCuso, line 97: On second thought, I'll remove the admin role and use the lock's lockManagers instead.

smart-contract-extensions/contracts/KeyPurchaserFactory.sol
```diff
@@ -0,0 +1,110 @@
...
+    IPublicLockV7 _lock,
+    uint _maxKeyPrice,
+    uint _renewWindow,
+    uint _renewMinFrequency,
+    uint _msgSenderReward
+  ) public
+  {
+    // This require is not strictly necessary, but helps to maintain trust when surfacing these
+    // options on the frontend.
+    require(_lock.isLockManager(msg.sender), 'ONLY_LOCK_MANAGER');
```
NickCuso, line 78: I think we remove this require

NickCuso:
@nfurfaro I think I've made all the pending changes now. Ready for another review.

smart-contract-extensions/contracts/KeyPurchaserFactory.sol
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * @notice Deploys a new KeyPurchaser for a lock and stores the address for reference.
+   */
+  function deployKeyPurchaser(
+    IPublicLockV7 _lock,
+    uint _maxKeyPrice,
+    uint _renewWindow,
+    uint _renewMinFrequency,
```
nfurfaro, line 72: So for these 2 params:
`_renewWindow` is to specify at what point in time a key is eligible for renewal, and
`_renewMinFrequency` is to specify the min. renewal period.. ex: every year, or every month, etc..
Do I have this straight?

smart-contract-extensions/contracts/KeyPurchaserFactory.sol
```diff
@@ -0,0 +1,106 @@
...
+  ) public view
+    returns (address[] memory)
+  {
+    return lockToKeyPurchasers[_lock];
+  }
+
+  /**
+   * @notice Returns the number of available keyPurchasers for a given lock.
+   */
+  function getKeyPurchaserCount(
```
nfurfaro, line 99: This could maybe be considered redundant (or at least a front-end concern) as the length is available by calling `getKeyPurchasers` and just checking the retuned array's `length` property. Maybe it's worth keeping for convenience though...

smart-contract-extensions/contracts/KeyPurchaserFactory.sol
nfurfaro: Just saw the comments in the first contract.

smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,263 @@
...
+   * before it expires for the renewal to happen.
+   */
+  uint public renewWindow;
+
+  /**
+   * @notice The earliest a renewal may be purchased after the previous purchase was done.
+   * @dev This typically would not apply, but helps to avoid potential abuse or confusion
+   * around cancel and transfer scenarios.
+   */
+  uint public renewMinFrequency;
```
nfurfaro, line 59: I can live with it as is, but the naming here is a bit confusing. It makes me think this is related to how often the key can be renewed. Something like `earliestRenewal` might have clearer semantics.

smart-contract-extensions/contracts/KeyPurchaser.sol
```diff
@@ -0,0 +1,263 @@
...
+    address payable _recipient,
+    address _referrer,
+    bytes memory _data
+  ) public view
+  {
+    uint purchasePrice = _readyToPurchaseFor(_recipient, _referrer, _data);
+    purchasePrice = purchasePrice.add(msgSenderReward);
+
+    // It's okay if the lock changes tokenAddress as the ERC-20 approval is specifically
+    // the token the endUser wanted to spend
```
nfurfaro, line 220: "for the token ..."

smart-contract-extensions/test/keyPurchaser.js
```diff
@@ -0,0 +1,427 @@
...
+        from: lockCreator,
+      })
+
+      // Make first purchase
+      await dai.approve(keyPurchaser.address, -1, { from: endUser })
+      await keyPurchaser.purchaseFor(endUser, constants.ZERO_ADDRESS, [], {
+        from: otherAccount,
+      })
+
+      // Ensure at least 1 second has past to avoid the min frequency check
```
nfurfaro, line 262: Ensure at least 1 second has passed ...

smart-contract-extensions/test/keyPurchaser.js
```diff
@@ -0,0 +1,427 @@
...
+      )
+      await reverts(
+        keyPurchaser.purchaseFor(endUser, constants.ZERO_ADDRESS, [], {
+          from: otherAccount,
+        }),
+        'OUTSIDE_RENEW_WINDOW'
+      )
+    })
+
+    describe('after enough time has past', () => {
```
nfurfaro, line 279: after enough time has passed

smart-contract-extensions/contracts/KeyPurchaser.sol
nfurfaro: Cancel that. After reading the tests I am clear on how this works/what it means now.

nfurfaro:
@NickCuso I sprinkled a few minor comments here and there, but overall this looks really good.

smart-contract-extensions/contracts/KeyPurchaserFactory.sol
NickCuso: This is a backup in case there are a large number of options.  Now that shouldn't actually be a realistic concern for this use case, but with Solidity the node will reject your attempt to read an array if the array is too large.

nfurfaro: got it.

PR #6054, opened 2020-02-25
Checkout keeps track of purchases in progress

cnasc:
# Description

One thing we need on the checkout is for the UI to react when a purchase begins. In particular, it should disable lock buttons once a purchase starts so that we don't have any duplicate purchases.

This PR adds that, and also takes the opportunity to remove some unneeded stuff from the CheckoutWrapper component.

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/Lock.tsx
+++ b/unlock-app/src/components/interface/checkout/Lock.tsx
@@ -166,27 +166,54 @@ export const LoadingLock = () => {
 
 interface LockProps {
   lock: RawLock
+  purchasingLockAddress: string | null
+  setPurchasingLockAddress: (lockAddress: string) => void
 }
 
-const lockKeysAvailable = (lock: RawLock) => {
-  if ((lock as any).unlimitedKeys) {
+interface LockKeysAvailableLock {
+  unlimitedKeys?: boolean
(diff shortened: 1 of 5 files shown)


unlock-app/src/components/interface/checkout/Lock.tsx
```diff
@@ -166,27 +166,54 @@ export const LoadingLock = () => {
...
+  outstandingKeys,
+}: LockKeysAvailableLock) => {
+  if (unlimitedKeys) {
     return 'Unlimited'
   }
 
   // maxNumberOfKeys and outstandingKeys are assumed to be defined
   // if they are ever not, a runtime error can occur
-  return (lock.maxNumberOfKeys! - lock.outstandingKeys!).toString()
+  return (maxNumberOfKeys! - outstandingKeys!).toString()
```
akeem: This may be an artifact of how things were before, but this looks a bit weird. We should probably move the value generation portion of this outside of the view component

cnasc: This actually is outside the component, just contained in the same file at the moment. My next project involves a bit of an overhaul on this piece, and in the process of doing that I'll look for checkout-related things to extract to utils

Issue #5837, opened 2020-01-27
v4 default tokenUri regressed


NickCuso:
**Describe the bug**
any v4 lock which does not define their own base URI will not be able to read the default url as expected

**To Reproduce**
https://rinkeby.etherscan.io/token/0x771e09a5bfef4c4b85d796a112d49e839c98d9da#readContract
reading the uri reverts

**Expected behavior**
it should use the default base uri from unlock.sol.

**Additional context**
Due to https://github.com/unlock-protocol/unlock/pull/4799

We need both versions of the call in Unlock.sol now.  Once that's done we can upgrade Unlock.sol and v4 locks will function correctly again.

PR #6056, opened 2020-02-25
Refactor upgrade tests

NickCuso:
# Description

Merges all the upgrade tests into a single file.  I tried to preserve all existing test, and a few small ones were added.  

Two benefits here:
 - It's easier to see breaking changes between versions
 - Adding a new version does not depend on a copy/paste process (which is hard to review)

# Issues

Fixes #
Refs #

# Checklist:


nfurfaro:
Does this refactor address the type of regression we saw in  Issue #5837?

NickCuso:
No.  But I believe this pattern makes adding new test cases like that easier to do going forward.

PR #5965, opened 2020-02-13
[WIP] Adding a keyManager for each key

nfurfaro:
# Description
This was a fairly large task with many changes. The most notable of them are:

- In order to `approve` someone to transfer or share your key for you, you must also make them the keyManager for your key.
- Both `shareKey` and `grantKeys` now have an extra parameter (`_keyManager` & `_keyManagers[]` respectively)
- new event added: `KeyManagerChanged`
- new modifier added: `onlyKeyManager`
- new functions added: `setKeyManagerOf(...)` & `getKeyManagerOf(...)`
- new mapping: `mapping (uint => address) internal keyManagerOf;`
- 1 new test file: `permissions/keyManager.js`
- modifications to a bunch of test files, mostly around adding the extra params to function calls, and updating the log indexes when reading emitted events.

# Issues

Fixes #5557 
Fixes #5753
Refs #

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [x] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [x] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #6072, opened 2020-02-26
[WIP] Key manager permissions part-1

### LINKED ISSUE #5753: Granted keys should not be transferable/cancellable by default
In the past we have discussed challenges around granted keys. I believe that, by default, keys which have been created thru the `grantKeys` step should not be transferable/cancellable by their owner.

This should leverage the [work done on this issue](https://github.com/unlock-protocol/unlock/issues/5557).

However the grantor (or lock owner) should be able to revoke that limitation, if they want.

nfurfaro:
# Description
This is part 1/3 for adding new keyManager permissions. Splitting up to simplify review.
- This PR supercedes #5965
- The next PR will include an update to setApprovalForAll and address the permissions around cancelation/refunds.

# Issues

Fixes #5753
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -38,7 +38,7 @@ contract MixinGrantKeys is
       _assignNewTokenId(toKey);
       _recordOwner(recipient, toKey.tokenId);
       // Assign the KeyManager
-      _setKeyManagerOf(toKey.tokenId, keyManager);
+      _setKeyManagerOf(toKey.tokenId, keyManager, false);
       toKey.expirationTimestamp = expirationTimestamp;
 
       // trigger event
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -21,7 +21,7 @@ contract MixinGrantKeys is
   function grantKeys(
     address[] calldata _recipients,
     uint[] calldata _expirationTimestamps,
-    address[] _keyManagers
+    address[] calldata _keyManagers
   ) external
     onlyKeyGranter
   {
@@ -38,7 +38,7 @@ contract MixinGrantKeys is
       _assignNewTokenId(toKey);
       _recordOwner(recipient, toKey.tokenId);
       // Assign the KeyManager
-      _setKeyManager(toKey.tokenId, keyManager);
+      _setKeyManagerOf(toKey.tokenId, keyManager);
       toKey.expirationTimestamp = expirationTimestamp;
 
       // trigger event
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -21,19 +21,14 @@ contract MixinGrantKeys is
   function grantKeys(
     address[] calldata _recipients,
     uint[] calldata _expirationTimestamps,
-    address _keyManager
+    address[] _keyManagers
   ) external
     onlyKeyGranter
   {
-    address keyManager;
-    if (_keyManager != address(0)) {
-      keyManager = _keyManager;
-    } else {
-      keyManager = msg.sender;
-    }
     for(uint i = 0; i < _recipients.length; i++) {
       address recipient = _recipients[i];
       uint expirationTimestamp = _expirationTimestamps[i];
+      address keyManager = _keyManagers[i];
 
       require(recipient != address(0), 'INVALID_ADDRESS');
 
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -20,10 +20,17 @@ contract MixinGrantKeys is
    */
   function grantKeys(
     address[] calldata _recipients,
-    uint[] calldata _expirationTimestamps
+    uint[] calldata _expirationTimestamps,
+    address _keyManager
   ) external
     onlyKeyGranter
   {
+    address keyManager;
+    if (_keyManager != address(0)) {
+      keyManager = _keyManager;
+    } else {
+      keyManager = msg.sender;
+    }
     for(uint i = 0; i < _recipients.length; i++) {
       address recipient = _recipients[i];
       uint expirationTimestamp = _expirationTimestamps[i];
@@ -35,6 +42,8 @@ contract MixinGrantKeys is
 
       _assignNewTokenId(toKey);
       _recordOwner(recipient, toKey.tokenId);
+      // Assign the KeyManager
+      _setKeyManager(toKey.tokenId, keyManager);
       toKey.expirationTimestamp = expirationTimestamp;
 
       // trigger event
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -37,12 +37,12 @@ contract MixinGrantKeys is
 
       _assignNewTokenId(toKey);
       _recordOwner(recipient, toKey.tokenId);
-      // Only assign the KeyManager for new keys
-      if(keyManagerOf[toKey.tokenId] == address(0)) {
+
+      // Only assign the KeyManager for new or expired keys
+      if(toKey.expirationTimestamp <= block.timestamp) {
         keyManagerOf[toKey.tokenId] = keyManager;
         emit KeyManagerChanged(toKey.tokenId, keyManager);
       }
-
       toKey.expirationTimestamp = expirationTimestamp;
 
       // trigger event
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -37,9 +37,12 @@ contract MixinGrantKeys is
 
       _assignNewTokenId(toKey);
       _recordOwner(recipient, toKey.tokenId);
-      // Assign the KeyManager
-      keyManagerOf[toKey.tokenId] = keyManager;
-      emit KeyManagerChanged(toKey.tokenId, keyManager);
+      // Only assign the KeyManager for new keys
+      if(keyManagerOf[toKey.tokenId] == address(0)) {
+        keyManagerOf[toKey.tokenId] = keyManager;
+        emit KeyManagerChanged(toKey.tokenId, keyManager);
+      }
+
       toKey.expirationTimestamp = expirationTimestamp;
 
       // trigger event
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -38,7 +38,8 @@ contract MixinGrantKeys is
       _assignNewTokenId(toKey);
       _recordOwner(recipient, toKey.tokenId);
       // Assign the KeyManager
-      _setKeyManagerOf(toKey.tokenId, keyManager);
+      keyManagerOf[toKey.tokenId] = keyManager;
+      emit KeyManagerChanged(toKey.tokenId, keyManager);
       toKey.expirationTimestamp = expirationTimestamp;
 
       // trigger event
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -38,7 +38,7 @@ contract MixinGrantKeys is
       _assignNewTokenId(toKey);
       _recordOwner(recipient, toKey.tokenId);
       // Assign the KeyManager
-      _setKeyManagerOf(toKey.tokenId, keyManager, false);
+      _setKeyManagerOf(toKey.tokenId, keyManager);
       toKey.expirationTimestamp = expirationTimestamp;
 
       // trigger event
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -35,23 +35,24 @@ contract MixinGrantKeys is
       Key storage toKey = keyByOwner[recipient];
       require(expirationTimestamp > toKey.expirationTimestamp, 'ALREADY_OWNS_KEY');
 
-      _assignNewTokenId(toKey);
-      uint toId = toKey.tokenId;
-      _recordOwner(recipient, toId);
+      uint idTo = toKey.tokenId;
 
-      // Set the key Manager if other than 0
-      if(keyManager != address(0)) {
-        keyManagerOf[toId] = keyManager;
-        emit KeyManagerChanged(toId, keyManager);
+      if(idTo == 0) {
+        _assignNewTokenId(toKey);
+        idTo = toKey.tokenId;
+        _recordOwner(recipient, idTo);
+        // trigger event
+        emit Transfer(
+          address(0), // This is a creation.
+          recipient,
+          idTo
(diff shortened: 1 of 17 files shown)


smart-contracts/contracts/mixins/MixinGrantKeys.sol
```diff
@@ -35,6 +37,12 @@ contract MixinGrantKeys is
 
       _assignNewTokenId(toKey);
       _recordOwner(recipient, toKey.tokenId);
+      // Only assign the KeyManager for new keys
+      if(keyManagerOf[toKey.tokenId] == address(0)) {
```
nfurfaro: hmm. I think this needs work.

nfurfaro: The intent was to prevent the use of `grantKeys` to be able to "take-over" an existing user's key. EG:
you have a valid key. I grant you a new key (which extends yours) but I also set myself as the keyManager for your key.

nfurfaro: but there's already a check for this above, so I will push a small change to simplify this.

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -49,6 +51,11 @@ contract MixinKeys is
   // Addresses are never removed by design to avoid abuses around referals
   address[] public owners;
 
+  // A given key has both an owner and a manager.
+  // The owner and manager may be the same address
+  // Each key can have at most 1 keyManager, but may have no manager.
```
NickCuso: What does no manager mean? Does that imply the keyOwner is the manager at that time? I think that could make sense but the current implementation I think effectively disables the key if the manager is 0

NickCuso: Also... not sure on this. But should keyManager be a field of the Key struct?

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -228,6 +243,35 @@ contract MixinKeys is
     return _ownerOf[_tokenId];
   }
 
+  /**
+  * @notice Get the key manager for the given tokenId
+  * @return The address of the manager for the given key
+  */
+  function getKeyManagerOf(
```
NickCuso: In general do we prefer getters like this or public variables?  We should aim to pick one for consistency.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -48,10 +48,12 @@ contract MixinTransfer is
   function shareKey(
     address _to,
     uint _tokenId,
-    uint _timeShared
+    uint _timeShared,
+    address _keyManager
   ) public
     onlyIfAlive
     onlyKeyOwnerOrApproved(_tokenId)
+    onlyKeyManager(_tokenId)
```
NickCuso: I think having multiple modifiers using the same params is a little hard to follow.  Can we merge these into a single modifier?

smart-contracts/contracts/mixins/MixinKeys.sol
nfurfaro: I don't feel strongly about this either way, but I thought we had discussed and decided that address(0) was a valid keyManager address.
The key would still work, but not be transferrable/shareable.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: This change is made already, just not part of this PR. (It's in with the changes to the `approve` & `setApprovalForAll` permissions.

smart-contracts/contracts/mixins/MixinKeys.sol
NickCuso: Okay that's a fair use case. Just ensure there's a comment to that effect - maybe on the setter. Something like 'setting to 0 disables transfers for this key'.

Hmm begs a question. What if I buy with CC, Unlock is the key manager. They set the keyManager to 0. A few months later I buy again - this time with crypto. ..but I still cannot transfer? I'll never get those permissions back?

smart-contracts/contracts/mixins/MixinTransfer.sol
NickCuso: okay that's fine. maybe I should focus on that one first. i'm not sure having both like this is valid

smart-contracts/contracts/mixins/MixinKeys.sol
nfurfaro: good point.of course, Unlock should probably have set themselves as the keyManager, but maybe supporting the zero address as a keyManager is overly complicated for the limited value it brings.

NickCuso: Maybe on purchase, if the user does not have a key or the key is expired ATM - they become the keyManager.  e.g. if the first purchase was with CC and then expired before we updated the keyManager - it seems that on next purchase the buyer reclaim that role.

smart-contracts/contracts/mixins/MixinKeys.sol
nfurfaro: yup. I think I should make this a public variable.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: NOw that the way forward with approvals is clear I could add it to this pr... it's dependent on this one anyway and doesn't change much. It should simplify review.

smart-contracts/contracts/mixins/MixinKeys.sol
nfurfaro: Yeah that makes sense. Also, just saw this:
>Also... not sure on this. But should keyManager be a field of the Key struct?

smart-contracts/contracts/mixins/MixinKeys.sol
NickCuso: I lean that way too. But a benefit of using getters instead is we can make variables private and therefore any use by other mixins is known to be contained to functions inside that contract.

smart-contracts/contracts/mixins/MixinTransfer.sol
NickCuso: Okay that works. I just hesitate to merge to master with something like this. An alternate approach we could consider is PRing small parts into a `sm-dev` branch. Then once the entire feature is working we PR that branch to master for a final sanity check.

smart-contracts/contracts/mixins/MixinKeys.sol
NickCuso: Yea I don't really know the tradeoffs. Begs the question - should we have a Key struct at all? And does the answer change knowing Solidity 0.6 is coming which includes abiV2 (which means we can publicly expose structs)

nfurfaro: Can we make that a new issue to track it? we can always refactor if we decide to change things up.

nfurfaro: >Maybe on purchase, if the user does not have a key or the key is expired ATM - they become the keyManager. e.g. if the first purchase was with CC and then expired before we updated the keyManager - it seems that on next purchase the buyer reclaim that role.

this has now been fixed.

smart-contracts/contracts/mixins/MixinGrantKeys.sol
```diff
@@ -35,6 +37,12 @@ contract MixinGrantKeys is
 
       _assignNewTokenId(toKey);
       _recordOwner(recipient, toKey.tokenId);
+
+      // Only assign the KeyManager for new or expired keys
+      if(toKey.expirationTimestamp <= block.timestamp) {
```
NickCuso: Hmm are you sure?  It seems this puts the CC use case at risk.

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -228,6 +243,24 @@ contract MixinKeys is
...
+  * setting to address(0) disables transfers for this key until it expires
+  * */
+  //
+  function setKeyManagerOf(
+    uint _tokenId,
+    address _keyManager
+  ) public
+    isKey(_tokenId)
+  {
+    require((msg.sender == keyManagerOf[_tokenId]) || (isLockManager(msg.sender) == true), 'SETKEYMANAGEROF_ACCESS_DENIED');
```
NickCuso: For consistent syntax / style don't include `== true` when checking a bool. The params here are also not needed.

smart-contracts/contracts/mixins/MixinKeys.sol
NickCuso: Could you add a comment to clarify what no manager means?  I thought we were talking about removing that. But the code still supports no manager.

I think I prefer no manager means the keyOwner.  Then we only set it if it's different - e.g. it's 0 for a normal purchase.

smart-contracts/contracts/mixins/MixinGrantKeys.sol
```diff
@@ -35,6 +37,12 @@ contract MixinGrantKeys is
...
+        keyManagerOf[toKey.tokenId] = keyManager;
```
NickCuso: If we want to say keyManager is not 0, we should add a require

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -85,6 +86,9 @@ contract MixinTransfer is
       _assignNewTokenId(toKey);
       iDTo = toKey.tokenId;
       _recordOwner(_to, iDTo);
+      // Assign the KeyManager
+      keyManagerOf[toKey.tokenId] = _keyManager;
```
NickCuso: This is not consistent with purchase. What if the iDTo owns an expired key? We should be consistent or simplify

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -48,10 +48,11 @@ contract MixinTransfer is
   function shareKey(
     address _to,
     uint _tokenId,
-    uint _timeShared
+    uint _timeShared,
+    address _keyManager
```
NickCuso: Do we really want this? I think we can simplify and exclude keyManager here.

Thin keyManager does not change in this scenario. And no keyManager means that the keyOwner has those rights.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -128,6 +132,8 @@ contract MixinTransfer is
     if (toKey.tokenId == 0) {
       toKey.tokenId = fromKey.tokenId;
       _recordOwner(_recipient, toKey.tokenId);
+      keyManagerOf[toKey.tokenId] = _recipient;
```
NickCuso: Again what about the expired scenario?

smart-contracts/contracts/mixins/MixinGrantKeys.sol
nfurfaro: What's the risk you see here? 
What I want to avoid is this scenario:
- you have a valid key which you purchased, and for which you are the keyManager
- I (the keyGranter) come along and grant you a new key which extends your existing key, but I also set myself as the keyManager of your key, in effect stealing it from you.

smart-contracts/contracts/mixins/MixinKeys.sol
nfurfaro: I will remove the `== true`. thx for the consistency reminder. 
not sure about the params though... `isLockManager` takes an address and `keyManagerOf` needs a uint. I know I must be missing something very obvious.

smart-contracts/contracts/mixins/MixinGrantKeys.sol
NickCuso: I buy with CC. I see a pending tx from Unlock protocol. I frontrun a crypto purchase. Then unlock buys and I cancel and refund, and chargeback - potentially doubling my investment

smart-contracts/contracts/mixins/MixinKeys.sol
nfurfaro: Looking back over our discussion, we did discuss the merits of both keeping/removing this.
>I think I prefer no manager means the keyOwner. Then we only set it if it's different - e.g. it's 0 for a normal purchase.

 I support this and will modify so that if 0 is passed as the keyManager address, then the  keyOwner is also the manager.

smart-contracts/contracts/mixins/MixinGrantKeys.sol
NickCuso: No that's not exactly right.  I think we should just trust the keyGranter to do the correct thing here.  That's a trusted role - much more so than a keyManager of course.

e.g. if they have 1 day left on the key and they use Unlock CC to reup another month, we should have the ability to take the keyManager role to prevent chargeback issues.

NickCuso: Since the keyGranter sets an expiration instead of adding time to it they already have permissions to revoke keys for all keyOwners without any refund. So this change does not seem to add more risk.

smart-contracts/contracts/mixins/MixinKeys.sol
NickCuso: This is minor of course, but I mean this is the same `msg.sender == keyManagerOf[_tokenId] || isLockManager(msg.sender)`

smart-contracts/contracts/mixins/MixinGrantKeys.sol
nfurfaro: hmm.  the keyGranter will be able to(as per our current permissions doc) expire any key, so I agree that this doesn't seem to pose any additional risk.

smart-contracts/contracts/mixins/MixinKeys.sol
nfurfaro: I think we had already migrated a handful of getters to public variables a ways back, so maybe go that way for consistency... 
one thing I should check is the size-diff between a public variable and a public getter function. NOte that if we think the benefits to using private vars outweighs the (assumed) size penalty I'm not against going that route.

NickCuso: I'll open a new issue - seems drafting a style guide for questions like this would be useful.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: the check for this block is:
`if (iDTo == 0) {...`
ensuring this only operates on new keys.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: the check for this block is:
`if (toKey.tokenId == 0) {...`
ensuring this only operates on new keys.

smart-contracts/contracts/mixins/MixinTransfer.sol
nfurfaro: Yeah we can simplify this. As we don't have a getter function for keyManager anymore, it probably makes the most sense to say that for all functions:
- if `address(0)` is passed as the keyManager, then the key owner is also the manager.(ie: `grantKeys`)
- if there's no param for keyManager in a function, then the key owner is also the manager.(ie: `shareKey`)

nfurfaro: >Thin keyManager does not change in this scenario. And no keyManager means that the keyOwner has those rights.

I just want to take an extra look at this. in this scenario, the existing keyManager is the one sharing the key. would you like to see the keyManager of the "shared" key here be the existing keyManager(of the parent key) or the keyOwner(of the child key)?

nfurfaro:
Closing as this is now implemented.

PR #6074, opened 2020-02-26
changing the init process of the mailbox

julien51:
# Description

This is the last but one change to support multiple configs in the paywall.
This removes the configuration setting from the mailbox constructor and moves it into the `init` method.

# Checklist:

### DIFF
--- a/paywall/src/data-iframe/Mailbox.ts
+++ b/paywall/src/data-iframe/Mailbox.ts
@@ -25,7 +25,6 @@ import {
   LocalStorageWindow,
   EventTypes,
 } from '../windowTypes'
-import { waitFor } from '../utils/promises'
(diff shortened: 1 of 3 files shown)


paywall/src/data-iframe/Mailbox.ts
```diff
@@ -213,6 +212,9 @@ export default class Mailbox {
     })
   }
 
+  /**
+   * TODO: consider moving all this to the constructor... this will avoid race conditions
```
cnasc: :+1: It looks like nothing here needs to be waited on, so it'll be a good future move to do this

cnasc: (although may not be needed in the long run since we're moving all of this stuff)

julien51: yeah :/

Issue #5489, opened 2019-12-16
Coupons or Discount Codes


julien51:
Locks have a price which can be changed by their owner.
The owner may also want to provide a Coupon or discount code which would let the recipient pay "less" than the actual price.

Conferences have a strong need for that feature: they want to give away discounted tickets to press, students or partners.

A lot could have multiple discount codes, each offering specific amounts. The lock owner should be able to add/remove discount codes.

The discount codes should remain "secret" so they cannot be supplied "in clear" when purchasinga key.

NickCuso:
Do we need to address the cancelAndRefund feature as well?  The refund could be > what they paid if they purchased with a discount code.

For implementation we were talking about using signatures where the signing account is the discount code itself.  So the `signing_account_private_key = keccak256(abi.encodePacked(lock_address, discount_code))`.  The public address for that account is registered in the contract and stored with a structure such as `mapping(address => uint) public discount_address_to_discount_basis_points`.  To redeem the key purchaser uses the signing_account in JS to sign a message locally, maybe `"authorizeDiscount"`.  They include this signature with the purchase transaction: `ecrecover` will provide the public address for the signing account, which we then use to look up the discount.

The lock owner could add, remove, or update entries in the discount mapping anytime.

julien51:
> Do we need to address the cancelAndRefund feature as well? The refund could be > what they paid if they purchased with a discount code.

I think for now we can leave that aside but warn creators that they should set the cancellation policy to take the coupons into account.

I think this approach makes sense, but I do not have a deep enough knowledge of `keccak256`. Is `keccak256` yielding a private key? 
We would absolutely need to sign the purchaser's address, otherwise, anyone could look at any transaction with a discount code and issue the same?

NickCuso:
I'm suggesting keccak256 to hash the lock's address and the discount code itself.  This way there are no assumptions / requirements on the length of the discount code string.

Yes we must sign the msg.sender so that it cannot be reused by another account, but the "authorizeDiscount for" portion is not necessary.

Note that with this approach people could attempt to brute force.  i.e. guess and checking can be done entirely in JS.  So if the discount code is trivial people will be able to determine it (e.g. if there's a `5 off` code people could check to see if `10 off`, `20 off`, or `50 off` works better).

julien51:
That's true but it is the case with any discount code system, crypto or not... and they have no way of knowing if it works before actually sending the transaction?

nfurfaro:
[...]

PR #6137, opened 2020-03-03
Add discount to the OnKeyPurchase hook

NickCuso:
# Description

Updates the OnKeySold hook to return a discount. This will simplify the discount code implementation and may make some other scenarios easier as well.

Also renames to `OnKeyPurchase`.

# Issues

Fixes #
Refs #5489

# Checklist:


smart-contracts/contracts/mixins/MixinEventHooks.sol
```diff
@@ -10,17 +11,17 @@ import './MixinLockCore.sol';
 contract MixinEventHooks is
   MixinLockCore
 {
-  ILockEventHooks public onKeySoldHook;
-  ILockEventHooks public onKeyCancelHook;
+  ILockKeySoldHook public onKeySoldHook;
+  ILockKeyCancelHook public onKeyCancelHook;
 
   function setEventHooks(
```
NickCuso: Maybe check each address is `address(0)` || `isContract`

NickCuso: Add a comment and add to IPublicLock

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -84,28 +84,56 @@ contract MixinPurchase is
     uint inMemoryKeyPrice = keyPrice;
     (discount, tokens) = unlockProtocol.computeAvailableDiscountFor(_recipient, inMemoryKeyPrice);
```
NickCuso: Hmm on second thought it seems we should apply the hook discount before calling unlockProtocol

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -84,28 +84,56 @@ contract MixinPurchase is
...
-    }
-
-    unlockProtocol.recordKeyPurchase(inMemoryKeyPrice, getHasValidKey(_referrer) ? _referrer : address(0));
-
     // We explicitly allow for greater amounts of ETH or tokens to allow 'donations'
     if(tokenAddress != address(0)) {
       require(_value >= inMemoryKeyPrice, 'INSUFFICIENT_VALUE');
       inMemoryKeyPrice = _value;
     }
-    // Security: after state changes to minimize risk of re-entrancy
```
NickCuso: I'm not so sure what the re-entrancy risks are at this point. I can't think of an attack but maybe we should add a `ReentrancyGuard` to most calls just in case.

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -84,28 +84,56 @@ contract MixinPurchase is
...
+    address _referrer,
+    bytes calldata _data
+  ) external view
+    returns (uint)
+  {
+    // Let's get the actual price for the key from the Unlock smart contract
+    uint inMemoryKeyPrice = keyPrice;
+    (uint discount, ) = unlockProtocol.computeAvailableDiscountFor(_recipient, inMemoryKeyPrice);
+
+    discount += _onKeySoldDiscount(_recipient, _referrer, inMemoryKeyPrice, _data);
```
NickCuso: This should change to match the change mentioned in `purchase`

NickCuso:
hmm should we rename this to `OnKeyPurchase`?

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -216,6 +224,17 @@ contract IPublicLock is IERC721Enumerable {
     bytes calldata _data
   ) external payable;
 
+  /**
+   * @notice returns the minimum price paid for a purchase with these params.
+   * @dev this considers any discount from Unlock or the OnKeySold hook.
+   */
+  function purchasePriceFor(
```
NickCuso: Maybe we move this to the `LockInspector` instead

NickCuso: Hmm but the hook might want to use msg.sender to determine the lock. So not sure we could move it to an extension...

NickCuso:
hmm the merge broke something..

NickCuso:
Should be fixed now

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -233,6 +233,17 @@ contract IPublicLock is IERC721Enumerable {
     bytes calldata _data
   ) external payable;
 
+  /**
+   * @notice returns the minimum price paid for a purchase with these params.
```
nfurfaro: Just to clarify, this is the minimum price to allow for tips, correct?

smart-contracts/contracts/mixins/MixinEventHooks.sol
```diff
@@ -10,44 +12,72 @@ import './MixinLockCore.sol';
...
-    address _onKeySoldHook,
+    address _onKeyPurchaseHook,
     address _onKeyCancelHook
   ) external
     onlyLockManager()
   {
-    onKeySoldHook = ILockEventHooks(_onKeySoldHook);
-    onKeyCancelHook = ILockEventHooks(_onKeyCancelHook);
+    require(_onKeyPurchaseHook == address(0) || _onKeyPurchaseHook.isContract(), 'INVALID_ON_KEY_SOLD_HOOK');
```
nfurfaro: I can't remember from the meeting yesterday why `_onKeyPurchaseHook == address(0)` is a valid condition here?

nfurfaro: Doh. nevermind. This is to Unset an existing hook.

NickCuso:
TODO from the meeting:
 - [ ] Change discount to keyPrice for purchase hook
 - [ ] Comment that lock address is msg.sender
 - [ ] Ensure that the keyPurchase hook has the tip
 - [ ] Single discount code deployment for all locks

smart-contracts/contracts/interfaces/IPublicLock.sol
NickCuso: Yup this should be clarified after the refactor.

smart-contracts/contracts/interfaces/hooks/ILockKeyPurchaseHook.sol
```diff
@@ -0,0 +1,50 @@
...
+   * @dev the lock's address is the `msg.sender` when this function is called via
+   * the lock's `purchasePriceFor` function
+   */
+  function keyPurchasePrice(
+    address from,
+    address recipient,
+    address referrer,
+    bytes calldata data
+  ) external view
+    returns (bool purchaseSupported, uint minKeyPrice);
```
nfurfaro: Just a refresher, but we return a bool here rather than reverting to keep options open for how failures are handled right?

NickCuso: Not sure if this is necessary... returning the bool is easier on JS integrations, but overkill for Solidity..

nfurfaro: Ok. I thought I remembered something( from the meeting about hooks )about wanting to allow a false value here to still be a valid response rather than just reverting. But I might be mixing this up with something else, that was a while ago and my mind has been elsewhere.

nfurfaro: I'm ok with a small optimization for js integrations here... it doesn't add much in the way of complexity or obfuscation.

nfurfaro:
This looks good Nick.

smart-contracts/contracts/interfaces/hooks/ILockKeyPurchaseHook.sol
NickCuso: I think I'm going to switch back to using revert instead.. I think it's more clear. This is a reasonable use case to revert a view call.

NickCuso: k switched, just waiting on ci

PR #6139, opened 2020-03-03
Update stripe components

cnasc:
# Description

In the process of implementing the stripe flow for the new checkout page, I found that stripe has deprecated their old library in favor of a new one. This new approach has some advantages (can load `stripe.js` without adding something to the head, no need to poll for the window variable, a _lot_ less boilerplate).

This PR explores replacing our existing stripe use with the new approach, and I'm pushing up a draft before committing further. The only things missing are some styling things, though it does need to go through some thorough manual testing.

One advantage of this approach is that now we can have a story for the stripe elements components, which would have been quite difficult before. We can also probably unit test them, though I haven't tried that yet.

Some of what's in here will also be foundational for the metadata form handling, in particular the use of `react-hook-form`.

# Checklist:


unlock-app/src/components/content/AccountContent.tsx
```diff
@@ -12,73 +11,42 @@ import { GridPadding, IframeWrapper } from '../interface/user-account/styles'
...
 import svg from '../interface/svg'
-
-interface StripeWindow {
-  Stripe?: stripe.StripeStatic
-}
-
-declare global {
-  interface Window extends StripeWindow {}
-}
+import { Card } from '@stripe/stripe-js'
```
cnasc: Typescript support is built-in now

julien51:
Great catch. Let me review this very soon.

unlock-app/src/components/interface/user-account/PaymentDetails.tsx
```diff
@@ -1,217 +1,96 @@
...
+        fontFamily: 'IBM Plex Sans, sans-serif',
+        fontSize: '16px',
+        lineHeight: '60px',
+      },
+    },
   }
+  
+  return (
+    <form onSubmit={handleSubmit(onSubmit)}>
+      <Input type="text" placeholder="Cardholder Name" name="Cardholder Name" ref={register({required: true})} />
```
cnasc: The `CardElement` actually collects all the important information here, including zip. We still gather the cardholder name, even though it's optional.

cnasc:
This is a bit further along now. Right now I'm getting `400` errors from locksmith when trying to set the token for a user. That probably indicates an issue with the signature, although everything looks good from that end at a glance.

julien51:
Would it be worth splitting to see what is going with the signature?

cnasc:
> Would it be worth splitting to see what is going with the signature?

Yeah, I think so. I'll pare this down and re-submit this afternoon

cnasc:
Closing in favor of #6157

Issue #6010, opened 2020-02-18
Lock contract size is almost maxed-out.


nfurfaro:
Creating this issue to track/discuss the issue and determine:

**1.)** Is this actually an issue for us?
  I've seen reference to using so-called "transparent" contracts to get around the size limitation. Does this apply to our architecture as it's currently implemented?

**2.)** Explore options for moving forward. There are many ways of both making the locks smaller, and limiting future growth. 

### References:
- https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md
- https://ethereum.stackexchange.com/questions/47539/how-big-could-a-contract-be
- https://github.com/ethereum/EIPs/issues/1538
- https://blog.polymath.network/solidity-tips-and-tricks-to-save-gas-and-reduce-bytecode-size-c44580b218e6

mudgen:
Let me know if anyone has any questions about ERC1538.

julien51:
Some recent cleanups have reduced the size of the contract.

PR #6143, opened 2020-03-04
[on hold till next version] Port getOwnersByPage to an extension

NickCuso:
# Description

Creates a `LockInspector` extension contract. I started with just `getOwnersByPage` but there may be other functions to port over - or just to add for convenience.

A small simplification to the lock contract and saves us a bit of space.

The tests are basically what we had previously, but I also removed the revert for this read-only call.

# Issues

Fixes #
Refs #6010 

# Checklist:


NickCuso:
Reduces PublicLock deployment by 90.5k gas. I'm unsure how much that is in terms of our byte limit.

NickCuso:
> UnlockUtils.sol

That's an internal library atm.  But the general point of should this be in smart-contracts or smart-contract-extensions is a good question. I'm torn on it.

Additionally maybe we do not merge this until after v7. We don't have a deploy process figured for extensions yet..

nfurfaro:
I think we should wait to merge till after v7.

PR #6147, opened 2020-03-04
Initial work on new paywall script

cnasc:
# Description

This PR introduces a new directory from which to work on the new paywall main window script, as well as a webpack config to build it.

Right now it just contains a `postmate` stub, and future work will involve porting over functionality from the old script as well as implementing newer functionality to meet our goals of size and simplicity.

# Checklist:

### DIFF
--- a/paywall/src/paywall-script/index.ts
+++ b/paywall/src/paywall-script/index.ts
@@ -1,0 +1,12 @@
+import Postmate from 'postmate'
+
+const baseUrl = 'http://localhost:3000/checkout'
+const rawConfig = (window as any).unlockProtocolConfig
+const encodedConfig = encodeURI(JSON.stringify(rawConfig))
+
+const handshake = new Postmate({
+  url: `${baseUrl}?paywallConfig=${encodedConfig}`,
+  classListArray: ['unlock-protocol-checkout'],
+})
+
+handshake.then(child => console.log(child))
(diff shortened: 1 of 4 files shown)


paywall/unlock.2.0.js.webpack.config.js
```diff
@@ -0,0 +1,38 @@
+/* eslint no-console: 0 */
+var path = require('path')
+const webpack = require('webpack')
+const configVariables = require('./environment')
+
+module.exports = () => {
+  return {
+    cache: false,
+    mode: 'development',
```
cnasc: Even with mode set to `development` or `none`, this takes about 10 seconds to build on my machine. It may be worth simply building this with `tsc`, which should be faster and it will be easier for us to implement automatic builds on changes. For the time being this works well enough.

paywall/src/paywall-script/index.ts
```diff
@@ -0,0 +1,12 @@
+import Postmate from 'postmate'
+
+const baseUrl = 'http://localhost:3000/checkout'
+const rawConfig = (window as any).unlockProtocolConfig
+const encodedConfig = encodeURI(JSON.stringify(rawConfig))
```
cnasc: These 3 values are the core functionality for setting up the connection to the child iframe. All we really need is a way to pick the base URL for a given environment and then it's all set.

paywall/src/paywall-script/index.ts
```diff
@@ -0,0 +1,12 @@
+import Postmate from 'postmate'
+
+const baseUrl = 'http://localhost:3000/checkout'
```
julien51: Do you want to use the ENV VAR set thru webpack?

julien51: You can replace this at compile time

Issue #6135, opened 2020-03-03
Implement communication between checkout and paywall script


cnasc:
To hook up the new checkout experience with the paywall script we embed, we need to set up a model so the two components can talk. Ideally this information exposed should be trivially extensible. 

### user information 

at the moment, just the address but we could conceivably want to provide more information in the future

The child will emit `userInformation` events as changes to this data occur.

Example value:
```javascript
{
  address: "0x...",
  // additional properties may be added to this object as needed
}
```

### close modal

When the child iframe emits the `closeModal` event, the parent window hides the child iframe.

### transaction information

When a transaction is initiated, the child iframe emits a `newTransaction` event containing information about the transaction. 

Example value:
```javascript
{
  hash: "0x...",
  // additional properties may be added to this object as needed
}
```

PR #6149, opened 2020-03-05
Implement dashboard side of postmate communication

cnasc:
# Description

This PR introduces the `usePostmateParent` hook, which performs the postmate handshake and returns the resulting `ChildAPI` once it resolves.

Additionally, this PR introduces a higher-level hook called 'useCheckoutCommunication` which wraps the `ChildAPI` with convenience functions for the messages we have defined.

# Issues

Refs #6135 (won't be done until we start listening on the paywall side)

# Checklist:


unlock-app/src/hooks/useCheckoutCommunication.ts
```diff
@@ -0,0 +1,43 @@
...
+  closeModal = 'checkout.closeModal',
+  transactionInfo = 'checkout.transactionInfo',
+}
+
+// This is just a convenience hook that wraps the `emit` function
+// provided by the parent around some communication helpers. Consumers
+// are intended to check the `ready` property to know when to start
+// using the hook. If they do not, runtime errors may occur if any of
+// these functions are called before the postmate handshake completes.
+export const useCheckoutCommunication = () => {
```
julien51: Probably Overkill but would there be a way to "cache" the calls until ready if but ready yet?

cnasc: We actually do something like that in the main window side of the current paywall: https://github.com/unlock-protocol/unlock/blob/c00cb0655289eec82a5f1f1e2a5e38f7e7a257ad/paywall/src/unlock.js/PostMessageEmitters/CheckoutIframeMessageEmitter.ts#L65

It's doable, but probably a bit more slippery when we're doing it in/with a hook

cnasc: I think it makes sense as a future enhancement

julien51: Thanks

PR #6151, opened 2020-03-05
New paywall script and checkout can communicate

### LINKED ISSUE #6135: Implement communication between checkout and paywall script

cnasc:
# Description

This PR hooks up the communication line between the "main window" hosting the paywall script, and the child iframe containing the new checkout UI. It's relatively small, but I'll leave some inline comments in key places.

This is a good milestone for us, because now that they're hooked up and we can auto-compile (though not hot reload) the script, iteration can be quick and reflective of actual use.

Next steps to really get this on the hot path include having a test page (like `NewDemoContent`) that injects the new script, and acting on the messages in a way beyond just logging them.

# Issues

Fixes #6135 

# Checklist:

### DIFF
--- a/paywall/package.json
+++ b/paywall/package.json
@@ -14,6 +14,7 @@
     "lint": "eslint --ext .tsx,.ts,.js src/",
     "svg-2-components": "./node_modules/@svgr/cli/bin/svgr --title-prop --no-dimensions --template src/components/interface/svg/template.js --no-dimensions -d src/components/interface/svg/ src/static/images/svg/",
     "build-checkout": "yarn build-unlock.1.0.js && yarn build-data-iframe.1.0.js",
+    "watch-checkout": "webpack --watch --config unlock.2.0.js.webpack.config.js",
     "build-unlock.1.0.js": "webpack --config unlock.1.0.js.webpack.config.js",
     "build-unlock.2.0.js": "webpack --config unlock.2.0.js.webpack.config.js",
     "build-unlock-web": "cross-env UNLOCK_ENV=prod webpack --config unlock-web-module.webpack.config.js",
--- a/paywall/src/paywall-script/index.ts
+++ b/paywall/src/paywall-script/index.ts
@@ -1,11 +1,27 @@
 import Postmate from 'postmate'
-import {
-  CheckoutEvents,
-  UserInfo,
-  TransactionInfo,
-} from '../../../unlock-app/src/hooks/useCheckoutCommunication'
 
 declare let __ENVIRONMENT_VARIABLES__: { unlockAppUrl: string }
+
+/**
+ * These type definitions come from `useCheckoutCommunication` in
+ * `unlock-app`. We'll have to keep them in sync manually because we
+ * don't have access to `unlock-app` files in the `paywall` docker
+ * image.
+ */
+export interface UserInfo {
+  address: string
+}
+
+export interface TransactionInfo {
+  hash: string
+}
+
+export enum CheckoutEvents {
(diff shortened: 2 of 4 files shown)


cnasc:
OK, this fails because I crossed the repo boundary while importing types. I'll fix that.

paywall/package.json
```diff
@@ -14,6 +14,7 @@
     "lint": "eslint --ext .tsx,.ts,.js src/",
     "svg-2-components": "./node_modules/@svgr/cli/bin/svgr --title-prop --no-dimensions --template src/components/interface/svg/template.js --no-dimensions -d src/components/interface/svg/ src/static/images/svg/",
     "build-checkout": "yarn build-unlock.1.0.js && yarn build-data-iframe.1.0.js",
+    "watch-checkout": "webpack --watch --config unlock.2.0.js.webpack.config.js",
```
cnasc: Pretty simple to have it automatically compile on change. The only downside is that this requires 2 terminal windows; otherwise the watcher blocks execution of the main paywall app.

paywall/src/paywall-script/index.ts
```diff
@@ -1,14 +1,31 @@
...
+  UserInfo,
+  TransactionInfo,
+} from '../../../unlock-app/src/hooks/useCheckoutCommunication'
 
 declare let __ENVIRONMENT_VARIABLES__: { unlockAppUrl: string }
 
 const { unlockAppUrl } = __ENVIRONMENT_VARIABLES__
 const rawConfig = (window as any).unlockProtocolConfig
-const encodedConfig = encodeURI(JSON.stringify(rawConfig))
+const encodedConfig = encodeURIComponent(JSON.stringify(rawConfig))
```
cnasc: This slight difference caused the config to not always encode in a reversible way.

paywall/src/paywall-script/index.ts
```diff
@@ -1,14 +1,31 @@
 import Postmate from 'postmate'
+import {
+  CheckoutEvents,
+  UserInfo,
+  TransactionInfo,
+} from '../../../unlock-app/src/hooks/useCheckoutCommunication'
```
cnasc: This turns out not to be kosher in the docker images, since this file is never copied over. I'd like to avoid duplicating this code, but I'll have to figure out the best way to do it.

unlock-app/src/hooks/usePostmateParent.ts
```diff
@@ -1,11 +1,11 @@
 import { useState, useEffect } from 'react'
-import { ChildAPI, Model } from 'postmate'
+import Postmate from 'postmate'
```
cnasc: It didn't like the destructuring import when actually running, so I changed it back to this

paywall/package.json
julien51: Would you mind updatting the README as well so we can keep track of that?

PR #6157, opened 2020-03-06
Minimal stripe update

cnasc:
# Description

Unlike #6139, this PR updates to the newer stripe components without touching the underlying form as much. It also completes the migration so that other usages of stripe work as before (mostly regarding the `Card` type).

# Checklist:


cnasc:
I want to test this on the staging draft URL before we merge

cnasc:
This is kind of weird:

image

The UI hangs in the "submitting" state, even though locksmith accepts the request and the app has the new card in the state. Nothing has really changed in the render code for SettingsContent so I'm not 100% sure what's going on there.

cnasc:
The particularly troublesome part of this is that I still get `400` responses from locksmith locally when attempting to save the card, so it's not easy to troubleshoot.

I've confirmed though that on staging, even after refreshing the page the settings page doesn't reflect the card that's in the redux state, so something definitely regressed on the render logic.

cnasc:
OK, pretty sure I've tracked down the bug. `accountReducer` was mutating the state variable when it received account updates, and that caused the card update not to be read. It was always a bug, it was just surfaced by this change (although I'm not sure exactly why)

cnasc:
Confirmed that the reducer fix keeps this operating normally.

Issue #5557, opened 2019-12-30
Key transfer rights


julien51:
In the current implementation of our keys, only their owner can transfer them.
We need a more flexible approach where a specific address can "own" the transfer rights and that address can be different from the key owner.

This enables these 2 scenarios:
1. Lending keys. As a user I want to be able to "lend" a key to someone else. By transferring ownership to them, but by keeping the transfer rights to myself, I can easily reclaim the key when I need it back!

2. key in CDP. A key has a secondary market value. As a user, I could transfer the key transfer rights to a CDP and get some of the key price back. If I maintain the collateralization ratio, then I can keep owning the key and pay it "back" as time passes. If I fail to pay it back, someone else can liquidate it and sell it on a secondary market.

julien51:
Similarly, it would be interesting to have a reflexion around the rights to change the beneficiary of a lock. Right now, it can be changed by the owner (or the beneficiary) but it would be interesting to limit the ability to change the beneficiary to another 3rd party address (multi sig for example!)

julien51:
We discussed this wih @akeem today. Another use case would be for the lock owner to delegate `grantKey` to a (trusted) 3rd party address.

Ideally, we'd need a very flexible solution where the lock owner can grant access to any function to any address, with a mechanism where it keep tracks of authorized users for functions y keeping track of hashes(function signature, user address) and a modifier on function could check if that mapping exists.

NickCuso:
The challenge here is for the CDP scenario we need to ensure that the original "owner" of the key no longer has permissions to transfer (vs just adding an account with transfer privileges).

Could we get clarification on the requirements to enable a CDP?  I'm not clear on what calls they will be making to our contract in order to make this possible.

julien51:
Let's not focus on the CDP use case for now.
I do think this will be important for "loans", but that needs to also include restrictions on cancellations as well.

This would also be (very) useful in the context of keys purchased thru credit card transactions. The keys purchased this way should only be transferable/cancellable by the purchaser (not the beneficiary) by default. This will make credit card purchases much safer.

NickCuso:
We could add a transfer/cancel hook, similar to the hook on purchase.  That callback could add transfer rules, reverting the tx when they don't have permissions.  Then we could launch a new contract with custom logic - e.g. I think for this use case we could say something like if purchased by x then apply the restriction.

julien51:
[...]

PR #6160, opened 2020-03-08
Updating cancellation permissions

nfurfaro:
# Description
This makes the  changes required to add the `keyManager` to `cancelAndRefund` & `cancelAndRefundFor`. It also makes some modifications to make things work:
- `keyOwnerToNonce` becomes `keyManagerToNonce`
- `consumeOffchainApproval` was changed to compare the signer's address to the keyManager rather than the keyOwner.
- an extra param `uint _tokenId` was added to both `cancelAndRefund` & `cancelAndRefundFor`. without it, it would be impossible to lookup the keyOwner and calculate the refund.

# Issues

Fixes #
Refs #5557

# Checklist:

### DIFF
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -258,7 +258,7 @@ contract IPublicLock is IERC721Enumerable {
   ) external view returns (uint);
 
   /**
-   * @dev Invoked by the lock owner to destroy the user's key and perform a refund and cancellation of the key
+   * @dev Invoked by the lock owner to expire the user's key and perform a refund and cancellation of the key
    * @param _keyOwner The key owner to whom we wish to send a refund to
    * @param amount The amount to refund the key-owner
    * @dev Throws if called by other than owner
@@ -270,13 +270,14 @@ contract IPublicLock is IERC721Enumerable {
   ) external;
 
    /**
-   * @dev Destroys the key managed by msg.sender and sends a refund
-   * based on the amount of time remaining.
+   * @dev allows the key manager to expire a given tokenId
+   * and send a refund to the keyOwner based on the amount of time remaining.
+   * @param _tokenId The id of the key to cancel.
    */
-  function cancelAndRefund(uint) external;
+  function cancelAndRefund(uint _tokenId) external;
 
   /**
-   * @dev Cancels a key managed by a different user and sends the funds to the msg.sender.
+   * @dev Cancels a key managed by a different user and sends the funds to the keyOwner.
    * @param _keyManager the key managed by this user will be canceled
    * @param _v _r _s getCancelAndRefundApprovalHash signed by the _keyManager
    * @param _tokenId The key to cancel
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -269,21 +269,24 @@ contract IPublicLock is IERC721Enumerable {
     uint amount
   ) external;
 
-  /**
-   * @notice Destroys the msg.sender's key and sends a refund based on the amount of time remaining.
+   /**
+   * @dev Destroys the key managed by msg.sender and sends a refund
+   * based on the amount of time remaining.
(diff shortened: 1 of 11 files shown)


smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -269,21 +269,24 @@ contract IPublicLock is IERC721Enumerable {
...
   ) external;
 
-  /**
-   * @notice Destroys the msg.sender's key and sends a refund based on the amount of time remaining.
+   /**
+   * @dev Destroys the key managed by msg.sender and sends a refund
+   * based on the amount of time remaining.
    */
-  function cancelAndRefund() external;
+  function cancelAndRefund(uint) external;
```
NickCuso: We should include the variable name here as well.

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -269,21 +269,24 @@ contract IPublicLock is IERC721Enumerable {
     uint amount
   ) external;
 
-  /**
-   * @notice Destroys the msg.sender's key and sends a refund based on the amount of time remaining.
+   /**
+   * @dev Destroys the key managed by msg.sender and sends a refund
```
NickCuso: instead of `by msg.sender` it's the token id provided right?

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -269,21 +269,24 @@ contract IPublicLock is IERC721Enumerable {
...
+   * based on the amount of time remaining.
    */
-  function cancelAndRefund() external;
+  function cancelAndRefund(uint) external;
 
   /**
-   * @dev Cancels a key owned by a different user and sends the funds to the msg.sender.
-   * @param _keyOwner this user's key will be canceled
-   * @param _v _r _s getCancelAndRefundApprovalHash signed by the _keyOwner
+   * @dev Cancels a key managed by a different user and sends the funds to the msg.sender.
```
NickCuso: Pretty sure I wrote this originally... but the comment is wrong. the funds are sent to the keyOwner, not msg.sender.

smart-contracts/contracts/interfaces/IPublicLock.sol
julien51:
```suggestion
   * @dev Expires the key managed by msg.sender and sends a refund
```

nfurfaro: done.

smart-contracts/contracts/interfaces/IPublicLock.sol
nfurfaro: done.

smart-contracts/contracts/interfaces/IPublicLock.sol
nfurfaro: good catch. fixed.

smart-contracts/contracts/interfaces/IPublicLock.sol
nfurfaro: @NickCuso So it's both. As per the permissions doc, `cancelAndRefund` is only callable by the keyManager of a given key. This is in line with what we had before, where a key owner could only cancel their own key. Now, someone can only cancel a key they manage (unless using metaTX and  `cancelAndRefundFor`
We need the token ID here because we have no way to look up the token given only the keyManager.

nfurfaro: Ah, but I could still change the comment to make it simpler. The implementation is as above though.

NickCuso: yup makes sense. but for clarity it's based on the tokenId since the keyManager may have control over may different keys

NickCuso: Maybe something like "allows the key manager to expire a given tokenId and send a refund to the keyOwner"

nfurfaro: Exactly. Fixed.

PR #6173, opened 2020-03-09
Paywall script lazy-loads and renders correctly

cnasc:
# Description

This PR hooks up the paywall communication a little more. Now, the `loadCheckoutModal` function is available, and it will load the iframe on the first click, and on subsequent clicks simply show the iframe that is already there.

This PR also adds some CSS and sets up the `NewDemoContent` page to use the new script when a query param is passed, which will allow us to more easily test this locally.

image

# Checklist:

### DIFF
--- a/paywall/src/components/content/NewDemoContent.tsx
+++ b/paywall/src/components/content/NewDemoContent.tsx
@@ -51,6 +51,7 @@ export default function NewDemoContent() {
     url.searchParams.get('unlockUserAccounts') === 'true'
   const useMetadataInputs: boolean =
     url.searchParams.get('metadataInputs') === 'true'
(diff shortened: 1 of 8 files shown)


paywall/src/components/content/NewDemoContent.tsx
```diff
@@ -51,6 +51,7 @@ export default function NewDemoContent() {
     url.searchParams.get('unlockUserAccounts') === 'true'
   const useMetadataInputs: boolean =
     url.searchParams.get('metadataInputs') === 'true'
+  const useNewCheckout: boolean = url.searchParams.get('newCheckout') === 'true'
```
julien51: Thanks! That's going to be useful!

cnasc:
Removed the Greyout component because it was causing some issues, replaced it with a transparent background for the time being.

image

PR #6180, opened 2020-03-10
Adding keyExpirationTimestampFor util

julien51:
# Description

This will allow us to keep the paywall embed script extremely small.

# Checklist:


julien51:
I defaulted to using `Fetch` which is  native in browsers rather than Axios to keep this code as small as possible.

cnasc:
I think using fetch is the right choice. We use it in a few other places, and I think we’re _very_ unlikely to run into problems. Absolute worst case there are small polyfills available.

PR #6186, opened 2020-03-11
Start to clean up paywall init script

cnasc:
# Description

This PR is a quick pass at adding some structure to the paywall init script. I'm going to add some tests for the various methods before this is merged.

# Checklist:


cnasc:
It's actually quite difficult to unit test the postmate handshake. I think we should defer that to the integration tests.

cnasc:
> Is this safe regarding the "old" paywall?

Yes, this does not modify anything currently in production. Additionally, the exposed interface is the same: you have an `unlockProtocolConfig` variable in your window, and when the script loads it uses that to instantiate the checkout.

PR #6187, opened 2020-03-11
Give LockManager access to grantKeys

nfurfaro:
# Description
I missed this in the previous PRs around permissions.

# Issues

Fixes #
Refs #5557

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinGrantKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinGrantKeys.sol
@@ -23,8 +23,8 @@ contract MixinGrantKeys is
     uint[] calldata _expirationTimestamps,
     address[] calldata _keyManagers
   ) external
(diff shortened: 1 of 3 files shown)


smart-contracts/contracts/mixins/MixinGrantKeys.sol
```diff
@@ -23,8 +23,8 @@ contract MixinGrantKeys is
     uint[] calldata _expirationTimestamps,
     address[] calldata _keyManagers
   ) external
-    onlyKeyGranter
```
NickCuso: Shouldn't we fix the modifier instead?

nfurfaro: Looking again, I think that's a better approach. I was thinking that There were some case where KeyGranter should be allowed access but not LockManager, but this isn't the case.

PR #6188, opened 2020-03-11
Paywall script unlocks page if user is found to already own a key

cnasc:
# Description

This PR updates the paywall init script to unlock the page on receiving an account address if that address already owns a key with an expiration date in the future. There will have to be some more work around this -- the checkout iframe needs to know which key that is so it can show the one that is already purchased and disable any extra purchases. I'll take care of that in a future PR.

# Checklist:

### DIFF
--- a/paywall/src/paywall-script/index.ts
+++ b/paywall/src/paywall-script/index.ts
@@ -103,6 +103,7 @@ export class Paywall {
     )
 
     if (timeStamps.some(val => val > new Date().getTime() / 1000)) {
+      // TODO: communicate to checkout iframe that there is already a valid key
       this.unlockPage()
     }
   }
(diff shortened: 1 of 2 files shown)


julien51:
Very cool to see this getting there!

paywall/src/paywall-script/index.ts
```diff
@@ -82,6 +87,27 @@ export class Paywall {
...
+        return keyExpirationTimestampFor(
+          readOnlyProvider,
+          lockAddress,
+          this.userAccountAddress!
+        )
+      })
+    )
+
+    if (timeStamps.some(val => val > new Date().getTime() / 1000)) {
+      // TODO: communicate to checkout iframe that there is already a valid key
```
julien51: I think it's better if the checkout iframe does this determination by itself, no?

cnasc: That actually would make it easier, and cleaner due to separation of concerns. That's how we'll handle it then!

PR #6190, opened 2020-03-11
Checkout component knows when a user already has a key

cnasc:
# Description

This PR is the mirror image of the last one on the paywall script side. In short, it queries Web3Service for keys associated with the locks on the paywall. If it returns any valid ones, they are reflected in the UI (and other key purchases are disabled).

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/Lock.tsx
+++ b/unlock-app/src/components/interface/checkout/Lock.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect } from 'react'
+import { KeyResult } from '@unlock-protocol/unlock-js'
 import { RawLock, Balances } from '../../../unlockTypes'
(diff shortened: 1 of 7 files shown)


unlock-app/src/components/interface/checkout/Lock.tsx
```diff
@@ -55,16 +58,18 @@ export const Lock = ({
 
   const canAfford = userCanAffordKey(lock, balances)
 
+  const keyForThisLock = activeKeys.find(key => key.lock === lock.address)
```
julien51: I wonder if this could not happen one level above? But I hguess this creates trouble to detect that a lock should be "disabled"?

cnasc: Yup, that's exactly the thought process I had. I was going to just pass a boolean, but for disabling other locks it was easiest to just pass the list down.

PR #6195, opened 2020-03-12
Handle NO_SUCH_KEY in keyExpirationTimestampFor

cnasc:
# Description

This PR adds a previously unhandled case in the `keyExpirationTimestampFor` util that caused the paywall to unlock when a user had never purchased a key for a given lock. This issue was that the smart contract returns a huge number used as a magic value to indicate when a user has no key and has never had a key. We were parsing this into an expiration timestamp that was guaranteed to be far in the future and we treated it as a valid key.

With the addition of a check, we now return an expiration of 0 in this case.

# Checklist:

### DIFF
--- a/paywall/src/utils/keyExpirationTimestampFor.ts
+++ b/paywall/src/utils/keyExpirationTimestampFor.ts
@@ -1,3 +1,6 @@
+export const NO_SUCH_KEY =
+  '0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000134841535f4e455645525f4f574e45445f4b455900000000000000000000000000'
+
 export const keyExpirationTimestampFor = async (
   provider: string,
   lock: string,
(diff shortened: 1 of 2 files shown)


paywall/src/utils/keyExpirationTimestampFor.ts
```diff
@@ -1,3 +1,6 @@
+export const NO_SUCH_KEY =
+  '0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000134841535f4e455645525f4f574e45445f4b455900000000000000000000000000'
```
akeem: Where does this value come from?

julien51: One of the smart contract versions returns `3963877391197344453575983046348115674221700746820753546331534351508065746944` which is `NO_SUCH_KEY`.

cnasc: It's MAX_INT on the smart contract

PR #6209, opened 2020-03-13
Adding a top-level store for the checkout iframe

cnasc:
# Description

This PR introduces a top-level store for the checkout iframe. I'm keeping this separate from the existing Redux store because there shouldn't be any cross-pollination between them. The basic step was to find all instances of `useState` and replace them with updates dispatched to the store. This lays the foundation for easier switching between views without dumping all our state (important for metadata).

# Checklist:

### DIFF
--- a/unlock-app/src/pages/_app.js
+++ b/unlock-app/src/pages/_app.js
@@ -15,7 +15,6 @@ import { WalletServiceContext } from '../utils/withWalletService'
 import { Web3ServiceContext } from '../utils/withWeb3Service'
 import { StorageServiceContext } from '../utils/withStorageService'
 import { StorageService } from '../services/storageService'
-import { CheckoutStoreProvider } from '../hooks/useCheckoutStore'
 
 import FullScreenModal from '../components/interface/FullScreenModals'
 import GlobalErrorConsumer from '../components/interface/GlobalErrorConsumer'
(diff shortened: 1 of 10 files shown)


cnasc:
This needs some additional tests before it passes, which I am working on now

unlock-app/src/pages/_app.js
```diff
@@ -155,9 +156,11 @@ The Unlock team
                 <Web3ServiceProvider value={web3Service}>
                   <WalletServiceProvider value={walletService}>
                     <ConfigProvider value={config}>
-                      <GlobalErrorConsumer>
-                        <Component {...pageProps} />
-                      </GlobalErrorConsumer>
+                      <CheckoutStoreProvider>
```
julien51: Would it make more sense to put that one level down, inside of a `FullCheckoutWrapper` component?

julien51: or maybe actually the existing `CheckoutContainer` ?

julien51: so that we keep the `App` stuff only for things which are "global"?

julien51:
I think this is generally good, but I have a tension around the use of `_app`. Could we wrap stuff inside of `CheckoutStoreProvider` at a "lower" level, like `CheckoutContainer`?

julien51:
This was it does not "pollute" the rest of the application?

cnasc:
Moved down a level from `_app`, ready for another review

PR #6211, opened 2020-03-13
saving transaction

julien51:
# Description

This is a WIP PR because I think it requires an upgrde to unlock-js to yield transaction data

I also need to write tests ;)

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/usePurchaseKey.ts
+++ b/unlock-app/src/hooks/usePurchaseKey.ts
@@ -2,6 +2,9 @@ import { useState, useContext } from 'react'
 import { WalletService } from '@unlock-protocol/unlock-js'
 import { RawLock } from '../unlockTypes'
 import { WalletServiceContext } from '../utils/withWalletService'
+import { ConfigContext } from '../utils/withConfig'
+import { StorageServiceContext } from '../utils/withStorageService'
+import { StorageService } from '../services/storageService'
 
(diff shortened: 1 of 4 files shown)


unlock-app/src/hooks/usePurchaseKey.ts
```diff
@@ -25,8 +72,21 @@ export const usePurchaseKey = (lock: RawLock, accountAddress: string) => {
...
+          storeTransactionFor(
+            config.services.storage.host,
+            accountAddress,
+            config.requiredNetworkId,
+            {
+              from: accountAddress,
+              lock: lock.address,
+              to: lock.address,
+              hash,
+              input: '', // TODO Is this required?
```
julien51: This piece will require an update from unlock-js. 
I'll move there for now!

unlock-app/src/services/storageService.js
```diff
@@ -53,19 +53,25 @@ export class StorageService extends EventEmitter {
...
+   * @paran {*} data (input of transaction, optional)
    */
   async storeTransaction(
     transactionHash,
     senderAddress,
     recipientAddress,
-    chain
+    chain,
+    beneficiaryAddress /** Used in the context of key purchase *for* */,
+    data
```
julien51: This was added because the paywall currently save that, so I assumed this was useful... maybe it won't but at least its there if we need it!

PR #6212, opened 2020-03-13
Basic optimistic unlocking

julien51:
# Description

This pull-request adds optimistic unlocking.
When the paywall script is loaded:
* if the user does not have a key, we fetch past transactions from locksmith and check if any transaction would result in a key, and unlock in that case
* then a transaction is sent back from the checkout iframe, we check if this transaction would result in a key and unlock in that case.

# Checklist:

### DIFF
--- a/paywall/src/utils/optimisticUnlocking.ts
+++ b/paywall/src/utils/optimisticUnlocking.ts
@@ -1,0 +1,73 @@
+/**
+ * Yields true if we should be optimistic for user and the locks
+ * @param locks
+ * @param user
+ */
+export const optimisticUnlocking = async (
+  locksmithUri: string,
+  locks: string[],
+  user: string
+) => {
+  const userTransactions = await getTransactionsForUserAndLocks(
+    locksmithUri,
+    user,
+    locks
+  )
+  const unlocked = await Promise.all(
+    userTransactions.map((transaction: any) => {
+      return willUnlock(
+        user,
+        transaction.recipient,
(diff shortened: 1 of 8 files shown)


paywall/src/utils/optimisticUnlocking.ts
```diff
@@ -0,0 +1,42 @@
...
+  locksmithUri: string,
+  locks: string[],
+  user: string
+) => {
+  const userTransactions = await getTransactionsForUserAndLocks(
+    locksmithUri,
+    user,
+    locks
+  )
+  console.log(userTransactions)
```
julien51: WIP ;)

locksmith/src/controllers/lockController.js
```diff
@@ -2,8 +2,8 @@ import LockOwnership from '../data/lockOwnership'
 
 const logger = require('../locksmithLogger')
 const lockOperations = require('../operations/lockOperations')
-const lockIconUtils = require('../../src/utils/lockIcon').default
-const { getBaseTokenData } = require('../../src/operations/metadataOperations')
+const lockIconUtils = require('../utils/lockIcon').default
+const { getBaseTokenData } = require('../operations/metadataOperations')
```
cnasc: You can rebase on master, this should be resolved now

cnasc:
Overall approach looks good, I think it just needs a rebase before merging

julien51:
@cnasc alright PTAL!

PR #6236, opened 2020-03-17
Fix compiler warning

### LINKED ISSUE #6233: Compile warning: unused `_keyManager`
**Describe the bug**
```
> Compilation warnings encountered:

    /home/unlock/smart-contracts/contracts/mixins/MixinKeys.sol:271:5: Warning: Unused function parameter. Remove or comment out the variable name to silence this warning.
    address _keyManager
    ^-----------------^
```

**To Reproduce**
yarn build

nfurfaro:
# Description
FIxes the use of params inside the `isKeyManager` internal function.

# Issues

Fixes #6233 
Refs #

# Checklist:


smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -272,8 +272,8 @@ contract MixinKeys is
   ) internal view
     returns (bool)
   {
-    if(keyManagerOf[_tokenId] == msg.sender ||
```
julien51: Should there be a pair of tests which make sure it passes with `_keyManager` and fails with `msg.sender` in the case where the sender is not the `_keyManager`?

nfurfaro: I could add this. This function is not tested directly yet.

nfurfaro: added some tests.

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -266,14 +266,14 @@ contract MixinKeys is
   * Returns true if _keyManager is the manager of the key
   * identified by _tokenId
    */
-  function isKeyManager(
+  function _isKeyManager(
     uint _tokenId,
     address _keyManager
   ) internal view
```
NickCuso: Should we make this public instead?

nfurfaro: I don't know. We already have the public mapping `keyManagerOf` available. This seems like something more needed internally maybe?

nfurfaro:
Yes, a good review of our permissions and maybe more testing before v7.

smart-contracts/contracts/mixins/MixinKeys.sol
nfurfaro: We can always make this public if the need arises. Will merge as-is for now.

PR #6239, opened 2020-03-17
usePurchaseKey now directly sets transaction hash in state

cnasc:
# Description

In support of the metadata form PR, so that the UI can update correctly, we need to track the transaction hash in the app-level state, not in a return value from the hook. This PR changes `usePurchaseKey` and its one consumer around to ensure that it does so.

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/usePurchaseKey.ts
+++ b/unlock-app/src/hooks/usePurchaseKey.ts
@@ -5,21 +5,23 @@ import { WalletServiceContext } from '../utils/withWalletService'
 import { ConfigContext } from '../utils/withConfig'
 import { StorageServiceContext } from '../utils/withStorageService'
 import { StorageService } from '../services/storageService'
+import { useCheckoutStore } from './useCheckoutStore'
+import { setTransactionHash } from '../utils/checkoutActions'
+import { TransactionInfo } from './useCheckoutCommunication'
 
-type TransactionHash = string | null
 type PurchaseError = Error | null
 
-export const usePurchaseKey = (lock: RawLock, accountAddress: string) => {
-  const [transactionHash, setTransactionHash] = useState(
-    null as TransactionHash
-  )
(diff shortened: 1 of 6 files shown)


unlock-app/src/utils/checkoutActions.ts
```diff
@@ -33,3 +37,13 @@ export const setPurchasingLockAddress = (
   kind: 'setPurchasingLockAddress',
   address,
 })
+
+interface SetTransactionHash {
+  kind: 'setTransactionHash'
```
akeem: It looks like an artifact of yestercode, but with definition of the type why do we also have the kind property of with a string value of its name?

unlock-app/src/hooks/usePurchaseKey.ts
```diff
@@ -41,10 +45,16 @@ export const usePurchaseKey = (lock: RawLock, accountAddress: string) => {
             accountAddress,
             transaction.data
           )
+
+          // emit transaction info to main window
+          emitTransactionInfo({
```
julien51: I wonder if this should eventually move to the `store`?

unlock-app/src/utils/checkoutActions.ts
cnasc: It's an implicit type predicate - each of these interfaces has a unique "kind;" this allows the compiler to infer the following:

```javascript
if (thing.kind = "setTransactionHash") {
  // thing is now understood to be of the type SetTransactionHash
}
```

This adds some help while editing (you'll get autocomplete for the relevant properties) and it also ensures you won't try to access something that isn't there.

unlock-app/src/hooks/usePurchaseKey.ts
cnasc: I was thinking of doing that, but didn't want to nest everything too deeply until we had the architecture solidified.

Issue #6199, opened 2020-03-12
Checkout iframe needs to be vetted and improved with unlock user accounts


cnasc:
Right now we only handle a subset of the necessary functionality for unlock user accounts to be first-class users of the new checkout. I'll update this issue with a list of required features as we enumerate what is missing.

julien51:
The flow should be this:

1. [ ] Once the user is not logged in, we show disabled locked and login button, and price should be in crypto (prices set on locks  by creators). Only show the log-in button if paywall config allows for `unlockUserAccounts` (this will be changed later though)

2. [ ] Once the user logs in we need to assess which lock can be purchased with CC. Implement a locksmith call for this (given a lock address, check whether it's in the list of approved locks, and return the price to be charged).

3. [ ] Show the $ price for locks which are CC approved and mark them as "purchasable".

4. [ ] Once the user clicks on the lock which is purchasable via CC, 
 a. [ ] If metadata needs to be collected, collect it and move to 4b. 
 b. [ ] show confirmation screen or CC form.

5. [ ] Once user submits, send a purchase call to locksmith and yield transaction hash to paywall script.

cnasc:
For the first task (list of locks when not logged in):

The code to do it is already there, but for some reason it doesn't show up in the iframe. Here's what I mean:

<img width="413" alt="image" src="https://user-images.githubusercontent.com/9300702/76897444-24b6c000-686a-11ea-9e5f-1de809426b23.png">
On the paywall

<img width="432" alt="image" src="https://user-images.githubusercontent.com/9300702/76897465-2e402800-686a-11ea-9473-b5ba0d6a963c.png">
Same config, but directly on the checkout page

Trying to figure out what's going on with that

cnasc:
Got it, had a `useEffect` only firing once so it didn't get the update to lock addresses.

cnasc:
I think this is all done now

PR #6252, opened 2020-03-18
adding a method which returns the pricing for credit card payment for any lock, if they have been approved

julien51:
# Description

adding a method which returns the pricing for credit card payment for any lock, if they have been approved.

cc @cnasc for API!

# Issues

Refs #6199

# Checklist:


locksmith/src/controllers/priceController.ts
```diff
@@ -15,6 +17,40 @@ namespace PriceController {
...
+      config.web3ProviderHost,
+      config.unlockContractAddress
+    )
+
+    // First, let's see if the lock was authorized for credit card payments
+    const isAuthorizedForCreditCard = await AuthorizedLockOperations.hasAuthorization(
+      lockAddress
+    )
+
+    if (!isAuthorizedForCreditCard) {
```
akeem: I think it might be worthwhile to throw an error for this case.  It should make it bit easier for the caller to react rather than doing a null check

locksmith/src/routes/price.ts
```diff
@@ -4,5 +4,6 @@ const router = express.Router()
 const priceController = require('../controllers/priceController')
 
 router.get('/:lockAddress', priceController.price)
+router.get('/fiat/:lockAddress', priceController.fiatPrice)
```
akeem: We could probably include the fiat price as at attribute under the original call, not sure if there is a  technical reason but this could make this a smaller change in  PR and a smaller change for callers to switch to

locksmith/src/controllers/priceController.ts
julien51: So, eventually I want us to be able to have multiple fiat currencies returned, so maybe instead of `null` we just yield an empty object?

locksmith/src/routes/price.ts
julien51: So the original call does not have the logic of not returning the price for locks which are not approved which I think is fine because we had other requirements. 
Here, I am trying to think forward of the approach we will eventually take where the front end will not care about what kind of lock that is, what kind of purchase (pruchaseFor or grantKey) that is... etc.

The front end will just display things and pass things to locksmith.

julien51: Also I do not want to "break" things/flows which are working

julien51:
I have no idea why these tests fail. They pass locally just fine :(

PR #6275, opened 2020-03-19
Unlockjs structure

akeem:
# Description

# Issues

Fixes #
Refs #

# Checklist:


docker/development/standup.sh
```diff
@@ -4,6 +4,6 @@ BLOCKTIME=${BLOCKTIME:-3}
 
 echo "using block time of ${BLOCKTIME} seconds"
 
-(node /standup/prepare-ganache-for-unlock.js &)
+# (node /standup/prepare-ganache-for-unlock.js &)
```
julien51: I like this approach but the docker image is used by more than just unlock-js (for exaple it is the one used for local dev). It is important that we have a way to run this script in more than just unlock-js

akeem: Agreed, this is currently in draft to shake things out. This looks like it will be a large change, id like to see if we can reach a state where the following occur:

1. The test manages its stand up/tear down
2. The development environment exists as it currently does
3. We untangle these concerns so that that they can develop independently. ie: the unlockjs test suite should be able to test an erc20 backed lock that doesnt currently exist in the development setup

unlock-js/package.json
```diff
@@ -42,7 +44,10 @@
     "@babel/cli": "7.8.4",
     "@babel/core": "7.8.7",
     "@babel/preset-env": "7.8.7",
+    "@unlock-protocol/unlock-js": "0.8.7",
```
julien51: I think we shold not require this dep here?

julien51: I am actually not sure how "circular" dev dependencies work. Maybe that's fine?

akeem:
> I think the approach is good, but I would love to avoid duplicating the setup files. Do you think we can "re-use" the ones in the `docker/dev` folder instead of duplicating them over?

@julien51 At the moment I think it would be more prudent to copy them rather than choose the wrong abstraction here. While serving similar use cases, the test setup for unlockjs should include only the minimal set it needs to function where as the development environment seems to include variable configuration and may include items that unlockjs does not require for its integration test

unlock-js/package.json
akeem: In the stand ups current form this is implicitly being used for its access to WalletService and latest.  

This is included as a development dependency to ensure that this change is a structural one to avoid the side effects of referencing the classes from the library itself now that it performs the test setup for itself. 

We can pull it but I think we should do it after this goes out to isolate any ripples

PR #6279, opened 2020-03-19
Adding UserAccountLock component

cnasc:
# Description

This PR adds the UserAccountLock component which will be used for credit card purchases. It needs a supporting hook to be added before it's useful, so this will stay as a draft until that is complete.

It is a spin on the existing lock component, and they share the underlying UI pieces, so keeping them in sync will not be difficult.

Ideally, the only custom piece we need is a credit card purchasekey hook. Everything else should behave the same as under a normal lock purchase.

# Checklist:


julien51:
> Ideally, the only custom piece we need is a credit card purchasekey hook. Everything else should behave the same as under a normal lock purchase.

Yes, that's exactly right.

PR #6280, opened 2020-03-19
Hook for user account key purchase

cnasc:
# Description

This is a hook that will support purchasing keys with a user account. The unit tests are not great here since nearly everything has to be mocked out, but we'll cover it with integration tests later.

# Checklist:


julien51:
Approach looks very good to me.

PR #6296, opened 2020-03-22
Communication via cookies

julien51:
# Description

In the context of a proxy based approach, we cannot use postMessage communications... but we can use cookies!

I am not yet 100% sure this is the right appoach. Maybe we can just *inject* a window listener at the proxy level which would perform the actions the user intends?

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Found a better approach, I think!

PR #6301, opened 2020-03-23
Remove the onlyKeyManager modifier

NickCuso:
# Description

Removing onlyKeyManager and using onlyKeyManagerOrApproved instead. I'm not sure there's ever a use case where we do not want both.

However after looking at this I think we should clear approval when the keyManager is changed. Will follow up with another PR.

# Issues

Fixes #
Refs #

# Checklist:


nfurfaro:
Based on this doc there are (currently) a number of such cases. We can review and see if this is still accurate.

https://docs.google.com/document/d/11bP7OjY2ei_NO1pzYOoTlrvQ1_9AtqXfmtgwZUYMwY0/edit?pli=1#heading=h.cnmy30f2sp2u

nfurfaro:
>However after looking at this I think we should clear approval when the keyManager is changed.
Yeah I agree here.

NickCuso:
Hmm I think that was a miss when we reviewed this.

Do you remember the rational?
To me it seems to violate the intent of approval - basically I'm approving an operator to act on my behalf. The could transfer the key to themselves and then cancel anyways

nfurfaro:
>The could transfer the key to themselves and then cancel anyways

True. Once they transfer to themselves they would become the keyManager anyway and gain access.

PR #6307, opened 2020-03-23
Pull provider correctly in useProvider

cnasc:
# Description

The previous iteration of `useProvider` actually just got the _name_ of the provider from the state. This PR updates it to use that name to get the _actual_ provider out of the config.

Additionally, this PR corrects a misnamed property in the purchase request typed data, which now allows user account purchases to proceed.

# Checklist:


cnasc:
> Seems fine, looking at the PR its not clear if there is something that would raise the flag if/when the call to `signKeyPurchaseRequestData` may be missing some required data

Yeah, unfortunately the file it's imported from does not have type definitions, so thee inference only goes as far as method names and not method signatures. Something to improve in the future.

PR #6320, opened 2020-03-24
adding configureUnlock

julien51:
# Description

Version 7 introduces a new way to configure the unlock contract so I needed to make this version specific rather than "global".

# Checklist:

### DIFF
--- a/unlock-js/src/v13/configureUnlock.js
+++ b/unlock-js/src/v13/configureUnlock.js
@@ -1,0 +1,27 @@
+/**
+ * Configures unlock version v13
+ * @param {PropTypes.lock} publicLockTemplateAddress
+ * @param {PropTypes.string} globalTokenSymbol
+ * @param {PropTypes.string} globalBaseTokenURI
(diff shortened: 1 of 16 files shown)


unlock-js/src/v13/configureUnlock.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Creates a lock on behalf of the user, using version v0
```
akeem: Does this comment reflect the code below? Is there a reason why we would be using v0 here rather than v13?

julien51: ha! Wrong comments. Let me fix.

julien51:
So the duplication is actually kind of necessary for now because we may/will get rid of some versions eventually and that's easier if each version is full self-contained.

julien51:
> Could definitely be an issue if the behavior is expected to change from version to version

Ha, yes I should have clarified that indeed v7 is different which is why this is moving from the shared "trunk" to the local versions.

PR #6343, opened 2020-03-26
paywall is available as module

julien51:
# Description

Created a module which exports the `isUnlocked` method which can be used in different contexts
This is alpha release... as there are things to imporve!
Feel free to chime in!

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Here is a script

```
const paywall = require('@unlock-protocol/paywall')

async function run() {
  const userAccountAddress = '0x62139339a2a3966c9660cab942e27f22b549f306'
  const paywallConfig = {
    locks: {
      '0xd3fCD4c1fA2952CF3B1a3f531C95b4b89F0974E6': {},
    },
  }
  const readOnlyProvider =
    'https://eth-mainnet.alchemyapi.io/jsonrpc/6idtzGwDtRbzil3s6QbYHr2Q_WBfn100'
  const locksmithUri = 'https://locksmith.unlock-protocol.com/'
  const unlocked = await paywall.default.isUnlocked(
    userAccountAddress,
    paywallConfig,
    {
      readOnlyProvider,
      locksmithUri,
    }
  )
  console.log({
    unlocked,
  })
}

run()

```

julien51:
i do not like the `paywall.default.isUnlocked` but not sure how to avoid it :/

paywall/src/utils/isUnlocked.ts
```diff
@@ -1,11 +1,5 @@
 import { keyExpirationTimestampFor } from './keyExpirationTimestampFor'
 import { optimisticUnlocking } from './optimisticUnlocking'
-
-declare let __ENVIRONMENT_VARIABLES__: {
```
julien51: This cannot work in a module... so we are now passing these values to the function

paywall/src/paywall-script/module.js
```diff
@@ -0,0 +1,5 @@
+import { isUnlocked } from '../utils/isUnlocked'
+
+export default {
```
julien51: Eventually, I think we could export the `Paywall` object too, which then front end applications could use directly!

paywall/src/paywall-script/webpack.config.js
```diff
@@ -0,0 +1,37 @@
...
+          exclude: /node_modules/,
+        },
+      ],
+    },
+    resolve: {
+      extensions: ['.ts', '.js'],
+    },
+    plugins: [
+      new webpack.DefinePlugin({
+        fetch: "require('node-fetch')",
```
julien51: This feels *very* dirty but basically we replace any mention of `fetch` with `require('node-fetch')` at build time.

paywall/src/paywall-script/module.js
cnasc: I believe if you don't do a default export you'll be able to reference it as 

```javascript
const paywall = require('@unlock-protocol/paywall')
paywall.isUnlocked // ...
```

Without having the default in between

julien51: but then how do I `export`?

cnasc: You just `export` without the default keyword

paywall/src/utils/isUnlocked.ts
```diff
@@ -14,10 +8,9 @@ declare let __ENVIRONMENT_VARIABLES__: {
  */
 export const isUnlocked = async (
   userAccountAddress: string,
-  paywallConfig: any
+  paywallConfig: any,
```
cnasc: https://github.com/unlock-protocol/unlock/blob/c65641fcd38078409b1b828645a924698482e335/paywall/src/unlockTypes.ts#L85

paywall/src/utils/isUnlocked.ts
```diff
@@ -14,10 +8,9 @@ declare let __ENVIRONMENT_VARIABLES__: {
...
+  { readOnlyProvider, locksmithUri }: any
```
cnasc: I think these should both be strings?

paywall/src/paywall-script/module.js
julien51: oh! of course! let me try!

cnasc:
LG overall with a few comments. I think it would be nice if we could also generate a `.d.ts` file to include in the module, would be a big benefit to any TS users on the consumption side.

paywall/src/paywall-script/module.js
julien51: worked :)

paywall/src/paywall-script/module.d.ts
```diff
@@ -0,0 +1,12 @@
+interface PaywallModuleConfig {
+  readOnlyProvider: string
+  locksmithUri: string
+}
+
+interface PaywallModule {
```
cnasc: I think you might want
```typescript
export function  isUnlocked(
    userAccountAddress: string,
    paywallConfig: string,
    config: PaywallModuleConfig
  ): Promise<boolean>
```

PR #6350, opened 2020-03-27
adding isLockManager method

julien51:
# Description

Since v7 does not have lock ownwers any more, we are adding a function to check if an address is the lock manager.
This will mostly be used in locksmith which relies on a user being a lock owner before allowing them to do some changes or accessing some data.

# Checklist:


julien51:
This is actually a smaller PR than it looks since almost all implementations for each versions are the same... except for v7.

PR #6365, opened 2020-03-30
using isLockManager in unlock-app

cnasc:
# Description

I believe this PR wraps up the last uses of `lock.owner` in `unlock-app`

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useIsLockManager.ts
+++ b/unlock-app/src/hooks/useIsLockManager.ts
@@ -9,17 +9,19 @@ export const useIsLockManager = (
   const web3Service: Web3Service = useContext(Web3ServiceContext)
 
   const [isLockManager, setIsLockManager] = useState(false)
+  const [loading, setLoading] = useState(true)
 
   async function getLockManagementStatus() {
     const ilm = await web3Service.isLockManager(lockAddress, accountAddress)
     setIsLockManager(ilm)
+    setLoading(false)
   }
 
   useEffect(() => {
     getLockManagementStatus()
   }, [lockAddress, accountAddress])
 
-  return { isLockManager }
+  return { isLockManager, loading }
 }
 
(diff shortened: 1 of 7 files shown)


unlock-app/src/hooks/useIsLockManager.ts
```diff
@@ -0,0 +1,25 @@
...
+export const useIsLockManager = (
+  lockAddress: string,
+  accountAddress: string
+) => {
+  const web3Service: Web3Service = useContext(Web3ServiceContext)
+
+  const [isLockManager, setIsLockManager] = useState(false)
+
+  async function getLockManagementStatus() {
+    const ilm = await web3Service.isLockManager(lockAddress, accountAddress)
```
julien51: Nit: Meh. not a great variable name...

unlock-app/src/hooks/useMembers.js
```diff
@@ -131,12 +135,17 @@ export const useMembers = (lockAddresses, viewer, filter) => {
     setLoading(true)
     const membersForLocksPromise = keyHolders.locks.map(async lock => {
       // If the viewer is not the lock owner, just show the members from chain
-      if (lock.owner.toLowerCase() !== viewer.toLowerCase()) {
+      const isLockManager = await web3Service.isLockManager(
```
julien51: I am a bit worried about "flashes" here since there is going to be a small latency to check `isLockManager`. This means we'll start rendering the "not a lock manager" branch and then will render the "a lock manager" branch.

Since the default value is false, I also wonder if ti will not re-render twice even in the case of the viewer not being the lock mamager?

Maybe it would be good to add a `const [loading, setLoading] = useState(true)` to the hook and return early while rendering?

julien51:
Thanks Chris!

PR #6379, opened 2020-03-31
Allow user accounts to submit metadata at checkout

cnasc:
# Description

What it says on the tin

# Checklist:


unlock-app/src/services/unlockProvider.js
```diff
@@ -64,6 +64,11 @@ export default class UnlockProvider extends providers.JsonRpcProvider {
     return sigUtil.personalSign(privateKey, { data })
   }
 
+  eth_signTypedData([_, data]) {
```
cnasc: This was the key missing piece, RPCProvider doesn't seem to implement this method so we needed an internal implementation

PR #6401, opened 2020-04-02
Re-enable smart contract tests after previously being disabled for failure

akeem:
# Description

We end up in a scenario where the environment flag was being flipped. It has been updated to explicitly check its value being true.

For the time being this is being run without the CI flag set until all other tests have been completed so that networking can be addressed

# Issues

Fixes #
Refs #

# Checklist:


akeem:
to be clear I am not sure why it started failing this morning

PR #6403, opened 2020-04-02
adding stripeCustomer

julien51:
# Description

To keep things simpler, we're moving the stripe customer id away from the user reference table into their own table.

# Checklist:


locksmith/src/models/stripeCustomer.ts
```diff
@@ -0,0 +1,11 @@
+import { Table, Column, Model } from 'sequelize-typescript'
+
+@Table({ tableName: 'StripeCustomer', timestamps: true })
```
akeem, line 3: I am not sure what is the scheme we currently have in place in the app, but this needs to match the value used in the migration

julien51: Doh. Sorry. thanks for catching it!

PR #6406, opened 2020-04-03
Providing authentication around  requesting cards by email address

akeem:
# Description

In order to close an information hole, callers are now required to sign this get request. Invalid scenarios will receive a 401 error code in response.

There is a bit swirling around in here, since the initial request is a GET we have kept that the same and expecting the personal signed version of the payload attached as a query parameter to the request.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/__tests__/controllers/userController/cards.test.ts
+++ b/locksmith/__tests__/controllers/userController/cards.test.ts
@@ -1,0 +1,118 @@
+import request from 'supertest'
+
+import app = require('../../../src/app')
+import Base64 = require('../../../src/utils/base64')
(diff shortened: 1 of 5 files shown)


julien51:
Let's not merge yet please!

julien51:
@cnasc rightfully noted that we need to update the "old" paywall/checkout for this to be merged first.

So I think this will have to come "last", once we have the new paywall in place...

julien51:
Actually we may never need this, because the new paywall will not use these routes at all.

akeem:
@julien51 ah okay, I guess give it a close if this is no longer required

Issue #6408, opened 2020-04-03
Signing the card retrieval request


julien51:
Locksmith is adding [signature verification to the API call](https://github.com/unlock-protocol/unlock/pull/6406) to list cards for a user. We need to make sure this call is signed by unlock-app.

cnasc:
One major question around how this will be handled:
Right now, we get cards right after we get an account. This means that once we start supporting cards for crypto wallet users they will be prompted for a signature every time they load unlock-app, whether they want to use a credit card or not (or whether it even makes sense for them to do so). How to we intend to minimize that impact?

PR #6416, opened 2020-04-03
SettingsContent retrieves cards with a hook

cnasc:
# Description

This PR adds a `useCards` hook to retrieve cards from locksmith. It passes along an authentication payload.

It also uses this hook in place of Redux in `SettingsContent`
# Issues

Fixes #
Refs #6408 (won't be resolved until we've removed all uses of the storageService method)

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useCards.ts
+++ b/unlock-app/src/hooks/useCards.ts
@@ -1,0 +1,90 @@
+import { useContext, useEffect, useState } from 'react'
+import { WalletService } from '@unlock-protocol/unlock-js'
+import { Card } from '@stripe/stripe-js'
+import { WalletServiceContext } from '../utils/withWalletService'
+import { ConfigContext } from '../utils/withConfig'
+import { Account } from '../unlockTypes'
+
+interface Config {
+  services: {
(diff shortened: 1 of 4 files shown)


unlock-app/src/hooks/useCards.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+}
+
+export const genAuthorizationHeader = (token: string) => {
+  return { Authorization: ` Bearer ${token}` }
+}
+
+// Taken from locksmith's userController/cards.test.ts
+export function generateTypedData(publicKey: string) {
```
julien51, line 21: 👍

julien51:
> (won't be resolved until we've removed all uses of the storageService method) 

Are there other places where this is being used now that we do this thru the hook?

cnasc:
> > (won't be resolved until we've removed all uses of the storageService method)
> 
> Are there other places where this is being used now that we do this thru the hook?

We use it in storageMiddleware to populate the account in Redux, and then we use that in both the new checkout and the current checkout

PR #6417, opened 2020-04-03
switching to using package for svg assets for unlock-app

julien51:
# Description

Next: unlock-protocol.com !
I kept the story to make sure we can easily see what is going on there...

# Checklist:


julien51:
ha, that's a great point... do you think we should remove that story?

cnasc:
> ha, that's a great point... do you think we should remove that story? I like the idea of being able to check that everything looks "good" in `unlock-app`.

I'm OK either way, I don't know that we update the SVGs frequently enough for it to be an issue

julien51:
ok, YOLO. We'll remove if it's too much of a pain eventually.

PR #6451, opened 2020-04-07
Replace uses of raw paywall config in CheckoutContent

cnasc:
# Description

`<CheckoutContent>` uses the paywall configuration object to determine what to render. It can receive this config in two ways: as a search param in the url (typically used in testing, when there is no parent page): `configFromSearch`, or as a message coming from the host page (when embedded into a page via the paywall script): `config`. 

Because the underlying code shouldn't care about how it receives the config, we choose which config to use (precedent going to the config that gets messaged into the page by the paywall) and store it in a variable called `paywallConfig`. However, there was an oversight where some child components referred to the raw `config` variable, which isn't present when running the checkout as a standalone page. This allowed us to enter an unrepresentable state and caused runtime errors.

This PR cleans up `<CheckoutContent>` so that all child components use the `paywallConfig` variable.

# Checklist:


julien51:
Thanks!

PR #6466, opened 2020-04-09
Provider cleanup

julien51:
# Description

This adds the missing part for the checkout components

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/middlewares/providerMiddleware.test.ts
+++ b/unlock-app/src/__tests__/middlewares/providerMiddleware.test.ts
@@ -2,10 +2,9 @@ import providerMiddleware, {
   changePassword,
   initializeUnlockProvider,
 } from '../../middlewares/providerMiddleware'
-import { SET_PROVIDER, providerReady } from '../../actions/provider'
+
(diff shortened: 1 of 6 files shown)


unlock-app/src/components/content/CheckoutContent.tsx
```diff
@@ -43,6 +44,11 @@ export const CheckoutContent = ({
   configFromSearch,
   errors,
 }: CheckoutContentProps) => {
+  const { loading } = useProvider()
+  if (loading) {
+    return <></>
```
cnasc, line 49: Do we want to return nothing here? That will likely cause an increase in perceived latency for the end user

julien51: That's a good point. I think the _pick provider_ logic will soon be moved to another place. I will address this in there.

cnasc:
Do you think we could split this up a bit? Maybe extend the hook first, then follow up with introducing the new capabilities?

julien51:
That's a good idea. I'll do that.

PR #6473, opened 2020-04-10
useCards now has a method to save cards

julien51:
# Description

Pushing this as is, as I am working on tests so you can validate the approach!

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/hooks/useCards.test.ts
+++ b/unlock-app/src/__tests__/hooks/useCards.test.ts
@@ -4,7 +4,7 @@ import { renderHook } from '@testing-library/react-hooks'
 import fetch from 'jest-fetch-mock'
 import { WalletServiceContext } from '../../utils/withWalletService'
 import { ConfigContext } from '../../utils/withConfig'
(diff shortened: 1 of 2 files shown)


unlock-app/src/components/content/SettingsContent.tsx
```diff
@@ -18,14 +18,16 @@ interface PaymentSettings {
...
+  const { cards, loading, saveCard } = useCards(address)
   if (loading) {
     return <Loading />
   }
   if (cards.length > 0) {
     return <PaymentMethods cards={cards} />
   }
-  return <PaymentDetails />
+  return (
+    <PaymentDetails saveCard={(token: string) => saveCard(address, token)} />
```
cnasc, line 29: Because we pass an address into useCards, do you think it would make sense to do this wrapping internal to the hook?

julien51: yes, I'll fix this!

julien51:
ok, I also removed the component updates and will add them to another PR for simplicity!
PTAL!

PR #6474, opened 2020-04-10
Providing implementation of keypricerUSD

akeem:
# Description

This pulls all the components together, given the address of a lock contract provide the price in USD when possible and throws when not. 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/__tests__/utils/keyPricer.test.ts
+++ b/locksmith/__tests__/utils/keyPricer.test.ts
@@ -1,4 +1,5 @@
 import KeyPricer from '../../src/utils/keyPricer'
+import PriceConversion from '../../src/utils/priceConversion'
 
 const standardLock = {
   asOf: 227,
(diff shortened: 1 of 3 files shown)


locksmith/__tests__/utils/keyPricer.test.ts
```diff
@@ -26,8 +41,29 @@ jest.mock('@unlock-protocol/unlock-js', () => ({
...
 
 const keyPricer = new KeyPricer('provider url', 'unlock contract address')
+const spy = jest.spyOn(PriceConversion.prototype, 'convertToUSD')
 
 describe('KeyPricer', () => {
+  describe('keyPriceUSD', () => {
+    describe('when the lock currency has an exchange rate on coinbase', () => {
+      it('returns the key price in USD', async () => {
+        expect.assertions(1)
+        await keyPricer.keyPriceUSD('an address')
```
cnasc, line 51: Should we check what the return value is from these?

akeem: @cnasc good question, the code generating the values has coverage under `PriceConversion` im proceeding as if this is an integration test. As long as this method calls the method with the correct params we'll assume its sorted of see a failure in a set of tests at the lower level

PR #6475, opened 2020-04-10
Add LockRoles extension

NickCuso:
# Description

An extension mixin to make it easier to interface with various roles from the lock.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contract-extensions/contracts/mixins/LockRoles.sol
+++ b/smart-contract-extensions/contracts/mixins/LockRoles.sol
@@ -1,0 +1,64 @@
+pragma solidity ^0.5.0;
+
+import 'unlock-abi-7/IPublicLockV7.sol';
+
+
+/**
+ * @notice Provides modifiers for interfacing with roles from a lock.
+ */
+contract LockRoles
+{
+  /**
+   * @notice Lock managers are admins for the lock contract.
+   */
+  modifier onlyLockManager(
+    IPublicLockV7 _lock
+  )
+  {
+    require(_lock.isLockManager(msg.sender), 'ONLY_LOCK_MANAGER');
+    _;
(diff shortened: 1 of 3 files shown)


smart-contract-extensions/contracts/mixins/LockRoles.sol
```diff
@@ -0,0 +1,64 @@
...
+  ) internal view
+    returns (bool)
+  {
+    address keyManager = _lock.keyManagerOf(_tokenId);
+    if(keyManager != address(0))
+    {
+      return keyManager == _user;
+    }
+    return _lock.isKeyOwner(_tokenId, _user);
+  }
```
nfurfaro, line 50: Could we use the existing `_isKeyManager()` function from `MixinKeys.sol` here?

smart-contract-extensions/contracts/mixins/LockRoles.sol
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * @notice The key manager role is required to transfer or cancel a key.
```
nfurfaro, line 53: An `approved` address should have this right also, correct?

smart-contract-extensions/contracts/mixins/LockRoles.sol
NickCuso: Yea, longer term I think we should change that to public.

smart-contract-extensions/contracts/mixins/LockRoles.sol
NickCuso: I removed this modifier. We don't have a use case for it yet and it was getting complicated.

PR #6488, opened 2020-04-13
adding a way to delete a card

julien51:
# Description

While working on the credit card UI I found that it would be important to have a way to remove payment info.
Rather than rely on Stripe, we remove the customer id info. This guarantees we can no longer use the card.

# Checklist:

### DIFF
--- a/locksmith/src/operations/stripeOperations.ts
+++ b/locksmith/src/operations/stripeOperations.ts
@@ -59,7 +59,35 @@ export const saveStripeCustomerIdForAddress = async (
   }
 }
 
+/**
+ * Method which delets the stripe customer id for an address
+ */
+export const deletePaymentDetailsForAddress = async (
+  publicKey: ethereumAddress
+) => {
+  const normalizedEthereumAddress = Normalizer.ethereumAddress(publicKey)
+
+  // First, let's delete the StripeCustomer
+  const deletedStripeCustomer = await StripeCustomer.destroy({
+    where: { publicKey: { [Op.eq]: normalizedEthereumAddress } },
+  })
+
+  // Then, update UserReference
+  // TODO: deprecate when all stripe_customer_id UserReferences have been moved to use StripeCustomer
+  const [updatedUserReference] = await UserReference.update(
+    {
+      stripe_customer_id: null,
+    },
+    {
+      where: { publicKey: { [Op.eq]: normalizedEthereumAddress } },
+    }
+  )
+
+  return deletedStripeCustomer > 0 || updatedUserReference > 0
+}
+
 export default {
-  saveStripeCustomerIdForAddress,
+  deletePaymentDetailsForAddress,
   getStripeCustomerIdForAddress,
+  saveStripeCustomerIdForAddress,
 }
--- a/locksmith/__tests__/controllers/userController/cards.test.ts
+++ b/locksmith/__tests__/controllers/userController/cards.test.ts
@@ -6,6 +6,7 @@ import ethJsUtil = require('ethereumjs-util')
 import models = require('../../../src/models')
 import sigUtil = require('eth-sig-util')
 import UserOperations = require('../../../src/operations/userOperations')
+import StripeOperations = require('../../../src/operations/stripeOperations')
 
 const { UserReference } = models
 const { User } = models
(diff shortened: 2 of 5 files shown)


locksmith/src/operations/stripeOperations.ts
```diff
@@ -59,7 +59,35 @@ export const saveStripeCustomerIdForAddress = async (
...
+  const [updatedUserReference] = await UserReference.update(
+    {
+      stripe_customer_id: '',
+    },
+    {
+      where: { publicKey: { [Op.eq]: normalizedEthereumAddress } },
+    }
+  )
+
+  return deletedStripeCustomer > 0 || updatedUserReference > 0
```
akeem, line 86: What are some of the reasons that this would not be  true? It looks like `updatedUserReference` is initially stored into its own array, here it looks like we are checking to see if the result is greater than zero is this design intentional?

unlock-app/src/components/interface/user-account/PaymentDetails.tsx
```diff
@@ -12,17 +12,18 @@ import { Input, Label, Select, Button } from '../checkout/FormStyles'
...
-import { useProvider } from '../../../hooks/useProvider'
-import { StorageServiceContext } from '../../../utils/withStorageService'
-import { StorageService } from '../../../services/storageService'
 
 const { stripeApiKey } = configure()
 const stripePromise = loadStripe(stripeApiKey)
 
-export const PaymentDetails = () => {
+interface PaymentDetailsProps {
+  saveCard: (stripeToken: string) => any
```
akeem, line 20: There are a few references to `saveCard` here, is this in support of the feature to delete cards or is this from some unrelated work?

If it is related to this feature, i'd like to suggest renaming this to better reflect the usage

locksmith/src/operations/stripeOperations.ts
```diff
@@ -59,7 +59,35 @@ export const saveStripeCustomerIdForAddress = async (
...
+/**
+ * Method which delets the stripe customer id for an address
+ */
+export const deletePaymentDetailsForAddress = async (
+  publicKey: ethereumAddress
+) => {
+  const normalizedEthereumAddress = Normalizer.ethereumAddress(publicKey)
+
+  // First, let's delete the StripeCustomer
+  const deletedStripeCustomer = await StripeCustomer.destroy({
```
akeem, line 71: Not sure of the exact impetus for this feature but, for the sake of completeness the card should probably be removed from stripe, I do believe that their console allows for charges to be created there.

locksmith/src/operations/stripeOperations.ts
```diff
@@ -59,7 +59,35 @@ export const saveStripeCustomerIdForAddress = async (
...
+    where: { publicKey: { [Op.eq]: normalizedEthereumAddress } },
+  })
+
+  // Then, update UserReference
+  // TODO: deprecate when all stripe_customer_id UserReferences have been moved to use StripeCustomer
+  const [updatedUserReference] = await UserReference.update(
+    {
+      stripe_customer_id: '',
```
akeem, line 79: Not sure if this is being set to empty string due to a table constraint not allowing  NULL or what but this generally creates down stream pain where people need to have knowledge about this rather than a value existing or being NULL.

If possible we should set this to NULL

julien51:
Dang the front end code is actually not related and was removed :/

locksmith/src/operations/stripeOperations.ts
julien51: Sequlize will return the number of rows affected. If this is 0 it means the row was not changed.

julien51: The second object in the array is the list of rows affected.

unlock-app/src/components/interface/user-account/PaymentDetails.tsx
julien51: Sorry this is irrelevant to this PR :/ I did `git checkout` from a branch which had this.

locksmith/src/operations/stripeOperations.ts
julien51: This is something that I was thinking we should do at another level.

locksmith/src/operations/stripeOperations.ts
julien51: good point. let me check if NULL is possible.

locksmith/src/operations/stripeOperations.ts
julien51: I'll ad this code.

julien51: It requires a bit more work since we currently do not keep track of the `source` specificially, so that means we would need to list all sources and remove them all.

julien51:
also changed the `''` to `null`. PTAL ;)

PR #6497, opened 2020-04-14
Freeze data-iframe and unlock.1.0 at current state

cnasc:
# Description

This PR adds the current deployed production versions of `data-iframe.1.0.min.js` and `unlock.1.0.min.js` to the repository and stops building these scripts as part of the deploy process. The effect is that the current deployed versions are frozen, and we can in future PRs delete the code used to produce them.

Here's the process I used to get these files (in the root paywall dir):
`curl https://paywall.unlock-protocol.com/static/unlock.1.0.min.js --output src/static/unlock.1.0.min.js `

`curl https://paywall.unlock-protocol.com/static/data-iframe.1.0.min.js --output src/static/data-iframe.1.0.min.js`

Because configuration is baked into these built artifacts, they will only work in a mainnet context.

We should test carefully against production locks after this is deployed.

# Checklist:


cnasc:
On second thought, this will break the remaining integration tests based on the legacy paywall, since they don't run in a mainnet context

PR #6506, opened 2020-04-14
Initial pass at functionality allowing service fee & charging for connected account

akeem:
# Description

An initial pass at test coverage around using a connected account to handle charging a customer and taking a service fee. 

In its current state, it could probably be tightened up but we can probably build up on this

# Issues

Fixes #
Refs #

# Checklist:


akeem:
/ rebase

locksmith/__tests__/payment/paymentProcessor.test.ts
```diff
@@ -78,6 +78,25 @@ describe('PaymentProcessor', () => {
       }
     )
 
+    await UserReference.create(
+      {
+        emailAddress: Normalizer.emailAddress(
+          'connected_account_user@example.com'
+        ),
+        stripe_customer_id: 'cus_H669IyGrYp85kA',
```
julien51, line 86: Can we not use this but the `StripeCustomer` table?
Also, since these card numbers are going to be per lock, maybe we need to improave that table to store stripe_customer_id per api_key?

julien51: The reason I am asking this s that we have to assume that this credit card support will be applicable to all users (including these with a crupto wallet), not just unlock account users (for which we use the `UserReference` table)

locksmith/src/payment/paymentProcessor.ts
```diff
@@ -105,6 +105,46 @@ export class PaymentProcessor {
...
+        const charge = await this.stripe.charges.create(
+          {
+            amount: keyPriceUSD,
+            currency: 'USD',
+            customer: stripeCustomerId,
+            metadata: { lock, publicKey },
+            application_fee_amount: applicationFee,
+          },
+          {
+            stripe_account: connectedStripeId,
```
julien51, line 137: Nice!

locksmith/__tests__/payment/paymentProcessor.test.ts
akeem: Since this test setup reflects the state of the application now, lets use this to finish out the functionality for this feature with the scenario that we currently have. From there we should have a relatively straightforward path to receive customer ids and api keys from elsewhere.

I'd have to take a closer look at the StripeCustomer table to see if modifications seem appropriate but it feel like it would be out of scope of those code change, adding the initial version of this method to the code base without any downstream callers

PR #6510, opened 2020-04-14
adding the ability to set a defaultValue on metadata field

### LINKED ISSUE #6509: default value on metadata field
Creators can require some metadata to be supplied when checking out.
The team at EthGlobal asked if there was a way to supply a default value in order to skip steps for their users who might have already entered some data in their application.

julien51:
# Description

The team at ethGlobal asked for this ;)

# Issues

Fixes #6509

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/MetadataForm.tsx
+++ b/unlock-app/src/components/interface/checkout/MetadataForm.tsx
@@ -10,11 +10,25 @@ interface Props {
   onSubmit: (metadata: UserMetadata) => void
 }
 
+interface DefautltValues {
+  [key: string]: string
+}
+
 export const MetadataForm = ({ fields, onSubmit }: Props) => {
   // We can also destructure the `errors` field here and use it for
   // validation -- we'll have to consider how to handle the different
   // kinds of errors so that we can show the right message
-  const { handleSubmit, register } = useForm()
+
+  const defaultValues = {} as DefautltValues
+  fields.forEach(field => {
+    if (field.name && field.defaultValue) {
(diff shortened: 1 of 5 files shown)


julien51:
image

unlock-app/src/components/interface/checkout/MetadataForm.tsx
```diff
@@ -10,11 +10,21 @@ interface Props {
   onSubmit: (metadata: UserMetadata) => void
 }
 
+interface DefautltValues {
```
cnasc, line 13:
```suggestion
interface DefaultValues {
```

unlock-app/src/unlockTypes.ts
```diff
@@ -172,6 +172,7 @@ export interface KeyholdersByLock {
 
 export interface MetadataInput {
   name: string
+  defaultValue: string
```
cnasc, line 175: Do we want to require a defaultValue?

julien51: no, it should be optional. good catch.

julien51:
PTAL @cnasc

cnasc:
I think this looks good overall, but there are 2 test cases we should probably add to https://github.com/unlock-protocol/unlock/blob/master/unlock-app/src/__tests__/components/interface/checkout/MetadataForm.test.tsx

- Verify that the form can be submitted with an unchanged default value
- Verify that a default value can be overwritten by the user when filling out the form

(and maybe while we're in there we can fix the interface name typo)

julien51:
Added PTAL!

PR #6511, opened 2020-04-14
names in the config will override default lock name

julien51:
# Description

This was reported by an early of the new paywall :)
We now make sure to overide the name of the lock with the config.

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/usePaywallLocks.ts
+++ b/unlock-app/src/hooks/usePaywallLocks.ts
@@ -1,11 +1,12 @@
 import { useContext, useState, useEffect } from 'react'
 import { Web3Service } from '@unlock-protocol/unlock-js'
 import { Web3ServiceContext } from '../utils/withWeb3Service'
-import { RawLock } from '../unlockTypes'
+import { RawLock, PaywallConfig } from '../unlockTypes'
 
 export const usePaywallLocks = (
   lockAddresses: string[],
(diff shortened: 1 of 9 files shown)


julien51:
Hum. I think there might a normalization here... annoying :/

unlock-app/src/hooks/usePaywallLocks.ts
```diff
@@ -31,6 +32,14 @@ export const usePaywallLocks = (
 
       // getLock doesn't always return the lock address apparently
       lock.address = lockAddresses[index]
+
+      // Use the config's overwrite for the name if applicable
+      Object.keys(config.locks).forEach(configLockAddress => {
+        // The address may differ from the config to the lock received from web3Service.
+        if (configLockAddress.toLowerCase() === lock.address.toLowerCase()) {
```
julien51, line 39: I think this solves the normalization issue.

julien51: weird. It looks like it does not :/

cnasc: We have a `foreach` over the locks inside a `foreach` over the locks here. I think we can simplify this by just using the lock address already available inside each iteration of the existing loop to index into the config and pull the override name.

julien51: Good point!

PR #6512, opened 2020-04-15
not requiring a user anymore to store a card.

julien51:
# Description

We do not require the existence of a user account to store a stripe customer id.

# Checklist:

### DIFF
--- a/locksmith/src/payment/paymentProcessor.ts
+++ b/locksmith/src/payment/paymentProcessor.ts
@@ -51,31 +51,28 @@ export class PaymentProcessor {
       const user = await this.findUserByPublicKey(publicKey)
       const stripeCustomerId = await getStripeCustomerIdForAddress(publicKey)
 
-      if (user && stripeCustomerId) {
+      // If we already have a stripe customer id
+      if (stripeCustomerId) {
         await this.stripe.customers.createSource(stripeCustomerId, {
           source: token,
         })
 
         return true
       }
-      if (user && !stripeCustomerId) {
-        const customer = await this.createStripeCustomer(
-          user.emailAddress,
-          token
-        )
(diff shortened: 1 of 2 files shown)


locksmith/src/payment/paymentProcessor.ts
```diff
@@ -51,27 +51,26 @@ export class PaymentProcessor {
...
+        user ? user.emailAddress : ''
+      )
+      return !!(await saveStripeCustomerIdForAddress(publicKey, customer.id))
     } catch (e) {
       return false
     }
   }
 
-  createStripeCustomer(emailAddress: string, token: string) {
+  createStripeCustomer(token: string, emailAddress?: string) {
```
julien51, line 73: I had to switch these 2 params becuase `emailAddress` needs to be optional.

akeem: If this is the reason for taking this approach, lets take a different approach to achieving this. Lets create a type for the arguments so that the ordering doesnt matter.

Inside the body of the method you can proceed as you see fit

locksmith/src/payment/paymentProcessor.ts
```diff
@@ -51,27 +51,26 @@ export class PaymentProcessor {
...
-          user.emailAddress,
-          token
-        )
-        return !!(await saveStripeCustomerIdForAddress(publicKey, customer.id))
-      }
-      return false
+
+      const customer = await this.createStripeCustomer(
+        token,
+        user ? user.emailAddress : ''
```
akeem, line 65: In the case of no user available is it okay to pass empty string here?

julien51: yes

julien51:
This is a rather modest change, would it be ok to merge with the fix  to the param so tha Chris can work on that todat?

PR #6524, opened 2020-04-16
Using the Authenticate component

julien51:
# Description

This adds the authenticate component to most places but in the checkout component.
I had to rewrite a few tests.
I also noticed that even though we do not have a "flash" of the login form anymore, we now have a flash of the "not enabled in provider" error, which makes sense since once we have a provider it takes a few more steps to have an account. I'll work on a change to useProvider for this.

# Checklist:


unlock-app/src/components/content/KeychainContent.tsx
```diff
@@ -38,21 +38,19 @@ export const KeychainContent = ({
...
-            qrEmail={qrEmail}
-          />
-        </BrowserOnly>
-      )}
-      {!account && (
-        // Default to log in form. User can toggle to signup.
-        <LogInSignUp login />
-      )}
+      <Authenticate unlockUserAccount>
+        {account && (
```
cnasc, line 42: In this case, if the children are rendered `account` is guaranteed to be there, right?

unlock-app/src/components/content/MembersContent.tsx
```diff
@@ -48,29 +48,31 @@ export const MembersContent = ({ account, network, lockAddresses }: Props) => {
...
-              value={MemberFilters.ACTIVE}
-              current={filter}
-              setFilter={value => setFilter(value)}
-            />
-            <Filter
-              value={MemberFilters.ALL}
-              current={filter}
-              setFilter={value => setFilter(value)}
+      <Authenticate>
+        {account && (
```
cnasc, line 52: Same

unlock-app/src/components/content/SettingsContent.tsx
```diff
@@ -38,12 +38,12 @@ export const SettingsContent = ({ account }: SettingsContentProps) => {
...
       </Head>
-
-      <Errors />
-      {!account && <LogInSignUp login />}
-      {account && account.emailAddress && <AccountInfo />}
-      {account && <PaymentSettings address={account.address} />}
-      {account && account.emailAddress && <EjectAccount />}
+      <Authenticate unlockUserAccount>
+        <Errors />
+        {account && account.emailAddress && <AccountInfo />}
```
cnasc, line 43: also same for these

cnasc:
I think this looks great overall, but maybe we can remove some of those render conditions now?

unlock-app/src/components/content/KeychainContent.tsx
julien51: yeah, but React complains if this is not there because the code below uses `account.address`.

I will clean that up in a future version.

unlock-app/src/components/content/MembersContent.tsx
julien51: same it uses `account.address` below and since react does not know that this cannot be rendered if account is null, it complains.

unlock-app/src/components/content/SettingsContent.tsx
julien51: same same

PR #6542, opened 2020-04-20
useProvider doesn't reinitialize when there is already a provider

cnasc:
# Description

This PR is a step toward allowing a provider hosted on the "main window" of the paywall (for example, web3modal) to be used from within the checkout.

1. useProvider does not reinitialize if a provider has already been set
2. useProvider will take an optional param which is the provider to use if none has already been set

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useProvider.ts
+++ b/unlock-app/src/hooks/useProvider.ts
@@ -53,7 +53,7 @@ export const useProvider = (providerAdapter?: any) => {
     const ethereumWindow = (window as unknown) as EthereumWindow
 
     if (providerAdapter) {
-      // If we are using a provider in the "main window" via the paywall script
+      // If we are using an "explicit" provider (rather than one which was "injected").
       setWeb3Provider(providerAdapter)
     } else if (ethereumWindow.ethereum) {
       // If there is an injected provider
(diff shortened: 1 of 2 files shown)


unlock-app/src/hooks/useProvider.ts
```diff
@@ -52,8 +52,11 @@ export const useProvider = () => {
 
     const ethereumWindow = (window as unknown) as EthereumWindow
 
-    // If there is an injected provider
-    if (ethereumWindow.ethereum) {
+    if (providerAdapter) {
+      // If we are using a provider in the "main window" via the paywall script
```
julien51, line 56:
```suggestion
      // If we are using an "explicit" provider (rather than one which was "injected").
```

unlock-app/src/hooks/useProvider.ts
```diff
@@ -28,7 +28,7 @@ interface Config {
   httpProvider: string
 }
 
-export const useProvider = () => {
+export const useProvider = (providerAdapter?: any) => {
```
julien51, line 31: nit: can we stick to `provider` here?

I think this functionality will be used for more than just the paywall-passed provider. For example I think this is who we will also suppport web3Modal in unlock-app(outside of the pywall context)

PR #6546, opened 2020-04-20
Messy first pass at delegating provider from main window to checkout

cnasc:
# Description

This doesn't work yet

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/content/CheckoutContent.tsx
+++ b/unlock-app/src/components/content/CheckoutContent.tsx
@@ -7,7 +7,11 @@ import queryString from 'query-string'
 import { UnlockError } from '../../utils/Error'
 import BrowserOnly from '../helpers/BrowserOnly'
 import { pageTitle } from '../../constants'
-import { useCheckoutCommunication } from '../../hooks/useCheckoutCommunication'
+import {
+  useCheckoutCommunication,
+  TransactionInfo,
+  UserInfo,
+} from '../../hooks/useCheckoutCommunication'
 import { checkoutMachine, CheckoutState } from '../../stateMachines/checkout'
 import CheckoutWrapper from '../interface/checkout/CheckoutWrapper'
 import CheckoutContainer from '../interface/checkout/CheckoutContainer'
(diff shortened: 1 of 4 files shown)


unlock-app/src/components/content/CheckoutContent.tsx
```diff
@@ -40,14 +44,23 @@ interface CheckoutContentProps {
   errors: UnlockError[]
 }
 
-export const CheckoutContent = ({
+export const CheckoutContent = (props: CheckoutContentProps) => {
+  return (
+    <BrowserOnly>
```
cnasc, line 49: Needed to make all of this very ugly for the time being so that the server side rendering wouldn't break anything.

unlock-app/src/hooks/useCheckoutCommunication.ts
```diff
@@ -19,39 +20,61 @@ export enum CheckoutEvents {
...
     }
   }
 
+  const [config, setConfig] = useState<PaywallConfig | undefined>(undefined)
+  const [providerAdapter, setProviderAdapter] = useState<
+    ProviderAdapter | undefined
+  >(undefined)
+  parent = usePostmateParent({
+    setConfig: (config: PaywallConfig) => {
+      setProviderAdapter(new ProviderAdapter(pushOrEmit))
```
cnasc, line 66: when we do this for real, it will only be if the config specifies we're using a provider on the other end

unlock-app/src/components/content/CheckoutContent.tsx
julien51: Yeah I think there is no server side rendering which would make such sense here for now anyway.

cnasc:
To make this work there's a lot of synchronization that has to occur around things that may or may not ever be present.

The most likely path forward is a refactor of the checkout communication hook, since now it needs to support a more robust kind of bidirectional communication than it previously did.

In order to minimize the mental load here I'm going to try and pull things out that can stand independently of the main job.

PR #6548, opened 2020-04-20
adding grantKey to walletService

### LINKED ISSUE #6518: grantKey support inside of unlock-js
UnlockJs needs to support grantKey in order to imporve our credit card flow.

julien51:
# Description

This adds the ability to grant key.
The v0 smart contracts do not support this...

I will now write some unit tests to complete this.
And maybe I'll split it in small PRs.

# Issues

Fixes #6518

# Checklist:

### DIFF
--- a/unlock-js/src/__tests__/integration/walletServiceIntegration.test.js
+++ b/unlock-js/src/__tests__/integration/walletServiceIntegration.test.js
@@ -219,6 +219,64 @@ describe('Wallet Service Integration', () => {
           })
         })
 
+        if (['v0'].indexOf(versionName) === -1) {
+          describe('grantKey', () => {
+            let tokenId
+            let key
+            let keyGrantee
+            let transactionHash
+            beforeAll(async () => {
+              keyGrantee = accounts[7]
+
+              tokenId = await walletService.grantKey(
+                {
+                  lockAddress,
+                  recipient: keyGrantee,
+                },
+                (error, hash) => {
+                  transactionHash = hash
+                }
+              )
+              key = await web3Service.getKeyByLockForOwner(
+                lockAddress,
+                keyGrantee
+              )
+            })
+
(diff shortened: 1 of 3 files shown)


julien51:
I did this TDD using the integration tests to build things.

unlock-js/src/__tests__/integration/walletServiceIntegration.test.js
```diff
@@ -219,6 +219,64 @@ describe('Wallet Service Integration', () => {
           })
         })
 
+        if (['v0'].indexOf(versionName) === -1) {
```
julien51, line 222: v0 does not have support for `grantKeys`

julien51:
Ok, this will need a serious rebase once #6551 #6552 #6553  #6554 #6555 #6556 #6557 #6558 !

unlock-js/src/walletService.js
```diff
@@ -279,6 +283,16 @@ export default class WalletService extends UnlockService {
     return version.purchaseKey.bind(this)(params, callback)
   }
 
+  /**
+   * Grants a key to an address
+   * @param {function} callback : callback invoked with the transaction hash
+   */
+  async grantKey(params = {}, callback) {
```
akeem, line 290: It might be useful to note what params might consist of in the method documentation. It looks lockAddress is all that is required up top, but since its is delegating the call else where callers wont know unless they review the code

julien51:
There are actually subtle differences ad we have 4 different implementations :/

unlock-js/src/walletService.js
julien51: That's a good point. We need to be better at this. versions actually require/enable different params. for example in v7, one can pass the lockManager, but other versions do not have that concept.

I think there will need to have a broard effort around docs.

PR #6562, opened 2020-04-21
CheckoutContent cleanup

cnasc:
# Description

This PR cleans up CheckoutContent a bit, mostly by factoring out the inner component to a separate file. There are no real functional changes at this point, but this will make it easier when we have to juggle the provider and the checkout communication as part of the delegated provider work.

# Checklist:


unlock-app/src/hooks/useProvider.ts
```diff
@@ -55,6 +55,7 @@ export const useProvider = (providerAdapter?: any) => {
     if (providerAdapter) {
       // If we are using an "explicit" provider (rather than one which was "injected").
       setWeb3Provider(providerAdapter)
+      dispatch(providerReady())
```
julien51, line 58: 🚀

PR #6565, opened 2020-04-21
Add cleaner remote method call implementation

cnasc:
# Description

This is a slightly cleaner swing at an implementation for the checkout to make remote method calls against a provider in the host window.

Basically, when the provider receives a method call, it:

- Stores the callback for the method call in a table keyed by request id
- Sends a message to the parent iframe containing the method call

When the parent iframe has the result of the method call, it passes it back down (not implemented on paywall script side yet):
- Calls `resolveMethodCall` on the child iframe with the result, which:
  - pulls the relevant callback from the table
  - deletes the callback from the table
  - applies the callback to the result

This needs testing, and to do it properly we'll need to implement both ends of this. There is some remaining mystery about the method call format, because they don't always come in the format implied by the type signatures from `ethers`.

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/hooks/useCheckoutCommunication.test.ts
+++ b/unlock-app/src/__tests__/hooks/useCheckoutCommunication.test.ts
@@ -3,6 +3,8 @@ import { renderHook, act } from '@testing-library/react-hooks'
 import {
   useCheckoutCommunication,
   CheckoutEvents,
+  waitingMethodCalls,
+  resolveMethodCall,
 } from '../../hooks/useCheckoutCommunication'
 
 let emit = jest.fn()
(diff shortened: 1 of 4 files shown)


cnasc:
Trying to keep the scope of these changes minimal, so while my manual testing will be end-to-end, the final PR will not actually integrate these capabilities into the app. Integration will follow in a later PR.

paywall/src/components/content/NewDemoContent.tsx
```diff
@@ -85,6 +85,7 @@ export default function NewDemoContent() {
         'You have reached your limit of free articles. Please purchase access to continue reading',
       noWallet: 'Please use a web3 enabled browser!',
     },
+    useDelegatedProvider: true,
```
cnasc, line 88: this should be removed before merging

cnasc:
This now seems to work. It needs a little more cleanup and validation, but it's ready for review

paywall/src/components/content/NewDemoContent.tsx
julien51: Pleaase remove ;p

paywall/src/paywall-script/index.ts
```diff
@@ -163,6 +181,16 @@ export class Paywall {
     this.checkKeysAndLock()
   }
 
+  handleMethodCallEvent = async ({ method, params, id }: MethodCall) => {
+    // TODO: handle getting provider
```
julien51, line 185: Do you want to do that in a later PR?
Let's have a discussion about this tomorrow but I think in practice we will need to use the injected provider by default anyway :/

cnasc:
This is now rebased down to the minimum set of changes, tests to come and then it will be ready to go.

cnasc:
Unit tests added, verified manually. @julien51 PTAL if you have any additional thoughts before we merge

PR #6566, opened 2020-04-21
Exposing the  baseTokenURI  storage field on the Lock contract

nfurfaro:
# Description
This is a refactoring of our existing `tokenURI` function in `PublicLock.sol`, motivated by the need to be able to get the custom `baseTokenURI` even when no keys have been sold yet, which was previously impossible to do. the refactored version of the function handles all possible combinations of 2 primary conditions:
1.) Is the `tokenId` 0 or not?
  - passing 0 will now return the baseTokenID (custom or global)

2.) Has a custom `baseTokenURI` been set?
  - if NO, the `globalBaseTokenURI` + the lock address is returned.
  - if YES, the `baseTokenURI` is returned, without the lock address (the address is redundant as a differentiator when a custom URI has been set).

This also moves us in the right direction regarding issue 5836 (Do not revert read-only calls)

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinLockMetadata.sol
+++ b/smart-contracts/contracts/mixins/MixinLockMetadata.sol
@@ -91,7 +91,11 @@ contract MixinLockMetadata is
    view
    returns(string memory)
   {
-    return baseTokenURI;
+    if(bytes(baseTokenURI).length == 0) {
+      return unlockProtocol.globalBaseTokenURI();
+    } else {
+      return baseTokenURI;
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


nfurfaro:
@NickCuso Did you have some ideas around how we could add support for a public `baseTokenURI` getter function for v7? If not, I'll continue with adding it for v8.

smart-contracts/contracts/mixins/MixinLockMetadata.sol
```diff
@@ -31,7 +31,7 @@ contract MixinLockMetadata is
   string private lockSymbol;
 
   // the base Token URI for this Lock. If not set by lock owner, the global URI stored in Unlock is used.
-  string private baseTokenURI;
+  string public baseTokenURI;
```
nfurfaro, line 34: @NickCuso Although this is the simplest way to expose this data publicly, Julien would like the function to fall back to getting the `globalBaseTokenURI` from Unlock if this is not set. I'll make the changes to implement that for v8, unless you have an idea of how we can do this with extensions?

NickCuso: Hmm we could create an extension that reads the tokenUri and removes the address and token to extract the base uri. But it's probably better to build this for v8 as that's a little hacky.

On a related note, it seems that if the lock has a base uri the token uri shouldn't also add the lock's address. That may be redundant for some use cases and if it is needed, the lock's address could be included when setting the uri

nfurfaro: Hmm. maybe so. But in order to guarantee that each key has a globally unique tokenURI (which I think we should ensure), the lock addresses seems to be the simplest way to do this. 
I guess it would be fine to omit the lock's address when a custom URI has been set, as long as each lock has a unique custom URI. Most of the time this would probably be the case.

smart-contracts/contracts/mixins/MixinLockMetadata.sol
```diff
@@ -86,6 +86,14 @@ contract MixinLockMetadata is
     }
   }
 
+  function getBaseTokenURI()
+   external
+   view
+   returns(string memory)
+  {
+    return baseTokenURI;
```
NickCuso, line 94: hmm why did you switch from a public variable to this function?

nfurfaro:
Dang. 🤦‍♂️I forgot to add the logic

Julien wants this to return the globalBaseTokenURI if the lock-level URI
has not been set. But I didn't implement it.

PR #6581, opened 2020-04-23
Add util to enable provider in paywall

cnasc:
# Description

To help with using the delegated provider, this PR introduces some utils around getting and enabling injected providers.

# Checklist:

### DIFF
--- a/paywall/src/utils/enableInjectedProvider.ts
+++ b/paywall/src/utils/enableInjectedProvider.ts
@@ -1,0 +1,40 @@
+type Enabler = {
+  enable: () => Promise<void>
+}
+
+interface Web3Window {
+  // Present with more recent injected providers
+  ethereum?: Enabler
+  // Present with legacy injected providers
+  web3?: {
+    currentProvider: Enabler
+  }
+}
+
+export const getProvider = (window: Web3Window): Enabler | undefined => {
+  // Using a "modern" dapp browser
+  if (window.ethereum) {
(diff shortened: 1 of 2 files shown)


paywall/src/utils/enableInjectedProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+  // Browser is not web3 capable
+  return undefined
+}
+
+export const enableInjectedProvider = async (provider: Enabler | undefined) => {
+  // This should not occur; if there is no provider, the paywall
+  // script should prevent the checkout iframe from trying to connect
+  // to one.
+  if (!provider) {
+    throw new Error('Fatal: no web3 provider found.')
```
julien51, line 34: Should that throw? I worry this will break the flow for people with no web3 provider even though they _could_ use their unlock account on the checkout side?

julien51: I guess the checkout side will not call enable...

cnasc: This code path should only execute when there is an injected provider, that call is mainly there as a final sanity check.

Basically, the paywall script will check whether or not there is a provider at construction time. Then when launching the iframe, it will include that information so the iframe knows whether or not it has access to a delegated provider. If it does, it will propagate method calls upward. If it doesn't, it will behave like it does now when there is no available provider and allow login.

PR #6582, opened 2020-04-23
Paywall script gets and stores injected provider on construction

cnasc:
# Description

Starting to integrate provider handling in the paywall script.

# Checklist:


paywall/src/paywall-script/index.ts
```diff
@@ -47,6 +52,8 @@ export class Paywall {
 
   lockStatus?: string
 
+  provider?: Enabler
```
cnasc, line 55: We have to decide how to signal the presence or absence of this provider to the checkout. We can intercept the setconfig call and include the presence of the provider as a value in the config that would be for internal use only which would limit the number of things we have to wait for

julien51: yes, I think we can add it to the config.

PR #6588, opened 2020-04-23
Updating the new constraint to be a unique constraint rather than a primary key constraint

akeem:
# Description

We've noticed the current migration failing in place due to one of the columns in the compound key containing null values. This adjustment should achieve the current intended affect of stopping duplicate rows from being created and creating some time to determine if there should be place holders. It may make sense to reference another table for the connected_account_id and have some place holder row for the existing values currently null today.

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Thanks Akeem. Sorry for the delay! I had to go pick groceries!

PR #6618, opened 2020-04-29
[draft] Adding Paywall object to module

cnasc:
# Description

This PR starts to expose the paywall object from the paywall module. We need to determine how this gets built (ie with which config variables)

# Checklist:

### DIFF
--- a/paywall/.eslintignore
+++ b/paywall/.eslintignore
@@ -6,4 +6,5 @@
 /src/out
 /.cache
 /src/paywall-script/dist/
+/src/paywall-script/module.d.ts
 *.min.js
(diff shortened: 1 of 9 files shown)


julien51:
The configuration variable is a great question. Now that I think about it I wonder if these should be configurable? Maybe adding anothe argument to the `Paywall` constructor?

cnasc:
@julien51 I think we're in pretty good shape here now. Before bumping the version and publishing we should do a smoke test just to make sure everything works as expected.

Once this is in place, we can extend to allow a provider to be passed in the constructor.

cnasc:
Going to have to figure out what's up with eslint here

cnasc:
`eslint-disable` appears not to be sufficient to resolve this

cnasc:
This is confirmed working on local dev, ready for review

PR #6622, opened 2020-04-29
Paywall object takes an optional provider at construction

### LINKED ISSUE #6397: Support web3 provider passed from the paywall application
One we are done refactoring outr paywall, it will be interesting and useful to see if the paywall application can itself "yield" a provider to the unlock checkout application.

This will open the door to letting creators support a provider of their choice.

cnasc:
# Description

This small PR allows the Paywall object to take an optional parameter indicating a provider to use (such as from Web3Modal). In the absence of that provider, the paywall will use the injected provider from the window (if available).

This should be everything, but tomorrow I'm going to take some time to test thoroughly on a local environment and use that as an example of how to use this new feature with Web3Modal in a blog post.

# Issues

Fixes #6397 

# Checklist:

### DIFF
--- a/paywall/src/paywall-script/Paywall.ts
+++ b/paywall/src/paywall-script/Paywall.ts
@@ -77,9 +77,10 @@ export class Paywall {
 
   moduleConfig: ModuleConfig
 
-  constructor(paywallConfig: any, moduleConfig: ModuleConfig) {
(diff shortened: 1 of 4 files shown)


cnasc:
@julien51, this PR has been updated with a change so that the `unlockProtocol` window variable is only set when we load from the published script, not when the paywall is imported as a module. PTAL, and we should test that everything works as expected in both scenarios

PR #6747, opened 2020-06-18
Migrate to the UniswapV2 oracle

### LINKED ISSUE #6314: Upgrade to Uniswap v2
https://twitter.com/UniswapExchange/status/1242183677590810626

We are most interested in the new price oracle feature.
Need to confirm they gain liquidity before we make the swap.

NickCuso:
# Description

# Issues

Fixes #6314
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/Unlock.sol
+++ b/smart-contracts/contracts/Unlock.sol
@@ -29,11 +29,10 @@ pragma solidity 0.5.17;
 import '@openzeppelin/contracts-ethereum-package/contracts/ownership/Ownable.sol';
 import '@openzeppelin/upgrades/contracts/Initializable.sol';
 import 'hardlydifficult-ethereum-contracts/contracts/proxies/Clone2Factory.sol';
-import 'hardlydifficult-ethereum-contracts/contracts/interfaces/IUniswapExchange.sol';
 import './interfaces/IPublicLock.sol';
 import './interfaces/IUnlock.sol';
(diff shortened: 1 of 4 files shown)


smart-contracts/contracts/Unlock.sol
```diff
@@ -203,14 +217,11 @@ contract Unlock is
...
       address tokenAddress = IPublicLock(msg.sender).tokenAddress();
-      if(tokenAddress != address(0)) {
-        // If priced in an ERC-20 token, find the supported uniswap exchange
-        IUniswapExchange exchange = uniswapExchanges[tokenAddress];
-        if(address(exchange) != address(0)) {
-          valueInETH = exchange.getTokenToEthInputPrice(_value);
-        } else {
-          // If the token type is not supported, assume 0 value
-          valueInETH = 0;
+      if(tokenAddress != address(0) && tokenAddress != weth) {
```
julien51, line 220: Does that mean we will nto account for locks in Weth? I think we should also use them?

NickCuso: UniswapV2 doesn't support ETH under the hood. So we are converting to weth, and if the tokenAddress is ETH or WETH then we skip Uniswap and use the value directly in the else block below.

julien51: I think in that case we should just set `valueInETH = _value`

NickCuso: Correct, that's what we do in the else block below

PR #6785, opened 2020-07-07
UDT: Accept minter as a param on initialization

NickCuso:
# Description

Modifies the initialize method to accept the minter as a param instead of using msg.sender.
The trouble with msg.sender is that will be the ProxyAdmin contract address and not the user as expected.

Once merged we should deploy a new UDT proxy. I deleted the ZOS entries so there will be no conflicts.

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Do you want to go ahead and deploy on rinkeby and then transfer ownership to the multisig?

NickCuso:
Yes @julien51 thanks. You could do mainnet as well

Issue #4282, opened 2019-07-24
Mint Unlock Discount Tokens


NickCuso:
We have an upgradable ERC-20 implementation for the Unlock Discount Token.  Once we have #3711 implemented we can begin minting UDT with each Key purchase.

Just need to lock down the formula we want to start with.

After this is complete we can talk about applying a discount based on the user's holdings.

julien51:
It's now time to work on this!
[This wiki page](https://github.com/unlock-protocol/unlock/wiki/The-Unlock-Discount-Tokens) provides the description of the Unlock Discount Token.

The minting process (creation of new tokens) happens through a single mechanism: referrals.
Every time a new key is purchased if there is a referrer the Unlock smart contract's `recordKeyPurchase` method will be invoked by the lock.

This method already implements the logic which consists of increasing the `grossNetworkProduct`. We now need to add the logic of granting tokens to the `referrer`.

There are 2 rules which help determine the number of tokens minted for the referrer.

1. Gas limit
In order to avoid Sybil attacks (lock owners referring themselves), the value of tokens earned is AT MOST equal to the value of gas spent.
In other words, if someone spent 0.001 Eth worth of gas, they should earn an amount of UDT which is at most 0.001 Eth.

The value of UDT should be retrieved from a Uniswap pool. Unlock Inc will provide liquidity for such a pool, with a very low price, assuming that the tokens actually have little value until the "discount claiming" mechanism is implemented.

2. Token inflation
Additionally, in order to guarantee "scarcity" of the Unlock Discount Tokens, the number of tokens minted/granted is such that the supply of tokens increases AT MOST by half of the increase in GDP resulting from the purchase. In other words, if the `grossNetworkProduct` increases by 1% when a referral is made, the number of tokens minted should grow the supply by at most 0.5%.

julien51:
Here are things we need to clarify/set before we set this in montion:

[ ] What is the address of the ERC20 UDT contract?
[ ] We should pre-mine at least 1 token (in order to perform the calculation when it comes to supply.
[ ] We should set a Uniswap pool for UDT/Eth.

julien51:
Here are the tasks which I am identifying for this to be fully implemented:

- [ ] Retrieve the UDT price in Eth from the Uniswap pool
- [ ] Compute Eth spent for gas when purchasing a key (gas amount could be hardcoded while gas price needs to be actual value)
- [ ] Determine max amount of tokens to be granted based on gas spent
[...]

PR #6786, opened 2020-07-07
Mint UDT

NickCuso:
# Description

Mint UDT as per the [wiki](https://github.com/unlock-protocol/unlock/wiki/The-Unlock-Discount-Tokens)

After upgrading we will need to:
 - Mint UDT to fund the Uniswap pool
 - Add Unlock.sol as a UDT minter (and maybe remove the original account)
 - Approve the uniswapRouter & create a UDT - WETH UniswapV2 pool
 - Make a trade on Uniswap so that the oracle can determine the exchange rate
 - Call `configUnlock` on Unlock to set the UDT address and the estimated gas for a purchase
 - Call `setOracle` on Unlock to add support for UDT

# Issues

Fixes #
Refs #4282

# Checklist:

### DIFF
--- a/docker/docker-compose.ci.yml
+++ b/docker/docker-compose.ci.yml
@@ -5,7 +5,7 @@ services:
     image: trufflesuite/ganache-cli:v6.9.1
     expose:
       - 8545
-    command: ['-h=0.0.0.0', '-m=hello unlock save the web', '-i=1984']
+    command: ['-h=0.0.0.0', '-m=hello unlock save the web', '-i=1984', '-e 1000000']
 
   ganache-standup: # Sets up the ganache instance with the required smart contracts and balances
     image: ganache-standup
--- a/smart-contracts/contracts/Unlock.sol
+++ b/smart-contracts/contracts/Unlock.sol
@@ -241,13 +241,16 @@ contract Unlock is
         {
           // Get the value of 1 UDT (w/ 18 decimals) in ETH
           uint udtPrice = udtOracle.updateAndConsult(udt, 10 ** 18, weth);
-          uint gasCost = estimatedGasForPurchase * tx.gasprice;
-          uint tokensToMint = gasCost / udtPrice;
+
+          // tokensToMint is either == to the gas cost times 1.25 to cover the 20% dev cut
+          uint tokensToMint = (estimatedGasForPurchase * tx.gasprice).mul(125 * 10 ** 18) / 100 / udtPrice;
+          // or tokensToMint is capped by percent growth
           uint maxTokens = IMintableERC20(udt).totalSupply().mul(valueInETH) / 2 / grossNetworkProduct;
           if(tokensToMint > maxTokens)
           {
             tokensToMint = maxTokens;
           }
+
           if(tokensToMint > 0)
           {
             // 80% goes to the referrer, 20% to the Unlock dev - round in favor of the referrer
--- a/smart-contracts/contracts/Unlock.sol
+++ b/smart-contracts/contracts/Unlock.sol
@@ -33,6 +33,8 @@ import './interfaces/IPublicLock.sol';
 import './interfaces/IUnlock.sol';
 import '@openzeppelin/contracts-ethereum-package/contracts/utils/Address.sol';
 import 'hardlydifficult-eth/contracts/protocols/Uniswap/IUniswapOracle.sol';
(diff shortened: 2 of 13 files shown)


NickCuso:
@julien51 before I work on the tests - could you spot check this to confirm I got the core concepts correct?

smart-contracts/contracts/Unlock.sol
```diff
@@ -119,15 +130,19 @@ contract Unlock is
...
-  function initializeWeth(
-    address _weth
+  function configure(
+    address _udt,
+    address _weth,
+    uint _estimatedGasForPurchase
   ) public
     onlyOwner
   {
+    udt = IMintableERC20(_udt);
```
julien51, line 143: Shouldn't this be ?
```
     udt = _udt;
```

smart-contracts/contracts/Unlock.sol
```diff
@@ -229,8 +244,39 @@ contract Unlock is
...
+      if(_referrer != address(0))
+      {
+        IUniswapOracle udtOracle = uniswapOracles[address(udt)];
+        if(address(udtOracle) != address(0))
+        {
+          // Get the value of 1 UDT (w/ 18 decimals) in ETH
+          uint udtPrice = udtOracle.updateAndConsult(address(udt), 10 ** 18, weth);
+          uint gasCost = estimatedGasForPurchase * tx.gasprice;
+          uint tokensToMint = gasCost / udtPrice;
+          // If the delta is < 0.01% then no tokens will be minted
```
julien51, line 261: Could you explain this please? 
We should mint (even a very small amount!) for any referral (even when the network is very very large)

smart-contracts/contracts/Unlock.sol
```diff
@@ -229,8 +244,39 @@ contract Unlock is
...
+          uint percentGNPChangeInBP = BASIS_POINTS_DEN.mul(valueInETH) / grossNetworkProduct;
+          uint maxTokens = percentGNPChangeInBP.mul(udt.totalSupply()) / BASIS_POINTS_DEN;
```
julien51, line 263: We should divide `percentGNPChangeInBP` by 2, because the supply can only increase at half the speed of the GNP.

smart-contracts/contracts/Unlock.sol
NickCuso: It's like this because I declared `udt` as that type instead of an address. It's really just a style question, the casting doesn't impact gas.

smart-contracts/contracts/Unlock.sol
NickCuso: To resolve this and the following point, here's a new approach. Does this look right?
```
uint maxTokens = udt.totalSupply().mul(valueInETH) / 2 / grossNetworkProduct
```

julien51: yes, I think that works!

NickCuso:
Ready for review

docker/docker-compose.ci.yml
```diff
@@ -5,7 +5,7 @@ services:
     image: trufflesuite/ganache-cli:v6.9.1
     expose:
       - 8545
-    command: ['-h=0.0.0.0', '-m=hello unlock save the web', '-i=1984']
+    command: ['-h=0.0.0.0', '-m=hello unlock save the web', '-i=1984', '-e 1000000']
```
julien51, line 8: What is this change for?

smart-contracts/test/UnlockDiscountToken/mintingTokens.js
```diff
@@ -0,0 +1,207 @@
...
+        value: await lock.keyPrice(),
+      })
+      const transaction = await web3.eth.getTransaction(tx.tx)
+      // using estimatedGas instead of the actual gas used so this test does not regress as other features are implemented
+      gasSpent = new BigNumber(transaction.gasPrice).times(estimateGas)
+    })
+
+    it('referrer has some UDT now', async () => {
+      const actual = await udt.balanceOf(referrer)
+      assert.notEqual(actual.toString(), 0)
```
julien51, line 140: 👏

docker/docker-compose.ci.yml
NickCuso: The test funds 40 ETH into the uniswap pool. The default account balance is only 100. This increases the default account balance so we don't run out of ETH.

PR #7087, opened 2021-05-05
UDT on side chains

julien51:
# Description

UDT on mainnet is the only place where the supply of UDT increases (using a log curve): new UDT are minted for every new key purchases, but fewer and fewer are minted.

On side chains, we want to be able to distribute UDT as well, but we cannot do this in a way which would be inflationary. The approach is then to transfer some UDT from mainnet (minted ad-hoc through governance) to the side chain (we expect all EVM compatible chains to have a way to transfer ERC20 from mainnet, called bridges.

Once these UDT are on the side chains, they are assigned to the Unlock contract on this side chain. Then, the contract, instead of mining them, just distribute them using an asymptotic function. [See more info](https://docs.unlock-protocol.com/governance/the-unlock-token/side-chains-and-layer-2).

For a GDP growth of 𝝙, the reward ratio `𝝙/(2 * (1+𝝙))`. This number is between 0 and 0.5 for any 𝝙 between 0 and Infinity. This means that the "reserve" is never fully distributed even though the number of tokens distributed gets smaller and smaller.

# Checklist:


smart-contracts/contracts/Unlock.sol
```diff
@@ -242,23 +242,47 @@ contract Unlock is
...
-          // tokensToMint is either == to the gas cost times 1.25 to cover the 20% dev cut
-          uint tokensToMint = (estimatedGasForPurchase * tx.gasprice).mul(125 * 10 ** 18) / 100 / udtPrice;
-          // or tokensToMint is capped by percent growth
-          uint maxTokens = IMintableERC20(udt).totalSupply().mul(valueInETH) / 2 / grossNetworkProduct;
-          if(tokensToMint > maxTokens)
+          // tokensToDistribute is either == to the gas cost times 1.25 to cover the 20% dev cut
+          uint tokensToDistribute = (estimatedGasForPurchase * tx.gasprice).mul(125 * 10 ** 18) / 100 / udtPrice;
+
+          // or tokensToDistribute is capped by network GDP growth
+          uint gdpGrowth = valueInETH / grossNetworkProduct;
```
NickCuso, line 249: You need to rework this approach so it only depends on whole numbers. If you don't need much precision, maybe basis points would be sufficient. e.g. 10000 * valueInETH / GNP. at the end you can then divide the 10000 back out.

Also impacts some of below.

NickCuso (commented):
For testing, I think setting the network id impacts chainId

```
    network_id: 666,
```

this is an option under config -> network -> development.

Then to test both changes, maybe you can have another network like mainnetDev with a different chainId. Then tests could branch on the current chainId.

Not 100% but something like that might work.

smart-contracts/contracts/Unlock.sol
julien51: Ha! That explains why you had no "intermediate" number calculation in the existing code. I think precision is important because we're quickly going to talk about very small numbers (like a 1 DAI transaction on a 1Bn GDP is veery small but should still earn its "fair" share of UDT).

julien51:
I will look at ways to do that. Another approach I was thinking about was to have a "flag" on `Unlock.sol` which sets `isMainnet = true` but that would require a constructor/config change which means probably a new v9... and i want to avoid it if we can to reduce complexity.

Wdyt?

smart-contracts/contracts/Unlock.sol
NickCuso: Best way if you can rework the formulas so that all division is done once, at the very end -- e.g. `(a * b + c) / (x * y)`

Wolfgram is helpful for finding ways to reorder terms.

NickCuso:
yea a flag would work as well, but it's unfortunate to have a flag intended for testing in the contract.

hardhat does have a way of striping out code when doing a non-dev build. but that's a big migration.
there are some solidity macro options out there you might consider.

julien51:
> yea a flag would work as well, but it's unfortunate to have a flag intended for testing in the contract.

Oh I would want to do the flag for mainnet vs side chains (xdai, polygon, bsc, optimism... etc).
It looks like ganache does not support `block.chainid` so the flag seems to be the best option at the moment :/

nfurfaro:
Hey, re: adding a flag to `Unlock.sol`:
Not exactly the same scenario, but I've recently been adding a `chainIndex` parameter to some token contracts. The idea is that the chain of minting gets embedded in the `tokenId` to prevent the possibility of collisions if tokens are mintable on more than 1 chain.

smart-contracts/contracts/Unlock.sol
nfurfaro: Another possibility here is to look at how floating-point numbers are handled in the Locked.fyi hook.

julien51:
Thanks @nfurfaro This is maybe something we should do?

nfurfaro:
@julien51 the `chainIndex` idea is really only useful if we anticipate minting of tokens on more than 1 chain (ie: eth-mainnet && xDai/Polygon). Otherwise it won't help any.

julien51:
> @julien51 the chainIndex idea is really only useful if we anticipate minting of tokens on more than 1 chain (ie: eth-mainnet && xDai/Polygon). Otherwise it won't help any.

So I was actually wondering if it would be interesting for someone to create the same lock on multiple chains, ideally with the same address?

julien51:
Hum. Getting an error that I did not expect.

```
Error: Contract Unlock is abstract and cannot be deployed.
    at LocalController.checkCanAdd (/home/unlock/smart-contracts/node_modules/@openzeppelin/cli/src/models/local/LocalController.ts:103:13)
    at contractsData.forEach (/home/unlock/smart-contracts/node_modules/@openzeppelin/cli/src/scripts/add.ts:12:16)
    at Array.forEach (<anonymous>)
    at add (/home/unlock/smart-contracts/node_modules/@openzeppelin/cli/src/scripts/add.ts:11:17)
    at zosDeploy (/home/unlock/smart-contracts/migrations/2_deploy_unlock.js:10:3)
    at deployUnlock (/home/unlock/smart-contracts/migrations/2_deploy_unlock.js:24:9)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

julien51:
Strangely enough this error is not happening locally.

julien51:
Found the issue! i needed to update the interface for the function which sets the config :) 
I also changed the version to 9! Next: see how tests work and write some for 9!

smart-contracts/contracts/Unlock.sol
```diff
@@ -289,7 +307,7 @@ contract Unlock is
   ) external pure
     returns (uint16)
   {
-    return 8;
+    return 9;
```
julien51, line 310: Upgraded version!

smart-contracts/package.json
```diff
@@ -15,6 +15,15 @@
     "@openzeppelin/contracts-ethereum-package": "2.5.0",
     "@openzeppelin/test-helpers": "0.5.6",
     "@openzeppelin/upgrades": "2.8.0",
+    "@unlock-protocol/unlock-abi-0": "1.0.4",
```
julien51, line 18: All of these are the same as before but renamed.

smart-contracts/test/Lock/behaviors/lockBehaviors.js
```diff
@@ -35,9 +35,13 @@ contract('Lock / lockBehaviors', (accounts) => {
 
         // approve spending (ignored if the test pass does not use ERC-20)
         for (let i = 0; i < accounts.length; i++) {
-          await this.testToken.approve(this.lock.address, '-1', {
-            from: accounts[i],
-          })
+          await this.testToken.approve(
+            this.lock.address,
+            await this.lock.keyPrice.call(),
```
julien51, line 40: using `keyPrice` instead of `-1`

smart-contracts/test/Lock/purchaseTip.js
```diff
@@ -33,7 +33,7 @@ contract('Lock / purchaseTip', (accounts) => {
         lock = locks.FIRST
 
         // Approve spending
-        await testToken.approve(lock.address, -1, {
+        await testToken.approve(lock.address, tip, {
```
julien51, line 36: only approving the right amount

smart-contracts/contracts/Unlock.sol
```diff
@@ -242,22 +246,42 @@ contract Unlock is
...
-          uint tokensToMint = (estimatedGasForPurchase * tx.gasprice).mul(125 * 10 ** 18) / 100 / udtPrice;
-          // or tokensToMint is capped by percent growth
-          uint maxTokens = IMintableERC20(udt).totalSupply().mul(valueInETH) / 2 / grossNetworkProduct;
-          if(tokensToMint > maxTokens)
+          // tokensToDistribute is either == to the gas cost times 1.25 to cover the 20% dev cut
+          uint tokensToDistribute = (estimatedGasForPurchase * tx.gasprice).mul(125 * 10 ** 18) / 100 / udtPrice;
+
+          // or tokensToDistribute is capped by network GDP growth
+          uint maxTokens = 0;
+          if (chainId > 1)
```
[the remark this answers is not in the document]
julien51: Exactly.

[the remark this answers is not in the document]
julien51: I was thinking of using a flag like `isMainNet` but thought it would be more flexible to use the chainId as we may in the future want to have different behaviors based on the chain id (not jist mainnet vs non mainnet)

NickCuso (approved):

smart-contracts/test/UnlockDiscountToken/transferingTokens.js
```diff
@@ -0,0 +1,224 @@
...
+  })
+
+  it('owner starts with 0 UDT', async () => {
+    assert.equal(
+      new BigNumber(await udt.balanceOf(await unlock.owner())).toFixed(),
+      '0'
+    )
+  })
+
+  it('unlock has some 0 UDT', async () => {
```
nfurfaro, line 128: the title of the test is confusing. is it some or 0 ?

nfurfaro (approved):
One minor comment, but looks good!

PR #7411, opened 2021-08-06
V9 fixes

julien51:
# Description

This PR includes 2 fixes for the current contract which are not changing the ABI and hence won't impact the front-ends.

1. First, we now return the lock address as part of the create lock function. This enables other contracts to call the factory and keep track of the address of the contract they deployed.
2. We remove requirements on the referrer address. We already have a protection against self-mining in the form the gas spent cap. 

# Checklist:

### DIFF
--- a/smart-contracts/contracts/Unlock.sol
+++ b/smart-contracts/contracts/Unlock.sol
@@ -149,7 +149,7 @@ contract Unlock is
     uint _maxNumberOfKeys,
     string memory _lockName,
     bytes12 _salt
-  ) public
+  ) public returns(address)
   {
     require(publicLockAddress != address(0), 'MISSING_LOCK_TEMPLATE');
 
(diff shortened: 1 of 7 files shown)


smart-contracts/contracts/Unlock.sol
```diff
@@ -182,6 +182,7 @@ contract Unlock is
 
     // trigger event
     emit NewLock(msg.sender, newLock);
+    return newLock;
```
nfurfaro, line 185: We could perhaps add a return parameter to this function, ie:
```
function createLock(
    uint _expirationDuration,
    address _tokenAddress,
    uint _keyPrice,
    uint _maxNumberOfKeys,
    string memory _lockName,
    bytes12 _salt
  ) public returns(address)
  {
```

nfurfaro (commented):
Looks good overall. 1 comment addded

julien51:
Thanks a lot @nfurfaro !

Issue #7427, opened 2021-08-17
Mono repo improvements


julien51:
We currently use a mono repo approach but we're still duplicating too many dependencies between the various projects.
Additionally, some sub-folders such as unlock-js and unlock-assets are dependencies to others and we should simplify how they are being imported.

Eventually the goal is to make it much easier for anyone to run these applications as well as reduce our build/test time.

clemsos:
Reading various docs and articles, a few other solutions exist outside of the combo yarn workspaces + lerna. 

- Lerna supports yarn natively and yarn recommends lerna in its [workspaces docs](https://classic.yarnpkg.com/en/docs/workspaces/#toc-how-does-it-compare-to-lerna). Both seems to work fine but I read lots of complains discussing how slooooww lerna is
-  [Microsoft Rush](https://github.com/microsoft/rushstack/tree/master/apps/rush) seems to work much faster and less trouble with [yarn hoisting](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/)
- [NX](https://nx.dev/) but it seems to address tasks for projects with much larger teams with automations, multiple languages, etc - see [here](https://nx.dev/latest/react/guides/lerna-and-nx)
- use pnpm with [workspaces](https://pnpm.io/workspaces), much faster, less code and deps

My takeaway from reading all this and previous experience with monorepos 

1) lerna is kind of bloated
2) moving to pnpm will really increase build speed - but may require to remove yarn from existing
3) rush and nx are both worth exploring more. Nx seesm to have steeper learning curve but provide top-nocth integration with ms stack (VS code, github, etc)

clemsos:
After some testing, it looks like nx addresses several needs that aren't cover by yarn/lerna - for instance managing global eslint settings, using optimized test runners etc - see their [blog post](https://blog.nrwl.io/lerna-yarn-nx-faster-build-times-better-dev-ergonomics-2ec28463d3a5). There is also documented ways to run CIs / deploy using Circle or [Github actions](https://medium.com/ngconf/deploying-nx-affected-apps-from-github-actions-af28f8822ff3).

The only remaining question is the "migration" process. Nx, lerna, etc all have quite opinionated ideas about folders structure i.e. all code living in a `/packages` subfolder. Adding this tools will require quite a lot of changes in how the repo is organized - and therefore rewrite CIs etc . 

[...]

PR #7446, opened 2021-08-24
Yarn workspaces and upgrade to v2

clemsos:
# Description

- [x] yarn workspaces
- [x] linter conf as package
- [ ] fix various issues preventing install -- see `yarn dlx @yarnpkg/doctor`
- [ ] upgrade to yarn v2  

# Issues

Refs #7427

# Checklist:


clemsos:
The recommendation for yarn v2 is to check `.yarn/cache` into the monorepo so the build time get down to a few seconds. In our case this is ~500M of files so not sure yet if this is the best way to go as it could also slow down things quite a lot ( cloning indexing etc).

Need to be reassess

julien51:
Oh wow! Maybe we can discard some of the (heaviest) dependencies we have? any way to assess them?

clemsos:
yep I will look into that

clemsos:
> Oh wow! Maybe we can discard some of the (heaviest) dependencies we have? any way to assess them?

to answer this, it seems that won't be necessary after all. What will be good is to rely on yarn cache on rebuild. There is sth called "plug n play" on yarn 2 and if we manage to activate the cache, that will be like few secs only to install packages once its setup.

The main thing now is to create a proper multi-stage docker image to build directly from monorepo, with stage 1 being the current core image, and other targets building different packages and apps (using docker buildkit ).

If we can use yarn cache properly then it will be much faster. Also we can also bundle all alpine/apk deps inside stage 1 so we dont get to rebuild all these either. Stages per services can then be used directly in compose, which is one less hassle.

There is a [yarn prod-install plugin](https://gitlab.com/Larry1123/yarn-contrib/-/blob/master/packages/plugin-production-install/README.md#example-dockerfile-advanced) that helps package things for Docker.

julien51:
😱 all of the conflicts :(

clemsos:
yes that's a lot of conflicts, but again we can get rid of all the locks so it is going to be fine.

I am in the middle of cleaning the deps anyway, running some utils to check deps. What works best so far 

- `yarn workspaces foreach dlx depcheck`
- `yarn dlx @yarnpkg/doctor`

julien51:
Thanks for the clarification here! Maybe one thing we need to understand is see why we have more than 500MB of deps? Any clear culprit that we can remove or cleanup?

clemsos:
> Thanks for the clarification here! Maybe one thing we need to understand is see why we have more than 500MB of deps? Any clear culprit that we can remove or cleanup?

yes thats a better approach. Here 

```sh
$ du -sh ./node_modules/* | sort -nr | grep '\dM.*' | head -n20

142M    ./node_modules/@netlify
141M    ./node_modules/@storybook
113M    ./node_modules/truffle
 95M    ./node_modules/@openzeppelin
 87M    ./node_modules/@vue
 70M    ./node_modules/hardlydifficult-ethereum-contracts
 66M    ./node_modules/aws-sdk
 63M    ./node_modules/@truffle
 61M    ./node_modules/mcl-wasm
 61M    ./node_modules/ethers
 58M    ./node_modules/typescript
 58M    ./node_modules/microbundle-crl
 54M    ./node_modules/detective-typescript
 54M    ./node_modules/@firebase
 49M    ./node_modules/ethereumjs-testrpc
 46M    ./node_modules/solidity-coverage
 45M    ./node_modules/ganache-core
 43M    ./node_modules/hardhat
 33M    ./node_modules/hardlydifficult-eth
 30M    ./node_modules/next
```

what we can do

1. cleanup `smart-contracts-extension`:  remove `@vue/cli`(unused), migrate to hardhat (remove truffle + ganache), remove deprec `@openzeppelin/upgrades`
1. replace netlify deployment by docker-based solution. It should be possible to dockerize the cli and build from an image instead of a node dep. That will remove the `@netlify` and `detective-typescript`
2. remove unused ref `@openzeppelin/cli` in smart-contracts. Will also remove firebase

I think we can already cut down the size to >150M with this

julien51:
Sounds good!

Please, do move smart-contract-extensions to hardhat first! I think that's a low hanging fruit and you have experience with i already :)

I can look at the`@netlify` stuff and the `@openzeppelin/cli`.

I will also look into storybooks and see if we can make it ligther.

clemsos:
ok lots of learning here, also some complexity. I am gonna close this and split the problems into smaller PRs

PR #7501, opened 2021-09-02
removed dependency for netlify as it is too heavy

julien51:
# Description

Per @clemsos 's discovery work in https://github.com/unlock-protocol/unlock/pull/7446 we identify that the netlify dependency is quite large. This PR removes it as dependency and just installs it in the deploy script.
Another option could have been to use their docker image, but I am not sure this would be significantly better.

# Checklist:


julien51:
Great catch!

julien51:
Finally! PTAL and I'll merge!

clemsos:
good to go I think !

PR #7505, opened 2021-09-03
Create shared eslint config dep

clemsos:
# Description

Here we init the use of yarn workspaces and move the base eslint config into its own package.

# Issues

Relates to #7427

# Checklist:


clemsos:
Locksmiths tests are failing because of a breaking change in jest (removal of `setImmediate`) as discussed https://github.com/testing-library/react-testing-library/issues/918

julien51:
Ok, storybook issue solved. Both `paywall` and `unlock-protocol.com` were not really using it in a useful way so I removed it from them. That said, I'd love to find a mechanism to make sure we can run storybook in a "shared" way in all applications...  Maybe for later!

julien51:
For Ganache, I commented on the relevant issue. 
For locksmith, what do you think is the right way forward?

julien51:
When looking a bit closer at the test failing for `unlock-app` I think it is actually a transient error which I have seen happen in the past. I think it's because the tests are generating private keys which does take a bit of time and may indeed trigger the timeout.

clemsos:
> For locksmith, what do you think is the right way forward?

Need to polyfill `setImmediate` as show [here](https://docs.expo.dev/guides/using-nextjs/#polyfill-setimmediate)

clemsos:
the `setImmediate` issue is fixed but there are a bunch of new errors now. All seems related to ganache as well, so likely to share some common root

julien51:
hahahha, getting there! It looks like only the deploy scripts are failing!

clemsos:
> hahahha, getting there! It looks like only the deploy scripts are failing!

hehe all these build times teaching me patience

scripts/deploy.sh
```diff
@@ -9,11 +9,17 @@ TARGET=$3
 COMMIT=$4
 BRANCH=$5
 IS_FORKED_PR=$6
-NPM_SCRIPT="yarn run deploy-$TARGET"
 REPO_ROOT=`dirname "$0"`/..
 BASE_DOCKER_COMPOSE=$REPO_ROOT/docker/docker-compose.yml
 DOCKER_COMPOSE_FILE=$REPO_ROOT/docker/docker-compose.ci.yml
 
+if [ "$SERVICE" = "unlock-protocol-com" ] ; then
```
julien51, line 16: We really need to rename that service. Wdyt? In a later PR maybe?

.circleci/config.yml
```diff
@@ -3,9 +3,11 @@ version: 2
 jobs:
   unlock-protocol-com-tests:
     machine:
-      image: ubuntu-1604:202004-01
+      image: ubuntu-1604:202104-01
```
julien51, line 6: I wonder if there is a way to re-use that so we DRY things up a bit?

Dockerfile
```diff
@@ -0,0 +1,91 @@
+# syntax = docker/dockerfile:experimental
+
+# default value
+ARG BUILD_DIR=unlock-js
```
julien51, line 4: ok to keep or should we just throw an error if no `BUILD_DIR` is set?

scripts/deploy.sh
clemsos: this exists because the folder name is `unlock-protocol.com` but the docker service name is `unlock-protocol-com`. Another option will be to hyphenate all names by replacing special chars by dashes.

docker/docker-compose.ci.yml
```diff
@@ -35,18 +35,18 @@ services:
...
+      args:
+        BUILD_DIR: smart-contract-extensions
       cache_from:
         - smart-contract-extensions
         - unlockprotocol/smart-contract-extensions:master
-    depends_on:
-      - ganache
 
   smart-contracts: # Image with our contracts
-    image: smart-contractss
```
julien51, line 46: 🤦

docker/docker-compose.ci.yml
```diff
@@ -57,10 +57,13 @@ services:
...
     build:
       context: ../
-      dockerfile: ./docker/unlock-app.dockerfile
+      args: 
+        BUILD_DIR: unlock-app
       cache_from:
         - unlock-app
         - unlockprotocol/unlock-app:master
+    depends_on:
+      - ganache-integration
```
julien51, line 66: Needed?

docker/test.env
```diff
@@ -19,7 +19,7 @@ PAYWALL_HOST=paywall
 PAYWALL_PORT=3001
 PAYWALL_SCRIPT_URL=http://paywall:3001/static/paywall.min.js
 PAYWALL_URL=http://paywall:3001
-postgres_db-locksmith_test
+postgres_db=locksmith_test
```
julien51, line 22: 🤦

locksmith/src/app.ts
```diff
@@ -1,3 +1,4 @@
+import 'setimmediate'
```
julien51, line 1: Maybe add a quick comment about the fact that this is a shim?

julien51:
> hehe all these build times are teaching me patience

It already feels much faster though!

scripts/deploy.sh
julien51: Let's rename the folder in a later PR

clemsos:
Yep time is down from 18m 56s to 13m. I am sure there still tons of possible improvements, esp on the docker side

packages/eslint-config/test/index.test.js
```diff
@@ -0,0 +1,15 @@
+/* eslint strict: 0, global-require: 0 */
+
+'use strict';
+
+test('entry point parse', () => {
+    expect(() => require('..')).not.toThrow()
+})
+
+test('contains some lint instructions', () => {
```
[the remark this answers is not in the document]
clemsos: solen from airbnb haha

julien51 (approved):
I think this is looking very good! Let's fix a few small things and ship it!

docker/docker-compose.ci.yml
clemsos: yes i think so, we better make sure ganache is running before we start the tests -- even tho jest is likely slower to start

clemsos:
cool! added a few changes here and there.  Actually looking at the logs the CI time was taking 30m before we start removing deps last week -- so its down ~60% :)

Issue #7433, opened 2021-08-20
OAuth provider endpoint


julien51:
In order to provide an easier way for 3rd party applications to integrate Unlock, we want to provide an OAuth endpoint. Like all front-end software that we provide, it is not required and is provided just as a convenience mechanism.
One of its core benefits is that it handles transparently Unlock accounts as well as wallets. It will also include an optional checkout step as part of the normal OAuth flow. 

For users, it means they will be offered to "Connect through Unlock" on 3rd party sites, such as Wordpress or Ghost blogs, and their local accounts will then be mapped to an Ethereum address. The 3rd party sites can then easily verify that these users own a valid membership by querying RPC endpoints.

We are implementing a flow that is very similar to the one described by @aaronpk [on this page](https://aaronparecki.com/oauth-2-simplified/).

Task list:

* [ ] Authorization endpoint
    * [x] Update the checkout endpoint of `unlock-app` to support OAuth parameters
    * [x] Implement the flow where the user denies
    * [x] Implement the flow where the user approves
        * [x] When users are already Unlock users
        * [x] When users use their own crypto wallet
        * [x] When users create an Unlock account
    * [ ] Generate [Authorization code on backend](https://www.oauth.com/oauth2-servers/pkce/authorization-request/)
    * [ ] Implement [PKCE for additional security](https://www.oauth.com/oauth2-servers/pkce/)
    * [ ] Implement checkout from as part of OAuth connection 
* [x]  Identification endpoint 
     * [x] Implement [verification of the tokens](https://www.oauth.com/oauth2-servers/authorization/the-authorization-response/) provided in the `locksmith` application

julien51:
One addition we want to add to these URLs is the ability for the implementer to prompt the purchase of memberships (locks) once the user has successfully logged in. We would do that by allowing the implementer to provide a `paywallConfig` object as part of this flow.

julien51:
Let's close this for now as we're moving on with SIWE.

PR #7518, opened 2021-09-08
Oauth handler

julien51:
# Description

We want to add an OAuth endpoint to enable implementing application to easily offer their users to "connect with Unlock".
The OAuth flow includes 2 steps:
1. Getting a code from the front-end
2. Accessing the user's address by redeeming the code with a backend service.

This PR introduces this flow.
We are leveraging the existing checkout UI in `unlock-app` to let the user share their address with a 3rd party service and then we're introducing the locksmith API call to retrieve the user's address.

This is a first iteration and more iterations will come later as we add improvements to the flow.

Ref: #7433

image

### DIFF
--- a/unlock-app/src/components/interface/checkout/Checkout.tsx
+++ b/unlock-app/src/components/interface/checkout/Checkout.tsx
@@ -104,10 +104,6 @@ export const Checkout = ({
   const [existingKeys, setHasKey] = useReducer(keysReducer, {})
   const [selectedLock, selectLock] = useState<any>(null)
   const [savedMetadata, setSavedMetadata] = useState<any>(false)
-  // state change
-  useEffect(() => {
-    setState(defaultState)
-  }, [defaultState])
 
   // state change
   useEffect(() => {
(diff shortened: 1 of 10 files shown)


unlock-app/src/hooks/useProvider.ts
```diff
@@ -46,10 +46,11 @@ export const useProvider = (config: any) => {
       const _account = await _walletService.getAccount()
       let _signedMessage
       if (messageToSign) {
-        _signedMessage = utils.base64.encode(
```
julien51, line 49: No beed to base64 encode the message here.

unlock-app/src/components/interface/checkout/Checkout.tsx
```diff
@@ -104,10 +104,6 @@ export const Checkout = ({
   const [existingKeys, setHasKey] = useReducer(keysReducer, {})
   const [selectedLock, selectLock] = useState<any>(null)
   const [savedMetadata, setSavedMetadata] = useState<any>(false)
-  // state change
```
julien51, line 107: This was duplicated! (see line 109)

PR #7566, opened 2021-09-20
shared config package

### LINKED ISSUE #7555: Shared network config
Right now, all of our apps have their own "config" for networks:
* [unlock-app](https://github.com/unlock-protocol/unlock/blob/master/unlock-app/src/config.js#L58)
* [paywall](https://github.com/unlock-protocol/unlock/blob/master/paywall/src/paywall-script/networkConfigs.ts)
* [locksmith](https://github.com/unlock-protocol/unlock/blob/master/locksmith/src/networks.ts)
* [unlock-js](https://github.com/unlock-protocol/unlock/blob/master/unlock-js/examples/networks.js)

Now that we have a sane way of doing packages, we should isolate that logic in its own package that's imported by other applications!

clemsos:
# Description

A shared package containing info about the different networks where Unlock is deployed

# Issues

Fixes #7555 

# Checklist:


julien51 (approved):
Do u want to add the import inside the packages in another PR?

packages/networks/tsconfig.json
```diff
@@ -0,0 +1,73 @@
+{
```
julien51, line 1: I wonder if this should be shared at the root of the repo? wdyt?

clemsos: yep the best way is to have a `ts-config` package that we can import in the different workspaces. I think we keep it local for now then will add a package in another PR

clemsos:
> Do u want to add the import inside the packages in another PR?

hmmm makes more sense to have imports inside this one so we can run the tests against the new config

PR #7594, opened 2021-10-01
Public Lock bump version to v9

clemsos:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/test/Unlock/upgrades.js
+++ b/smart-contracts/test/Unlock/upgrades.js
@@ -21,7 +21,7 @@ const artifactsPath = path.resolve(
   'past-versions'
 )
 
-const versionsCount = 9
+const versionsCount = 10
 let unlock
 
 contract('Unlock / upgrades', async (accounts) => {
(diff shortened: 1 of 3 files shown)


clemsos:
after starting to dig into PublicLock upgrade to v10 in #7574 , I figured out we will have to upgrade Unlock as well. The reason for this is that we are moving to solidity `^0.8.0` and Unlock import PublicLock -so we have to upgrade sol version there too. Also we take the occasion to bump all oz contracts versions, which will affect the API (i.e. access controls / roles management)

So better we bump both versions together

smart-contracts/test/Unlock/upgrades.js
```diff
@@ -173,10 +173,17 @@ contract('Unlock / upgrades', async (accounts) => {
...
-                if (versionNumber == 2) {
+                const templateVersion = await lock.publicLockVersion()
+                const version =
+                  typeof templateVersion !== 'number'
+                    ? templateVersion.toNumber()
+                    : templateVersion
+                if (versionNumber === 2) {
                   // version 2 had a bug: we forgot to bump the lock version
                   assert.equal(version, 1)
+                } else if (versionNumber === 9) {
```
julien51, line 184: At some point we should make these better because in practice Unlock should be able to deploy any version of PublicLock (this is going to be part of the work on making PublicLock upgradable!)

PR #7761, opened 2021-11-11
Upgradeable publiclock

clemsos:
# Description

This add supports to `Unlock` for deployment of upgradeable locks using OZ proxy pattern.

See #6682 for a more detailed description

# Issues

Fixes #
Refs #

# Checklist:

- [ ] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

Issue #6682, opened 2020-05-18
Upgradable Public Lock


julien51:
The Public Lock contracts are currently *not* upgradable neither by us, nor by their owner or admin.
Ideally, at some point, we should make them upgradable by their admins, if they want to.

julien51:
Let's revisit this!

The plan would be this:
* change how `deployLock` works on the Unlock contract so that it deploys a proxy for locks ([either transparent or UUPS](https://docs.openzeppelin.com/contracts/4.x/api/proxy#transparent-vs-uups)). The Unlock contract will be the proxy admin for all locks. (See if it makes sense to still use the [minimal proxy](https://blog.openzeppelin.com/deep-dive-into-the-minimal-proxy-contract/) to deploy the proxy contract.
* In Unlock, instead of keeping track of a single template, we should have an array of all supported templates. The `deployLock` function should always deploy the last item of the array above.
* In Unlock, we add an `upgradeLock` function that takes 2 arguments, the lock address, and the template address. This function should do the following:
  * Verify that the `msg.sender` is one of the lock managers on this specific lock (fail if not)
  * Verify that the template address is "supported" (in the array above) and fail if not
  * Verify that the version of the template address is strictly one version larger than the current version of the lock, and fail if not
  * proceed to the upgrade of the lock!

clemsos:
For reference, WIP here https://github.com/clemsos/sol-upgrade-pattern/blob/master/contracts/ProxyDeployer.sol

julien51:
Closing as this was merged!

PR #7765, opened 2021-11-12
Rebased upgradeable publiclock

julien51:
# Description

This add supports to `Unlock` for deployment of upgradeable locks using OZ proxy pattern.

See #6682 for a more detailed description

NB : This is #7761 but rebased.

# Issues

Fixes #
Refs #

# Checklist:


smart-contracts/contracts/Unlock.sol
```diff
@@ -126,7 +148,55 @@ contract Unlock is
...
+  }
+
+  /**
+  * @dev Registers a new PublicLock template immplementation
+  * The template is identified by a version number
+  * Once registered, the template can be used to upgrade an existing Lock
+  */
+  function addLockTemplate(address impl, uint16 version) public onlyOwner {
+    require(impl != address(0), "impl address can not be 0x");
+    require(version != 0, "version can not be 0");
```
julien51, line 190: I wonder how "useful" that is?
Maybe we should check if this is indeed a smart contract address and if this is indeed a `PublickLock`? but it's probably hard to do?

smart-contracts/contracts/Unlock.sol
```diff
@@ -126,7 +148,55 @@ contract Unlock is
...
+    require(_publicLockVersions[impl] == 0, "address already used by another version");
```
julien51, line 191: I wonder if we should let the admin "overwrite" a function. Maybe it's fine since it's an admin thing anyway?

smart-contracts/contracts/Unlock.sol
```diff
@@ -126,7 +148,55 @@ contract Unlock is
...
+    require(_publicLockImpls[version] == address(0), "version already assigned");
```
julien51, line 192: same?

smart-contracts/contracts/Unlock.sol
```diff
@@ -126,7 +148,55 @@ contract Unlock is
...
+
+    _publicLockVersions[impl] = version;
+    _publicLockImpls[version] = impl;
+    if (publicLockLatestVersion < version) publicLockLatestVersion = version;
+
+    // TODO: init template to prevent possible attack
```
julien51, line 198: 👍  but that means these params have to be passed as `data` to this function?

julien51: it should be doable on the FE

smart-contracts/contracts/Unlock.sol
```diff
@@ -160,8 +231,13 @@ contract Unlock is
...
     }
-    address payable newLock = payable(address(uint160(publicLockAddress.createClone2(salt))));
-    IPublicLock(newLock).initialize(
+
+    // default to latest implementation
+    address impl = _publicLockImpls[publicLockLatestVersion];
+
+    // TODO: more flexible approche to init w params
+    bytes memory data = abi.encodeWithSignature(
+      'initialize(address,uint256,address,uint256,uint256,string)',
```
julien51, line 240: same as above. we need to pass a `data` blob that matches the actual `initialize` call + its params (the same way the DAO proposals are built I guess!)

smart-contracts/contracts/Unlock.sol
```diff
@@ -180,6 +260,39 @@ contract Unlock is
...
+
+    // check version
+    IPublicLock lock = IPublicLock(lockAddress);
+    uint16 currentVersion = lock.publicLockVersion();
+    require( version == currentVersion + 1, 'version error: only +1 increments are allowed');
+
+    // make our upgrade
+    address impl = _publicLockImpls[version];
+    TransparentUpgradeableProxy proxy = TransparentUpgradeableProxy(lockAddress);
+    proxyAdmin.upgrade(proxy, impl);
```
julien51, line 283: I assume this will fail for "old" (non upgradable) locks? It would be good to actually have a way to find "upgradable locks" so the UI only shows an upgrade button when the lock is indeed upgradable? (or do we just say any lock prior to 10 is not upgradable?)

smart-contracts/contracts/Unlock.sol
julien51: And would not be that different from how multisg wallets and/pr Governor contracts actually work (submitting a tx that includes a tx to be sent if approved by other members)

smart-contracts/contracts/Unlock.sol
clemsos: there is no proper way I know of in Solidity to check if an address actually exists or if it is a contract. You can check but that will open up to some possible attacks (by faking a contract).

clemsos: regarding the usefulness, it is more of my own coding convention to check the validity of a function params. That may come from too much Javascript lately :P Agree that as Solidity is strongly typed that may not be so useful

smart-contracts/contracts/Unlock.sol
clemsos: not sure what do you mean by overwrite?

smart-contracts/contracts/Unlock.sol
clemsos: yep agree. The main issue is that we will have to trigger the upgrade off-chain

smart-contracts/contracts/Unlock.sol
clemsos: yes the >10 solution seems the most straightforward

smart-contracts/contracts/Unlock.sol
julien51: ok I think since this is `onlyOwner` we can "trust" that they do good things? (and allow them to overwrite below always). So worst case scenario: they waste gas?

julien51: So basically I would remove the checks

smart-contracts/contracts/Unlock.sol
julien51: Well let the owner (re)set an existing version...

smart-contracts/contracts/Unlock.sol
julien51: wdym?

smart-contracts/contracts/Unlock.sol
julien51: On the FE only then. so we save gas :)

smart-contracts/contracts/Unlock.sol
clemsos: yep fine with me

smart-contracts/contracts/Unlock.sol
```diff
@@ -126,51 +148,80 @@ contract Unlock is
...
+  * @return address of the create lock
   */
   function createLock(
-    uint _expirationDuration,
-    address _tokenAddress,
-    uint _keyPrice,
-    uint _maxNumberOfKeys,
-    string memory _lockName,
-    bytes12 _salt
+    bytes memory data
```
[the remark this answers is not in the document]
julien51: > It means you will be able to create locks on behalf of others... bad.

Can you expand on this? Why is this bad?

Agree that 3 seems a bit "cleaner" but I am not sure it is bad to let people create locks on behalf of someone else?

clemsos:
What I mean is you lose the signature itself, which is the whole point of using crypto. So you can "impersonate" others without them knowing it. 

Let's say I create a lock with your address and distribute content that goes against your views or is plain illegal (say child porn). That's associât with you, plus you still receive money from it.

OTOH I see how it can be interesting to create a lock on behalf of someone else i.e. support groups for an individual ala GoFundMe etc

clemsos:
Note that you could also just disable the lock altogether as well, as this is your address associated with it.

I guess what I mean is that this is a non technical decision

julien51:
So technically you can still view who is the deployer (whoever sent the tx). 
Also, now I can deploy a lock, add you as a LockManager and renounce my own LockManager role and be in the same situation?

So all in all, as much as I understand what you're worried about, in practice it is quite easy to see who really deployed a lock anyway?

julien51:
So, in conclusion, I think there is no need for any of 1, 2 and 3. We should just add a param to the function that creates the lock to set the first lockManager as the msg.sender 

Note: be careful about the `beneficiary` that is set to `msg.sender` now :) (let's [look at all the mixins](https://github.com/unlock-protocol/unlock/blob/master/smart-contracts/contracts/PublicLock.sol#L42) and make sure they are initialized correctly!)

clemsos:
This seems good to go BUT all the various calls to `createLock` in libs like UnlockJS havent been updated. Maybe better in another PR?

smart-contracts/contracts/Unlock.sol
```diff
@@ -149,6 +146,8 @@ contract Unlock is
   {
     // We must manually initialize Ownable
     UnlockOwnable.__initializeOwnable(_unlockOwner);
+    // add a proxy admin on deployment
+    _deployProxyAdmin();
```
julien51, line 150: 👍

smart-contracts/contracts/interfaces/IUnlock.sol
```diff
@@ -29,6 +37,18 @@ interface IUnlock
     bytes12 _salt
   ) external returns(address);
 
+  /**
+  * @notice Upgrade a lock to a specific version
+  * @dev only available for publicLockVersion > 10 (proxyAdmin /required)
+  * @param lockAddress the existing lock address
+  * @param version the versin number you are targeting
```
julien51, line 44:
```suggestion
  * @param version the version number you are targeting
```

smart-contracts/contracts/interfaces/IUnlock.sol
```diff
@@ -123,6 +158,11 @@ interface IUnlock
     address payable _publicLockAddress
   ) external;
 
+  /**
+   * @dev This checks wether the sender is a manager of the specified lock
+   */
+  function isLockManager(address lockAddress, address _sender) external view returns(bool isManager);
```
julien51, line 164: Do we need this in the interface?

smart-contracts/test/Lock/createLockWithInfiniteKeys.js
```diff
@@ -16,14 +17,16 @@ contract('Lock / createLockWithInfiniteKeys', () => {
...
         60 * 60 * 24 * 30, // expirationDuration: 30 days
-        web3.utils.padLeft(0, 40),
+        web3.utils.padLeft(0, 40), // token address
         web3.utils.toWei('1', 'ether'), // keyPrice: in wei
         constants.MAX_UINT, // maxNumberOfKeys
-        'Infinite Keys Lock',
-        '0x000000000000000000000000'
-      )
+        'Infinite Keys Lock', // name
+        // '0x000000000000000000000000',
```
julien51, line 26: Do you know what this was?

julien51: Oh that was the salt, correct? 
Do we want to keep it? I think the goal was to "pre-compute" the address of the deployed lock, but I don't know if we can do it with this new approach...

julien51: Ok after offline convo I think we can safely get rid of the hashes :)
We will have to see how the lock proxies are deployed to see if we can compute their address before deployment but that's a front-end issue that we should solve there.

julien51:
LGTM!

Now I will update docs :)

julien51:
Merging now!

Issue #7726, opened 2021-11-01
Cache user address in front-end application


julien51:
Unlock's front-end application (`unlock-app`) requires the user to login in order to authenticate them. However, the user information is only available while the session is ongoing. Now that unlock actually supports an explicit "log out", it would be useful to cache/persist locally the user's address so that when the user refreshes the page, they don't have to login again. Similarly, when a user visits an Unlock page again, they should be logged-in by default...

We should consider using the browser's local storage to store all account info (including Unlock accounts' info like email!).

I suspect the best way to do that is to look at `useProvider.ts` and see how it can write its state to localStorage and read from localStorage when it is loaded.

julien51:
Assigning to @kalidiagne

PR #7775, opened 2021-11-16
Add cache user feature for autologin

kalidiagne:
# Description

Required to https://github.com/unlock-protocol/unlock/issues/7726 

When user log-in, info are stored in localStorage, following the path `${@unlock-app}.${key}`,

This will be used to autologin when the user visits the page next time if the is any data stored in localStorage.
otherwise have to follow the normal flow to login.

In particular, added `authenticateWithProvider`, defined in `useAuthenticaHandler`, that receives the name of the provider we want to use to login as a parameter. That's allowed in the future to use the same function and have the assurance we are calling an existing provider, and also provide more simplicity.

# Issues

Fixes #
- [x] flash effect during autologin
- [x] autologin

Refs #

# Checklist:


unlock-app/src/components/interface/LoginPrompt.tsx
```diff
@@ -108,6 +116,19 @@ const LoginPrompt = ({
     await authenticateIfNotHandled(ethereum)
   }
 
+  const tryAutoLogin = useCallback(() => {
```
julien51, line 119: I wonder if that logic wouldn't be better off in a different location? I worry that this might create a flash?

kalidiagne: Checked also simulating a slow network connection and no flash effect. as you suggested I can move it to a different location like `DashboardContent`, and also include a loader to prevent any eventual flash

julien51 (commented):
What about trying to see if we have account info in the `<Authenticate>` component?

unlock-app/src/components/interface/LoginPrompt.tsx
julien51: Great idea and good news! 
I think it would be better in the `<Authenticate>` component as it is happening at the highest level and can set the values in the provider directly? Wdyt?

julien51:
Hey @kalidiagne , what's the status on this? Are you making progress? Thanks!

kalidiagne:
Hi @julien51! , all done! just pushed all and updated the description!

kalidiagne:
> Hey @kalidiagne , what's the status on this? Are you making progress? Thanks!

Hi @julien51! , all done! just pushed all and updated the description!

julien51:
Nice! Let me look into that! In the meantime, could you check why tests/linting is not passing?
Thanks!

unlock-app/src/components/content/DashboardContent.js
```diff
@@ -1,5 +1,5 @@
 import styled from 'styled-components'
-import React, { useContext, useState } from 'react'
+import React, { useContext, useState, useCallback } from 'react'
```
julien51, line 2: Let's clean that up please 🧹

unlock-app/src/components/interface/Authenticate.jsx
```diff
@@ -143,7 +151,7 @@ export const Authenticate = ({
             email={email}
             encryptedPrivateKey={encryptedPrivateKey}
           >
-            {children}
+            {!isLoading ? children : "Loading..."}
```
julien51, line 154: Do you mind using the `<Loading>` component we use in other places?

julien51 (commented):
I really like this new approach! I'll do a more thorough review a bit later today or tomorrow but I think we're looking good here!

unlock-app/src/components/content/DashboardContent.js
kalidiagne: @julien51 fixed, and removed other non-used vars.
anyway, we can think about removing `no-unused-vars` from the `rulesToIgnore` in `.eslintrc`

kalidiagne:
cool! i'll keep an eye on the tests and linting.
it should be fixed now

kalidiagne:
@julien51 oh, running tests in local blocks completely my machine.
 i'll add more commints with single fixs

julien51:
> @julien51 oh, running tests in local blocks completely my machine.

Hum, that is bad :/ Any idea why?

julien51:
Also you should be able to run `yarn lint` locally to help fix these... Let me know if you need help! Thanks!

kalidiagne:
All fixed! also add some little refactoring

julien51:
@kalidiagne This is very cool and the code is very clean (left a few comments).
When I tried locally I was faced with this 
image


I think it is because my localStorage has `isMetaMask` stored :) which is very unique but makes me think we could have some kind of failover mechanism where if the value is not one we expect, we just delete it and move on?

julien51:
Alright! Looks good to me @kalidiagne !
I will ask @clemsos to give a look as well to get his opinion.

kalidiagne:
> @kalidiagne This is very cool and the code is very clean (left a few comments). When I tried locally I was faced with this image
> 
> I think it is because my localStorage has `isMetaMask` stored :) which is very unique but makes me think we could have some kind of failover mechanism where if the value is not one we expect, we just delete it and move on?

Oh that's a good idea! thanks for the tip, I see you already add the edit thk!

julien51:
Merging now! Thanks a lot @kalidiagne

PR #7796, opened 2021-11-19
Gas refund on purchase

### LINKED ISSUE #7774: Gas refund from lock
We want the locks to be able to refund gas to the `msg.sender` of the transaction.
Right now any address can purchase a membership on behalf of someone else (because the purchase function has a `beneficiary` address. 
We need to add the ability for the lock owner to set a percentage of the key price to be refunded to the sender. By default it should be 0.

This will enable better key purchases through credit cards, as well as cross-chain transactions.

clemsos:
# Description

Set a percentage of the key price to be refunded to the sender on purchase

# Issues

Fixes #7774
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -215,12 +215,12 @@ interface IPublicLock
   /**
   * @dev Set a percentage of the key price to be refunded to the sender on purchase
   */
-  function setGasRefundPercentage(uint8 _percent) external;
+  function setGasRefundBasisPoints(uint128 _basisPoints) external;
   
   /**
   * @dev Returns percentage be refunded to the sender on purchase
   */
-  function gasRefundPercentage() external view returns (uint8 percentage);
+  function gasRefundBasisPoints() external view returns (uint128 basisPoints);
 
   /**
    * @notice returns the minimum price paid for a purchase with these params.
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -213,6 +213,16 @@ interface IPublicLock
   ) external payable;
 
   /**
+  * @dev Set a percentage of the key price to be refunded to the sender on purchase
+  */
+  function setGasRefundPercentage(uint8 _percent) external;
+  
+  /**
+  * @dev Returns percentage be refunded to the sender on purchase
+  */
+  function gasRefundPercentage() public view external (uint8 percentage);
+
+  /**
    * @notice returns the minimum price paid for a purchase with these params.
    * @dev this considers any discount from Unlock or the OnKeyPurchase hook.
    */
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -220,7 +220,7 @@ interface IPublicLock
   /**
   * @dev Returns percentage be refunded to the sender on purchase
   */
-  function gasRefundPercentage() public view external (uint8 percentage);
+  function gasRefundPercentage() external view returns (uint8 percentage);
 
   /**
    * @notice returns the minimum price paid for a purchase with these params.
--- a/smart-contracts/contracts/mixins/MixinPurchase.sol
+++ b/smart-contracts/contracts/mixins/MixinPurchase.sol
@@ -29,14 +29,14 @@ contract MixinPurchase is
   * @dev Set a percentage as basis point (10000th) of the key price to be refunded to the sender on purchase
   */
 
-  function setGasRefundBasisPoints(uint128 _basisPoint) external onlyLockManager {
-    _gasRefundBasisPoints = _basisPoint;
+  function setGasRefundBasisPoints(uint128 _basisPoints) external onlyLockManager {
+    _gasRefundBasisPoints = _basisPoints;
   }
   
   /**
   * @dev Returns percentage as basis point (10000th) to be refunded to the sender on purchase
   */
-  function gasRefundBasisPoints() external view returns (uint128 percentage) {
+  function gasRefundBasisPoints() external view returns (uint128 basisPoints) {
     return _gasRefundBasisPoints;
   }
 
--- a/smart-contracts/contracts/mixins/MixinPurchase.sol
+++ b/smart-contracts/contracts/mixins/MixinPurchase.sol
@@ -6,7 +6,6 @@ import './MixinKeys.sol';
 import './MixinLockCore.sol';
 import './MixinFunds.sol';
 
-
 /**
  * @title Mixin for the purchase-related functions.
  * @author HardlyDifficult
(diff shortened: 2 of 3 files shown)


smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -21,6 +20,24 @@ contract MixinPurchase is
 {
   event RenewKeyPurchase(address indexed owner, uint newExpiration);
 
+  event GasRefunded(address indexed receiver, uint refundedAmount, address tokenAddress);
```
clemsos, line 23: A specific event is emitted so things can be tracked, but we can also just track transfer event to save gas

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -106,6 +123,19 @@ contract MixinPurchase is
...
+    if (_gasRefundPercentage != 0) {
+      uint toRefund = _gasRefundPercentage * pricePaid / 100;
+      if(tokenAddress != address(0)) {
+        IERC20Upgradeable token = IERC20Upgradeable(tokenAddress);
+        token.transferFrom(address(this), msg.sender, toRefund);
+      } else {
+        (bool success, ) = msg.sender.call{value: toRefund}("");
+        require(success, "Refund failed.");
+      }
+      emit GasRefunded(msg.sender, toRefund, tokenAddress);
```
clemsos, line 137: Gas refund as a percentage of the key price means that gas is in ETH while key price will be refunded in any sent token. Does that make sense?

smart-contracts/contracts/mixins/MixinPurchase.sol
julien51: Let's keep the event for now.

smart-contracts/contracts/mixins/MixinPurchase.sol
julien51: Yes and I think that's the only way as we don't know that the lock will have eth to cover the gas... The "submitters" will be able to assess whether it is worth for them to issue the tx based on the conversion rate of a given token.

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -21,6 +20,24 @@ contract MixinPurchase is
...
   event RenewKeyPurchase(address indexed owner, uint newExpiration);
 
+  event GasRefunded(address indexed receiver, uint refundedAmount, address tokenAddress);
+
+  uint8 private _gasRefundPercentage = 0; // default to 0
+
+  /**
+  * @dev Set a percentage of the key price to be refunded to the sender on purchase
+  */
+  function setGasRefundPercentage(uint8 _percent) external onlyLockManagerOrBeneficiary {
```
julien51, line 30:
```suggestion
  function setGasRefundPercentage(uint8 _percent) external onlyLockManager {
```

Let's limit this to the lock manager.

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -21,6 +20,24 @@ contract MixinPurchase is
...
+    _gasRefundPercentage = _percent;
+  }
+  
+  /**
+  * @dev Returns percentage be refunded to the sender on purchase
+  */
+  function gasRefundPercentage() external view onlyLockManagerOrBeneficiary returns (uint8 percentage) {
```
julien51, line 37: I think `uint8` is too small! We need to go up to 10,000 (basis points). Also this should not have the `onlyLockManagerOrBeneficiary` modifier :)

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -106,6 +123,19 @@ contract MixinPurchase is
     {
       onKeyPurchaseHook.onKeyPurchase(msg.sender, _recipient, _referrer, _data, inMemoryKeyPrice, pricePaid);
     }
+
+    // refund gas
+    if (_gasRefundPercentage != 0) {
+      uint toRefund = _gasRefundPercentage * pricePaid / 100;
```
julien51, line 129: `/100` will not work. We need to do this in basis points (check methods like the penalties for an example as it is also a %)

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -212,6 +212,16 @@ interface IPublicLock
     bytes calldata _data
   ) external payable;
 
+  /**
+  * @dev Set a percentage of the key price to be refunded to the sender on purchase
+  */
+  function setGasRefundPercentage(uint8 _percent) external;
```
julien51, line 218: `uint8` is too small :)

smart-contracts/contracts/mixins/MixinPurchase.sol
julien51: Also we should probably check that it is lower (or equal even tho that would not make a lot of sense) to 10,000.

smart-contracts/test/Lock/purchaseGasRefund.js
```diff
@@ -0,0 +1,143 @@
...
+    lock = locks.FIRST
+  })
+
+  describe('gas refund percentage', () => {
+    it('get set properly', async () => {
+      await lock.setGasRefundPercentage(25)
+      assert.equal(await lock.gasRefundPercentage(), 25)
+    })
+
+    it('can not be set if caller is not lock manager or beneficiary', async () => {
```
julien51, line 29: No need to allow the beneficiary...

smart-contracts/test/Lock/purchaseGasRefund.js
```diff
@@ -0,0 +1,143 @@
...
+      await lock.addLockManager(accounts[5], { from: accounts[0] })
+      await lock.setGasRefundPercentage(25, { from: accounts[5] })
+      assert.equal(await lock.gasRefundPercentage(), 25)
+    })
+  })
+
+  describe('purchase with gas refund', () => {
+    let tx
+
+    beforeEach(async () => {
```
julien51, line 54: Let's also write tests for an ERC20 just to make sure :)

julien51 (changes requested):
A few changes are required :)

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: moved to basis points

smart-contracts/test/Lock/purchaseGasRefund.js
clemsos: yep added

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -125,8 +127,8 @@ contract MixinPurchase is
     }
 
     // refund gas
-    if (_gasRefundPercentage != 0) {
-      uint toRefund = _gasRefundPercentage * pricePaid / 100;
+    if (_gasRefundBasisPoints != 0) {
+      uint toRefund = _gasRefundBasisPoints * pricePaid / BASIS_POINTS_DEN;
```
julien51, line 131: 👍 for reusing BASIS_POINTS_DEN!

julien51:
LGTM!

julien51 (approved):
💯

PR #7797, opened 2021-11-19
Add key manager to lock purchase

### LINKED ISSUE #7663: Add keyManager to `purchase` function
As part of the publicLock new version (10!) let's make sure we add the ability to pass an address of `keyManager` on the `purchase` function.
This will let us improve our credit card purchase mechanism further!

clemsos:
# Description

Allow to set a keyManager during lock creation or duration extension

# Issues

Fixes #7663
Refs #

# Checklist:


julien51:
I think our `/rebase` github action does not work :(

clemsos:
Note that this introduces breaking changes on the purchase function

julien51:
Yup. We will have to update unlockjs :)

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -86,14 +93,24 @@ contract MixinPurchase is
...
     } else {
       // This is an existing owner trying to renew their expired key
       newTimeStamp = block.timestamp + expirationDuration;
       toKey.expirationTimestamp = newTimeStamp;
 
-      // reset the key Manager to 0x00
-      _setKeyManagerOf(idTo, address(0));
+      if(_keyManager != address(0) && keyManagerOf[idTo] != _keyManager) {
+          _setKeyManagerOf(idTo, _keyManager);
+      } else {
```
julien51, line 110: I don't think we need this anymore. Basically whoever will "pay" for that key (call the `purchase` function) will get to decide who the manager is and I think that's fine.

julien51 (approved):
One very small simplification but I think we're looking good after that!

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: you mean not reseting the key manager to address zero?

also one possible case is someone willing to extend their key and keep the existing key manager. We could check id a key manager already exists if address zero is passed, so we just keep the existing key manager, instead of erasing it ?

julien51: I think the best way is to just set it to whatever the caller of the `purchase` transaction decided. This will bring clarity and make things easier to manager.

clemsos:
Here is seems we have an interesting case: gh-actions is testing the merged state, while circle is testing the PR code as it is.

julien51 (changes requested):
I just realized that we should NOT reset the key manager in that case"
* if there is one
* if the key is not expired
Because if we reset it in that case, someone can actually "steal" funds from the ley manager.

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -86,14 +91,17 @@ contract MixinPurchase is
       // This is an existing owner trying to extend their key
       newTimeStamp = toKey.expirationTimestamp + expirationDuration;
       toKey.expirationTimestamp = newTimeStamp;
+
+      // set key manager
+      _setKeyManagerOf(idTo, _keyManager);
```
julien51, line 96: So actually in tat case, we should NOT reset the key manager.

clemsos: yep makes sense as we are not using `msg.sender` here

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -48,7 +48,9 @@ contract MixinPurchase is
   * (_value is ignored when using ETH)
   * @param _recipient address of the recipient of the purchased key
   * @param _referrer address of the user making the referral
+  * @param _keyManager optional address to grant managing rights to a specific address on creation
   * @param _data arbitrary data populated by the front-end which initiated the sale
+  * @notice when called for an existing and non-expired key, the `_keyManager` param will be ignored
```
julien51, line 53: 👍

julien51 (approved):

PR #7815, opened 2021-11-24
changed image path within reddit-ama blog post

ccarfi:


julien51:
The CLA bot does not like you! It never happens to me and it already happened twice to you!

Issue #7805, opened 2021-11-21
Hardhat module to deploy Unlock + Locks locally


julien51:
When using Unlock in a smart contract we need to make it easier for people who are deploying locks locally to test things out.
For this we should introduce a hardhat plugin package that developers can easily install and that offer the ability to deploy unlock locally, as well as locks and make it easy to use Unlock in their own projects!

We are attaching a 500 DAI bounty to this issue! Please get in touch if you're interested :)

julien51:
For this we can probably leverage @unlock-protocol/contracts!

julien51:
Related/relevant #7659

PR #7821, opened 2021-11-25
Serialize and duplicate a Lock

clemsos:
Contracts and tooling to clone / replicate a lock 

# Description

# Issues

Fixes #
Refs #

# Checklist:


smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -125,6 +125,13 @@ interface IPublicLock
    */
   function numberOfOwners() external view returns (uint);
 
+  /**
+  * A function which returns a subset of the keys for this Lock as an array
+  * @param _page the page of key owners requested when faceted by page size
+  * @param _pageSize the number of Key Owners requested per page
+  */
+  function getOwnersByPage(uint _page, uint _pageSize) external view returns (address[] memory);
```
julien51, line 133: Hum. I am not sure I understand why we need this?

smart-contracts/contracts/utils/LockSerializer.sol
```diff
@@ -0,0 +1,176 @@
...
+      numberOfOwners,
+      totalSupply,
+      keyOwners,
+      expirationTimestamps
+    );
+
+    return serializedLock;
+  }
+
+  function deployLock(address unlockAddress, Lock memory _lock, bytes12 _salt) public returns (address newLockAddress) {
```
julien51, line 154: I don't think we need this. we can deploy from JS using the Unlock contract

julien51:
So I think this is a good approach but there is no need to re-introduce the method to list all key owners IMO (at least it would not matter because well the locks are already deployed and have that method!)

We also need to make sure we get the lock managers off chain

clemsos:
My idea is that it may be useful to have a serializer that works entirely off-chain (except for managers currently). For instance, this can allow to duplicate a lock which is something users may want (for instance to migrate towards an upgradeable lock?).

smart-contracts/contracts/interfaces/IPublicLock.sol
clemsos: yep removed

smart-contracts/scripts/lock/samples.js
```diff
@@ -0,0 +1,69 @@
+const { ethers } = require('hardhat')
+const Locks = require('../../test/fixtures/locks')
```
clemsos, line 2: This script/task fetch data from test fixtures and deploy the locks locally. This is is very useful for local development

smart-contracts/contracts/utils/LockSerializer.sol
```diff
@@ -0,0 +1,156 @@
...
+    // ownership
+    uint256 totalSupply = lock.totalSupply();
+    uint numberOfOwners = lock.numberOfOwners();
+    
+    // keys
+    address[] memory keyOwners = new address[](totalSupply);
+    address[] memory keyManagers = new address[](totalSupply);
+    uint[] memory expirationTimestamps = new uint[](totalSupply);
+    
+    for (uint256 tokenId = 0; tokenId < totalSupply; tokenId++) {
```
clemsos, line 124: to list all key owners we just loop through the whole token supply - so no need for the deprec `getOwnersByPage`

clemsos:
we now have a script to deploy locks (no cli task tho), so this relates to #7805 it seems

smart-contracts/contracts/interfaces/IUnlock.sol
```diff
@@ -21,20 +21,23 @@ interface IUnlock
   function proxyAdminAddress() external view;
```
julien51, line 21: Are changes in this file expected?

smart-contracts/contracts/utils/LockSerializer.sol
```diff
@@ -0,0 +1,158 @@
...
+    uint256 freeTrialLength;
+    uint256 refundPenaltyBasisPoints;
+    uint256 transferFeeBasisPoints;
+  }
+
+  struct LockMetadata {
+    // metadata
+    string name;
+    string symbol;
+    // string baseTokenURI; // private?
```
julien51, line 66: Hum. Is there no way to export it?

smart-contracts/package.json
```diff
@@ -43,6 +43,7 @@
     "hardlydifficult-eth": "1.1.4",
     "hardlydifficult-ethereum-contracts": "0.11.1",
     "lint-staged": "11.2.6",
+    "node-fetch": "^3.1.0",
```
julien51, line 46: used?

smart-contracts/contracts/interfaces/IUnlock.sol
julien51: We cannot change create lock.. the Unlock contract has already been deployed and this would break other people's integrations :(

smart-contracts/scripts/deployments/lock.js
```diff
@@ -0,0 +1,99 @@
...
+    tx = await unlock.createLock(
+      expirationDuration,
+      tokenAddress,
+      keyPrice,
+      maxNumberOfKeys,
+      name,
+      salt
+    )
+  } else {
+    const calldata = await createLockHash({
```
julien51, line 37: Ho nice, does that mean we support v10 ?

smart-contracts/scripts/deployments/lock.js
```diff
@@ -0,0 +1,99 @@
...
+  const { args } = events.find(({ event }) => event === 'NewLock')
+  const { newLockAddress } = args
+
+  // eslint-disable-next-line no-console
+  console.log(
+    `LOCK DEPLOY > deployed to : ${newLockAddress} (tx: ${transactionHash})`
+  )
+
+  // get versionof the original lock
+  const publicLockVersion = 8 // await unlock.publicLockVersion()
```
julien51, line 54: We probably have v7 locks too on xDAI?

smart-contracts/scripts/lock/clone.js
```diff
@@ -0,0 +1,80 @@
...
+    unlockAddress,
+    unlockVersion,
+    salt: web3.utils.randomHex(12),
+  })
+
+  // eslint-disable-next-line no-console
+  console.log('CLONE LOCK > add key owners...')
+  const newLock = PublicLock.attach(newLockAddress)
+  const { keyOwners, expirationTimestamps, keyManagers } = serializedLock
+  const keyTx = await newLock.grantKeys(
```
julien51, line 33: We probably need to paginate here. I worry this won't work for very large locks (with more than 100 members)

smart-contracts/scripts/lock/clone.js
```diff
@@ -0,0 +1,80 @@
...
+    const txs = Promise.all(
+      managers.map((manager) => newLock.addLockManager(manager))
+    )
+    const { events } = await txs.wait()
+    const added = events.map(({ event }) => event.name === 'LockManagerAdded')
+    added.forEach(({ args }) => {
+      // eslint-disable-next-line no-console
+      console.log(`LOCK CLONE > ${args[0]} added as lock manager.`)
+    })
+  }
```
julien51, line 66: If more than 1 was added, let's also make sure we remove ourselves as lockManager.

smart-contracts/scripts/lock/clone.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// execute as standalone
```
julien51, line 69: Oh and let's make sure we update the beneficiary?

julien51 (commented):
A few things missing but this is going on the right direction.

smart-contracts/contracts/interfaces/IUnlock.sol
clemsos: hmmm yes it was just to be consistent with what is already in the repo. Presently the Unlock contract in the repo has already been updated to use upgradeable locks, so I thought we may as well update the interface too - and use  IUnlockV8 where needed

smart-contracts/package.json
clemsos: yes used to fetch data from the graph in the `lock:managers` task

smart-contracts/scripts/deployments/lock.js
clemsos: yes it should support both version. Also there is no v10, we only reached v9 with the upgradeable pattern. PublicLock is at v10 tho

smart-contracts/scripts/deployments/lock.js
clemsos: ok will look into supporting this too then

smart-contracts/scripts/lock/clone.js
clemsos: yep , this is more a generic task/script but this will be added in the locksmith route script

smart-contracts/contracts/interfaces/IUnlock.sol
julien51: Yes, but that is actually the thing I wanted to chat about with you regarding the upgrade :/ Let's circle back via voice.

smart-contracts/scripts/lock/clone.js
julien51: That's a good question and i don't know. I think the lock with most users on xdai is this one https://blockscout.com/xdai/mainnet/address/0xac1fceC2e4064CCd83ac8C9B0c9B8d944AB0D246/transactions

We should try with it and see what happens

smart-contracts/package.json
clemsos: replaced by cross-fetch btw

clemsos:
To get the `baseTokenURI` is tricky because solidity does not really support any advanced string manipulations... Even checking is a substring exists is non-trivial. i will propose we can return all tokenURIs then can do the computation off-chain.

I am not sure what is a reliable way to detect if `baseTokenURI` is custom or default, but I guess we can just check against the common ones we use to see if it is any different.

julien51 (approved):

PR #7886, opened 2021-12-05
feat: Add fleek deployment

### LINKED ISSUE #7198: Deploy all front-end applications to IPFS
Our front-end code is currently deployed using a cloud service called Netlify but they are perfectly "static" applications. This means that they can/should also be deployed on IPFS so that they remain accessible at all points, without relying on Unlock Inc's hosting bills :)

We want to have all of our builds to be seamlessly deployed to IPFS, pinned to our Pinata server.
We want to deploy as part of our CI process using the following rules:
- if the CI is running as part of a PR: deploy to IPFS and (do not pin it?). Do not point the IPNS to it
- if CI is running as part of a "master" build deploy to IPFS, pin it forever in our pinata server. Point to a "staging" domain using IPNS
- if CI is running as part of our "production" build, deploy to IPFS, pin it forever in our pinata server. Point our production domain to it using IPNS.

Right now, all front-end applications are deployed thru our CircleCI config. ([see this for example](https://github.com/unlock-protocol/unlock/blob/master/.circleci/config.yml#L178)). This [script then calls](https://github.com/unlock-protocol/unlock/blob/master/scripts/deploy.sh) a yarn command `yarn run deploy-netlify`, but should also probably just call `yarn run deploy-ipfs` :)

We are attaching a 500 DAI bounty to this to which we add a 300 DAI bonus if completed by July 31st!

searchableguy:
# Description

This PR is adding fleek as a deployment target for the IPFS support.

Currently, fleek cli doesn't support passing the path to .fleek file so we cannot add a staging deployment as well. I raised this as an issue to the fleek team and they are looking into adding an option to pass a .fleek file. So in the future, we can have `.fleek.staging` for the staging environment and modify the script to deploy. 

image

if not, I could probably write a small script to do it ourselves. I looked into their fleek-cli. It packages files and then uploads them to an endpoint with API key in the header. Doesn't look too complex. 

We will need to provide `FLEEK_API_KEY` as an environment variable with the appropriate key when we enable the deployment. 

# Issues

Fixes #7198 

# Checklist:


.prettierrc
```diff
@@ -1,5 +1,5 @@
 {
   "trailingComma": "es5",
   "singleQuote": true,
-  "semi": false,
```
julien51, line 4: Nit: remove this change from this PR?

unlock-app/next-env.d.ts
```diff
@@ -0,0 +1,6 @@
+/// <reference types="next" />
```
julien51, line 1: Nit: remove from this PR?

yarn.lock
```diff
@@ -13800,7 +13800,7 @@ __metadata:
   bin:
     asc: bin/asc
     asinit: bin/asinit
-  checksum: fda292f6b3b8b4a9bd4457c9143e50145f8473f3d664e48c057638a9bb1c8e890cd4de4e38e18dd4334240394621cfd9d01afee344ab049c5a7072fe855329b5
+  checksum: 8a407db6179addfaa5fcc637543a755ecc22ae8dd5e8259492305afd0d2bb73eae019d4178f79b515f75648d71dfda2c1ceb0c7421bff7817aebb2e2f5c8bfe7
```
julien51, line 13803: Any idea why this changed?

.github/disabled-workflows/_fleek.yml
```diff
@@ -0,0 +1,36 @@
...
+        shell: bash
+        id: check_changes
+      - name: Check if forked PR
+        run: |
+          echo "::set-output name=is_forked::$(${{github.repository_owner != 'Unlock Protocol' && github.event.pull_request.merged == true}})
+      - name: Set target environment
+        run: |
+          echo "::set-output name=target::$(${{ contains(github.base_ref, 'production') || input.service == 'unlock-protocol-com' }} && 'prod' || 'staging')"
+      - name: Deploy to fleek
+        # Only allow production deployment for now
```
julien51, line 33: 👍

julien51 (commented):
Looks good! I think it is ok to only deploy the production code to fleek/IPFS.
Have you been able to deploy things successfully?

What do you recommend we do when it comes to the API KEY?

yarn.lock
searchableguy: The checksum wasn't matching on yarn install so it was failing. I think they changed the GitHub release. I forced yarn to update the checkup in lock file.

searchableguy:
For FLEEK_API_KEY, I would recommend adding it in the github secrets so we can reference it in our deployment CI. 

Deployment was working fine locally but hasn't been tested in CI.

searchableguy:
One issue to resolve after this if we are going to use ENS or HNS based domain is to change how we are linking to the app in the unlock-protocol-com site.  Right now, it's an environment variable directly pointing to `https://app.unlock-protocol.com/dashboard`. We should make the domain part dynamic and only set the subdomain part like `app` so we can infer the domain name to build our URL.

Other options might be more complicated to add but may provide more flexibility. I will let you decide on the approach.

yarn.lock
julien51: Ok thanks for the clarification

julien51:
This is starting to look very good! I will now see how to add the secret to the Github action setup! Thanks a bunch.

julien51:
Alright! I added the `FLEEK_API_KEY` variable to the repo secrets
image

julien51:
So I assume we are good to merge for now, which is exciting.

Regarding the domain issue. Let's see afterward, once we have the deployments working consistently. I would assume for now they will use the pre-configured domain in fleek like `unlock-protocol.on.fleek.co` ?

julien51:
Let me know if it is ready to do a final review please!

PR #7889, opened 2021-12-05
Remove manual copy function due to moving files directly into public

searchableguy:
# Description

This PR removes `moveStaticFiles` which is made redundant by https://github.com/unlock-protocol/unlock/commit/84f2d039190d34c0feea7f3089c1a45a80361842, I believe. This is causing build for unlock-app to fail.

PS: The `${PORT:=3000}` added to yarn start end won't work. Do you want to load PORT from environment and provide 3000 as default? If so, you will need to use another utility package to do it or write a shell script.  I have reverted it for now.

# Checklist:


julien51 (approved):
Thanks a lot!

PR #7910, opened 2021-12-13
Support custom Erc721 balanceof hook

### LINKED ISSUE #7021: Support for 3rd party NFT
We want to introduce a hook called `onBalanceOf` that can be configured optionally by a lock manager.
This hook will be called _after_ our logic that checks whether the user does indeed have a key and should pass as arguments the lock address, the user address, and the timestamp of the user's key.

clemsos:
# Description

Allow lock managers to define a custom contract to define how `balanceOf` (i.e. `hasValidKey`) behaves.

# Issues

Fixes #7021
Refs #

# Checklist:


smart-contracts/contracts/test-artifacts/TestEventHooks.sol
```diff
@@ -78,4 +78,16 @@ contract TestEventHooks is ILockKeyPurchaseHook, ILockKeyCancelHook
...
+
+  
+  function hasValidKey(
+    address _lockAddress,
+    address _from,
+    uint _expirationtimeStamp
+  ) external view
+    returns (bool isValidKey)
+  { 
+    return _expirationtimeStamp < block.timestamp;
```
clemsos, line 90: That test case is a bit weak - it only uses one param. Not sure how to improve it, maybe by using some more conditions?

smart-contracts/contracts/interfaces/hooks/ILockBalanceOfHook.sol
```diff
@@ -0,0 +1,26 @@
...
+{
+
+  /**
+   * @notice If the lock owner has registered an implementer then this hook
+   * is called every time balanceOf is called
+   * @param lockAddress the address of the current lock
+   * @param keyOwner the potential owner of the key for which we are retrieving the `balanceof`
+   * @param expirationTimestamp the key expiration timestamp
+   */
+  function hasValidKey(
```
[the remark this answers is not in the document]
julien51: Oh yes, I think we should use only `balanceOf` and get rid of `hasValidKey`.

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -148,7 +148,16 @@ contract MixinKeys is
     public
     view
     returns (bool)
-  {
+  { 
+    // use hook if it exists
+    if(address(onBalanceOfHook) != address(0)) {
```
julien51, line 153: So I think a more powerful approach would be to still have 
`balanceOf = keyByOwner[_keyOwner].expirationTimestamp > block.timestamp;` and then, if the hook is defined, call it with all the params you currently have AND `balanceOf`, so that the 3rd party hook can decide on its logic completely.

smart-contracts/contracts/mixins/MixinLockCore.sol
```diff
@@ -205,16 +207,19 @@ contract MixinLockCore is
...
     address _onKeyPurchaseHook,
     address _onKeyCancelHook,
+    address _onBalanceOfHook,
     address _onTokenURIHook
   ) external
     onlyLockManager()
   {
     require(_onKeyPurchaseHook == address(0) || _onKeyPurchaseHook.isContract(), 'INVALID_ON_KEY_SOLD_HOOK');
     require(_onKeyCancelHook == address(0) || _onKeyCancelHook.isContract(), 'INVALID_ON_KEY_CANCEL_HOOK');
+    require(_onBalanceOfHook == address(0) || _onBalanceOfHook.isContract(), 'INVALID_ON_BALANCEOF_HOOK');
```
julien51, line 217:
```suggestion
    require(_onBalanceOfHook == address(0) || _onBalanceOfHook.isContract(), 'INVALID_ON_BALANCE_OF_HOOK');
```

smart-contracts/contracts/test-artifacts/TestEventHooks.sol
julien51: Oh, maybe we can make it just return true for an hardcoded address that does not have a valid membership?

julien51 (changes requested):
Let's change a little bit how to hook works so that we pass what the lock _would_ return to the hook as well, and the hook can decide to "override" the behavior?

clemsos:
thanks for the inputs. I ended up rewriting the logic a bit and focused the hook on `hasValidKey` instead of `balanceOf`. I think in our case it makes more sense - `balanceOf` is more of an ERC721 artefact. 

The key validity is now passed to the hook which can chose to use it or not. I added a better test case where "special members" get infinite use cases, while returning default result for others adresses.

smart-contracts/contracts/interfaces/hooks/ILockValidKeyHook.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.5.17 <0.9.0;
+
+
+/**
+ * @notice Functions to be implemented by a balanceOf Hook.
```
julien51, line 6:
```suggestion
 * @notice Functions to be implemented by a hasValidKey Hook.
```

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -147,9 +147,19 @@ contract MixinKeys is
...
+    // use hook if it exists
+    if(address(onValidKeyHook) != address(0)) {
+      isValid = onValidKeyHook.hasValidKey(
+        address(this),
+        _keyOwner,
+        keyByOwner[_keyOwner].expirationTimestamp,
+        isValid
+      );
+    }    
   }
```
julien51, line 163: Are we missing a return statement?

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -178,6 +178,7 @@ interface IPublicLock
   function setEventHooks(
     address _onKeyPurchaseHook,
     address _onKeyCancelHook,
+    address _onValidKeyHook,
```
julien51, line 181: Let's be consistent in the naming 

```suggestion
    address _onHasValidKeyHook,
```

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -311,6 +312,8 @@ interface IPublicLock
 
   function onKeyCancelHook() external view returns(address);
   
+  function onValidKeyHook() external view returns(bool);
```
julien51, line 315:
```suggestion
  function onHasValidKeyHook() external view returns(bool);
```

smart-contracts/contracts/test-artifacts/TestEventHooks.sol
```diff
@@ -85,6 +85,32 @@ contract TestEventHooks is ILockKeyPurchaseHook, ILockKeyCancelHook, ILockTokenU
...
+
+  function hasValidKey(
+    address _lockAddress,
+    address _from,
+    uint, // removed var name to silence "unused" warning
+    bool _hasValidKey
+  ) external view
+    returns (bool isValidKey)
+  { 
+    // special members should always have access to content, even when the key is expired
```
julien51, line 106:
```suggestion
    // special members should always have access to content, even when the key is expired or they don't have one
```

julien51 (approved):

smart-contracts/contracts/mixins/MixinKeys.sol
clemsos: nope. It turns out that Solidity will declare and return the variable named in the `returns` declaration, so the `returns (bool isValid)` is enough for to return it. I have seen it used elsewhere in the Unlock contracts already

julien51:
Let's do this! Feel free to merge when ready :D

Issue #6088, opened 2020-02-27
Allow callbacks to Unlock.sol to fail


NickCuso:
**Describe the bug**
Unlock.sol could be upgraded to disable a lock today.

**Expected behavior**
Allow the call to fail.  This means both do not revert if Unlock reverts and cap the gas so it does not force the tx to fail.

nfurfaro:
I'm not sure why this failed yet. I have tried to reproduce the failure locally with no luck. The test passes as it's supposed to.

NickCuso:
hmm i think you meant to post this under https://github.com/unlock-protocol/unlock/issues/6283

julien51:
This is done

PR #7914, opened 2021-12-14
Try/catch on Unlock callbacks

### LINKED ISSUE #7885: try/catch on Unlock callbacks
The PublicLock contract currently calls back to Unlock when a [new purchase is made](https://github.com/unlock-protocol/unlock/blob/master/smart-contracts/contracts/mixins/MixinPurchase.sol#L89).

I believe this creates a "dependency" on the Unlock contract that I think can be reduced by wrapping this in a `try`/`catch` block so that if it fails, the key purchase is still successful.

When caught we should emit an event to indicate/surface the error. 

Additionally for now we should remove the call to compute the discount as this would save gas.

clemsos:
# Description

Locks could start failing by changes in Unlock, as every lock call back Unlock when a key is purchased. By wrapping Unlock calls in try/catch we make sure that locks stand on their own and can live unaffected by changes in Unlock contract

# Issues

Fixes #7885 and #6088
Refs #

# Checklist:


smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -177,9 +189,17 @@ contract MixinPurchase is
...
+      try unlockProtocol.computeAvailableDiscountFor(_recipient, minKeyPrice) 
+      returns (uint unlockDiscount, uint unlockTokens)
+      {
+        require(unlockDiscount <= minKeyPrice, 'INVALID_DISCOUNT_FROM_UNLOCK');
+        minKeyPrice -= unlockDiscount;
+      } catch {
+        // fail silently?
+        // adding an event will require writing to state
+        // also why wasnt this function returning anything?
+        // emit MissingUnlock(address(this), address(unlockProtocol));
```
clemsos, line 201: firing an event here will require to remove the `view` modifier from  the function - and likely affects gas. Maybe better to wrap all calls to `recordKeyPurchase` instead?

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -98,14 +100,24 @@ contract MixinPurchase is
       emit RenewKeyPurchase(_recipient, newTimeStamp);
     }
 
+    
     (uint inMemoryKeyPrice, uint discount, uint tokens) = _purchasePriceFor(_recipient, _referrer, _data);
     if (discount > 0)
     {
-      unlockProtocol.recordConsumedDiscount(discount, tokens);
+      try unlockProtocol.recordConsumedDiscount(discount, tokens) 
+      {}
```
julien51, line 108: Hum? What's this? Some Solidity feature that I don't know about?

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -21,6 +21,8 @@ contract MixinPurchase is
   event RenewKeyPurchase(address indexed owner, uint newExpiration);
 
   event GasRefunded(address indexed receiver, uint refundedAmount, address tokenAddress);
+  
+  event MissingUnlock(address indexed lockAddress, address unlockAddress);
```
julien51, line 25: Let's use a better name? Something like "UnlockCallFail"?

smart-contracts/contracts/mixins/MixinPurchase.sol
julien51: Indeed, I think we would need to wrap more in `recordKeyPurchase`... but also I think for now we can actually remove this `computeAvailableDiscountFor` completely since we don't support it and that locks will be upgradable so we can easily add it back later on (if we need it).
Wdyt?

smart-contracts/contracts/mixins/MixinPurchase.sol
julien51: Also, I think we can cleanup `recordConsumedDiscount` because it is not used... so let's simply remove it for now? (that'll save gas!)

julien51 (commented):
Nicely done!
I think we can cleanup the `recordConsumedDiscount` and `computeAvailableDiscountFor` for now. This will reduce gas for things that are not used anyway?

Wdyt?

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: yes, lets removed them

smart-contracts/contracts/Unlock.sol
```diff
@@ -259,25 +259,6 @@ contract Unlock is
...
   }
 
-  /**
-   * This function returns the discount available for a user, when purchasing a
-   * a key from a lock.
-   * This does not modify the state. It returns both the discount and the number of tokens
-   * consumed to grant that discount.
-   * TODO: actually implement this.
-   */
-  function computeAvailableDiscountFor(
```
julien51, line 269: I think we need to keep it, but just not call it anymore from the latest version of Unlock, otherwise "old" locks will break :/

julien51: Let's make sure we add a comment to show it as deprecated but kept for backward compat

smart-contracts/contracts/Unlock.sol
```diff
@@ -365,24 +346,6 @@ contract Unlock is
     }
   }
 
-  /**
-   * This function will keep track of consumed discounts by a given user.
-   * It will also grant discount tokens to the creator who is granting the discount based on the
-   * amount of discount and compensation rate.
-   * This function is invoked by a previously deployed lock only.
-   */
-  function recordConsumedDiscount(
```
julien51, line 374: Same as above

julien51 (approved):
Sorry I was not very clear. We need to keep the functions in Unlock.sol, but remove the calls to them from the PublicLock contract's latest version because if we removed them from Unlock then we'd break the old locks.

clemsos:
> Sorry I was not very clear. We need to keep the functions in Unlock.sol, but remove the calls to them from the PublicLock contract's latest version because if we removed them from Unlock then we'd break the old locks.

yes... silly me. I was wondering why the upgrade tests were failing. Will ad back

smart-contracts/contracts/Unlock.sol
```diff
@@ -259,6 +259,26 @@ contract Unlock is
...
+   */
+  function computeAvailableDiscountFor(
+    address /* _purchaser */,
+    uint /* _keyPrice */
+  )
+    public
+    pure
+    returns (uint discount, uint tokens)
+  {
+    // TODO: implement me
```
julien51, line 278: Let's remove the comment.

smart-contracts/contracts/Unlock.sol
```diff
@@ -346,6 +366,26 @@ contract Unlock is
...
+   * This function is invoked by a previously deployed lock only.
+   */
+  function recordConsumedDiscount(
+    uint _discount,
+    uint /* _tokens */
+  )
+    public
+    onlyFromDeployedLock()
+  {
+    // TODO: implement me
```
julien51, line 384: Let's remove the comment.

smart-contracts/contracts/Unlock.sol
```diff
@@ -346,6 +366,26 @@ contract Unlock is
...
+    totalDiscountGranted += _discount;
+    return;
+  }
+
```
julien51, line 388: Actually let's not do that because we don't need to keep track of that for now. (saves gas!)

julien51 (approved):
Small cleanups and then LG!

julien51 (approved):

PR #7915, opened 2021-12-14
Set max number of keys

### LINKED ISSUE #7899: Increase max number of keys
We should let the lock manager of a lock increase the max number of keys. I believe the current arbitrary limitation does not actually matter much because a lock manager/creator could easily create another lock to circumvent the limit.

clemsos:
# Description

A simple function to set the max number of keys

# Issues

Fixes #7899
Refs #

# Checklist:


julien51 (approved):
💯

PR #7927, opened 2021-12-17
Non expiring memberships

### LINKED ISSUE #7430: Non expiring memberships
Several projects/teams and creators have asked if it would make sense to create non-expiring memberships.
I think we should explore that. This creates challenges around transfers (which cannot be partial) or cancellations but after that, but I wonder if there are other gotchas!

@clemsos Would you be able to do a bit of research on the impact that this would have on our lock contracts?

clemsos:
# Description

Add support for non-expiring memberships. 

To create a non-expiring membership, just pass `0` as `_expirationDuration` during the creation of the lock.  
NB: the lock `expirationDuration` will be set to ` type(uint).max`

# Issues

Fixes #7430
Refs #

# Checklist:


clemsos:
On a sidenote, it seems like the `expirationDuration` is not something that can be changed. Maybe it will be good to allow lock managers to modify it using a simple setter ?

julien51:
>On a sidenote, it seems like the expirationDuration is not something that can be changed. Maybe it will be good to allow lock managers to modify it using a simple setter ?

That's an interesting idea. I wonder about the artifacts of that: what happens to refunds.... etc? But I think it is probably ok (same kind of change as a price change), so please maybe add a new PR for this?

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -73,7 +73,12 @@ contract MixinPurchase is
       // refresh the cached value
       idTo = toKey.tokenId;
       _recordOwner(_recipient, idTo);
-      newTimeStamp = block.timestamp + expirationDuration;
+      // check for a non-expiring key
+      if (expirationDuration == type(uint).max) {
```
julien51, line 77: Hum I thought you set `0` for the unlimited duration keys?

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -83,13 +88,20 @@ contract MixinPurchase is
         idTo
       );
     } else if (toKey.expirationTimestamp > block.timestamp) {
+      // prevent re-purchase of a valid non-expiring key
+      require(expirationDuration != type(uint).max, 'A valid non-expiring key can not be purchased twice');
```
julien51, line 92: actually now that I think about it I think we should block the double purchase only if the existing key is set to expire in maxuint.
So basically we should allow double purchase if/when the existing key expires in the future (or has expired) but is less than maxuint. This would cover cases where the duration has been changed to maxuint. wdyt?

clemsos: yes good

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: nope, the non-expiring keys are using the same process, so they are set from `expirationDuration` which is `type(uint).max`. That way we continue to check validity the same way i.e. `block.timestamp < key.expirationTimestamp`

clemsos:
Hmm can not reproduce that failing test locally so far...

clemsos:
ok so it turns out this fails because the original test had a mistake.

https://github.com/unlock-protocol/unlock/blob/b3588227d5cdd2968158044de2ee54ad6cc7851a/smart-contracts/test/Lock/timeMachine.js#L76-L84

the `timestampAfter` value is derived from Solidity's `block.timestamp` which is expressed [in seconds](https://docs.soliditylang.org/en/latest/units-and-global-variables.html?highlight=block#block-and-transaction-properties) while js `Date.now()` is [in ms](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Date/now). 

So if we correct the `/ 1000` operator then the test fails. 

The formulation os what is tested is not really clear and I am not so sure what is the expected behaviour here.

clemsos:
There is an upgrade test for PublicLockV6 that is failing because of a call that ERC1820 that seems to not support the ts modification we are doing in non-expiring test suite, namely this line

https://github.com/unlock-protocol/unlock/blob/462ac0e8496cac359c63712e7b2ac343d5ce5a80/packages/contracts/src/contracts/PublicLock/PublicLockV6.sol#L2376

clemsos:
> There is an upgrade test for PublicLockV6 that is failing because of a call that ERC1820 that seems to not support the ts modification we are doing in non-expiring test suite, namely this line

silly me was resetting the chain so ERC1820 was deleted... This was sth I attempted to reset time at block levels but that part didnt work out. It did wiped out all deployed contracts though (including ERC1820).

julien51 (approved):
LGTM!

Merci!

PR #7928, opened 2021-12-17
Add storage gaps to PublicLock

### LINKED ISSUE #7919: Adding `__gap` before upgradable locks
Since we're moving forward with upgradable locks, we need to make our mixins "future proof". OpenZeppelin uses `__gap` variables at the end of each contract so that we can add/change storage inside of the mixins without completely changing the structure of these contracts when we want to upgrade them.

All mixins used in `PublicLock` should have a `__gap` method of say 1,000 uint256 slots.

See details on https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0d40f705a7d4a42ff622ae3a0e1a90305fc5b93e/docs/modules/ROOT/pages/upgradeable.adoc#storage-gaps

clemsos:
# Description

For the safety of future upgrades on `PublicLock` contract, we add [storage gaps](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0d40f705a7d4a42ff622ae3a0e1a90305fc5b93e/docs/modules/ROOT/pages/upgradeable.adoc#storage-gaps) at the end of our mixins.

# Issues

Fixes #7919
Refs #

# Checklist:


julien51 (approved):
LGTM!

clemsos:
It seems like 1000 is a bit excessive - it will mean adding 1000 variables per mixin... Not sure how much overhead it adds -- see https://github.com/openzeppelin/openzeppelin-contracts-upgradeable/issues/47

clemsos:
Thank you very much @Amxx for the help 🙏🏻  
We'll stick to 1000 then :)

julien51:
Merci @Amxx !

frangio:
I actually think using too big a gap may add to deployment cost (but not runtime cost). The reason is that a bigger slot number requires more bytecode to push onto the stack. Whether it's 1000 or 50 may not make a big difference though because it will likely be above 1 byte anyway.

clemsos:
Thanks @frangio ! Makes sense about deployment costs. It won't be really an issue for us, as we just intend to deploy the template once -- creating new locks will just deploy a proxy using the template as implementaion.

PR #7935, opened 2021-12-21
Fix mistake in time machine test

clemsos:
# Description

As discovered in #7927 , the current test for `_timeMachine` had a small mistake.

https://github.com/unlock-protocol/unlock/blob/b3588227d5cdd2968158044de2ee54ad6cc7851a/smart-contracts/test/Lock/timeMachine.js#L76-L84

the `timestampAfter` value is derived from Solidity's `block.timestamp` which is expressed [in seconds](https://docs.soliditylang.org/en/latest/units-and-global-variables.html?highlight=block#block-and-transaction-properties) while js `Date.now()` is [in ms](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Date/now). If we correct ` Date.now() / 1000` then the test fails (as demonstrated here).

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/test/Lock/timeMachine.js
+++ b/smart-contracts/test/Lock/timeMachine.js
@@ -1,3 +1,4 @@
+const { ethers } = require('hardhat')
 const BigNumber = require('bignumber.js')
 
 const unlockContract = artifacts.require('Unlock.sol')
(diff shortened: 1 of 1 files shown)


julien51 (approved):

julien51:
Lgtm

PR #7980, opened 2022-01-04
add createUpgradeableLock

### LINKED ISSUE #7884: createUpgradableLock instead of createLock
We have recently changed the signature of `createLock` to support upgradable locks.
I believe this is not a good idea because other contracts might be using `createLock` "as it is" and might break because of that change.
So I propose the following: 

1. We restore the `createLock` as it was and that would deploy non-upgradable versions. Additionally, we emit an event `DEPRECATED` to indicate that this method should NOT be used (ideally we'd find a way to point to the right doc in the code...
2. We create a`createUpgradableLock` that actually deploys upgradable locks! We should make sure this function also takes the actual version number because the content of the `data` will change based on the version and we want 3rd party contracts/apps to "keep working" even when we introduce new versions. It is ok for them to deploy "older" versions as we will have the upgrade path anyway...

**Additional context**

clemsos:
# Description

Add `createLock` legacy signature back into the Unlock contract and add `createUpgradeableLock`

# Issues

Fixes #7884
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/Unlock.sol
+++ b/smart-contracts/contracts/Unlock.sol
@@ -193,7 +193,7 @@ contract Unlock is
   }
 
   /**
-  * @notice Create lock
+  * @notice Create lock (legacy)
   * This deploys a lock for a creator. It also keeps track of the deployed lock.
   * @param _expirationDuration the duration of the lock (pass 0 for unlimited duration)
   * @param _tokenAddress set to the ERC20 token address, or 0 for ETH.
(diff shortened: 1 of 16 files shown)


smart-contracts/contracts/Unlock.sol
```diff
@@ -195,6 +195,35 @@ contract Unlock is
   /**
   * @notice Create lock
   * This deploys a lock for a creator. It also keeps track of the deployed lock.
+  * @param _tokenAddress set to the ERC20 token address, or 0 for ETH.
```
julien51, line 198: Remove or add the other params please :)

julien51 (approved):

PR #8003, opened 2022-01-05
feat(unlock-app): Grant keys

julien51:
# Description

This PR adds the ability to grant keys to a lock from the UI. The granter needs to be a lock manager and/or a key manager to be able to grant these keys and can grant them one by one.
I would love to get your feedback as I iron things out!

https://share.getcloudapp.com/rRuNnQe4

# Checklist:


julien51:
For now, the button shows for everyone which is not ideal. I will see if we can only show it for folks who are lock manager or key granters.
Additionally I want to add support for ENS resolution.

searchableguy:
Should we use https://www.npmjs.com/package/@unstoppabledomains/resolution for resolving ENS addresses?

julien51:
We already do ENS resolution in the [app with ethers!](https://github.com/unlock-protocol/unlock/blob/b95f61390ae6a6e28b0bb043d9edab2b34532c9a/unlock-app/src/hooks/useEns.js) 

I need to think about the UI...

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,274 @@
...
+  setIsOpen: (isOpen: boolean) => void
+  lockAddresses: string[]
+}
+
+export const GrantKeysDrawer = ({
+  isOpen,
+  setIsOpen,
+  lockAddresses,
+}: GrantKeysDrawerInterface) => {
+  const { network } = useContext(AuthenticationContext) // TODO: use the actual lock network instead of the currently connected network
```
clemsos, line 214: seems the same case in `GrantKeyForm`?

julien51: Yup.

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,274 @@
...
+/**
+ * Form part
+ * @returns
+ */
+const GrantKeyForm = ({ onGranted, lock }: GrantKeyFormProps) => {
+  const { account, network } = useContext(AuthenticationContext)
+  const config: any = useContext(ConfigContext)
+
+  const walletService = useContext(WalletServiceContext)
+  const web3Service = useContext(Web3ServiceContext)
```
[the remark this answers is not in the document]
julien51: I think this is a good pattern and probably worth doing as part of a much larger rewrite?

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+  const [transaction, setTransaction] = useState<string>('')
+  const [loading, setLoading] = useState(false)
```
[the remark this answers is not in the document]
julien51: That's a good idea. But there again I would want to make it a more "global" pattern, like:

```javascript
const {setLoading, setReady, setError, isLoading, isReady, isError} = useComponentState()
```

Does it make sense?

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,274 @@
...
+      const existingExpiration =
+        await web3Service.getKeyExpirationByLockForOwner(
+          lock.address,
+          recipient,
+          network
+        )
+      if (existingExpiration > new Date().getTime() / 1000) {
+        alert(
+          'This address already owns a valid key. You cannot grant them a new one.'
+        )
```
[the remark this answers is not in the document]
julien51: You know what, that's a great idea. I am going to add am `alert` component that we can use instead of the native ones. I'll open another PR and use it there.

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,274 @@
...
+          )}
+
+          {!errors.keyManager && (
+            <p className="text-xs italic -mt-4">
+              If set the key manager has the transfer and cancellation rights
+              for the recipient&apos;s key. If you leave empty, your address
+              will be set as manager.
+            </p>
+          )}
+        </div>
```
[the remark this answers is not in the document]
julien51: Yes. I think we need much better `Input` component. Probably worth thinking about normalizing their behavior across the app?

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,274 @@
...
+              rel="noopener noreferrer"
+              href={config.networks[network].explorer.urls.transaction(
+                transaction
+              )}
+            >
+              ↗
+            </a>
+          )}
+        </LoadingButton>
+      )}
```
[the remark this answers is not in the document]
julien51: Yes!

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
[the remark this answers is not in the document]
searchableguy: Yup

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,274 @@
...
+      alert(
+        'There was an error and the transaction could not be sent. Please refresh the page and try again.'
+      )
+      setTransaction('')
+    }
+    setLoading(false)
+  }
+
+  return (
+    <form onSubmit={handleSubmit(onSubmit)} className="w-full max-w-lg">
```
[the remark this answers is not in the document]
julien51: At this point our dashboard is actually desktop only which is not ideal...

Thomasb1986 (approved):

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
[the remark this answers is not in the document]
julien51: Built and replaced!

julien51:
Ok, ready for a final review @searchableguy ! Thanks!

julien51:
Alright I think it is ready for a final look @searchableguy @clemsos ! Thanks!

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,316 @@
...
+        title: 'Error',
+        body: 'There was an error and the transaction could not be sent. Please refresh the page and try again.',
+      })
+      setTransaction('')
+    }
+    setLoading(false)
+  }
+
+  const addressFieldChanged = (name: string) => {
+    return async (event: any) => {
```
searchableguy, line 136: You can type using React types. React.ChangeEvent<HTMLInputElement> or something.

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,316 @@
...
+      const address = await getAddressForName(event.target.value)
+      if (address) {
+        // @ts-expect-error
```
searchableguy, line 139: Why the ts-expect error?

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,316 @@
...
+              network
+            )
+          }
+        })
+      )
+      setLock(locks[lockAddresses[0]])
+      setLocks(locks)
+    }
+    loadLocks(lockAddresses)
+  }, [lockAddresses.join('')])
```
searchableguy, line 267: I kind of feel like this could be a separate hook of its own. I think this is reusable but for now, it's okay.

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+  const handleLockChanged = (evt: any) => {
```
searchableguy, line 269: again type the events whenever you can.

unlock-app/src/hooks/useEns.js
```diff
@@ -1,38 +1,52 @@
 import { ethers } from 'ethers'
 import { useState, useEffect, useContext } from 'react'
 import { AuthenticationContext } from '../contexts/AuthenticationContext'
+import configure from '../config'
+
+const config = configure()
+
+export const getNameOrAddressForAddress = async (address) => {
```
searchableguy, line 8: Type the address please? It makes it easy to use these functions elsewhere when you know what input and output they expect.

unlock-app/src/hooks/useEns.js
```diff
@@ -1,38 +1,52 @@
...
+    }
+    return address
+  } catch (error) {
+    // Resolution failed. So be it, we'll show the 0x address
+    console.error(`We could not resolve ENS for ${address}`)
+    return address
+  }
+}
+
+export const getAddressForName = async (name) => {
```
searchableguy, line 24: same, type the name

unlock-app/src/unlockTypes.ts
```diff
@@ -110,6 +110,7 @@ export enum KeyStatus {
 }
 
 export interface Lock {
+  canGrant?: any
```
searchableguy, line 113: Typey typey please

unlock-app/src/components/interface/checkout/FormStyles.tsx
```diff
@@ -96,6 +97,33 @@ export const LoadingButton = ({ children, ...props }: LoadingButtonProps) => (
   </Button>
 )
 
+interface TransactionPendingButtonProps {
+  network: number
+  transaction: any
```
searchableguy, line 102: why is transaction an any? 

I like the refactor though. It could more generalized.

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
```diff
@@ -0,0 +1,316 @@
...
+import Alert from '../../interface/Alert'
+
+interface GrantKeyFormProps {
+  lock: Lock
+  onGranted: (granted: boolean) => void
+}
+
+// Prevents re-rendering when time changes!
+const now = new Date().getTime()
+
```
searchableguy, line 28: Can you please explain this part a little more? 

The way I see it is, this will only run once when the component file is loaded and stick to the same now. So this won't change regardless of the component mount and unmount. Am I correct? 

I feel this is an unnecessary optimization and could be made better.

searchableguy (changes requested):
Just few more changes and then it's good :smile: Let me know if I'm being too picky or demanding.

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
julien51: Because it complains about the type :/ 

```
Argument of type 'string' is not assignable to parameter of type '"recipient" | "expiration" | "keyManager"'.ts
```
I am not sure what the best way but I fixed it with 

```
return setValue(
          name as 'recipient' | 'expiration' | 'keyManager',
          address,
          {
            shouldValidate: true,
            shouldDirty: true,
          }
        )
```

unlock-app/src/components/interface/checkout/FormStyles.tsx
julien51: Fixed type!

I think it could be a great addition to the `ui` package eventually?

unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
julien51: You are correct. The challenge is that if I put it in the lifecycle of the component then of course it changes all the time (and creates perpetual re-renders). The solution would be to put it in a `useEffect` hook but that seems pretty verbose for something not really that critical, because it in the end the user will pick an expiration in "days" more than seconds anyway? so the most precise time does not matter that much?

julien51:
> Just few more changes and then it's good 😄 Let me know if I'm being too picky or demanding.

No that's good! I need to be better :)
I was thinking maybe we could have stricter rules about types, but I wonder how they are going to break the existing code..,. and I don't want to spend too much time adding "past" types, so doing this incrementally is probably the right approach!

PR #8008, opened 2022-01-07
Fix Unlock v10 storage layout compatibility

clemsos:
# Description

When attempting an upgrade of Unlock on Rinkeby against the previous version of the following errors was thrown by the OZ upgrades lib.

```
╰─$ yarn hardhat upgrade:prepare --contract contracts/Unlock.sol --proxy 0xd8c88be5e8eb88e38e6ff5ce186d764676012b0b --network rinkeby                                                                                   1 

Deploying new implementation contracts/Unlock.sol on rinkeby.
An unexpected error occurred:

Error: New storage layout is incompatible

@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol:39: Renamed `initialized` to `_initialized`

@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol:44: Renamed `initializing` to `_initializing`

@openzeppelin/contracts-upgradeable/utils/ContextUpgradeable.sol:31: Renamed `______gap` to `__gap`
```

The reason for this is we upgraded the OZ libs when moving to Solidity v8 ([here](https://github.com/unlock-protocol/unlock/blame/d13ae8515b10b2414995a8e4e8fb73b2ffc0a278/smart-contracts/contracts/Unlock.sol#L30)) and the variables were  renamed in the new OZ version. We were using the now deprec (`@openzeppelin/upgrades`) lib.

This PR includes a fix, which consists on checking the relevant OZ contracts directly in our repo (`ContextUpgradeable.sol` and `Initializable.sol`) and revert back the variables names so the layout is compatible again.

### DIFF
--- a/smart-contracts/contracts/Unlock.sol
+++ b/smart-contracts/contracts/Unlock.sol
@@ -27,18 +27,18 @@ pragma solidity ^0.8.2;
  *  b. Keeping track of GNP
  */
 
-import '@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol';
 import '@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol';
 import '@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol';
 import 'hardlydifficult-eth/contracts/protocols/Uniswap/IUniswapOracle.sol';
 import './utils/UnlockOwnable.sol';
+import './utils/UnlockInitializable.sol';
 import './interfaces/IPublicLock.sol';
 import './interfaces/IMintableERC20.sol';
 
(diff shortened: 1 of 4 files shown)


clemsos:
Hey @Amxx @frangio ! 

We were having an issue with compatibility in variable names when upgrading from the old `@openzeppelin/upgrades` to `@openzeppelin/contracts-upgradeable`, so I made a (naive?) fix : checking in the files in the repo and modifying what the problematic variables.

Does that looks legit to you? Will there be a better approach?

Many thanks again for all the help 🙏🏻🙏🏻

julien51 (approved):
LGTM

PR #8014, opened 2022-01-10
feat(websub): Add websub controller and helper methods

### LINKED ISSUE #7316: Async job to identify Unlock transactions
One of the building blocks which we miss right now is the ability to trigger things when:
1. new locks are deployed
2. new memberships are purchased

We need to build a basic service that would leverage our hosts (Heroku) cron job feature and our subgraph to extract these things. The job should be able to ping a 3rd party service such as zappier with details so that we can fan-out to other applications such as a Twitter bot or our Discord... or Zapier.

## Registring hooks:

For this, we will use an API based on the [WebSub W3C recommendation](https://www.w3.org/TR/websub/), where we are hosting our own "hub" at `/api/1/hooks/`. (`1` is the current API version)
3rd party application will be able to register hooks on these 2 endoints:
* `/api/1/hooks/<network>/locks` for new locks deployed where `network` is the network on which the lock is deployed. `*` can be used to register a hook on locks being deployed on any network.
* `/api/1/hooks/<network>/locks/<lock address>/keys` for new keys being purchased on the lock deployed on `network` at `lock address`.

We will add more endpoints later on, for example to listen on lock property changes or key property changes.

## Triggering hooks

We will run a cronjob that runs every 1 minute. The job's workflow is this:
* for every supported network (see @unlock-protocol/networks), we will list all of the locks deployed within a (configurable) time frame of 1 hour to get started, using our subgraphs.
* for every lock in that list, prepare an update to be posted as a JSON object of lock (using the subgraph's schema). 
* identify all registered hooks
* check whether the hook has already been successfuly notified for that specific update in a dedicated `hook_triggers` table (that needs to be added)
* if the notification already happened, skip
* if the notification did not happen, perform the WebSub notification. 
* if the notification succeeds, store that state in the table.
* if the notification failed, store it in the table, but with a "failed" status and an indication of the "attempt" number. 
* drop any notification that has failed more than x times (x should be configurable but defaults to 5).

(This is WIP, please ask any clarifying question!)

searchableguy:
# Description

Controller for websub subscription 

# Checklist:


searchableguy:
One thing I'm confused about is how to actually handle the topic. 

The spec mandates that the user must provide a topic URL but if the user is directly making subscription request to the hook endpoint with valid params, then that *becomes* the topic. 

Should we have a single hub point for registering subscriptions for all our topic endpoints like for listening to lock addresses or key changes?

locksmith/src/controllers/hookController.ts
```diff
@@ -0,0 +1,131 @@
+// eslint-disable-next-line import/no-unresolved
+import { Response } from 'express-serve-static-core'
+import { networks } from '@unlock-protocol/networks'
+import * as z from 'zod'
```
julien51, line 4: Very interesting! I did not know about zod!

locksmith/src/controllers/hookController.ts
```diff
@@ -0,0 +1,131 @@
...
+      logger.error(
+        `Subscriber failed to confirm intent for callback: ${hub.callback} on topic: ${hub.topic}.`
+      )
+      return
+    }
+    const updatedHook = await hook.save()
+    return updatedHook
+  }
+
+  return hook
```
julien51, line 72: Should we add a case for when a hook already exists but is just extended?

julien51: In that case, we should also reconfirm intent BTW, and change the secret if needed.

julien51 (commented):
I think it would be worth writing a few test scenarios:
* new susbcription
* renewing/expanding
* unsubscribing
etc

locksmith/src/controllers/hookController.ts
searchableguy: Maybe I will just let people update all values instead of just mode :thinking: instead.

julien51: Let's remain compliant with websub please :)

searchableguy: okay, will add a case for increasing expiration date.

searchableguy: Actually, let's leave this for later addition.

locksmith/src/controllers/hookController.ts
searchableguy: Yeah, it's a good way to declare schema which you can enforce/parse for as well as infer the type for it.

locksmith/src/controllers/hookController.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Check if the subscriber request is valid or not. This will post a challenge to subscriber to confirm the intent of the request.
+export async function isValidHubIntent(hub: z.infer<typeof Hub>) {
```
julien51, line 74: This needs to happen asynchronously per https://www.w3.org/TR/websub/#x5-1-2-subscription-response-details

locksmith/src/controllers/hookController.ts
```diff
@@ -0,0 +1,131 @@
...
+    return
+  }
+  return networks[network]
+}
+
+export async function subscriptionHandler(req: Request, res: Response) {
+  try {
+    const hub = await Hub.parseAsync(req.body.hub)
+    const hook = await subscribe(hub, req.params)
+    return res.status(200).json(hook?.toJSON())
```
julien51, line 127: Let's make sure we follow the spec https://www.w3.org/TR/websub/#x5-1-2-subscription-response-details

julien51:
I think we should test the compliance with the websub spec here.

locksmith/src/controllers/hookController.ts
searchableguy: I'm confused what do you mean here by happening async. I went through the spec too and I still have no idea.

searchableguy:
I was hoping to get something working and then making it more compliant by testing against the webhubrocks compliance suite.

There aren't many implementation of websub in practice though to take clues from. Most of them are unmaintained or not proper.

locksmith/src/controllers/hookController.ts
searchableguy: Okay, I got this should happen after responding to the subscription request with accepted or denied.

searchableguy: fixed

searchableguy:
should be good to go!

locksmith/src/controllers/hookController.ts
julien51: Hum. Please make sure you keep track of it?

locksmith/src/controllers/hookController.ts
```diff
@@ -0,0 +1,146 @@
...
+    // We check the hub schema here to make sure the subscriber is sending us the correct data.
+    const hub = await Hub.parseAsync(req.body.hub)
+    // We check the network here to make sure the subscriber is sending to the right network endpoint.
+    const network = getSupportedNetwork(req.params.network)
+    if (!network) {
+      res.status(400).send('Unsupported network')
+      return
+    }
+    // We send the accepted request to the subscriber and then validate the intent of the subscriber as well as persist the subscription.
+    res.status(202).send('Accepted')
```
julien51, line 134: 👍

locksmith/src/controllers/hookController.ts
```diff
@@ -34,12 +38,8 @@ export async function subscribe(
...
       throw intentError
     }
-
-    const expiration = new Date()
-    expiration.setSeconds(
-      expiration.getSeconds() + (hub.lease_seconds ?? 60 * 60 * 24 * 30)
-    )
-
+    // 10 days by default
+    const expiration = getExpiration(hub.lease_seconds || 864000)
```
julien51, line 42: Let's move the `|| 864000` inside `getExpiration` and also make sure that the actual expiration is at most 90 days.

locksmith/src/controllers/hookController.ts
```diff
@@ -53,15 +53,24 @@ export async function subscribe(
     return createdHook
   }
 
-  // If changing mode, reconfirm the intent and update
-  if (hook?.mode !== hub.mode) {
+  // If changing mode or lease, reconfirm the intent and update
+  if (hook?.mode !== hub.mode || hub.lease_seconds) {
```
julien51, line 57: They can also update the secret... (important to support that for security reasons...)

locksmith/src/controllers/hookController.ts
```diff
@@ -53,15 +53,24 @@ export async function subscribe(
...
     hook.mode = hub.mode
+
+    const expiration = hub.lease_seconds
+      ? getExpiration(hub.lease_seconds)
```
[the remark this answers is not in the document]
searchableguy: that would cause the expiration to change every time someone makes any change though. Should we allow changing expiration on mode to "unsubscribe" too?

locksmith/src/controllers/hookController.ts
searchableguy: okay, will modify and add a check for extending expiration.

locksmith/src/controllers/hookController.ts
searchableguy: okay will add that.

locksmith/src/controllers/hookController.ts
```diff
@@ -0,0 +1,203 @@
...
+      hub: {
+        mode: req.body.hub.mode,
+        topic: req.body.hub.topic,
+        reason: error.message,
+      },
+    })
+  }
+}
+
+export function publisherHandler(req: Request, res: Response) {
```
searchableguy, line 162: One thing though, should we verify if the lock address they are trying to get is valid. It would be a request to the subgraph and throw if the lock param is present.

julien51: We could eventually but I think for now it is ok to go without it.

julien51 (approved):

PR #8012, opened 2022-01-10
Separate Unlock/PublicLock versions logic

clemsos:
# Description

This is a continuation of #8006, separating further the versioning logic of Unlock and PublicLock in the `unlock-js` lib.

These are only internal changes that aims at facilitating support for future versions, such as Unlock v10 (see #7941)

# Issues

Fixes #
Refs #

# Checklist:

- [ ] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #8017, opened 2022-01-10
feat(unlock-js): add new Unlock v10 upgradeable logic

### LINKED ISSUE #7941: UnlockJS upgrade
As we are getting ready to ship the newest version of `Unlock` and `PublicLock`, it is important that we add support for these in `unlock-js`.

clemsos:
# Description

This adds Unlock V10 function to deploy upgradeable contracts and upgrade them

rebased ~~NB: to be rebased once #8012 is merged.~~

# Issues

Fixes #7941
Refs #

# Checklist:


clemsos:
here are the function added in v10:

```
╰─$ diff IUnlockV9.sol IUnlockV10.sol | grep function                                                                                                                        1 ↵
>   function initializeProxyAdmin() external;
>   function proxyAdminAddress() external view;
>   * @dev this call is passed as encoded function - for instance:
>   function createUpgradeableLock(
>   function upgradeLock(
>    * @notice [DEPRECATED] Call to this function has been removed from PublicLock > v9.
>    * @notice [DEPRECATED] Call to this function has been removed from PublicLock > v9.
>   function addLockTemplate(address impl, uint16 version) external;
>   function publicLockImpls(uint16 _version) external view;
>   function publicLockVersions(address _impl) external view;
>   function publicLockLatestVersion() external view;
>   function __initializeOwnable(address sender) external;

```

Is there any other function we should add to unlock js?

julien51:
I think `publicLockLatestVersion` and `publicLockImpls` will be useful to show if there are available upgrades?

clemsos:
> I think `publicLockLatestVersion` and `publicLockImpls` will be useful to show if there are available upgrades?

Yep but they are simple getters. Isn't The js lib exposing the getters through abi already? Or do we need to add a specific method ?

julien51:
Ha! I don't think we expose the getters in UnlockJS, but that may actually be a good idea?

clemsos:
just realized in 97e5796 that the Lock test logic was flawned. Good thing is we were already testing Unlock against multiple Lock versions. The bad thing is tests are failing now - but that shouldnt be too much a problem to fix

packages/unlock-js/src/Unlock/v10/createUpgradeableLock.js
```diff
@@ -0,0 +1,109 @@
+import ethersUtils from '../../utils'
```
julien51, line 1: I actually wonder if we should just keep `createLock` that calls the contracts' `createUpgradeableLock`, because I don't think we should allow people to call non upgradable locks?

packages/unlock-js/src/Unlock/v10/index.js
```diff
@@ -0,0 +1,20 @@
+import abis from '../../abis'
+import createUpgradeableLock from './createUpgradeableLock'
+import upgradeLock from './upgradeLock'
+import { parseGetters } from '../../parser'
+
+import v9 from '../v9'
+
+const { createLock, configureUnlock } = v9
+
+const getters = parseGetters(abis.Unlock.v10.abi)
```
julien51, line 10: Hum, so I am not sure this is good enough because `unlock-js` does abstract away a lot of the solidity formats...

Maybe a better approach would be to expose the ethers contract classes in unlock-js letting anyone who needs it just access what they need thru the contracts?

julien51 (commented):
I think it could be easier to split a bit more because there are a lot of things going on here :/

packages/unlock-js/src/Unlock/v10/createUpgradeableLock.js
clemsos: > I think it could be easier to split a bit more because there are a lot of things going on here :/

Yep I splitted it in smaller PRs so its easier to review

clemsos: yep for most part it will be easier to just keep the `createLock` as it is. The main reason why I added the `createUpgradeableLock` is because there is a slight difference : we explicitly pass the `lockOwner` in params, while `createLock` uses `msg.sender`. 

I thought that was enough of a change to add it here - as that logic could be needed. 

We can also remove the `createUpgradeableLock` altogether ofc

julien51: Ok, let's just remove `createUpgradeableLock`.

julien51:
Please add a test for that one.

packages/unlock-js/scripts/compressAbi.js
```diff
@@ -13,22 +13,24 @@ const path = require('path')
 /* eslint-disable import/no-extraneous-dependencies */
 const abis = require('@unlock-protocol/contracts')
 
-const unlockVersions = ['v4', 'v6', 'v7', 'v8', 'v9']
+const unlockVersions = ['v4', 'v6', 'v7', 'v8', 'v9', 'v10']
 const publicLockVersions = ['v4', 'v6', 'v7', 'v8']
```
julien51, line 17: Hum, we should add tests for `publicLockVersions` `v9` as well, correct?

julien51 (approved):
🚀

packages/unlock-js/scripts/compressAbi.js
clemsos: actually we still don't have support for publicLock v9 in unlock-js, so thats another thing we should add

clemsos:
for latest fixes, merging/rebasing with #8049 will be useful

packages/unlock-js/scripts/compressAbi.js
clemsos: added v9 to #8050

clemsos:
rebase needed when #8059 will be merged

clemsos:
yet another rebase with #8066

PR #8029, opened 2022-01-12
add view/pure function to unlock v10 exports

clemsos:
# Description

We define a way to expose getters - i.e `view` and `pure` functions in solidity contracts - in the unlock-js interface.

This can be added once #8017 is merged

# Issues

Fixes #
Refs #

# Checklist:


clemsos:
from @julien51 

> Hum, so I am not sure this is good enough because `unlock-js` does abstract away a lot of the solidity formats...
>
> Maybe a better approach would be to expose the ethers contract classes in unlock-js letting anyone who needs it just access what they need thru the contracts?
>

clemsos:
closed as we can use directly the ethersjs contract that is already exposed

PR #8030, opened 2022-01-12
Unlock js test all PublicLock versions

clemsos:
# Description

Modifies the test logic to make sure we actually test all Unlock versions against all PublicLock versions.

This will increase the number of tests as we will have 4 unlock contracts * 4 lock templates * 3 different locks * and about 30-40 tests for each - which means a total of 1500+ tests

NB: need a rebase when #8034 is merged

# Issues

Fixes #
Refs #

# Checklist:


clemsos:
One of the main issue now is that we have to set the lock template but there is no `setLockTemplate` function in v4. I have commented out v4 for now but needs to bring it back. 

Also it is likely we can bring back v6 as there is erc1820 for hardhat

julien51:
Let's remove support for v6 locks since there is no "alive" lock in v6 anymore.

packages/unlock-js/src/PublicLock/v7/purchaseKey.js
```diff
@@ -1,135 +0,0 @@
-import utils from '../../utils'
```
julien51, line 1: good catch!

julien51 (approved):
Let's ship it!

PR #8032, opened 2022-01-13
feat(websub): Add websub server

searchableguy:
# Description

This adds our websub server. I haven't tested it yet so any feedback is very welcome. 

This implementation is very naive and first pass. I plan to add support for only sending diffs and trying failed hooks on server restarts in later iteration.

# Checklist:

### DIFF
--- a/locksmith/src/websub/jobs/locks.ts
+++ b/locksmith/src/websub/jobs/locks.ts
@@ -1,0 +1,21 @@
+import { Lock } from '../../graphql/datasource'
+import { Hook } from '../../models'
+import { TOPIC_LOCKS } from '../topics'
+
+import { networkMapToFnResult, notifyHook } from '../helpers'
+
+export async function notifyOfLocks(hooks: Hook[]) {
+  const subscribed = hooks.filter((hook) => {
+    const path = new URL(hook.topic).pathname
+    return TOPIC_LOCKS.test(path)
+  })
+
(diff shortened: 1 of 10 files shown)


locksmith/src/websub/jobs/locks.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export async function notifyOfLocks(hooks: Hook[]) {
+  const subscribed = hooks.filter((hook) => {
+    const path = new URL(hook.topic).pathname
+    return TOPIC_LOCKS.test(path)
+  })
+
+  const lockSource = new Lock()
+  const networkToLocksMap = await networkMapToFnResult((network) =>
+    lockSource.getLocks({ first: 25 }, Number(network))
```
julien51, line 15: i think we need to get rid of the 25 limit here?

julien51 (commented):
Good progress

locksmith/src/websub/jobs/locks.ts
julien51: Oh that's a good point actually. Let's keep it like this for now and think about better ways to handle this. I think keep track of a "last time checked" and using it + ⍙ can provide a good mechanism. Wdyt?

julien51 (approved):
In the interest of time let's approve and merge but I would love to see some tests for this as well!

PR #8042, opened 2022-01-17
fix(locksmith/graphql): Add network in the constructor for key and lock

searchableguy:
# Description

This fixes an issue where a method in our datasource class for key and lock was responsible for adding the base URL which meant other methods couldn't access the baseURL unless that method was invoked first. I've moved the network up as an argument in constructor instead so every method can access it and refactored the code to reflect this change.

# Checklist:

### DIFF
--- a/locksmith/__tests__/graphql/datasource/key.test.ts
+++ b/locksmith/__tests__/graphql/datasource/key.test.ts
@@ -1,0 +1,38 @@
+import { Key } from '../../../src/graphql/datasource'
+
+const TEST_KEY_ID = '0x0035807ec068737973d603a1f7e0fc1b12a78bf8-1'
+// Rinkeby
+const TEST_NETWORK = 4
+describe('Key', () => {
+  describe('getKey', () => {
+    describe('when data is returned for the data source', () => {
+      it('return the requested data', async () => {
+        expect.assertions(1)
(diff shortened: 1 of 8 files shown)


locksmith/__tests__/graphql/datasource/key.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const TEST_KEY_ID = '0x0035807ec068737973d603a1f7e0fc1b12a78bf8-1'
+// Rinkeby
+const TEST_NETWORK = 4
+describe('Key', () => {
+  describe('getKey', () => {
+    describe('when data is returned for the data source', () => {
+      it('return the requested data', async () => {
+        expect.assertions(1)
+        // Instead of mocking, we make request to a known key on rinkeby network
```
searchableguy, line 11: Let me know if you want to instead mock. I think actually simulating the request will be more robust as the mock responses can be anything regardless of what the function does and considering we don't have fully typed datasource classes, that is a high possibility.

searchableguy: If not, do let me know if I should pick lock and key address under unlock.

julien51: Agreed regarding mocking. My main concern with actual requests is latency though... 
I defer to what you think is best for now!

PR #8043, opened 2022-01-17
feat(websub): Add discord websub example

searchableguy:
# Description

This adds a discord webhook example. This will help us test the websub.

# Checklist:


searchableguy:
ugh, this one depends on websub routes PR which is getting lengthy. Let me try to push some more tests to that one so we can merge it.

searchableguy:
@julien51  should be good to go now.

julien51 (commented):
Since this is an example of application, let's make sure it can "stand alone", by adding a README, and its own package.json please. We will be able to use it as a reference implementation!

locksmith/src/websub/jobs/keys.ts
```diff
@@ -19,6 +19,6 @@ export async function notifyOfKeys(hooks: Hook[]) {
     const data = networkToLocksMap
       .get(Number(hook.network))
       .filter((key: any) => key.lock.id === hook.lock)
-    notifyHook(hook, data)
+    notifyHook(hook, { data })
```
julien51, line 22: Let's isolate these changes in another PR please.

julien51 (approved):
Small nit, but for consistency with other packages/folders, let's use `websub-discord` for the folder!

locksmith/examples/websub_discord/src/server.ts
```diff
@@ -0,0 +1,63 @@
...
+
+app.post('/callback/locks', websubMiddleware, async (req) => {
+  const embeds: MessageEmbed[] = []
+  const locks: any[] = req.body?.data
+  if (!locks.length) {
+    return
+  }
+  for (const lock of locks) {
+    const embed = new MessageEmbed()
+    embed.setTitle('New Lock')
```
julien51, line 26: I think it would be cool to add the network as well? So we know on what network this new lock is? And ideally a link to the block explorer for the network?

locksmith/examples/websub_discord/src/server.ts
```diff
@@ -0,0 +1,63 @@
...
+  const keys: any[] = req.body?.data
+
+  if (!keys.length) {
+    return
+  }
+
+  for (const key of keys) {
+    const embed = new MessageEmbed()
+    embed.setTitle('New Key')
+    embed.addField('Key Id', key.id)
```
julien51, line 49: Showing the lock and network + link to block explorer could be useful!

julien51 (approved):
Small changes but I am excited to try this out!

locksmith/examples/websub_discord/src/server.ts
searchableguy: I'm going to change this part later. It's just to see if things work correctly for testing purposes right now.

My plan for this is to eventually become something that people can easily self host or we can host it and have others provide us with their discord webhook creditionals to push the updates. 

This PR is already getting big as it is.

locksmith/examples/websub-discord/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "websub-discord",
   "packageManager": "yarn@3.1.0-rc.7",
   "dependencies": {
+    "@unlock-protocol/networks": "workspace:./packages/networks",
```
julien51, line 5: Hum since we would then this example to be standalone (and I assume reproducible) let's not use the workspace here but hardcode some values in a config file?

locksmith/examples/websub-discord/src/server.ts
```diff
@@ -1,7 +1,8 @@
 import { WebhookClient, MessageEmbed } from 'discord.js'
 import express from 'express'
+import { networks } from '@unlock-protocol/networks'
 import { config } from './config'
-import { chunk } from './util'
+import { chunk, NETWORK_COLOR } from './util'
```
julien51, line 5: I love the idea of having colors by networks!

julien51 (approved):
One final change!

locksmith/examples/websub-discord/package.json
searchableguy: Is it possible to deploy @unlock-protocol/networks package to npm?

Other people might benefit from it too and we can use that directly here.

searchableguy: I don't want to copy over the networks package to example because that mean maintaining two copies of the same. 

If we could publish networks package to npm, we could use @unlock-protocol/networks directly from npm in subsequent PR in future to make it completely standalone.

PR #8044, opened 2022-01-17
feat(websub): Add a model to keep track of sent keys and locks via hooks

searchableguy:
# Description

This adds a ProcessedHookItems table to keep track of sent keys and locks on different networks. 

# Checklist:


julien51:
So I think this will come bite us back pretty early. Let's store each id in a unique row and just use an index ti keep track of the latest?

searchableguy:
@julien51 Can you clarify it a bit more?

we will store each network as a separate row.

searchableguy:
I think one way this would work if we will just create a row for every time the cron job is run in the hook notifier with sent locks and keys. When checking next time, just get the last created row and compare. Would that work?

To get all the keys, we would need to get all the rows and aggregate the ids for locks, and keys from them though. I assume this will work out since that won't be a common operation.

locksmith/migrations/20220117151038-create-hook-notifier.js
```diff
@@ -0,0 +1,34 @@
...
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
+      },
+      network: {
+        allowNull: false,
+        type: Sequelize.INTEGER,
+      },
+      sentKeyIds: {
```
julien51, line 15: This will make it hard to get the last one processed...

julien51:
I think we should have 1 row per "object" (key or lock) that indicates _when_ it was processed. On each run we may write multiple rows.

locksmith/migrations/20220117151038-create-hook-notifier.js
julien51: Let's keep track of a `type` (key or lock for now) and a objectId (that is either the key id or the lock id).

searchableguy:
Okay done. Sequelize will automatically create updated and created fields on saving the processedHookItem. That will store the *when*.

julien51 (approved):

julien51 (commented):
Please make sure you add the right indices on type, objectId and the createdAt timestamp?

julien51:
and also on the network id! It will make querying much faster :)

searchableguy:
Actually, now that I think about. We don't need that many indexes. 

The ID field is auto incrementing and has its own index due to it being primary key. 

We can use this for all the situation created At will be required for. 

Network and type wouldn't benefit from having an index. The possible number of values there are limited and will be in future.

The only one that is worth having an index on is object ID. 

I'm wondering if it should be a composite index of network + object ID.

Afaik, key and lock IDs should be unique across networks, right?

The collison possibility is low but that's only the two possible type of object we are tracking right. We might add something network specific in future so in that case, a composite index of network and ID might be more robust.

What are your thoughts?

julien51:
To get the "latest" of a kind on a network I think we will sort by the following:
* `network`
* `type`
* `createdAt`
I also imagine we will want to check if a specific object has been processed. I agree the risk of collision is very low, so we need an index on `objectId` as well.

PR #8062, opened 2022-01-19
feat(websub): Add signature to headers when notifying subscribers

searchableguy:
# Description

Add signature to our headers using the secret from hook. 

The w3c doesn't mandate that we use sha1 but does mention that many implementation use it despite it being vulnerable to maintain compatibility. Do we want to do that or improve on the security there? (I chose the later for this PR). I can do sha1 though.

# Checklist:

### DIFF
--- a/locksmith/__mocks__/websub/subscriber.ts
+++ b/locksmith/__mocks__/websub/subscriber.ts
@@ -5,15 +5,17 @@ import { createSignature } from '../../src/websub/helpers'
 const handlers = [
   rest.post('http://localhost:4000/callback', (req, res, ctx) => {
     const signature = req.headers.get('x-hub-signature')
+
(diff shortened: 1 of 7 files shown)


julien51 (approved):
I would feel better with some tests here but in the interest of time, I will approve and expect that youll add a few before merging!

searchableguy:
I'm using msw to intercept network request and mock responses from a server as if it is a real subscriber for notify. We could refactor and improve on this approach to test all parts of locksmith later.

julien51:
Hum, I think a better appraoch would be to isolate the signature logic.

searchableguy:
good to go now? @julien51

julien51 (approved):
Thank you!

PR #8068, opened 2022-01-20
feat(websub): Support sending only diffs for locks job

searchableguy:
# Description

First pass implementation. 

# Checklist:


locksmith/src/websub/jobs/keys.ts
```diff
@@ -1,19 +1,42 @@
 import { Key } from '../../graphql/datasource'
-import { Hook } from '../../models'
+import { Hook, ProcessedHookItem } from '../../models'
 import { TOPIC_KEYS } from '../topics'
 
 import { networkMapToFnResult, notifyHook } from '../helpers'
 
+async function getUnprocessedKeys(network: number) {
+  const keySource = new Key(network)
+  const processedKeys = await ProcessedHookItem.findAll({
```
searchableguy, line 9: I thought of limiting this to not strain the database. What do you think of a limit of getting the last 100 items which fullfil this?

julien51: So I would do it in reverse. Rather than get object that have been processed, I would get a list of object "toProcess" from the subgraph and then look which of them have already been processed and only keep the ones which have not. Additionnaly if _all of them_ have not been processed, than it means we can use the subgraph to retrieve more keys and try to process them.

What do you think?

locksmith/src/websub/jobs/keys.ts
```diff
@@ -1,19 +1,42 @@
...
+    where: {
+      type: 'key',
+      network,
+    },
+    order: [['id', 'DESC']],
+  })
+  const keys = await keySource.getKeys({ first: 25 })
```
julien51, line 16: See my comment above, but I think we should start with this, then check these 25 (let's make that a constant at he top of the file) are _really_ new and only process these.
If the 25 are indeed _really_ new, then get the next 25... etc.
This way I think we cannot "miss" any?

searchableguy: Though, this will require changes in the datasource for adding offset. It doesn't support it currently.

julien51: Ha! I think you can use `skip`: https://thegraph.com/docs/en/developer/graphql-api/

searchableguy: Yeah and there is a PR for adding support for that.

locksmith/src/websub/jobs/locks.ts
```diff
@@ -1,19 +1,71 @@
...
+        },
+      },
+      order: [['id', 'DESC']],
+    })
+
+    const onlyUnprocessed = locks.filter(
+      (lock: any) => !processedLocks.find((item) => item.objectId === lock.id)
+    )
+
+    items.push(onlyUnprocessed)
```
julien51, line 36: I wonder if this would not cause a pretty big memory object when we could just start processing these, but let's maybe improve later.

locksmith/src/websub/jobs/keys.ts
```diff
@@ -1,23 +1,75 @@
...
+
+    page += 1
+  }
+  return items.flat()
+}
+
+async function getUnprocessedKeys(network: number) {
+  const results = await fetchUnprocessedKeys(network)
+
+  await ProcessedHookItem.bulkCreate(
```
julien51, line 50: Can we move that to _after_ we performed the notification? So that if the job dies for any reason we're not "missing" things.

locksmith/src/websub/jobs/locks.ts
```diff
@@ -1,19 +1,71 @@
...
+
+    page += 1
+  }
+  return items.flat()
+}
+
+export async function processNewLocks(network: number) {
+  const results = await fetchUnprocessedLocks(network)
+
+  await ProcessedHookItem.bulkCreate(
```
julien51, line 50: Same as above, let's move this to _after_ we have notified

locksmith/src/websub/jobs/locks.ts
```diff
@@ -1,21 +1,80 @@
...
+    })
+
+    await ProcessedHookItem.bulkCreate(processedHookItems)
+
+    page += 1
+  }
+}
+
+export async function notifyOfLocks(hooks: Hook[]) {
+  const subscribedHooks = filterHooksByTopic(hooks, TOPIC_LOCKS)
```
julien51, line 71: Same as on the other PR let's add a note to retrieve the hooks *inside* the `notifyHooksOfAllUnprocessedLocks` function so that we only retrieve hooks if/when there applicable ones.

julien51 (approved):

PR #8074, opened 2022-01-24
feat(websub): Improve websub keys job to only process diffs

searchableguy:
# Description

# Checklist:

### DIFF
--- a/locksmith/src/websub/jobs/keys.ts
+++ b/locksmith/src/websub/jobs/keys.ts
@@ -1,13 +1,9 @@
 import { Op } from 'sequelize'
+import { networks } from '@unlock-protocol/networks'
 import { Key } from '../../graphql/datasource'
 import { Hook, ProcessedHookItem } from '../../models'
 import { TOPIC_KEYS } from '../topics'
-
-import {
-  networkMapToFnResult,
-  notifyHook,
-  filterHooksByTopic,
-} from '../helpers'
+import { notifyHook, filterHooksByTopic } from '../helpers'
 
 const FETCH_LIMIT = 25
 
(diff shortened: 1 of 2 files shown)


locksmith/src/websub/jobs/keys.ts
```diff
@@ -1,24 +1,76 @@
...
+  let page = 0
+  const items = []
+
+  while (true) {
+    const keys = await fetchUnprocessedKeys(network, page)
+    // If empty, break the loop and return as there are no more new keys to process.
+    if (!keys.length) {
+      break
+    }
+    items.push(keys)
```
julien51, line 46: Unfortunately that is still not the right approach and won't scale well.
For each batch of keys, we should notify the items in that batch, then save the batch and _finally_, fetch the next batch.
This way we won't miss any if for any reason something goes wrong somwhere.

julien51 (commented):
Let's change the approach per above please...

locksmith/src/websub/jobs/keys.ts
searchableguy: Got it.

locksmith/src/websub/jobs/keys.ts
```diff
@@ -1,24 +1,77 @@
...
+
+    page += 1
+  }
+}
+
+export async function notifyOfKeys(hooks: Hook[]) {
+  const subscribedHooks = filterHooksByTopic(hooks, TOPIC_KEYS)
+  for (const network of Object.values(networks)) {
+    if (network.id !== 31337) {
+      const hooksFilteredByNetwork = subscribedHooks.filter(
```
julien51, line 71: I wonder if this will scale well, but we can start with this.
I think a good next step would be to retrieve the hooks only after we have `fetchUnprocessedKeys`. wdyt?

julien51 (approved):

PR #8088, opened 2022-01-27
feat(locksmith): adding email notification on new keys

julien51:
# Description

Adding a mechanism to automatically send an email notification when a new key is sold on a lock.

# Checklist:

## Release Note Draft Snippet

Automated emails sent when new key is sold and has a metadata field set for the owner.

### DIFF
--- a/locksmith/config/config.js
+++ b/locksmith/config/config.js
@@ -2,6 +2,7 @@ const urlParser = require('url')
 
 const config = {
   database: {
+    logging: false,
     dialect: 'postgres', // sequelize v4 needs this
   },
   stripeSecret: process.env.STRIPE_SECRET,
--- a/locksmith/config/config.js
+++ b/locksmith/config/config.js
@@ -11,6 +11,9 @@ const config = {
     '0x08491b7e20566b728ce21a07c88b12ed8b785b3826df93a7baceb21ddacf8b61',
   metadataHost: process.env.METADATA_HOST,
   logging: false,
+  services: {
+    wedlocks: 'http://localhost:1337'
+  },
 }
 
 if (Boolean(process.env.ON_HEROKU)) {
@@ -40,5 +43,10 @@ if (process.env.DATABASE_URL) {
     dialect: 'postgres',
   }
 }
+
+if (process.env.UNLOCK_ENV === 'prod' || process.env.UNLOCK_ENV === 'staging') {
+  config.services.wedlocks = 'https://wedlocks.unlock-protocol.com/.netlify/functions/handler'
+} 
+
 
 module.exports = config
--- a/locksmith/src/operations/wedlocksOperations.ts
+++ b/locksmith/src/operations/wedlocksOperations.ts
@@ -1,4 +1,5 @@
 import fetch from 'cross-fetch'
+import * as Normalizer from '../utils/normalizer'
 import { UserTokenMetadata } from '../models'
 import config from '../../config/config'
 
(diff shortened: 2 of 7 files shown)


locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ * Resolves when all new keys have been processed
+ * @param keys
+ */
+export const notifyNewKeysToWedlocks = async (keys: any[]) => {
+  for (const key of keys) {
+    await notifyNewKeyToWedlocks(key)
+  }
```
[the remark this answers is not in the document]
julien51: Actually we do. If we have too many keys we will send everything "at once" to wedlocks (or even to our own database) so Its better to do this sequentially.

locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -0,0 +1,80 @@
...
+ * @param params
+ * @param attachments
+ * @returns
+ */
+export const sendEmail = async (
+  template: string,
+  failoverTemplate: string,
+  recipient: string,
+  params: Params = {},
+  attachments: Attachment[] = []
```
searchableguy, line 29: A nit pick, I prefer an object as an argument when there are too many of these positional argument. 

It will also be easier to build the payload that way so not hugely repetitive.

locksmith/src/websub/jobs/keys.ts
```diff
@@ -39,6 +40,9 @@ async function notifyHooksOfAllUnprocessedKeys(hooks: Hook[], network: number) {
       break
     }
 
+    // Notify all new keys to wedlocks!
+    await notifyNewKeysToWedlocks(keys)
+
     await Promise.all(
       hooks.map(async (hook) => {
         const data = keys.filter((key: any) => key.lock.id === hook.lock)
```
searchableguy, line 48: Can you group both of those functions so they fire off concurrently.

locksmith/config/config.js
```diff
@@ -2,6 +2,7 @@ const urlParser = require('url')
 
 const config = {
   database: {
+    logging: false,
```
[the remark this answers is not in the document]
julien51: Oh this is the logging on the DB, I am not sure it is useful?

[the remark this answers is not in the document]
julien51: Very noisy in the logs and can easily be re-enabled to deploy

locksmith/config/config.js
```diff
@@ -41,4 +45,9 @@ if (process.env.DATABASE_URL) {
   }
 }
 
+if (process.env.UNLOCK_ENV === 'prod' || process.env.UNLOCK_ENV === 'staging') {
+  config.services.wedlocks = 'https://wedlocks.unlock-protocol.com/.netlify/functions/handler'
```
searchableguy, line 49: How do you stop someone else from spamming that specific endpoint? 

Afaik, there are no easy way to get static IPs on heroku so you can't rely on IP whitelist.

julien51: For now we don't. Since we only send emails from pre-defined templates, the value for 3rd party folks is pretty little. That said we had a plan to "sign" the API requests. I think once that gets a probem we should absolutely do it!

locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -0,0 +1,80 @@
...
+  })
+
+  if (userTokenMetadataRecord?.data?.userMetadata?.protected?.email) {
+    await sendEmail(
+      `keyMined-${key.owner.address}`,
+      'keyMined',
+      userTokenMetadataRecord.data.userMetadata.protected.email,
+      {
+        keychainUrl: '',
+      }
```
searchableguy, line 77: since it's referenced twice. Move it to a variable above.

like 

```ts
const email = userTokenMetadataRecord?.data?.userMetadata?.protected?.email

if(email) { sendEmail({ email, }) }
```

locksmith/src/websub/jobs/keys.ts
```diff
@@ -39,16 +40,17 @@ async function notifyHooksOfAllUnprocessedKeys(hooks: Hook[], network: number) {
       break
     }
 
-    await Promise.all(
-      hooks.map(async (hook) => {
+    await await Promise.all([
```
searchableguy, line 43: Should use Promise.allSettled. Promise.all will fail if any of those fail. I don't think those tasks depend on each other so one shouldn't disrupt the other for whatever reason. WDYT?

locksmith/src/operations/wedlocksOperations.ts
julien51: Great idea!

PR #8089, opened 2022-01-27
fix(ci): Add third build directory parameter to heroku script and CI deployment

searchableguy:
# Description
Websub exists as part of locksmith directory but we are deploying it separately. So the existing way to reuse service name as a build directory parameter for caching purposes in docker breaks. This PR adds an explicit build-dir argument to CI which is optional but can be configured to change build directory. 

This also fixes websub deployment which is broken as a reason.

# Checklist:


clemsos (approved):
good to go !

PR #8091, opened 2022-01-27
feat(websub): Add failure limit on responding to hooks

searchableguy:
# Description

This adds a failure limit where if a hook had 3 non-successful failure to process event, we unsubscribe it.

# Checklist:

### DIFF
--- a/locksmith/src/websub/helpers.ts
+++ b/locksmith/src/websub/helpers.ts
@@ -3,6 +3,7 @@ import crypto from 'crypto'
 import fetch from 'cross-fetch'
 import { AbortController } from 'node-abort-controller'
 import { setTimeout, clearTimeout } from 'timers'
+import { Op } from 'sequelize'
 import { Hook, HookEvent } from '../models'
 import { logger } from '../logger'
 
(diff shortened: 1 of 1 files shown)


locksmith/src/websub/helpers.ts
```diff
@@ -70,10 +71,34 @@ export async function notifyHook(hook: Hook, body: unknown) {
...
+      limit: 3,
+    })
+
+    const checkBadHealth = previousHookEventsForCurrentHook.every(
+      (event) => event.state !== 'success'
+    )
+
+    // If three attempts were not successful, unsubscribe the hook and return the pending event
+    if (checkBadHealth) {
+      logger.warn(`Unsubscribing hook (${hook.id}) due to being unresponsive.`)
```
julien51, line 96: Ha no, we should not unsubscribe, just stop retrying. The expiration date will fix this.
It is important because if for some reason a subscriber is down for a few minutes at the time of a notification we don't want to assume they're down "for good". If we did this they would need to resubscribe to everything once they come back up and that would cause extra load.

julien51 (commented):
Let's not unsubscribe hooks arbitrarily. The expiration date will handle that garbage collection when someone is down for too long.

locksmith/src/websub/helpers.ts
searchableguy: okay!

julien51 (approved):

PR #8170, opened 2022-02-01
feat(ui): Add docs, build script, and make it available.

searchableguy:
# Description

This makes our UI package available to other packages in the workspace for usage. I've added some docs on how to use it.

# Checklist:


julien51 (approved):

package.json
```diff
@@ -36,7 +36,7 @@
   ],
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "packages": "yarn workspaces foreach --from '{@unlock-protocol/types,@unlock-protocol/networks,@unlock-protocol/paywall,@unlock-protocol/unlock-js,@unlock-protocol/contracts}'",
+    "packages": "yarn workspaces foreach --from '{@unlock-protocol/types,@unlock-protocol/networks,@unlock-protocol/paywall,@unlock-protocol/unlock-js,@unlock-protocol/contracts,@unlock-protocol/ui}'",
```
julien51, line 39: I wonder if there would be a way to use `@unlock-protocol/*`? Maybe @clemsos knows?

julien51 (approved):

package.json
searchableguy: Yeah, there is. 

But I think current way is fine. It allows us to keep packages that we don't use (for example, ui wasn't used until now) from building and causing unnecessary delay in getting things start for others.

PR #8223, opened 2022-02-14
feat(unlock-app) formatting infinite durations

julien51:
# Description

Taking into account that we now have locks with infinite duration for keys!

# Checklist:

## Release Note Draft Snippet

Support for non expiring keys in the UI!


unlock-app/src/utils/durations.js
```diff
@@ -42,6 +42,10 @@ export function durations(seconds, intervals) {
  * @param {number} seconds
  */
 export function durationsAsTextFromSeconds(seconds) {
+  if (seconds < 1) {
+    // 0 or infinite return an empty string
+    return ''
+  }
```
searchableguy, line 48: This require changing existing test which is failing because we have one for 0.5 seconds.

packages/unlock-js/src/PublicLock/v7/getLock.js
```diff
@@ -16,7 +16,12 @@ export default async function (address, provider) {
   const attributes = {
     name: (x) => x,
     keyPrice: (x) => x,
-    expirationDuration: parseInt,
+    expirationDuration: (value) => {
+      if (utils.isInfiniteDuration(value)) {
+        return -1
+      }
+      return parseInt(value)
```
searchableguy, line 23: Would recommend explicitly adding the radix here in parseInt.

searchableguy (changes requested):

unlock-app/src/utils/durations.js
julien51: HA! Good point! Let me fix

PR #8232, opened 2022-02-15
feat(unlock-protocol): Add feed for blog

searchableguy:
# Description

Adds feed and resolves carfi's request to include content in our generated feed.

# Issues

Fixes #
Refs #

# Checklist:


searchableguy:
:thinking: Seems like jest doesn't like new es modules.

julien51:
Ha! Let's not add multiple feeds. This is not useful. We only need one (and we already have it!).
Also no need to add a new dependency here just manually add the `<content>` in the blog.js file :)

searchableguy:
Those feeds are in different format though. Depending on what kind of reader the person use, it may or may not have support for everything which is why I added it but I remove the additional feeds. 

I think we should keep atom format, remove the rest. WDTY?

searchableguy:
I'm aware of the existing function but I think it's better to use a well tested library for generating the feed than using template strings which is why I changed the function. It allows us to move formats more easily.

julien51 (approved):
Let's clean up the old file that generated the blog please :)

PR #8249, opened 2022-02-18
feat(smart-contracts): multiple key purchases

### LINKED ISSUE #8100: Multiple purchase
Our locks currently only support purchasing one key at a time. 
We need to add support for the ability to purchase multiple keys at a time. We should still have the same mechanism where the buyer is the payer but of course they could/should set multiple recipient addresses.

The base use case is for conferences like EthCC who want to be able to offer the ability to people to purchase multiple keys at once for multiple attendees (like someone in the company buys keys for all of their team members).

clemsos:
# Description

Add support for purchasing multiple keys at once in a lock. 

This introduces a new api for the `purchase` function that now accepts arrays of recipients, referrers, and key managers as param.

# Issues

Fixes #8100 
Refs #

# Checklist:


smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -57,94 +57,106 @@ contract MixinPurchase is
...
-      // emit missing unlock
-      emit UnlockCallFailed(address(this), address(unlockProtocol));
+      if(address(onKeyPurchaseHook) != address(0)) {
+        onKeyPurchaseHook.onKeyPurchase(
+          msg.sender, 
+          _recipient, 
+          _referrers[i], 
+          _data, 
+          inMemoryKeyPrice, 
+          0 // tokenAddress == address(0) ? _value : msg.value
```
clemsos, line 148: This is one I am unsure of so far. How is this intended to be used?

julien51:
cc @pscott =)

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -57,94 +57,106 @@ contract MixinPurchase is
   */
   function purchase(
     uint256 _value,
-    address _recipient,
-    address _referrer,
-    address _keyManager,
+    address[] memory _recipients,
+    address[] memory _referrers,
+    address[] memory _keyManagers,
     bytes calldata _data
```
julien51, line 63: Let's add that as an array as well please?

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -57,94 +57,106 @@ contract MixinPurchase is
   */
   function purchase(
     uint256 _value,
```
julien51, line 59: Let's move this in the array as well.

julien51 (commented):
@clemsos Let's really make it _exactly_ as if the `purchase` function was called n times.
The `value` could be different for each purchaser (different users might provide different tips)

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -57,94 +57,106 @@ contract MixinPurchase is
...
-        newTimeStamp = type(uint).max;
+    require(_recipients.length == _referrers.length, 'INVALID_REFERRERS_LENGTH');
+    require(_recipients.length == _keyManagers.length, 'INVALID_KEY_MANAGERS_LENGTH');
+
+    uint totalPriceToPay;
+
+    for (uint256 i = 0; i < _recipients.length; i++) {
+      // check recipient address
+      address _recipient = _recipients[i];
+      require(_recipient != address(0), 'INVALID_ADDRESS');
```
julien51, line 76: I actually think we could lift this requirement since the key manager could transfer that key anyway.

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -57,94 +57,106 @@ contract MixinPurchase is
...
+          address(0), // This is a creation.
+          _recipient,
+          idTo
+        );
+      } else if (toKey.expirationTimestamp > block.timestamp) {
+        // prevent re-purchase of a valid non-expiring key
+        require(toKey.expirationTimestamp != type(uint).max, 'A valid non-expiring key can not be purchased twice');
+
+        // This is an existing owner trying to extend their key
+        newTimeStamp = toKey.expirationTimestamp + expirationDuration;
```
julien51, line 111: I think we should check if `if(expirationDuration == type(uint).max) {` here too... if the duration was changed after thet previous purchase...

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -57,94 +57,106 @@ contract MixinPurchase is
...
+      }
 
-    
-    uint inMemoryKeyPrice = _purchasePriceFor(_recipient, _referrer, _data);
+      // price      
+      uint inMemoryKeyPrice = _purchasePriceFor(_recipient, _referrers[i], _data);
+      totalPriceToPay = totalPriceToPay + inMemoryKeyPrice;
+
+      // call Unlock contract to record GNP
+      try unlockProtocol.recordKeyPurchase(inMemoryKeyPrice, _referrers[i])
```
julien51, line 134: is that were we should check the gas limit too?

smart-contracts/contracts/mixins/MixinPurchase.sol
julien51: This is to show the actual price "submitted" by the user (either the `msg.value` or the `_value` for this specific purchase. (user can submit more than required...).

This is why it is important to also make sure the `_value` is in an array... The FE will build the `msg.value` to be the sum of all `_values`.

clemsos:
> @clemsos Let's really make it _exactly_ as if the `purchase` function was called n times. The `value` could be different for each purchaser (different users might provide different tips)

ok added `_values` as an array.

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: the gas limit could be checked against a max number of purchases that can be made at once, so we check as `require(_recipients.length == 20, ''20_PURCHASES_MAX");`

julien51: No no i meant the attack vector identified during the audit :) where unlock is compromised but could not break the locks themselves.

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: but what if the keyManager is invalid too?  

In practice, it is likely the same array will be passed for both recipients and keyManagers. If the mistake is in both then there are no ways to revert it and the key will be lost.

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: oh you mean if the lock duration was set to infinite while the loop is executed. yep added

julien51 (approved):
Let's ship it!

clemsos:
ok tests are finally passing when capping the 1recordKeyPurchase` at 200k gas

PR #8259, opened 2022-02-22
chore(ci): refactor gh actions to use separate file for different envs

clemsos:
# Description

The current structure of gh actions workflows is confusing, and the use of the `workflow_run` condition makes it even more so. The current PR aims at simplifying things by splitting into 3 main files for pr, master and prod describing what should happen in each context.

# Issues

Fixes #
Refs #

# Checklist:


.github/workflows/master.yml
```diff
@@ -0,0 +1,33 @@
...
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  run-all-tests:
+    uses: ./.github/workflows/_tests.yml
+
+  deploy-locksmith-staging:
```
julien51, line 12: We should also deploy all netlify services, correct?

.github/workflows/production.yml
```diff
@@ -0,0 +1,33 @@
+name: "Production Branch"
+
+on:
+  push:
+    branches:
+      - production
+
+jobs:
+  run-all-tests:
+    uses: ./.github/workflows/_tests.yml
```
julien51, line 10: Same as above?

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,58 @@
...
+    needs: run-tests
+    uses: ./.github/workflows/_netlify.yml
+    with:
+      service: unlock-protocol-com
+      target-env: staging
+    secrets:
+      SITE_ID: ${{ secrets.UNLOCK_PROTOCOL_COM_NETLIFY_STAGING_AUTH_TOKEN }}
+      AUTH_TOKEN: ${{ secrets.UNLOCK_PROTOCOL_COM_NETLIFY_STAGING_AUTH_KEY }}
+  
+  # deploy-paywall-app
```
julien51, line 53: ha! let's make sure we have these :)

julien51 (commented):
ok I like this new approach. Do you want to add the tasks to deploy all front-ends too?

They should be pretty similar?

.github/workflows/master.yml
clemsos: yep, fixing the workflow for PRs first, then will add

.github/workflows/_netlify.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Restore docker
+        run: .github/actions/docker-cache/restore.sh ${{ runner.temp }}/docker_cache/cache.tar
+      - name: "Export secrets to env"
+        id: filter_secrets
+        run: | 
+          SERVICE_NAME="${{ inputs.service }}"
+          UPCASE_SERVICE="${SERVICE_NAME^^}"
+          PREFIX="${UPCASE_SERVICE//-/_}_NETLIFY_${{ inputs.target-env }}"
+          PREFIX="${PREFIX^^}" # uppercase
+          echo "${PREFIX}"
```
julien51, line 90: 🧹

julien51 (approved):
Still looks good

.github/workflows/_netlify.yml
```diff
@@ -0,0 +1,97 @@
...
+      SITE_ID:
+        required: true
+      AUTH_TOKEN:
+        required: true
+      # unlock-app only
+      BASE64_WEDLOCKS_PUBLIC_KEY:
+        required: false
+      STRIPE_KEY:
+        required: false
+      UNLOCK_GA_ID:
```
clemsos, line 30: do we need to keep that as an env var?

PR #8262, opened 2022-02-22
chore(ci): remove unnecessary env vars from Netlify config files

clemsos:
# Description

There are several things that get set by ENV vars when we should really rely on config files instead. This also solved issues with Github actions in #8259

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/next.config.js
+++ b/unlock-app/next.config.js
@@ -11,15 +11,31 @@ dotenv.config({
 })
 
 // TODO renames these: URLs need to be URLs, hosts need to be hosts... etc
+const staging = {
+  locksmithHost: 'https://rinkeby.locksmith.unlock-protocol.com',
+  unlockStaticUrl: 'https://unlock-protocol.com',
+  unlockAppUrl: 'https://staging-app.unlock-protocol.com',
+  paywallUrl: 'https://staging-paywall.unlock-protocol.com',
+}
+
+const production = {
+  locksmithHost: 'https://locksmith.unlock-protocol.com',
+  unlockStaticUrl: 'https://unlock-protocol.com',
+  unlockAppUrl: 'https://app.unlock-protocol.com',
+  paywallUrl: 'https://paywall.unlock-protocol.com',
+}
+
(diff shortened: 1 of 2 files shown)


julien51 (approved):

searchableguy:
This is a bit confusing. Why are we allowing loading from environment variables on staging but not on production?

clemsos:
> This is a bit confusing. Why are we allowing loading from environment variables on staging but not on production?

Yes agree this is kind of ugly. The reason is to maintain support for actual dev workflow that relies on a `.env` file.

The best way will be to remove environment variables entirely and just have a simple config file. Wdyt?

searchableguy:
There is one problem with that. You will need more than two config (staging, prod) for fleek too because that will be hosted on a different domain. cc @julien51 

Another problem is development. I guess you are thinking of overriding the staging with local environment variables which should work fine. Nextjs actually automatically loads up environment variables inside `.env.local`. We don't need to depend on dotenv and our own setup for using them as config.

searchableguy:
Personally, I would suggest if we are putting config values in a file - then remove loading up environment variables in staging and rather have one more `dev` config which uses environment variables and default values.

It should be run by default if no unlockEnv is set to nothing or `dev`. WDTY?

clemsos:
Yes, agree that it makes more sense to have a dev section in the config and not using env vars.

What about fleek deployment? How do you think we should handle this?

clemsos:
also it seems that the dotenv deps exists to import a ENV file from the root of the repo. Will be good to get rid of this as well, and have a better way to set our URLS in config and secrets in exported ENV vars.

one reason I can think this pattern exists is that the config file can be shared across the different repos, so we avoid duplicating code. Now with yarn there is no way to import a file outside the current folder (i.e. `unlock-app`) by using a relative path. This needs to the idea of having a shared "config " package with URLs etc, which is sort of a rabbit hole too

clemsos:
@searchableguy just updated the config pattern following our discussion. Looks good to me as it is now, also knowing it will be handled in the website rewriting

PR #8337, opened 2022-03-02
feat(smart-contracts): mimick ownable

### LINKED ISSUE #8276: Owner
Several applications "expect" our lock contracts to have an "owner" and among them, OpenSea. 
We don't have one right now but we have the concept of lock manager.
We should add a `setOwner` method callable _only_ by a lock manager to set an address as "owner" and trigger an `OwnershipTransferred(address previousOwner, address newOwner)` event. The default owner should be the deployer.

Additionally we should add both `owner() → address` and  `isOwner() → bool` that will return the address of the current owner and true of the caller is an owner respectively.

clemsos:
# Description

Provides an helpers to mimick the `Ownable` contract ABI. The `Ownable` logic is used by many 3rd party services to determine contract ownership - e.g. who is allowed to edit metadata on Opensea.

NB: This logic is NOT used internally by the Unlock Protocol and is made available only as a convenience helper.

# Issues

Fixes #8276

# Checklist:


julien51 (approved):
💯

julien51:
Perfect :D

julien51:
One important thought here: how can we make sure existing locks can "safely" be upgraded with this?

smart-contracts/contracts/mixins/MixinRoles.sol
```diff
@@ -78,5 +84,33 @@ contract MixinRoles is AccessControlUpgradeable {
...
+    address _previousOwner = _convenienceOwner;
+    _convenienceOwner = account;
+    emit OwnershipTransferred(_previousOwner, account);
+  }
+
+  function isOwner(address account) public view returns (bool) {
+    return _convenienceOwner == account;
+  }
+
   uint256[1000] private __safe_upgrade_gap;
```
julien51, line 115: So I expect the Gap here to be reduced?

julien51 (changes requested):

smart-contracts/contracts/mixins/MixinRoles.sol
clemsos: Yes, I thought we will do that after other changes where done. But for sure we need a streamlined way to assess the storage continuity

julien51: Let's actually do that on every PR because I worry we may forget eventually?

smart-contracts/test/Lock/upgrades.js
```diff
@@ -0,0 +1,82 @@
...
+      'A neat upgradeable lock!',
+    ]
+
+    lock = await upgrades.deployProxy(PublicLockPast, args)
+    await lock.deployed()
+  })
+
+  it('should allow upgrade', async () => {
+    // deploy new implementation
+    await upgrades.upgradeProxy(lock.address, PublicLockLatest)
```
clemsos, line 80: Just a very basic test case `yarn hardhat test test/Lock/upgrades.js`

julien51:
So  after I think it is ok to create a Mixin and have this be isolated from the rest...

julien51 (approved):
Thanks for adding the upgradability test!

julien51:
Huhuhuh `trying to deploy a contract whose code is too large`. That's _very_ bad news.

clemsos:
> Huhuhuh `trying to deploy a contract whose code is too large`. That's _very_ bad news.

ok this appears to be a real issue with the contract size increase. It can be solved by using a lower value in the compiler `runs` optimizer but still should be careful about it ?

julien51:
Hum. I am quite worried about that because I suspect we will bump into the same issue as we add more things?

smart-contracts/hardhat.config.js
```diff
@@ -30,7 +30,7 @@ const { getHardhatNetwork } = require('./helpers/network')
 const settings = {
   optimizer: {
     enabled: true,
-    runs: 200,
+    runs: 20,
```
julien51, line 33: This does not leave us a lot of margin :/ wdyt?

clemsos: so from trial and error, we can get as low as 140 but ofc thats not a viable solution. Filed an issue for now

Issue #8361, opened 2022-03-07
Adding the ability to skip metadata collection when all fields are optional


julien51:
In a checkout flow [like this one](https://app.unlock-protocol.com/checkout?redirectUri=https://blackpool.finance/events/nft-barcelona3-28-2022?booked=true&paywallConfig={%22pessimistic%22:true,%22locks%22:{%220x89e975EA43E0Cfe338205e016BFFaeFeFdbc3511%22:{%22network%22:137,%22name%22:%22NFT%20Barcelona%22}},%22icon%22:%22https://blackpool.finance/favicon.ico%22,%22callToAction%22:{%22default%22:%22The%20traditional%20BlackPool%20NFT%20Auction,%20to%20take%20place%20in%20the%20incredible%20Valkiria%20Hub%20in%20Barcelona.%20RSVP%20below!%20If%20you%20show%20up%20to%20the%20event,%20your%20ticket%20enables%20you%20to%20participate%20in%20a%20raffle%20for%20a%20physical%20airdrop%20at%20the%20bar!%22},%22metadataInputs%22:[{%22name%22:%22email%22,%22type%22:%22email%22,%22required%22:false,%22public%22:false}]}), we want to allow the user to "skip" the metadata collection phase as all fields are optional.

The "Skip" button should _only_ be visible if all fields are optional. Alternatively it could just be a small text button below the "Save and Continue" button.

kalidiagne:
@julien51 In the case some fields are required, what will be the action that "skip" will have? 
In the current PR opened I covered only the first case

kalidiagne:
Resolved with https://github.com/unlock-protocol/unlock/pull/8369

PR #8369, opened 2022-03-08
feat(unlock-app) -  skip metadata collection when all fields are optional

kalidiagne:
# Description
- Added the ability to skip metadata collection when all fields are optional, in this case a show button will be visible under the "Save and Continue" button, and when the skip button is clicked `setUserMetadataData` will not be executed
- Fix and improvement on popup styling
# Issues
https://github.com/unlock-protocol/unlock/issues/8361 

# Fixes
- Remove box-shadow on checkout modal, to add a black transparent backdrop like others checkouts popups 
- Fixed some style to center the modal

| Before | After | 
| -- | -- | 
| 8361-before  |  8361-after |

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/CheckoutContainer.tsx
+++ b/unlock-app/src/components/interface/checkout/CheckoutContainer.tsx
@@ -16,21 +16,27 @@ export const CheckoutContainer: React.FunctionComponent<Props> = ({
 
 export const Container = styled.div`
   position: fixed;
-  height: 100%;
-  width: 100%;
+  height: 100vh;
+  width: 100vw;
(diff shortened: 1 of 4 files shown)


julien51:
Also, if I understand this correcly, pressing the `skip` button does not advance the to the next step, does it? I think we should just advance once the use clicks on `Skip`.

kalidiagne:
@julien51 changed the "skip" button styling and also updated the screen for reference.
actually, by pressing the `Skip` button, it advances to the next step without asking for a signature

julien51 (approved):
Let's merge this!

Issue #8264, opened 2022-02-22
Dashboard should not allow number of keys to be smaller 


julien51:
The smart contracts do not support setting a smaller number of keys than the existing number of keys.
Our front-end currently fails silently on that. We should add a validation error on that field.

PR #8263, opened 2022-02-22
adding blog post about websub

julien51:
# Description

All in the title!

# Issues

Fixes #
Refs #

# Checklist:

- [ ] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #8386, opened 2022-03-11
fix(unlock-app) #8263 - fix not allow number of keys to be smaller

kalidiagne:
# Description
This prevent the possibility to set smaller number of keys than the existing number of keys

# Issues

Fixes #
- https://github.com/unlock-protocol/unlock/issues/8264

fix-8264

# Checklist:

### DIFF
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -94,18 +94,14 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
   }
 
   const toggleCurrency = () => {
-    if (lockInForm.currencyContractAddress) {
-      return dispatch({
-        change: [
-          { name: 'currencyContractAddress', value: null },
-          { name: 'currencySymbol', value: null },
-        ],
-      })
-    }
+    const erc20Address = lockInForm.currencyContractAddress
+      ? null
+      : erc20.address
+    const erc20Symbol = lockInForm.currencyContractAddress ? null : erc20.symbol
     dispatch({
       change: [
-        { name: 'currencyContractAddress', value: erc20.address },
-        { name: 'currencySymbol', value: erc20.symbol },
+        { name: 'currencyContractAddress', value: erc20Address },
+        { name: 'currencySymbol', value: erc20Symbol },
       ],
     })
   }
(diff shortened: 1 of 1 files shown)


searchableguy (changes requested):
image


The unlimited button won't work because that uses infinity string symbol which is converted into -1 sadly with current validation.

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -131,9 +127,12 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
         }
         break
       case 'maxNumberOfKeys':
-        if (value !== UNLIMITED_KEYS_COUNT && !isPositiveInteger(value)) {
+        if (!isPositiveInteger(value)) {
```
julien51, line 130: Actually I think we should let creators deploy locks with 0 keys since they can either change it later (useful when someone wants to deploy everything and test things without "selling" memberships right away) or even only ever "grant" keys to anyone.

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -131,9 +127,12 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
...
           return 'The number of keys needs to be greater than 0'
         }
+        if (parseInt(value, 10) < lock.outstandingKeys) {
```
julien51, line 133:
```suggestion
        if (parseInt(value, 10) <= lock.outstandingKeys) {
```

In our next contract upgrade we will authorize this to be equal to the number of keys... so let's make sure the front-end can support that please :)

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -131,9 +127,12 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
         }
         break
       case 'maxNumberOfKeys':
-        if (value !== UNLIMITED_KEYS_COUNT && !isPositiveInteger(value)) {
+        if (!isPositiveNumber(value) && value !== INFINITY) {
```
kalidiagne, line 130: @searchableguy fixed based on your suggestion about the `Infinity` value

kalidiagne (commented):
updated the PR following the suggestions provided

julien51 (approved):

searchableguy (approved):
LGTM

One nitpick - we don't tell users they cannot set keys lower than their current quantity. The error isn't deplayed even though its returned.

PR #8374, opened 2022-03-09
feat(smart-contracts): extract `extend` key duration as a standalone fonction

clemsos:
# Description

This PR introduces a new `extend` function which complement `purchase` and allow users to directly extend duration of a given key

# Issues

Relates to #7938 

# Checklist:

- [ ] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

Issue #7938, opened 2021-12-23
Multiple memberships


julien51:
At this point, one of the cornerstones of the protocol's implementation is that a single address can only ever own a single membership/NFT. This made sense based on discussions we had with publishers in the early days of Unlock when they all expressed the fact that it was actually how things worked in the "legacy" membership model: one (email address) can only own a single Netflix account for example.

Another benefit is that it avoids having someone "waste" money by having 2 memberships for the same thing. Indeed, in the case of a monthly membership, having 2 memberships means that the user is paying twice as much per month.

This decision also affects how "transfers" work. Since one can only have a single membership, transfers are just transfers of time, as the actual NFT of the sender remains in possession of the sender (unless the recipient does not have one), and only the time off of their membership gets transferred.

Over the last few months, we have seen people (most recently Tales of Ronin) ask to support situations where an address would own more than 1 membership. 

It would be interesting to explore this and assess what impact this has on the rest of the protocol's behavior.

PR #8397, opened 2022-03-14
feat(smart-contracts): allow multiple keys ownership

clemsos:
# Description

Allow users to own multiple keys

~~TODO: needs to rebase once #8395 and #8374 are merged~~

# Issues

Fixes #
Refs #7938 

# Checklist:


smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -274,16 +265,16 @@ interface IPublicLock
...
+   * transfer to another account.  This is pro-rated so the fee goes 
+   * down overtime.
+   * @dev Throws if _tokenId is not have a valid key
+   * @param _tokenId The id of the key check the transfer fee for.
    * @param _time The amount of time to calculate the fee for.
    * @return The transfer fee in seconds.
    */
   function getTransferFee(
-    address _keyOwner,
+    address _tokenId,
```
julien51, line 277:
```suggestion
    uint _tokenId,
```

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -31,11 +31,21 @@ contract MixinKeys is
...
-  // Each owner can have at most exactly one key
-  // return 0 values when missing a key
+  // DEPREC: dont use
   mapping (address => Key) internal keyByOwner;
 
+  // store all keys: tokenId => token
+  mapping(uint256 => Key) private _keys;
+  
+  // store ownership: owner => list of tokenIds
+  mapping(address => mapping(uint256 => uint256)) private _ownedKeyIds;
```
julien51, line 41: Do we really need `uint256` for the index for a user's token ids?
I wonder if we could use something significantly smaller and/or if it matters?

```suggestion
  mapping(address => mapping(uint256 => uint256)) private _ownedKeyIds;
```

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -31,11 +31,21 @@ contract MixinKeys is
...
+  
+  // store indexes: owner => list of tokenIds
+  mapping(uint256 => uint256) private _ownedKeysIndex;
```
julien51, line 44: needed?

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -31,11 +31,21 @@ contract MixinKeys is
...
+
+  // Mapping owner address to token count
+  mapping(address => uint256) private _balances;
```
julien51, line 47: Wow. is there no way to know the size of a mapping?

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -131,57 +153,65 @@ contract MixinKeys is
...
-      uint newTimeStamp
+      uint newTimestamp
     )
   {
-    Key memory key = getKeyByOwner(_recipient);
+    uint expirationTimestamp = _keys[_tokenId].expirationTimestamp;
 
     // prevent extending a valid non-expiring key
-    require(key.expirationTimestamp != type(uint).max, 'CANT_EXTEND_NON_EXPIRING_KEY');
+    require(expirationTimestamp != type(uint).max, 'A valid non-expiring key can not be purchased twice');
```
julien51, line 200: Let's stay consustent with the rest of the error messages :)

```suggestion
    require(expirationTimestamp != type(uint).max, 'CANT_EXTEND_NON_EXPIRING_KEY');
```

julien51 (commented):
First pass. I'll do more tonight!

smart-contracts/contracts/mixins/MixinKeys.sol
clemsos: yes I think it is better to keep all in uint, it does not make much difference as we are not doing calculations on it

smart-contracts/contracts/mixins/MixinKeys.sol
clemsos: yes this is need to keep track of where a token is in array so when a token is removed, we can swap the last token at its index and we don't end up with an empty value in the array

smart-contracts/contracts/mixins/MixinKeys.sol
clemsos: nope, not in solidity

smart-contract-extensions/package.json
```diff
@@ -34,7 +34,7 @@
     "@nomiclabs/hardhat-web3": "2.0.0",
     "@openzeppelin/contracts": "4.5.0",
     "@openzeppelin/test-helpers": "0.5.15",
-    "@openzeppelin/upgrades": "2.8.0",
+    "@openzeppelin/upgrades": "^2.8.0",
```
julien51, line 37:
```suggestion
    "@openzeppelin/upgrades": "2.8.0",
```

julien51:
As discussed, let's add a `migrate` function in another PR that should be called upon the upgrade BUT also independently. This function should do the following:
- keep track of the `schemaVersion`, if the `schemaVersion` matches the lock version, then there is nothing to do!
- since we only have sequential upgrades, we will have a new implementation of the `migrate` function for each version which is great
- this function should be "resumable" (ie. callable multiple times in order to avoid hitting the gas limit).
- the function should be called as part of the function to upgrade the lock but should not hit the gas limit to make sure we can safe upgrade.
- while not migrated, let's make sure we disable some functions that might alter the state of the keys arrays?
- once migrated, let's update `schemaVersion`.

clemsos:
has to add #8433 to make upgrades/migrate test possible

smart-contracts/contracts/Unlock.sol
```diff
@@ -312,6 +312,12 @@ contract Unlock is
     TransparentUpgradeableProxy proxy = TransparentUpgradeableProxy(lockAddress);
     proxyAdmin.upgrade(proxy, impl);
 
+    // let's upgrade the data schema
+    // we (arbitrarly) cap the migration process to 100 keys to prevent running out of gas
+    // for lock that have more than 100 active keys, the `migrateKeys` function will need
+    // to be called again after the upgrade is done
+    lock.migrateKeys(100);
```
julien51, line 319: Let's make it agnostic.

clemsos:
- [x] Make sure lock upgrades fail if the  schema is too old: check that publicLockVersion matches schemaVersion
schemaVersion should be 0 or equal to publicLockVersion

smart-contracts/contracts/Unlock.sol
```diff
@@ -312,6 +312,12 @@ contract Unlock is
     TransparentUpgradeableProxy proxy = TransparentUpgradeableProxy(lockAddress);
     proxyAdmin.upgrade(proxy, impl);
 
+    // let's upgrade the data schema
+    // we (arbitrarly) cap the migration process to 100 keys to prevent running out of gas
```
julien51, line 316:
```suggestion
```

smart-contracts/contracts/Unlock.sol
```diff
@@ -312,6 +312,12 @@ contract Unlock is
...
+    // for lock that have more than 100 active keys, the `migrate` function will need
```
julien51, line 317:
```suggestion
```

smart-contracts/contracts/Unlock.sol
```diff
@@ -312,6 +312,12 @@ contract Unlock is
...
+    // to be called again after the upgrade is done
```
julien51, line 318:
```suggestion
    // to be called again after the upgrade is done
```

smart-contracts/contracts/mixins/MixinERC721Enumerable.sol
```diff
@@ -54,8 +54,7 @@ contract MixinERC721Enumerable is
   ) public view
     returns (uint256)
   {
-    require(_index < balanceOf(_keyOwner) && _keyOwner != address(0), 'ONLY_ONE_KEY_PER_OWNER');
-    return getTokenIdFor(_keyOwner);
+    return getKeyOfOwnerByIndex(_keyOwner, _index);
```
julien51, line 57: Maybe we can remove `tokenOfOwnerByIndex` ?

smart-contracts/contracts/mixins/MixinGrantKeys.sol
```diff
@@ -24,43 +24,19 @@ contract MixinGrantKeys is
     uint[] calldata _expirationTimestamps,
     address[] calldata _keyManagers
   ) external {
+    _onlyIfAlive();
```
julien51, line 27: Good catch!

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -31,9 +31,11 @@ contract MixinKeys is
...
   event KeyManagerChanged(uint indexed _tokenId, address indexed _newManager);
 
-  // Keys
-  // Each owner can have at most exactly one key
-  // return 0 values when missing a key
+  event KeysMigrated(
+    uint updatedRecordsCount
+  );
+
+  // DEPREC: dont use
```
julien51, line 38:
```suggestion
  // Deprecated: don't use this anymore as we know enable multiple keys per owner.
```

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -65,6 +67,18 @@ contract MixinKeys is
   // which is reset on transfer.
   mapping (address => mapping (address => bool)) private managerToOperatorApproved;
 
+  // store all keys: tokenId => token
+  mapping(uint256 => Key) private _keys;
+  
+  // store ownership: owner => list of tokenIds
```
julien51, line 73:
```suggestion
  // store ownership: owner => array of tokens owned by that owner
```

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -102,25 +131,91 @@ contract MixinKeys is
...
+        ||
+        schemaVersion == 0
+      ),
+      'SCHEMA_VERSION_NOT_CORRECT'
+    );
+
+    uint updatedRecordsCount;
+    uint recordsToUpdate = totalSupply();
+
+    // default to 100 when sent from Unlock
```
julien51, line 160:
```suggestion
    // default to 100 when sent from Unlock, as this is called by default in the upgrade script.
    // If there are more than 100 keys, the migrate function will need to be called again until all keys have been migrated.
```

smart-contracts/contracts/mixins/MixinLockCore.sol
```diff
@@ -231,5 +246,7 @@ contract MixinLockCore is
     return IERC20Upgradeable(tokenAddress).approve(_spender, _amount);
   }
 
-  uint256[1000] private __safe_upgrade_gap;
+
+  // decreased from 1000 to 999 when adding `schemaVersion` in v10
```
julien51, line 250: 👍

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -38,75 +38,66 @@ contract MixinTransfer is
   uint public transferFeeBasisPoints;
 
   /**
-  * @notice Allows the key owner to safely share their key (parent key) by
-  * transferring a portion of the remaining time to a new key (child key).
-  * @param _to The recipient of the shared key
-  * @param _tokenId the key to share
+  * @notice Allows the key owner to safely transfer a portion of the remaining time 
+  * from their key to a new keyx
```
julien51, line 42:
```suggestion
  * from their key to a new key
```

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -38,75 +38,66 @@ contract MixinTransfer is
...
+  * @param _tokenIdFrom the key to share
+  * @param _to The recipient of the shared time
   * @param _timeShared The amount of time shared
   */
   function shareKey(
```
julien51, line 47: As I am thinking about it we need a "mergeKeyInto" function that takes the time from a key and assigns to another key. Because here I cannot share key with someone who already has a key and add them time.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -174,29 +137,18 @@ contract MixinTransfer is
   }
 
   /**
-   * @notice An ERC-20 style transfer.
-   * @param _value sends a token with _value * expirationDuration (the amount of time remaining on a standard purchase).
-   * @dev The typical use case would be to call this with _value 1, which is on par with calling `transferFrom`. If the user
-   * has more than `expirationDuration` time remaining this may use the `shareKey` function to send some but not all of the token.
+   * @notice An ERC-721 style transfer.
```
julien51, line 140: That's actually an ERC20 style transfer :) as ERC721 does not have transfers.

```suggestion
   * @notice An ERC-20 style transfer.
```

julien51: We added it back in the day to make it easy to transfer from MetaMask which has ERC20 transfers.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -174,29 +137,18 @@ contract MixinTransfer is
...
+   * @param _tokenId the Id of the token to send
+   * @param _to the destination address
+   * @return success bool success/failure of the transfer
    */
   function transfer(
```
julien51, line 145: So let's keep the "value" and make it call `shareKey` because this is what it does in practice. Transfers time off of a single membership. ie: 1 => transfer the whole time, 0.5 transfer 50% of it... etc.

smart-contracts/hardhat.config.js
```diff
@@ -33,7 +33,7 @@ const { getHardhatNetwork } = require('./helpers/network')
 const settings = {
   optimizer: {
     enabled: true,
-    runs: 200,
+    runs: 90,
```
julien51, line 36: 😨

julien51 (commented):
Ok I gave a thorough look at the contracts and found a small number of minor issues.

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -102,25 +131,91 @@ contract MixinKeys is
...
-   * Get a key owned by a specific address
-   * @return The key owned by `_keyOwner`
+  * Migrate data from the previous single owner => key mapping to 
+  * the new data structure w multiple tokens.
+  * @dev when all record schemas are sucessfully upgraded, this function will update the `schemaVersion`
+  * variable to the latest/current lock version
   */
-  function getKeyByOwner(
-    address _keyOwner
+  function migrate() public {
```
julien51, line 145: So we actually need to pass a value here as a blob so we can re-use it. If we don't pass a value this will revert for locks with too many keys.

smart-contracts/contracts/Unlock.sol
```diff
@@ -312,6 +312,12 @@ contract Unlock is
...
+    lock.migrate();
```
julien51, line 319:
```suggestion
    lock.migrate(0);
```

We need to pass a value... but it will be ignored by the `migrate` function anyway

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -102,25 +131,91 @@ contract MixinKeys is
...
+        schemaVersion == 0
+      ),
+      'SCHEMA_VERSION_NOT_CORRECT'
+    );
+
+    uint updatedRecordsCount;
+    uint recordsToUpdate = totalSupply();
+
+    // default to 100 when sent from Unlock
+    uint _length;
```
julien51, line 161: We should actually cast the "blob" param to the migrate function here.

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -102,25 +131,91 @@ contract MixinKeys is
...
+    require(
+      (
+        (schemaVersion == publicLockVersion() - 1)
+        ||
+        schemaVersion == 0
+      ),
+      'SCHEMA_VERSION_NOT_CORRECT'
+    );
+
+    uint updatedRecordsCount;
```
julien51, line 157: This would be the blob?

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -102,25 +131,91 @@ contract MixinKeys is
...
+    for (uint256 i = 0; i < _length; i++) {
+      // tokenId starts at 1
+      uint tokenId = i + 1;
+      address keyOwner = _ownerOf[tokenId];
+      Key memory k = keyByOwner[keyOwner];
+
+      // make sure key exists
+      if(k.tokenId != 0 && k.expirationTimestamp != 0) {
+        // copy key in new mapping
+        _keys[i +1] = Key(k.tokenId, k.expirationTimestamp);
```
julien51, line 175:
```suggestion
        _keys[i + 1] = Key(k.tokenId, k.expirationTimestamp);
```

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -102,25 +131,91 @@ contract MixinKeys is
...
+        
+        // delete token from previous owner
+        delete keyByOwner[keyOwner];
```
julien51, line 178: should we also delete from `keyByOwner` ?

smart-contracts/test/LockSerializer/serialize.js
```diff
@@ -5,7 +5,7 @@ const compareValues = require('./_compareValues')
 const unlockContract = artifacts.require('Unlock.sol')
 const getProxy = require('../helpers/proxy')
 
-contract('LockSerializer', () => {
+contract.skip('LockSerializer', () => {
```
julien51, line 8:
```suggestion
contract('LockSerializer', () => {
```

smart-contracts/test/Unlock/migrateLockData.js
```diff
@@ -0,0 +1,368 @@
...
+    assert.equal(await lock.publicLockVersion(), pastVersion)
+  })
+
+  describe('data / schema migration with more than 100 records', () => {
+    let keyOwners
+    let tokenIds
+    let expirationTimestamps
+
+    beforeEach(async () => {
+      const [, generousBuyer] = await ethers.getSigners()
```
julien51, line 121: :D

smart-contracts/test/Unlock/migrateLockData.js
```diff
@@ -0,0 +1,368 @@
...
+
+      // create 110 random wallets
+      keyOwners = await Promise.all(
+        Array(110)
```
julien51, line 125: Ok so I think for full testing can we make an even larger number so we make sure that "pagination" works? 
Like 500? ( I assume this will bump into the gas limit?)

smart-contracts/test/Unlock/migrateLockData.js
```diff
@@ -0,0 +1,368 @@
...
+            .connect(generousBuyer)
+            .extend(0, 1, web3.utils.padLeft(0, 40), [], {
+              value: keyPrice,
+            }),
+          'MIGRATION_REQUIRED'
+        )
+      })
+    })
+
+    describe('relaunch remaining data migration', () => {
```
julien51, line 274: I want to test when we have to call this function twice... (or 3 times if we count the migration)

smart-contracts/contracts/mixins/MixinERC721Enumerable.sol
clemsos: yes, just `tokenOfOwnerByIndex` is part of standard erc721 from OZ

smart-contracts/contracts/mixins/MixinTransfer.sol
clemsos: yes we need a `mergeKeys(uint _tokenIdFrom, uint _tokenIdTo)` that will destroy the from key and add remaining time to the new one. Will do in another PR - I think will add `burn` first so we can make use of it

smart-contracts/hardhat.config.js
clemsos: yep changed back

smart-contracts/test/Unlock/migrateLockData.js
clemsos: Ok I added the tests for this

smart-contracts/test/Unlock/migrateLockData.js
clemsos: yep

clemsos:
one note about how the `migrate(bytes calldata _calldata)` currently works. As we don't know the number of keys to be migrated (no way to know how many have already been migrated), we need to pass the index of the final record to update.

That means if I pass `migrate(100)` that will upgrade the first 100 records. Then to migrate the next 100, I need to pass `migrate(200)` - that will loop from 0 to 200 and ignore the first 100 as they have already been migrated.

We have to loop through the entire array every time so we can be sure that all records have been migrated, and then update the `schemaVersion`.

it is a bit counter-intuitive but seems the only way to make this works imo.

smart-contracts/contracts/mixins/MixinTransfer.sol
clemsos: but then wont that defeat the purpose of having both `shareKey` and `transfer`? What is the goal of having both?

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -159,28 +161,31 @@ contract MixinKeys is
...
     
     // default to 100 when sent from Unlock, as this is called by default in the upgrade script.
     // If there are more than 100 keys, the migrate function will need to be called again until all keys have been migrated.
     if( msg.sender == address(unlockProtocol) ) {
-      upToIndex = 100;
+      nbRecordsToUpdate = 100;
     } else {
       // decode param
-      (upToIndex) = abi.decode(_calldata, (uint));
+      (startIndex, nbRecordsToUpdate) = abi.decode(_calldata, (uint, uint));
```
julien51, line 182: 👍

julien51 (approved):
Ship it!

PR #8010, opened 2022-01-07
refactor(unlock-app) using a globalwrapper

julien51:
# Description

Splitting the GlobalWrapper logic from https://github.com/unlock-protocol/unlock/pull/8007 for clarity.
# Issues

Fixes #
Refs #

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [x] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #8410, opened 2022-03-17
fix #8389 - keep user logged in

### LINKED ISSUE #8389: Keep the user logged in
Our checkout modal currently [does not keep the user](https://github.com/unlock-protocol/unlock/blob/master/unlock-app/src/pages/checkout.js#L19) "logged in" (i.e. it does not remember the wallet used to connect and/or the Unlock account between sessions). We need to change this. 

The reason we enabled `skipAutoLogin` is because it was creating issues when the checkout modal was configured to request the user to sign. Basically, it was skipping the request to sign even though it is required by the configuration because the signature request is too tightly coupled with the logic to connect the wallet.

So a first step should be to "isolate" the requirement to sign from the "connect" wallet. This way we can prompt the user to sign something even when they are already logged in.

A second issue will be for Unlock account users. When these users log-in we load the private key from the server and use the password to decrypt it on the front-end. The private key is then in memory for the duration of the session but if the session is terminated (tab closed), then the private key is "flushed". This means we currently cannot keep the Unlock users "logged in", even though that would be a useful step. We need to identify a way to do that.

kalidiagne:
# Description
This PR fixes problem with `autologin`, in detail
- Fix sign for a message that's shown immediately on opening the page
- If the user previously logged with `Unlock`, the login form will be shown if restart the flow 

# Issues

Fixes #8389 
Refs #

https://user-images.githubusercontent.com/20865711/159791097-2efd898f-e28b-4481-b318-bb0b8f13d6be.mov

# Checklist:

### DIFF
--- a/packages/unlock-js/src/Unlock/v10/createLock.js
+++ b/packages/unlock-js/src/Unlock/v10/createLock.js
@@ -35,7 +35,7 @@ export default async function (lock, callback) {
     maxNumberOfKeys = ETHERS_MAX_UINT
   }
   if (expirationDuration === -1) {
-    expirationDuration = ETHERS_MAX_UINT
+    maxNumberOfKeys = ETHERS_MAX_UINT
   }
 
   const decimalKeyPrice = await _getKeyPrice(lock, this.provider)
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -23,7 +23,6 @@ import {
   isNotEmpty,
   isPositiveInteger,
   isPositiveNumber,
-  isPositiveIntegerOrZero,
   isLTE,
 } from '../../utils/validators'
 
@@ -88,17 +87,9 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
     validateAndDispatch(target, [{ name, value: value * (60 * 60 * 24) }])
   }
 
-  const handleUnlimitedNumbersOfKeys = () => {
+  const handleUnlimitedClick = () => {
     dispatch({
       change: [{ name: 'maxNumberOfKeys', value: UNLIMITED_KEYS_COUNT }],
-    })
-  }
-
-  const handleUnlimitedDuration = () => {
-    dispatch({
-      change: [
-        { name: 'expirationDuration', value: ONE_HUNDRED_YEARS_IN_SECONDS },
-      ],
     })
   }
 
@@ -149,13 +140,6 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
     return ''
   }
 
-  const expirationDurationValue =
-    lockInForm?.expirationDuration === ONE_HUNDRED_YEARS_IN_SECONDS
-      ? INFINITY
-      : isPositiveIntegerOrZero(lockInForm.expirationDuration)
-      ? lockInForm.expirationDuration / (60 * 60 * 24)
-      : ''
-
   return (
     <form method="post" onSubmit={handleSubmit}>
       <FormLockRow>
@@ -175,22 +159,16 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
           </FormLockName>
           <FormLockDuration>
             <input
-              type="text"
+              type="number"
               step="1"
               inputMode="numeric"
               name="expirationDuration"
               onChange={handleChangeExpirationDuration}
-              value={expirationDurationValue}
+              defaultValue={lockInForm.expirationDuration / (60 * 60 * 24)}
               required={isNew}
               disabled={!isNew}
             />{' '}
             days
-            {lockInForm?.expirationDuration !== ONE_HUNDRED_YEARS_IN_SECONDS &&
-              isNew && (
-                <LockLabelUnlimited onClick={handleUnlimitedDuration}>
-                  Unlimited
-                </LockLabelUnlimited>
-              )}
           </FormLockDuration>
           <FormLockKeys>
             <input
@@ -205,7 +183,7 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
               required={isNew}
             />
             {lockInForm?.maxNumberOfKeys !== UNLIMITED_KEYS_COUNT && (
-              <LockLabelUnlimited onClick={handleUnlimitedNumbersOfKeys}>
+              <LockLabelUnlimited onClick={handleUnlimitedClick}>
                 Unlimited
               </LockLabelUnlimited>
             )}
--- a/unlock-app/src/components/creator/CreatorLockForm.js
+++ b/unlock-app/src/components/creator/CreatorLockForm.js
@@ -23,7 +23,6 @@ import {
   isNotEmpty,
   isPositiveInteger,
   isPositiveNumber,
-  isPositiveIntegerOrZero,
   isLTE,
 } from '../../utils/validators'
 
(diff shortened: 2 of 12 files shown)


unlock-app/src/components/interface/checkout/Checkout.tsx
```diff
@@ -147,11 +158,8 @@ export const Checkout = ({
   }
 
   const setCheckoutState = (state: string) => {
-    if (!state || state === 'connect' || state === 'loading') {
-      setShowBack(false)
-    } else {
-      setShowBack(true)
-    }
+    const showBack = !(!state || state === 'connect' || state === 'loading')
```
julien51, line 161:
```suggestion
    const showBack = state && state !== 'connect' && state !== 'loading'
```

packages/unlock-js/src/Unlock/v10/createLock.js
```diff
@@ -35,7 +35,7 @@ export default async function (lock, callback) {
     maxNumberOfKeys = ETHERS_MAX_UINT
   }
   if (expirationDuration === -1) {
-    maxNumberOfKeys = ETHERS_MAX_UINT
+    expirationDuration = ETHERS_MAX_UINT
```
julien51, line 38: Do you mind moving this to another PR please?

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -87,12 +88,20 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
     validateAndDispatch(target, [{ name, value: value * (60 * 60 * 24) }])
   }
 
-  const handleUnlimitedClick = () => {
+  const handleUnlimitedNumbersOfKeys = () => {
```
julien51, line 91: same as above :)

julien51 (commented):
Let's move the changes on the unlimited supply to another PR please?

unlock-app/src/components/interface/checkout/CardConfirmationCheckout.tsx
```diff
@@ -59,10 +59,16 @@ export const CardConfirmationCheckout = ({
     checkingRecipient,
   } = useAdvancedCheckout()
 
-  const totalPrice: number = Object.values(
-    lock.fiatPricing.usd as number
-  ).reduce((s: number, x: number): number => s + x, 0) as number
-  const fee = totalPrice - lock.fiatPricing.usd.keyPrice
+  let totalPrice: number = 0
```
julien51, line 62: Please move to another PR (and/or rebase because I think I fixed this)

kalidiagne:
PR updated. I also added a little video on the description

julien51:
Thanks Kali!

Can you please test the following just to make sure:
- Checkout URL with no signature in it and user is using crypto wallet
- Checkout URL with no signature in it and user is using unlock account
- Checkout URL with signature required (add `messageToSign` to the `paywallConfig` object) and user is using crypto wallet)
- Checkout URL with signature required (add `messageToSign` to the `paywallConfig` object) and user is using unlock account

To test I think the best way is to connect these account on the dashboard (maybe keychain?).

packages/unlock-js/src/Unlock/v10/createLock.js
```diff
@@ -35,7 +35,7 @@ export default async function (lock, callback) {
     maxNumberOfKeys = ETHERS_MAX_UINT
   }
   if (expirationDuration === -1) {
-    expirationDuration = ETHERS_MAX_UINT
+    maxNumberOfKeys = ETHERS_MAX_UINT
```
julien51, line 38: Please let's move this into another PR :)

julien51:
Oh, also can you check that it is persistent on the "oauth" flow? 
You can test with https://staging-app.unlock-protocol.com/checkout?client_id=ouvre-boite.com&redirect_uri=https://ouvre-boite.com/ or http://localhost:3000/checkout?client_id=ouvre-boite.com&redirect_uri=https://ouvre-boite.com/

julien51 (approved):
LGTM. Please do a few more cleanups and manual tests tho just to make sure we're not misisng anything!

kalidiagne:
> Thanks Kali!
> 
> Can you please test the following just to make sure:
> 
> * Checkout URL with no signature in it and user is using crypto wallet
> * Checkout URL with no signature in it and user is using unlock account
> * Checkout URL with signature required (add `messageToSign` to the `paywallConfig` object) and user is using crypto wallet)
> * Checkout URL with signature required (add `messageToSign` to the `paywallConfig` object) and user is using unlock account
> 
> To test I think the best way is to connect these account on the dashboard (maybe keychain?).

Tested all this flows, on all of this is working well! 
please share with me if you find something wrong!

unlock-app/src/components/interface/checkout/Lock.tsx
```diff
@@ -55,11 +55,7 @@ export const Lock = ({
 
   const alreadyHasKey = (key: any) => {
     const now = new Date().getTime() / 1000
-    if (key && (key.expiration === -1 || key.expiration > now)) {
-      setHasValidKey(true)
-    } else {
-      setHasValidKey(false)
-    }
+    setHasValidKey(key && (key.expiration === -1 || key.expiration > now))
```
searchableguy, line 58: nit: Split this logic. It's not very readable.

unlock-app/src/pages/_app.tsx
```diff
@@ -18,8 +18,8 @@ const UnlockApp = ({ Component, pageProps }: AppProps) => {
   }, [])
 
   return (
-    <GlobalWrapper pageProps={pageProps}>
```
searchableguy, line 21: I'm a bit worried about this. Have you checked if we pass any props to page aside from the checkout page and confirmed there is nothing dependent on it.

searchableguy (approved):
Looks fine.

unlock-app/src/components/creator/CreatorLockForm.js
```diff
@@ -176,6 +176,7 @@ const CreatorLockForm = ({ hideAction, lock, saveLock }) => {
           <FormLockDuration>
             <input
               type="text"
+              inputMode="numeric"
```
julien51, line 179: Please move to another PR

unlock-app/src/components/interface/checkout/Checkout.tsx
```diff
@@ -108,12 +115,24 @@ export const Checkout = ({
   const [existingKeys, setHasKey] = useReducer(keysReducer, {})
   const [selectedLock, selectLock] = useState<any>(null)
   const [savedMetadata, setSavedMetadata] = useState<any>(false)
+  const [storedLoginEmail, setStoredLoginEmail] = useState<string>('')
```
julien51, line 118: Maybe we could move that inside the `useAutoLogin` hook?

julien51 (approved):
Ok, I really want to test this on a draft deploy by netlify and I am excited to see it live!

unlock-app/src/pages/_app.tsx
kalidiagne: is not used in other parts, I also checked the previous PR to see when this was introduced.

ref:
[7929](https://github.com/unlock-protocol/unlock/pull/7929)
[8010](https://github.com/unlock-protocol/unlock/pull/8010)

unlock-app/src/components/interface/checkout/Checkout.tsx
kalidiagne: I defined the `storedLoginEmail` state in this component so we don't lose the value, we use the hook only to get the value

julien51:
ok let's merge please!

PR #8430, opened 2022-03-22
feat(unlock-app) - add support for key cancellation

### LINKED ISSUE #8097: Add support for key cancellation
Our PublicLock contract support 2 ways to cancel a keys:
* for a lock manager, using expireAndRefund
* for a key manager, using cancelAndRefund

We should support these 2 calls in our front-end.

The members page should add a button in front of each active/valid membership that would let the lock manager call expireAndRefund. The button should open a modal that would let the manager set the amount that they want to refund (defaults to 0).

The keychain should be updated to add a button on the keys to allow a keyManager to cancel their key. We should disable the button if the key owner is not the key manager. Additionnaly the contract has a function to indicate what amount will be refunded. We should show this in a modal to the user when they click the button and show a confirmation button.

kalidiagne:
# Description

This PR adds the support for key cancellation, specifically:

- `expire and refund` button will show members page, if the user is `keyManager`
- added `cancenl and refund` button on keyChain button, this button shows only if the key is valid
## contract ref: 
https://rinkeby.etherscan.io/address/0xf8112a74d38f56e404282c3c5071eaaed0c29b40 

# Issues

Fixes #8097 
Refs #

| Description | Screen | 
|-- |-- | 
| Modal for keyManager to "cancel and refund"   |  Screenshot 2022-03-22 at 18 48 39 | 

**Cancel and refund flow** 

https://user-images.githubusercontent.com/20865711/159965625-f026c244-2a8b-4032-bfc1-ea4ef45fcf2b.mov

# Checklist:


searchableguy:
Can we please not add another CSS processor to the mix? 

We already have styled components and tailwind. Adding saas increases the complexity we have to manage.

If possible, please use tailwind or plain css for any new code you writing.

searchableguy (approved):
LGTM

searchableguy:
Uou have useAccount hook inside the cancelAndRefund modal
which is dependent on getting web3service which is passed from the provider above.

When you render the cancelAndRefund modal, the useAccount hook doesn't receive the web3service from the context above so there is a promise error.

PR #8440, opened 2022-03-24
feat(unlock-js): Migrate to typescript

searchableguy:
# Description

Work in progress typescript migration

# Issues

Fixes #
Refs #

# Checklist:


searchableguy:
Should we convert files in PublicLock, Unlock, and __tests___, etc folder to typescript as well?

To me, it doesn't seem it's worth it. We should write new files in typescript but leave it as it is because it won't benefit an external user. All external users will use our UnlockService which provides abstraction over those functions with types.

@clemsos  @julien51 WDYT?

packages/unlock-js/src/walletService.ts
```diff
@@ -63,7 +65,7 @@ export default class WalletService extends UnlockService {
    */
   // eslint-disable-next-line no-underscore-dangle
   // TODO: Do we need this???
-  async _handleMethodCall(methodCall) {
+  async _handleMethodCall(methodCall: any) {
```
kalidiagne, line 68: this is a Promise, in the line below we are using await on `methodCall`
```suggestion
  async _handleMethodCall(methodCall: Promise<any>) {
```

searchableguy: Can also be a non promise. The distinction doesn't matter when you use any.

packages/unlock-js/src/web3Service.ts
```diff
@@ -99,17 +105,17 @@ export default class Web3Service extends UnlockService {
    * and formats it to a string of ether.
    * Returns a promise with the balance
    */
-  async getAddressBalance(address, network) {
+  async getAddressBalance(address: string, network: number) {
     const balance = await this.providerForNetwork(network).getBalance(address)
-    return utils.fromWei(balance, 'ether')
+    return utils.fromWei(balance, 'ethers')
```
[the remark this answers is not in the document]
searchableguy: Thanks. Make sense.

kalidiagne (approved):

julien51:
Agreed. If it's not external, probably not worth doing right now.

locksmith/src/routes/index.ts
```diff
@@ -22,8 +22,7 @@ router.use((request, _, next) => {
     // When the route starts with the chain (deprecated?)
     chain = parseInt(match[1])
   } else if (request.query?.chain) {
-    // @ts-expect-error
-    chain = parseInt(request.query.chain)
+    chain = parseInt(String(request.query.chain))
```
kalidiagne, line 25: @searchableguy there is a possibility that this gives `NaN` if `request.query.chain` has a value thas is not a number 
```suggestion
    const value = String(request.query.chain)
    chain = !isNaN(value) ? parseInt(value) : undefined
```

what do you think?

searchableguy: Looks okay.

julien51 (approved):

PR #8449, opened 2022-03-29
Ccarfi guildxyz telegram

ccarfi:
token gating telegram with guild blog post


julien51:
Looks very good to me! I wonder if we should also add that as a tutorial on our end in the docs? Abs with links on the dashboard for easy setup.

julien51 (approved):

Issue #8380, opened 2022-03-10
Contract code potentially too large


clemsos:
As identified with latest `PublicLock` rewriting, the contract size is beginning to be too large - see for instance https://github.com/unlock-protocol/unlock/pull/8337 

Error thrown by hardhat when attempting deployment: `Error: Transaction reverted: trying to deploy a contract whose code is too large`

clemsos:
some reporting using https://github.com/ItsNickBarry/hardhat-contract-sizer

```
|  Contract Name                  ·  Size (KB)  ·  Change (KB)  │
 ··································|·············|················
 |  TestNoop                       ·      0.062  ·        0.062  │
 ··································|·············|················
 |  Initializable                  ·      0.063  ·        0.063  │
 ··································|·············|················
 |  Roles                          ·      0.085  ·        0.085  │
 ··································|·············|················
 |  SafeMath                       ·      0.085  ·        0.085  │
 ··································|·············|················
 |  CountersUpgradeable            ·      0.086  ·        0.086  │
 ··································|·············|················
 |  ECDSAUpgradeable               ·      0.086  ·        0.086  │
 ··································|·············|················
 |  MathUpgradeable                ·      0.086  ·        0.086  │
 ··································|·············|················
 |  Roles                          ·      0.086  ·        0.086  │
 ··································|·············|················
 |  SafeCastUpgradeable            ·      0.086  ·        0.086  │
 ··································|·············|················
 |  UnlockUtils                    ·      0.086  ·        0.086  │
 ··································|·············|················
 |  console                        ·      0.086  ·        0.086  │
 ··································|·············|················
 |  Clone2Factory                  ·      0.086  ·        0.086  │
 ··································|·············|················
 |  AddressUpgradeable             ·      0.086  ·        0.086  │
 ··································|·············|················
 |  CountersUpgradeable            ·      0.086  ·        0.086  │
 ··································|·············|················
 |  StringsUpgradeable             ·      0.086  ·        0.086  │
 ··································|·············|················
 |  TimersUpgradeable              ·      0.086  ·        0.086  │
 ··································|·············|················
 |  Address                        ·      0.086  ·        0.086  │
 ··································|·············|················
 |  Counters                       ·      0.086  ·        0.086  │
 ··································|·············|················
 |  StorageSlot                    ·      0.086  ·        0.086  │
 ··································|·············|················
 |  Strings                        ·      0.086  ·        0.086  │
 ··································|·············|················
[...]

PR #8452, opened 2022-03-29
feat(smart-contracts): support for recurring memberships

### LINKED ISSUE #8323: Recurring Key Purchase
In order to support automatically recurring memberships, we need to later our public lock contract. (no change expected in Unlock).

* _snapshotting_ of the price + period. We expect users to trigger the first purchase on automatically recurring memberships. It is critical as a signal that they _agree_ to the price + frequency of payments. For this we should then alter the `purchase` function so that it stores the `latestPurchasePriceAndDuration` for the user. We should store this as a number of wei (or similar `decimals` for ERC20 per second). 
* Introduction of a new function `renewMembershipFor` that will take the following arguments:
   * uint tokenId: the token to be renewed/extended
   * address referrer: the referrer of the transaction (eligible to receive UDT) 
  
   This function will then verify the following things:
   * if the lock is an ERC20 lock
   * if the owner of the tokenId has approved enough ERC20 for the renewal
   * if the tokenId is nearly expired (less than 5% of the default duration) or expired
   * if the current lock price and duration are still equal or lower than the `latestPurchasePriceAndDuration` performed by the owner of the tokenId.
   
   After this, it will behave similarly to the `purchase` function (in the context of an extension!) including extension of the timestamp (of course!), calling back the Unlock contract for UDT minting + GDP recording, as well as issue the gas refund. The keyManager should not be changed as the balance of funds is taken from the beneficiary themselves.

We assume no tipping on the renewal transactions, so no need to pass the `value` that the user is willing to pay" they are "charged" the current keyPrice.

clemsos:
# Description

Add a `renewMembershipFor` function to support recurring memberships.

TODO: rebase when #8453 is merged

# Issues

Fixes #8323
Refs #

# Checklist:


clemsos:
```Error: Transaction reverted: trying to deploy a contract whose code is too large```
😢

**Comment by julien51** [comment]
Dammit. Any idea on how you save the shark amount that we need to save?

File: smart-contracts/contracts/mixins/MixinPurchase.sol [hunk 1]
```diff
@@ -190,6 +213,42 @@ contract MixinPurchase is
...
+    uint _tokenId,
+    address _referrer
+  ) public {
+    _onlyIfAlive();
+    _lockIsUpToDate();
+    _isKey(_tokenId);
+
+    // check the lock
+    require(_originalDurations[_tokenId] != type(uint).max, 'NON_EXPIRING_LOCK');
+    require(_originalPrices[_tokenId] != 0, 'FREE_LOCK');
```
julien51, line 232: I think we should enable free locks to be automatically renewed too

clemsos:
> Dammit. Any idea on how you save the shark amount that we need to save?

with cleanup happening in #8457, we can pass the bat but this will keep happening #8380

clemsos:
contract too large again....

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -188,14 +211,51 @@ contract MixinPurchase is
...
+  */
+  function renewMembershipFor(
+    uint _tokenId,
+    address _referrer
+  ) public {
+    _lockIsUpToDate();
+    _isKey(_tokenId);
+
+    // check the lock
+    require(_originalDurations[_tokenId] != type(uint).max, 'NON_EXPIRING_LOCK');
```
julien51, line 228: maybe we can skip this to save on size?

julien51 (approved):

clemsos:
So I had to lower the optimizer runs from 200 to 80 make this PR pass the contract size limit. 

Also I had to add another PR #8541 to refactor the mock contracts so it does not import the entire `PublicLock` contract - that was making tests fail as it exceeds authorized code size. This PR can be rebased once #8541 is merged.

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: Yes it was some misconceptions about free locks not needed to be renewed but thats not necessarily the case

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: yes this will revert by overflow anyway as we will add to max uint. Let's keep it tho - explicit is always better

julien51:
🤞

PR #8461, opened 2022-03-30
fix(smart-contracts): make `_data` an array in lock purchase function

### LINKED ISSUE #8446: `data` for `purchase` should be an array
As I was working on the purchaseHook I realized that we should make sure that `data` is also an array of the same size as the other aguments and the hook should be called with the rigth item in the array.

clemsos:
# Description

# Issues

Fixes #8446
Refs 

# Checklist:


smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.5.17 <0.9.0;
+pragma experimental ABIEncoderV2;
```
julien51, line 3: Hum? 🧹

clemsos: Nope, has to keep. It is needed up to 0.8.x to support bytes array in syntax (the `_data` args is one)

julien51 (approved):

julien51 (approved):

PR #8469, opened 2022-03-31
fix(smart-contracts): use defaults for metadata

clemsos:
# Description

This commit modifies how the `baseTokenURI` and `baseSymbol` are fetched by setting defaults value (from Unlock) at lock creation instead of calling it on each `tokenURI` / `symbol` call.

This makes code lighter and more gas efficient, but also will prevent all lock URIs from being updated in case the main Unlock config changed. 

In case the main Unlock `globalTokenBaseURI` changes, lock managers will have to call `lock.setBaseTokenURI('')` (with an empty string) from their lock to reset to the latest base URI from Unlock.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/UnlockUtils.sol
+++ b/smart-contracts/contracts/UnlockUtils.sol
@@ -7,17 +7,6 @@ pragma solidity >=0.5.17 <=0.8.7;
 
 library UnlockUtils {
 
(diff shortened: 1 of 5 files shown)


julien51:
Let's archvive that for now...

PR #8575, opened 2022-04-06
fix(unlock-app) No more generate lock address

julien51:
# Description

Since v9 we are not able to generate addresses for locks ahead of time as they are deployed as proxies.
This PR removes that logic and fixes handling of "pending" locks.

# Checklist:


unlock-app/src/components/lock/Icon.js
```diff
@@ -172,8 +172,13 @@ IconModal.propTypes = {
...
   current: PropTypes.string.isRequired,
-  lockAddress: PropTypes.string.isRequired,
-  network: PropTypes.number.isRequired,
+  lockAddress: PropTypes.string,
+  network: PropTypes.number,
+}
+
+IconModal.defaultProps = {
+  lockAddress: '',
+  network: 0,
```
clemsos, line 181: is that related to this PR?

packages/unlock-js/src/unlockService.ts
```diff
@@ -139,7 +139,9 @@ export default class UnlockService {
       const contractVersion = await contract.publicLockVersion()
       version = parseInt(contractVersion, 10) || 0
     } catch (error) {
-      // This is an older version of Unlock which did not support publicLockVersion
+      console.error(
+        `We could not retrieve the version of the Unlock contract ${address} on this network.`
```
clemsos, line 143: what is the case where we could not retrieve the lock version? All locks have this, no ?

julien51: Generally I think it would be a misconfiguration... and I bumped into that yesterday :) Adding the error makes it not silent!

clemsos (approved):
lgtm

PR #8579, opened 2022-04-06
feat(unlock-js): support for Lock v10

clemsos:
# Description

This adds support for the latest `PublicLock` version for `purchaseKey` and the the new `extendKey`, as well as a bunch of there new features

# Issues

Fixes #
Refs #

# Checklist:


packages/unlock-js/src/PublicLock/v10/index.js
```diff
@@ -1,5 +1,6 @@
 import abis from '../../abis'
 import purchaseKeys from './purchaseKeys'
+import purchaseKey from './purchaseKey'
```
clemsos, line 3: I added both batch and single purchase of keys. The idea was to make it simpler on frontend. Mot sure if both are needed tho ?

julien51: That's a good question.
From my perspective, I want to make it as easy on the front-end to buy 1 or multiple keys for _all_ lock versions, so ideally we will move everything to just use `purchaseKeys` everywhere but of course it will just work for older version if there is a single key and would call `purchaseKey` internally.

julien51:
So, I think it would be useful to add `purchaseKeys` to all older versions (you should only have to add it once for the oldest supported version) and just make sure that it fails if there is more than 1 key or call `purchaseKey` with the single key.

Also, I think we should add tests for merge and extend?

julien51:
So, I think it would be useful to add `purchaseKeys` to all older versions (you should only have to add it once for the oldest supported version) and just make sure that it fails if there is more than 1 key or call `purchaseKey` with the single key.

Also, I think we should add tests for merge and extend?

clemsos:
need to rebase once #8586 is fixed

packages/unlock-js/src/PublicLock/v10/purchaseKeys.js
```diff
@@ -0,0 +1,185 @@
...
+        this.signer
+      )
+    }
+  }
+
+  // tx options
+  const purchaseForOptions = {}
+  if (erc20Address && erc20Address !== ZERO) {
+    // Since we sent the approval transaction, we cannot rely on Ethers to do an estimate, because the computation would fail (since the approval might not have been mined yet)
+    // purchaseForOptions.gasLimit = 400000
```
clemsos, line 116: with this gas limit, the transaction reverts

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -398,7 +398,7 @@ contract MixinKeys is
     _isKey(_tokenIdTo);
     
     // make sure there is enough time remaining
-    require(keyExpirationTimestampFor(_tokenIdFrom) - block.timestamp > _amount, 'NOT_ENOUGH_TIME');
+    require(keyExpirationTimestampFor(_tokenIdFrom) - block.timestamp >= _amount, 'NOT_ENOUGH_TIME');
```
julien51, line 401: huh. Good catch at the last minute!

packages/contracts/src/contracts/PublicLock/PublicLockV10.sol
```diff
@@ -2845,7 +2844,7 @@ contract MixinKeys is
     _isKey(_tokenIdTo);
     
     // make sure there is enough time remaining
-    require(keyExpirationTimestampFor(_tokenIdFrom) - block.timestamp > _amount, 'NOT_ENOUGH_TIME');
+    require(keyExpirationTimestampFor(_tokenIdFrom) - block.timestamp >= _amount, 'NOT_ENOUGH_TIME');
```
julien51, line 2847: pfew

packages/unlock-js/src/web3Service.ts
```diff
@@ -153,6 +153,67 @@ export default class Web3Service extends UnlockService {
     )
   }
 
+  /**
+   * Returns the key to the lock by the token Id.
+   * @param {PropTypes.string} lockAddress
+   * @param {PropTypes.number} tokenId
+   */
+  async getKeyByTokenId(lockAddress: string, tokenId: string, network: number) {
```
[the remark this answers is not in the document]
julien51: Will that just work for v10 though?

packages/unlock-js/src/PublicLock/v4/purchaseKeys.js
```diff
@@ -0,0 +1,45 @@
...
+    owners = [],
+    keyManagers = [],
+    keyPrices = [],
+    referrers = [],
+    data = [],
+  },
+  callback
+) {
+  return await Promise.all(
+    owners.map(async (owner, i) =>
```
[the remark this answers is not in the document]
clemsos: I had to move that logic to a wrapper because for ERC20 the allowance call was failing so we need to approve the entire amount before calling `purchaseKey`. Works properly now :)

packages/unlock-js/src/web3Service.ts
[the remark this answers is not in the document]
clemsos: yes it will throw if the version is < 10

julien51:
🎉

PR #8584, opened 2022-04-07
feat(locksmith): Add token revocation and move to jwt

searchableguy:
# Description

This PR moves to jwt instead of using session cookies for allowing more flexibility and add the ability to revoke tokens. 

The user will be issued a short-lived access token (a few minutes) and a refresh token which is revocable. The user can use the refresh token to get a new access token. The user or we have the ability to find refresh tokens associated with a wallet address and revoke them which will force the user to log in again to get a new refresh token. 

# Issues

Fixes #
Refs #

# Checklist:


julien51 (approved):

locksmith/package.json
```diff
@@ -49,6 +49,7 @@
     "apollo-server": "3.6.7",
     "apollo-server-express": "2.25.3",
     "color-scheme": "1.0.1",
+    "cookie-parser": "^1.4.6",
```
kalidiagne, line 52: @searchableguy need to pin version here?

locksmith/package.json
```diff
@@ -90,13 +91,15 @@
     "zod": "3.14.3"
   },
   "devDependencies": {
+    "@types/cookie-parser": "^1.4.2",
```
kalidiagne, line 94: @searchableguy need to pin the version here?

kalidiagne (approved):

PR #8596, opened 2022-04-11
feat(unlock-app) - fix expire and refund button and error message

kalidiagne:
# Description
Expire and refund button will not e shown if the key is already expired

# Issues

Fixes #
Refs #

# Checklist:


kalidiagne:
> I think you had similar check to isKeyValid elsewhere. See if you could reduce the duplicate code if there is.
> 
> Otherwise LGTM

Agree with you, we have a LOT! of this type of check, I'm thinking about writing a helper for that, also for other duplicated functions.

julien51 (approved):

PR #8605, opened 2022-04-12
feat(locksmith): Use lockAddress as tokenAddress in user metadata and other changes

searchableguy:
# Description

This fixes an earlier instance of using key ID in building the tokenAddress. 

Based on discussions, get user metadata route is removed.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/metadataController.ts
+++ b/locksmith/src/controllers/v2/metadataController.ts
@@ -19,11 +19,43 @@ const BulkUserMetadataBody = z.object({
   users: z.array(UserMetadataBody),
 })
 
+interface IsKeyOrLockOwnerOptions {
+  userAddress?: string
+  lockAddress: string
+  keyId: string
+  network: number
+}
+
 export class MetadataController {
   public web3Service: Web3Service
 
   constructor({ web3Service }: { web3Service: Web3Service }) {
     this.web3Service = web3Service
+  }
+
+  async #isKeyOrLockOwner({
+    userAddress,
+    lockAddress,
+    keyId,
+    network,
+  }: IsKeyOrLockOwnerOptions) {
+    if (!userAddress) {
+      return false
+    }
+    const loggedUserAddress = Normalizer.ethereumAddress(userAddress)
+    const isLockOwner = await this.web3Service.isLockManager(
+      lockAddress,
+      loggedUserAddress,
+      network
+    )
+
+    const keyOwner = await this.web3Service.ownerOf(lockAddress, keyId, network)
+
+    const keyOwnerAddress = Normalizer.ethereumAddress(keyOwner)
+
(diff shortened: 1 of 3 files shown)


julien51 (commented):
That looks good!

locksmith/src/routes/v2/metadata.ts
```diff
@@ -33,24 +33,20 @@ router.put(
   }
 )
 
-router.get(
-  '/:network/locks/:lockAddress/users/:userAddress/keys/:keyId',
-  (req, res) => {
-    metadataController.getUserMetadata(req, res)
-  }
-)
+router.get('/:network/locks/:lockAddress/users/:userAddress', (req, res) => {
```
julien51, line 36: I think we can get rid of that endpoint.

locksmith/src/controllers/v2/metadataController.ts
```diff
@@ -86,7 +84,7 @@ export class MetadataController {
 
   async getKeyMetadata(request: Request, response: Response) {
     try {
-      const { keyId } = request.params
+      const keyId = request.params.keyId.toLowerCase()
```
julien51, line 87: keyId should be a number? Why the lower casing?

locksmith/src/controllers/v2/metadataController.ts
```diff
@@ -170,37 +122,33 @@ export class MetadataController {
...
-      } else {
-        const [updatedLockMetadata, success] = await LockMetadata.upsert(
-          {
-            address: lockAddress,
-            chain: network,
-            data: {
-              ...metadata,
-            },
+      }
+      const [updatedLockMetadata, created] = await LockMetadata.upsert(
```
julien51, line 138: are we doing `upsert` even if this is a `updateLockMetadata` call?

locksmith/src/controllers/v2/metadataController.ts
```diff
@@ -211,63 +159,44 @@ export class MetadataController {
 
   async updateKeyMetadata(request: Request, response: Response) {
     try {
-      const { keyId } = request.params
+      const keyId = request.params.keyId.toLowerCase()
```
julien51, line 162: same q as above

julien51 (approved):
2 small questions and then LG!

locksmith/src/controllers/v2/metadataController.ts
searchableguy: This is how it is done on the previous endpoint. I wasn't sure if the key were ever going to be non-integer so copied it from there.

locksmith/src/controllers/v2/metadataController.ts
julien51: I thought we said we'd change but that's fine either way.

PR #8614, opened 2022-04-14
feat(locksmith): Improve error handling for v2 endpoints

searchableguy:
# Description

This improves the error message for bulk user metadata endpoint and standardizes the error response to JSON across other v2.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/metadataController.ts
+++ b/locksmith/src/controllers/v2/metadataController.ts
@@ -71,14 +71,16 @@ export class MetadataController {
       })
 
       if (!lockData) {
-        return response.status(404).send('No lock metadata found.')
+        return response.status(404).send({
+          message: 'No lock metadata found.',
+        })
       }
       return response.status(200).send(lockData.data)
     } catch (error) {
       logger.error(error.message)
-      return response
-        .status(500)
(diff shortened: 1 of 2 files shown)


locksmith/src/controllers/v2/metadataController.ts
```diff
@@ -235,10 +235,14 @@ export class MetadataController {
         return response.status(201).send(createdUserMetadata.data)
       }
 
-      return response.status(409).send('User Metadata already exists.')
+      return response.status(409).send({
+        message: "User metadata doesn't exist.",
+      })
```
kalidiagne, line 240: @searchableguy in this `409` case, the error message shouldn't be `[...] already exists`  ?

searchableguy: I believe `409` is the correct HTTP status code for a resource that already exists and cannot perform updates. 

Do you have another suggestion?

kalidiagne: so we need can change it to `User Metadata already exists.` (as before) instead of `User metadata doesn't exist` 👍

kalidiagne (approved):

PR #8623, opened 2022-04-20
feat(locksmith): Add support for buying multiple keys via stripe

searchableguy:
# Description

This PR adds support for buying multiple keys via stripe checkout. 

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/operations/lockOperations.ts
```diff
@@ -115,7 +115,8 @@ export async function updateLockMigrationsLog(
 
 export async function isSoldOut(
   address: string,
-  chain: number
+  chain: number,
+  keysNeeded = 10
```
kalidiagne, line 119: @searchableguy default value for `keysNeeded` shout not be equal to zero?

kalidiagne (approved):

locksmith/migrations/20220420120142-add-recipients-field-to-charge.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+module.exports = {
```
julien51, line 3: I think we should not use "charges" anymore, but that's fine for this PR.

locksmith/migrations/20220420111523-add-recipients-field-to-payment-intent.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn('PaymentIntents', 'recipients', {
+      type: Sequelize.ARRAY(Sequelize.STRING),
```
julien51, line 6: Let's use the `Sequelize.JSON` instead for compliance.

julien51 (approved):
Outside of the quick data model change I think it's looking good!

locksmith/migrations/20220420120142-add-recipients-field-to-charge.js
searchableguy: We do use it to save the details so I adapted it.

locksmith/src/operations/lockOperations.ts
searchableguy: The default is high to provide a reasonable estimation of *are there enough keys for me to buy?*

locksmith/migrations/20220420111523-add-recipients-field-to-payment-intent.js
searchableguy: okay

PR #8638, opened 2022-04-22
feat(unlock-app) - add sign message with metadata

kalidiagne:
# Description
Require to sign message before metadata request

# Issues

Fixes #
Refs #

# Checklist:


searchableguy:
I'm not sure why we want user to sign when there is no need for it to store metadata because we don't check for it in our endpoint.

kalidiagne:
@searchableguy `submitBulkRecipients` is called only in case of metadata,  added  before with this PR https://github.com/unlock-protocol/unlock/pull/8636/files

searchableguy:
Okay, I confirmed with @julien51 regarding what he was confused about. This PR is not needed but we need to work on making sure that user knows they can continue with saving metadata if there is an error first time. It is not 100% clear that user can click on the metadata button again after getting an error so maybe change name of the button next time or provide a better message.

Issue #8656, opened 2022-04-26
Backend and front-end support for automatic renewals (aka subscriptions)


julien51:
Now that all of our contracts have been upgraded on all chains, we can work on supporting the automatically recurring memberships on all networks!

## Backend 

We will leverage the `websub` application created by @searchableguy to add a polling mechanism that will run every 5 minutes (as it already does!) that will do the following:
* for each network, identify the keys that have expired in the last **15 minutes** (picking a larger window lets us make sure we do not miss any key).
* for each key, identify if it can be renewed. We first need to retrieve the lock's version. Anything below 10 should be skipped. Then, the best way to do that is to send a transaction to estimate gas for the `renewMembership` function. If the gas estimate fails, we should skip and try the next one. If the gas estimate succeeds, we should look at the gas cost. We should only renew keys for which we are getting a gas refund whose value cover the gas spent. We will query the lock to get the gas refund and check that it covers the actual gas spent. If yes, we renew. If not, we skip.

## Front-end

We will introduce a new param on the checkout URL paywallConfig, which we will call `approveAmount` and whose value is a number. It will represent the total amount required for the ERC20 approval. The checkout flow should include a message in the confirmation screen about the fact that the total approved amount will be larger than the actual cost, allowing for renewals.
Finally, we should then change the `purchaseKey` to take the `approveAmount` as an option. If it is unset, we need will use the `actualAmount` computed.

## Testing

Let's deploy a lock on Polygon with a key price of 1 USDC, a gas refund of 3ct and make sure we enable automatically recurring memberships. I will create a recipient address for which I will add 300 USDC and let's set this address as the beneficiary of the lock (so that the funds circulate back!)

julien51:
cc @ccarfi

PR #8671, opened 2022-04-27
feat(websub): add worker to renew expired key 

clemsos:
# Description

This PR creates a job to check keys that just expired and send a `lock.renewMembershipFor` tx automatically renew the key

# Issues

Relates to #8656 
Refs #

# Checklist:


locksmith/src/graphql/datasource/keysToRenew.ts
```diff
@@ -0,0 +1,56 @@
...
+    page: number
+  ): Promise<any[]> {
+    this.baseURL = networks[network].subgraphURI
+    const keysToRenewQuery = gql`
+      query Keys($start: Int!, $end: Int!, $skip: Int) {
+        keys(
+          skip: $skip
+          orderBy: expiration
+          orderDirection: desc
+          where: { expiration_gte: $start, expiration_lte: $end }
```
clemsos, line 19: @searchableguy do you know if it is possible to do use subquery in a where clause in graphQL? 

I would like to do sth like (as only Lock after v10 support recurring memberships)

```gql
where: { 
   expiration_gte: $start 
   expiration_lte: $end 
   lock {
    version_gte: 10
  }
}
```

searchableguy: Nope. The current subgraph doesn't have any such filter for the lock version on keys. Your best bet filtering the keys with the v10 lock version or query all the v10 locks, then add all the keys array together.

locksmith/src/websub/jobs/renewKeys.ts
```diff
@@ -0,0 +1,99 @@
+import { Op } from 'sequelize'
+import { networks } from '@unlock-protocol/networks'
+import { KeysToRenew } from '../../graphql/datasource'
+import { KeyRenewal } from '../../models'
+import { renewKey } from '../helpers'
+import { logger } from '../../logger'
+
+const FETCH_LIMIT = 25
+
+async function fetchKeysToRenew(network: number, page = 0) {
```
julien51, line 10: Mind adding a few tests for this please?

clemsos:
ok after a day of investigating, it turns out there is bug in the gas refund function https://github.com/unlock-protocol/unlock/issues/8679

clemsos:
hmm seems that all existing tests are passing but still having a FAIL exit code on CI. Did it ever happen to you before when developing websub @searchableguy ?

searchableguy:
@clemsos Because test coverage is low. You need to add more tests to cover the code you are adding. 

😄

julien51:
Please @clemsos do not decrease coverage. We need to add tests for now. We need to increase our confidence.

clemsos:
> Please @clemsos do not decrease coverage. We need to add tests for now. We need to increase our confidence.

yep sure. Some integration tests with a running node will be required for more extensive testing, so was thinking of working on integration tests on another PR. 

For now I have added a simple mock function for the renewal function which may increase coverage back to its level before.

locksmith/src/graphql/datasource/keysToRenew.ts
```diff
@@ -0,0 +1,62 @@
...
+    const first = 100 // 100 by page
+    const skip = page * first
+
+    try {
+      const response = await this.query(keysToRenewQuery, {
+        variables: { start, end, first, skip },
+      })
+
+      // return only locks that support recurring memberships
+      const renewableLocks = response.data.keys.filter(
```
julien51, line 52: 👍

locksmith/src/websub/helpers/renewKey.ts
```diff
@@ -0,0 +1,130 @@
...
+import { KeyRenewal } from '../../models'
+import { purchaserCredentials } from '../../../config/config'
+
+interface RenewKeyParams {
+  keyId: number
+  lockAddress: string
+  network: number
+}
+
+const abi = [
```
julien51, line 13: Let's not use the `abi` here but unlockJS which is the ability to return the lock contract at a specific address so that we don't have to maintain ABIs in multiple places.

locksmith/src/websub/helpers/renewKey.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  // We only renew keys for which we are getting a gas refund whose value cover the gas spent.
+  const gasRefund = await lock.gasRefundValue()
+
+  // estimate gas for the renewMembership function.
+  const gasLimit = await lock.estimateGas.renewMembershipFor(
+    keyId,
+    constants.AddressZero
+  )
+  if (gasRefund.lte(gasLimit)) {
```
julien51, line 88: That won't work because `gasLimit` is in "native currency" (ether on mainnet, matic on polygon...) vs `gasRefund` is in ERC20 tokens.

We need to do a conversion to $ of these 2.

Generally I think we should isolate that logic in its own function `isWorthRenewing(lock, keyId)` so that we can also "overwrite" its behavior. For example in the early days we should be fine as long as we don't spend more than say $0.03

locksmith/src/websub/helpers/renewKey.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  // We only renew keys for which we are getting a gas refund whose value cover the gas spent.
+  const gasRefund = await lock.gasRefundValue()
+
+  // estimate gas for the renewMembership function.
+  const gasLimit = await lock.estimateGas.renewMembershipFor(
+    keyId,
+    constants.AddressZero
```
julien51, line 86: Let's make sure we use an Unlock address there :) I am thinking we could add the Unlock multisig addresses per network in the network package?

locksmith/src/websub/helpers/renewKey.ts
```diff
@@ -0,0 +1,130 @@
...
+}: RenewKeyParams) {
+  // super complicated parsing to make ts happy ;-)
+  const [, networkConfig]: [string, NetworkConfig] = Object.entries(
+    networks
+  ).find(([, n]) => n.id === network) as [string, NetworkConfig]
+
+  // get RPC connection and signer
+  const { provider } = networkConfig
+  const rpc = new ethers.providers.JsonRpcProvider(provider)
+  const signer = new Wallet(purchaserCredentials, rpc)
```
julien51, line 123: Can we just reuse the dispatcher class please?

locksmith/src/websub/helpers/renewKey.ts
clemsos: yep added

locksmith/src/websub/helpers/renewKey.ts
clemsos: the dispatcher class rely on unlock-js but `renewMembershipFor` hasn't been added to it. Should I add it to unlock-js?

locksmith/src/websub/jobs/renewKeys.ts
clemsos: yep

clemsos:
Here is a Rinkeby lock that uses WEENUS erc20 for testing purposes

https://rinkeby.etherscan.io/address/0x88a2f53df06084377a3afd32e791004ccfcf03c6#readProxyContract

locksmith/src/websub/helpers/renewKey.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+interface ShouldRenew {
+  shouldRenew: boolean
+  gasRefund?: string
+  error?: string
+}
+
+// precision base for gas calculations
+const BASE = 1000000
```
julien51, line 22: Can you use a better name so we understand what this is for wthout looking at the code please :)

locksmith/src/websub/helpers/renewKey.ts
```diff
@@ -0,0 +1,138 @@
...
+
+const getGasFee = async (network: number) => {
+  const pricer = new KeyPricer()
+  const gasPrice = await pricer.gasFee(network, BASE)
```
julien51, line 26: This function is currently computing the gas fee using the hardcoded gas cost/limit for a purchase. It would be useful to change how this function works so it can take the gas cost/limit as a param so we can re-use it.

Ideally we should also remove the "hardcoded" gas cost...

julien51 (commented):
Progress! I think there are still a few things left to fix on the gas side.

julien51 (approved):
Let's ship it!

PR #8672, opened 2022-04-27
feat(unlock-app) - toggle show metadata

kalidiagne:
# Description
Multiple recipients with multiple fields for metadata can be very long, we need a way to prevent this. To solve this issue this PR introduce: 
- By default only the user address is show in the recipients list
- User can she/hide metadata values using button `show metata/hide metata`
- The buttons to show metata will be visible only when we have metata on recipients form 

# Issues

Fixes #
Refs #
### issue
Screenshot 2022-04-27 at 15 21 44
### solution 

https://user-images.githubusercontent.com/20865711/165529464-893f6de1-11eb-4895-84fe-e9efc2f8438d.mov

# Checklist:

### DIFF
--- a/unlock-app/src/__tests__/components/interface/checkout/MetadataForm.test.tsx
+++ b/unlock-app/src/__tests__/components/interface/checkout/MetadataForm.test.tsx
@@ -37,6 +37,8 @@ const recipientsObj = {
   recipients: [],
   loading: false,
   maxRecipients: 1,
+  minRecipients: 1,
+  hasMinimumRecipients: true,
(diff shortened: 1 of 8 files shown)


julien51:
Looking good!

searchableguy (approved):
While this is working well, a more isolated approach where the individual metadata rendering component is responsible for the toggle state would be preferable. With the current approach, we are rendering the whole tree on any single change and it increases the complexity of composing the component elsewhere.

julien51 (approved):

PR #8765, opened 2022-05-04
feat(unlock-app): add support for recurring payment

### LINKED ISSUE #8656: Backend and front-end support for automatic renewals (aka subscriptions)

clemsos:
# Description

This PR adds support for recurring payment (and ERC20 approval) to crypto checkout workflow

### Checkout

Screenshot 2022-05-05 at 14 02 39

### Dashboard (before enabling)

Screenshot 2022-05-05 at 13 46 25

### Dashboard (enabled)
Screenshot 2022-05-05 at 14 03 12

# Issues

Fixes #8656
Refs #

# Checklist:


clemsos:
splitted out unlock-js modif here #8767

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -119,6 +129,13 @@ export const CreatorLock = ({
     )
   }
 
+  const handleApproveRecurring = () => {
+    const MAX_ALLOWANCE = '10000000000000000'
```
julien51, line 133: Let's use MAXUINT just to make sure. We have the constant already.

unlock-app/src/components/creator/lock/AppStore.js
```diff
@@ -118,6 +118,16 @@ const AppStore = ({ lock }) => {
       persistentCheckout: true,
       icon: `${config.services.storage.host}/lock/${lock.address}/icon`,
     }
+
+    if (
+      lock.publicLockVersion >= 10 &&
+      lock.currencyContractAddress &&
+      lock.selfAllowance > 100000
```
julien51, line 125: Max u int :)

unlock-app/src/hooks/useLock.ts
```diff
@@ -197,7 +235,8 @@ export const purchaseKeyFromLock = async (
   referrer: string,
   setLock: (...args: any) => void,
   data: string,
-  callback: (...args: any) => void
+  callback: (...args: any) => void,
+  recurringPayments?: number
```
julien51, line 239: Let's make sure we always use the callback as the last argument.

unlock-app/src/hooks/useLock.ts
```diff
@@ -261,13 +301,15 @@ export const purchaseMultipleKeysFromLock = async (
   keyPrices: string[],
   owners: string[],
   data: string[],
-  callback: (...args: any) => void
+  callback: (...args: any) => void,
+  recurringPayments?: number[]
```
julien51, line 305: See above :)

clemsos:
@ccarfi here is the updated version

Screenshot 2022-05-05 at 18 55 02

julien51:
Let's move the unlock-js changes to that other PR please so this one is easier to review?

packages/types/src/types/unlockTypes.ts
```diff
@@ -146,6 +146,7 @@ export interface PaywallConfig {
   metadataInputs?: MetadataInput[]
   persistentCheckout?: boolean
   useDelegatedProvider?: boolean
+  approveAmount?: number
```
julien51, line 149: Can we move this to inside the ` PaywallConfigLocks` please ?

clemsos:
> Let's move the unlock-js changes to that other PR please so this one is easier to review?

Yes sure. The unlock-js changes are split in #8780 and #8767

packages/unlock-js/src/PublicLock/utils/parseLockGetters.js
```diff
@@ -66,9 +67,20 @@ export default async function (address, provider) {
     )
     const erc20Symbol = await getErc20TokenSymbol(update.tokenAddress, provider)
 
+    // get lock allowance of itself (for v10 recurring)
+    const erc20LockAllowance = await getAllowance(
```
julien51, line 71: Since this is only for v10 can we move that into the v10's `getLock` so that we don't slow things down for other locks?

unlock-app/src/components/creator/CreatorLock.js
```diff
@@ -74,9 +75,18 @@ export const CreatorLock = ({
...
+    updateKeyPrice,
+    updateMaxNumberOfKeys,
+    withdraw,
+    updateSelfAllowance,
+  } = useLock(lockFromProps, network)
+
+  const [isRecurring, setIsRecurring] = useState(
+    lock.publicLockVersion >= 10 &&
+      lock.currencyContractAddress &&
+      lock.selfAllowance === MAX_UINT
```
julien51, line 89:
```suggestion
     (lock.publicLockVersion && lock.selfAllowance === MAX_UINT)
```

julien51 (approved):
ok LG

julien51 (approved):

clemsos:
ok fixed the latest little things and tested again, seems to work fine 

Here is another recurring WEENUS lock with a 1-day duration for testing purposes on Rinkeby : https://rinkeby.etherscan.io/address/0x2f1F33BdD092EAc32A0a9cB92aBD81A29486B129#readProxyContract

Issue #8663, opened 2022-04-26
Epic: Ticket checkin refactored!


julien51:
This is going to be a big refactor for the ticketing use-case of Unlock. We may open smaller Github issues for items that deserve it and we should definitely open multiple small Pull Requests to cover this.

**The overarching goal is to improve our ticket's check-in experience.**

For reference, here is the current flow: 
1. The user goes to the keychain and clicks on the icon to generate a QR code for a specific key
2. They sign a message
3. This displays a QR code that can be scanned by anyone
4. The Verification page is loaded and shows info about the ticket, including whether it has been verified _previously_.
5. If the scanner is a lock manager and the ticket has not been verified previously, they can click a button to "mark as checked-in".
6. If the lock manager clicks the button, they are prompted to sign a message 
7. Ticket is marked as checked-in in our data store.

We want to take the same flow but improve multiple aspects and reduce friction.

## Adding the "verifier" role

The first step is to add the role of "verifier". The goal of this new role is to avoid requiring the "admin" role of lock manager for people who are marking tickets as checked in. 
To do this, we will have to perform the following things:
* [x] Add a verifier table to locksmith database with the following fields: `address`, `lockAddress`, `network` (for the lock), `lockManager` (address of the lock manager that granted that role), `createdAt`.
* [x] Add an API endpoint for a lock manager to list all verifiers for a specicifc lock address
* [x] Add an API endpoint for a lock manager to add a verifier to a lock
* [x] Add an API endpoint for a lock manager to remove a verifier for a lock
* [x] Add a new page on the `unlock-app` that will provide a UI to list, add and remove verifier for a specific lock. This page will be at `/verifiers?lockAddress=0x....` and will use the SIWE authentication to only show data to a lock manager.
All of these API endpoints must use the v2 auth framework implemented by @searchableguy  (SIWE).

## Signature of tickets/NFT is done by locksmith

In order to simplify the user flows, we want to add the ability to send the QR codes to users without requiring them to connect to the keychain first. As such, we need to : 
* [x] add an API endpoint to locksmith that will generate a signature for a payload that can be used to create ticket QR code  for a specific lock address and a specific token id. This API can only be called by the specific key owner.
* [x] #9053
* [ ] #9031

## Allowing Verification of tickets signed by locksmith's purchaser address

Right now, the verification only succeeds if the signer is the owner of the NFT. 
[...]

PR #8785, opened 2022-05-10
feat(locksmith) - ticketing check-in with verifier

kalidiagne:
# Description
In order to get add a new role "Verifier", people who are marking tickets as checked in we need to provided news API's 
**ref**: https://github.com/unlock-protocol/unlock/issues/8663 

* [x] Add a verifier table to locksmith database with the following fields: `address`, `lockAddress`, `network` (for the lock), `lockManager` (address of the lock manager that granted that role), `createdAt`.
* [x] Add an API endpoint for a lock manager to list all verifiers for a specicifc lock address
* [x] Add an API endpoint for a lock manager to add a verifier to a lock
* [x] Add an API endpoint for a lock manager to remove a verifier for a lock
* [x] Add a new page on the `unlock-app` that will provide a UI to list, add and remove verifier for a specific lock. 
* [x] All of these API endpoints must use the v2 auth framework

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -0,0 +1,185 @@
...
+import { Verifier } from '../../models/verifier'
+
+export class VerifierController {
+  public web3Service: Web3Service
+
+  constructor({ web3Service }: { web3Service: Web3Service }) {
+    this.web3Service = web3Service
+  }
+
+  private async checkIsLockManager(
```
searchableguy, line 14: Don't use `private` keyword for methods in typescript. If you want to make a method private, use `#` while defining the name. The private keyword in typescript doesn't offer any guarantee of the method not being accessible publicly.

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -0,0 +1,185 @@
...
+      return response.status(201)
+    } catch (error) {
+      logger.error(error.message)
+      return response.status(500).send({
+        message: 'There were some problems in getting the lock data.',
+      })
+    }
+  }
+
+  private async isVerifierAlreadyExits(
```
searchableguy, line 41: Ditto as above.

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -0,0 +1,185 @@
...
+      )
+
+      const alreadyExists = await this.isVerifierAlreadyExits(
+        lockAddress,
+        verifierAddress,
+        loggedUserAddress,
+        network
+      )
+
+      if (alreadyExists !== null) {
```
searchableguy, line 117: I would use `!alreadyExists` instead as a check because that would also protect against cases of undefined values.

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -0,0 +1,185 @@
...
+      } else {
+        await Verifier.destroy({
+          where: {
+            lockAddress,
+            verifierAddress,
+            lockManager: loggedUserAddress,
+            network,
+          },
+        })
+        response.status(201)
```
searchableguy, line 175: Return a `204` which implies empty after deletion.

searchableguy (changes requested):
LGTM. 

A few changes and tests need to be added.

You need to also add the newly created route into the index.ts route file for it to actually work.

locksmith/src/controllers/v2/verifierController.ts
julien51: Also I think we may already have a function for this in the code base? Probably better to isolate and move to a helper file?

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  private async checkIsLockManager(
+    lockAddress: string,
+    loggedUserAddress: string,
+    network: number,
+    response: Response
+  ) {
+    try {
+      const isLockOwner = await this.web3Service.isLockManager(
```
julien51, line 21:
```suggestion
      const isLockManager = await this.web3Service.isLockManager(
```

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -0,0 +1,185 @@
...
+        lockAddress,
+        loggedUserAddress,
+        network
+      )
+
+      if (!isLockOwner) {
```
julien51, line 27:
```suggestion
      if (! isLockManager) {
```

locksmith/src/controllers/v2/verifierController.ts
searchableguy: We do have a similar function for the metadata lock endpoint because only the lock manager can operate on that. A middleware abstracting the logic would be good, I think.

locksmith/migrations/20220510110232-create-verifier.js
```diff
@@ -0,0 +1,36 @@
...
+      },
+      lockAddress: {
+        type: Sequelize.STRING
+      },
+      network: {
+        type: Sequelize.NUMBER
+      },
+      lockManager: {
+        type: Sequelize.STRING
+      },
```
searchableguy, line 22: Just to double confirm here - default is allowNull. If you don't want anyone to push null on any of the fields there while creating the column, please set to `allowNull` to true.

searchableguy:
Don't forget to add the partial route in the routes file here https://github.com/unlock-protocol/unlock/blob/master/locksmith/src/routes/index.ts

locksmith/src/models/verifier.ts
```diff
@@ -0,0 +1,16 @@
+import { Table, Column, Model } from 'sequelize-typescript'
+
+@Table({ tableName: 'Verifiers', timestamps: true })
+export class Verifier extends Model<Verifier> {
+  @Column
+  verifierAddress!: string
+
+  @Column
+  userAddress!: string
```
julien51, line 9: I am confused here. what is the `userAddress`? Same as `verifierAddress` ?

julien51: I think we are missing `lockAddress` ?

locksmith/src/controllers/v2/verifierController.ts
julien51: That is a great idea @kalidiagne ^^

locksmith/src/models/verifier.ts
kalidiagne: added `lockAddress` and `verifierAddress` -> `address`

thk!

searchableguy:
Looks good to me. Need tests however

locksmith/src/routes/v2/verifier.ts
```diff
@@ -0,0 +1,30 @@
...
+import { lockManagerMiddleware } from '../../utils/lockManager'
+
+const router = express.Router({ mergeParams: true })
+const web3Service = new Web3Service(networks)
+
+const verifierController = new VerifierController({
+  web3Service,
+})
+
+router.get('/list/:network/:lockAddress', lockManagerMiddleware, (req, res) =>
```
searchableguy, line 14: You can put middleware on "/" instead and it will apply it to all routes that come after so you won't need to repeat it.

You should also add authenticatedOnly middleware since your API depends on authenticated users.

locksmith/__tests__/controllers/v2/verifierController.test.ts
```diff
@@ -0,0 +1,67 @@
+import { ethers } from 'ethers'
+import request from 'supertest'
+
+const app = require('../../../src/app')
+
+jest.setTimeout(600000)
+const lockAddress = '0x3F09aD349a693bB62a162ff2ff3e097bD1cE9a8C'
+
+describe('Verifier v2 endpoints for locksmith', () => {
```
searchableguy, line 9: You need to login and use the authorization header. 

Also only user Address which is a lock manager can be used here afaik.

locksmith/migrations/20220510110232-create-verifier.js
```diff
@@ -0,0 +1,40 @@
...
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable('Verifiers', {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER
+      },
+      address: {
```
julien51, line 11: We might want to add an index here?

locksmith/migrations/20220510110232-create-verifier.js
```diff
@@ -0,0 +1,40 @@
...
+        type: Sequelize.STRING,
+        allowNull: false
+      },
+      lockAddress: {
```
julien51, line 15: We might want to add an index here?

locksmith/migrations/20220510110232-create-verifier.js
```diff
@@ -0,0 +1,40 @@
...
+        type: Sequelize.STRING,
+        allowNull: false
+      },
+      network: {
+        type: Sequelize.INTEGER,
+        allowNull: false
+      },
+      lockManager: {
```
julien51, line 23: We might want to add an index here?

julien51 (approved):
Let's add a few indices please to make sure the queries are not too slow!

unlock-app/src/__tests__/utils/exportStatic.test.js
```diff
@@ -26,6 +26,7 @@ describe('exportStatic', () => {
       '/verification': { page: '/verification' },
       '/members': { page: '/members' },
       '/checkout': { page: '/checkout' },
+      '/verifiers': { page: '/verifiers' },
```
searchableguy: Remove from this PR

julien51: fixed :)

locksmith/src/routes/v2/verifier.ts
julien51: Nice! @kalidiagne fixed this :)

PR #8786, opened 2022-05-10
feat(unlock-app): Use next eslint rules

searchableguy:
# Description

This changes eslint rules with what nextjs recommends. The old config lacked important rules regarding the usage of hooks in react and a11y (I've fixed all errors and some warnings). There are still many warnings left.

https://reactjs.org/docs/hooks-rules.html 

The config package we have cannot be used as an extension without ignoring a significant number of rules. There are conflicts with typescript and prop usage as well. I've removed it in the short term because it needs changes to be used on the frontend.

This is consistent with what we have in unlock static site which also uses nextjs recommended rules.

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Hum, can we move that to `@unlock-protocol/eslint-config` so we have a single config for all of the apps please?

searchableguy:
We can eventually but right now, if I do that. I would have to fix all the other stuff which will have error as a result of adding new constraints.

searchableguy:
Closing in favor of https://github.com/unlock-protocol/unlock/pull/8849

PR #8791, opened 2022-05-11
cancelling toast messages if the promised is resolved in less than 300ms

### LINKED ISSUE #8783: Unnecessary toast pop up if user is already connected
**Describe the bug**

**To Reproduce**
1. Connect with your wallet once. 
2. Reload the dashboard page after you are connected.
3. You should see an unnecessary toast. 

Video: https://www.loom.com/share/ea5135f717e1483a81a337e083ddb469

**Expected behavior**
Toast pop up should not happen. It should delay the pop up by checking local storage to make sure it doesn't fire unnecessarily.

julien51:
# Description

In looking at #8783 I found that we probably show too many messages in general (not just for that use case). I decided to debounce all toast messages if they have been "triggered" in less than 300ms. My heuristic is that in that scenario it's probably not worth showing a message to the user.

For this I moved to using the Helper rather than the library directly. This also provides better isolation... should we want to replace that dependency one day!

Fixes #8783

# Checklist:

### DIFF
--- a/unlock-app/src/components/helpers/toast.helper.ts
+++ b/unlock-app/src/components/helpers/toast.helper.ts
@@ -38,7 +38,11 @@ export const ToastHelper: ToastHelperProps = {
   success: (message) => toast.success(message),
   error: (message) => toast.error(message),
   promise: async (promise, msgs, opts = {}) => {
+    const start = new Date().getTime()
     await toast.promise(promise, msgs, opts)
+    if (new Date().getTime() - start < 300) {
+      toast.remove() // This cancels the toast immediately
+    }
   },
   // TODO: we need to provide an errors pages 404/500
   redirectErrorPage: (page) => {
--- a/unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
+++ b/unlock-app/src/components/creator/members/GrantKeysDrawer.tsx
@@ -1,6 +1,5 @@
 import React, { useEffect, useState, useContext } from 'react'
 import { useForm } from 'react-hook-form'
-import { toast } from 'react-hot-toast'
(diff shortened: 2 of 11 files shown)


unlock-app/src/components/helpers/toast.helper.ts
```diff
@@ -38,7 +38,11 @@ export const ToastHelper: ToastHelperProps = {
   success: (message) => toast.success(message),
   error: (message) => toast.error(message),
   promise: async (promise, msgs, opts = {}) => {
+    const start = new Date().getTime()
     await toast.promise(promise, msgs, opts)
+    if (new Date().getTime() - start < 300) {
+      toast.remove() // This cancels the toast immediately
+    }
```
kalidiagne, line 45: not sure we remove the toast. before this line we have `await`, until the promise is not resolve we still see the toast with `loading` message

kalidiagne (approved):
LGTM! just leave a comment

searchableguy:
I'm not sure this is helpful. It will be more confusing for the user to have all messages they depend on gone suddenly after one action because toast can load while waiting for user approval.

The implementation is confusing to me. What your code will do right now is remove all previous toasts if the promise toasts is successful instantly. If the intention is to limit the number of toasts on screen, then there are better ways to approach it via limiting the number of toasts displayed directly.

julien51:
> I'm not sure this is helpful. It will be more confusing for the user to have all messages they depend on gone suddenly after one action because toast can load while waiting for user approval.

They are only gone if they would have been triggered for less than 300ms. My assumption is that anything that takes less than 300ms does not need a "visual" indication that it happened because the user triggered it.

PR #8830, opened 2022-05-17
fix(locksmith) fixing gas computations for renewals

### LINKED ISSUE #8803: locksmith: gasRefund should call oracle to check ERC20 value in USD
**Describe the bug**

The pricing of a renewable lock is done in ERC20 and the `setGasRefundValue` sets a value ti be refunded in the token of the lock. In the websub task, when we do check the gasPrice from the actual network to evaluate the cost of renewal (in USD) but, instead of checking against the lock's `gasRefund` price in USD, we use directly the ERC20 value. 

**To Reproduce**
1. Create a renewable lock priced in ERC20
2. Buy a key
3. Set gas refund in ERC20
4. Wait for the key to expire
5. Launch websub
6. Key is renewed even if gasPrice in USD is higher than ERC20 gas refund

**Expected behavior**
The key should be renewed only if the cost is lower or equals to gas refund

julien51:
# Description

Fixing gas computations + referrer

# Issues

Fixes #8803

# Checklist:

### DIFF
--- a/locksmith/src/fulfillment/dispatcher.ts
+++ b/locksmith/src/fulfillment/dispatcher.ts
@@ -150,9 +150,7 @@ export default class Dispatcher {
   async renewMembershipFor(
     network: number,
     lockAddress: string,
-    keyId: number,
-    referrer: string,
-    transactionOptions?: any
+    keyId: number
   ) {
     const walletService = new WalletService(networks)
     const provider = new ethers.providers.JsonRpcProvider(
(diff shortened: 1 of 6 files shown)


locksmith/src/fulfillment/dispatcher.ts
```diff
@@ -163,16 +161,13 @@ export default class Dispatcher {
       config.purchaserCredentials,
       provider
     )
+
     await walletService.connect(provider, walletWithProvider)
 
     // get lock
     const lock = await walletService.getLockContract(lockAddress)
 
-    // make sure reccuring payments are supported
```
julien51, line 171: This check happened earlier... no need to check again. (simulation would have failed)

PR #8849, opened 2022-05-20
feat(all): Simplify eslint and add support for next in the eslint-config-package

searchableguy:
# Description

Based on the discussion and conclusion in the meeting yesterday, I've proceeded with the following

1. Remove multiple different conflicting styles (eg - airbnb, standard) and use default recommended set.
2. Add typescript-recommended set which is provided by eslint-typescript plugin which we didn't add before.
3. Create a separate `next` config file for frontend rule storage. This extends from our base config file but adds frontend specific nextjs rules which won't make sense in other contexts.
4. Fix errors and reduce the number of errors by changing some ts rules to warn for now.

This PR should fix any errors but adds a lot of warnings which we could slowly fix over (mostly ts-related). 

# Issues

Fixes #
Refs #

# Checklist:


smart-contracts/scripts/lock/clone.js
```diff
@@ -11,7 +11,6 @@ async function main({
   const PublicLock = await ethers.getContractFactory('PublicLock')
   const lock = PublicLock.attach(lockAddress)
 
-  // eslint-disable-next-line no-console
```
clemsos, line 14: I like the `no-console` rule has it enforced we keep code clean. wdyt?

searchableguy: I don't think it will add value. We are using console in every package and app. Someone will end up disabling the rule to add a console log like they have everywhere already. 

Console log is relatively harmless so I think if we want it somewhere, explicitly enable it in eslint rules for that project.

clemsos: the idea behind making the use of console log explicit is to make sure it was intended, so we dont forget any dev-related logs that could potentially leak credentials - and keep tests clean as they sometimes are reviewed less thoroughly so logs tends to remain there. 

Anyways, not a dealbreaker just some thoughts

searchableguy: I will add a warning then (no error).

PR #8859, opened 2022-05-25
feat(unlock-app) - Login and authorized endpoints

kalidiagne:
# Description
Add login and authorize endpoints
- This PR is a result of small other PR's opened before to add login and authorize verification endpoints

https://user-images.githubusercontent.com/20865711/171684046-53a85b2e-33f6-409a-bbcf-69b35c840959.mov

# Issues

Fixes #
Refs #

# Checklist:


packages/unlock-js/src/locksmithService.ts
```diff
@@ -71,10 +71,11 @@ export class LocksmithService {
    */
   async login(message: string, signature: string) {
     const headers = new Headers()
-    headers.set('content-type', 'application/json')
     const response = await this.request('/v2/auth/login', {
       method: 'POST',
-      headers,
+      headers: {
+        'Content-Type': 'application/json',
+      },
```
searchableguy, line 78: I think you can keep the old code here. Just import Headers from `import fetch, { Headers } from 'cross-fetch'`. It is a polyfill issue.

unlock-app/src/services/storageService.ts
```diff
@@ -57,6 +58,26 @@ export class StorageService extends EventEmitter {
...
+    const siweMessage = LocksmithService.createSiweMessage({
+      uri: 'https://locksmith.unlock-protocol.com',
+      domain: this.host,
+      address,
+      chainId,
+      version,
+      statement: 'Authorize',
+    })
+    return siweMessage.prepareMessage()
+  }
```
searchableguy, line 71: I would not use more than 2 positional arguments.

unlock-app/src/services/storageService.ts
```diff
@@ -57,6 +58,26 @@ export class StorageService extends EventEmitter {
...
 
+  async getSiweMessage(address: string, chainId: number, version = '1') {
+    const siweMessage = LocksmithService.createSiweMessage({
+      uri: 'https://locksmith.unlock-protocol.com',
+      domain: this.host,
+      address,
+      chainId,
+      version,
+      statement: 'Authorize',
+    })
```
searchableguy, line 69: You need to provide a unique nounce every time. 

```ts
import { generateNounce } from "siwe"
const nounce = generateNounce()
const siweMessage = locksmithService.createSiweMessage({ nounce, ...rest })
```

unlock-app/src/hooks/useToken.ts
```diff
@@ -0,0 +1,5 @@
+import { useAppStorage } from './useAppStorage'
+
+export const useToken = () => {
+  const storage = useAppStorage()
```
searchableguy: Do you think you can use the same package I've used in unlock-protocol-com site. There should be one use-localstorage-hook or something for consistency. We should get rid of the this appStorage hook at some point because it doesn't deal with some edge cases in browser refocus and storage events. 


It will also be code we don't need to maintain ourselves.

packages/unlock-js/src/locksmithService.ts
```diff
@@ -1,7 +1,6 @@
-import fetch from 'cross-fetch'
+import fetch, { Headers } from 'cross-fetch'
```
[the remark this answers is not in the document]
searchableguy: Not in this instance.

Cross-fetch is a compatibility layer which switches to node-fetch when operated in node environment and fetch when in browser.

This library can be used in either environment hence cross-fetch is used.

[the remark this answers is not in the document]
kalidiagne: agree! removed from here, in all the other implemented parts i use native `fetch`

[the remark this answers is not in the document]
searchableguy: That would make the library not work in node context anymore, kali because fetch doesn't exist natively until node 18.

unlock-app/src/components/interface/verifiers/VerifiersList.tsx
```diff
@@ -61,12 +88,18 @@ export const VerifiersList: React.FC<VerifiersListProsps> = ({
   const onConfirmDeleteVerifier = async () => {
     setLoading(true)
     const isValid = ethers.utils.isAddress(selectedVerifier)
+
     try {
       if (isValid) {
-        const addVerifierUrl = `/verifier/${network}/${lockAddress}/${selectedVerifier}`
+        const addVerifierUrl = getEndpoint(
```
julien51, line 94: I wonder if this logic (the whole request) should be inside of the LocksmithService? wdyt @searchableguy

julien51 (approved):
I like it! 
One question and I think it would be great to "document" how to use the same pattern in other places in the app, but I think we can also merge as it does not touch "existing" code/features?

searchableguy (changes requested):
I think I need to discuss the approach here with julien. Can you hold off merging until then?

unlock-app/src/components/interface/verifiers/VerifiersList.tsx
searchableguy: Eventually, yes. For now, I will move this to storage service. 

I plan to migrate all methods from storage service to locksmith service and directly pass the instance of locksmith service from the library. That way, everyone can take advantage of it.

julien51 (approved):
:shipit:

searchableguy (approved):

PR #8860, opened 2022-05-25
fix(subgraph): handling upgraded locks

julien51:
# Description

Warning: this is a short-term fix and a pretty limited one but it solves the issue we have.
I think it also provides a way to see how we can rewrite for our subgraph in a way that would support for upgrades natively!

# Checklist:


clemsos (approved):
LGTM

PR #8870, opened 2022-05-30
fix(smart-contracts): remove ts rule to avoid `require`

clemsos:
# Description

All smart-contracts tests and scripts uses `require` so far so that rule creates noise

# Issues

Fixes #
Refs #

# Checklist:


searchableguy (approved):
Seems fine. I added this to warn because wanted us to move to import everywhere else but feels like es import will take time to become common in node.

clemsos:
Yes, perfect. The goal is to have import in smart contracts tooling too

PR #8966, opened 2022-06-03
feat(unlock-app) - add verifier check into verification page

kalidiagne:
# Description
This PR is part of https://github.com/unlock-protocol/unlock/issues/8663, in the specific adds: 

-  if the user loading the application is a verifier and/or lock manager, load and show the metadata associated with the key and the owner's address
   - add new endpoints to check if a viewer is a verifier of a specific lock 

- [x]  show the actual NFT image
Screenshot 2022-06-14 at 19 25 28

- [x] Includes  SIWE integration in order to avoid asking the verifier to sign messages over and over. Make sure the "Mark as checked-in button" is functional for Lock Manager and Verifiers.
- [x]  if the user loading the application is a verifier and/or lock manager, load and show the metadata associated with the key and the owner's address (as shown on the member list UI)
- [x] Add a new endpoint to locksmith that marks a ticket/key as checked in. It can only be called by a verifier (auth error otherwise)
- [ ] We should make it succeed if the signer is either the owner of the NFT or the locksmith's purchaser address.
- [ ]  Additionally the signed message should include the token Id, as we now support owning multiple tokens per address.

# Issues

Fixes #
Refs https://github.com/unlock-protocol/unlock/issues/8663

# Checklist:


unlock-app/src/services/storageService.ts
```diff
@@ -87,31 +88,38 @@ export class StorageService extends EventEmitter {
...
   get token() {
-    return this.accessToken
+    return this.accessToken || sessionStorage.getItem('token')
   }
 
   async refreshToken(token: string) {
     return this.locksmith.refreshToken(token)
   }
 
+  async loginPrompt({ walletService, address, chainId }: LoginPromptProps) {
```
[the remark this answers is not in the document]
searchableguy: You don't need to do this though. Refresh token is attached as an http cookie. 

If you don't have a token, simply call refresh token endpoint again. If that returns a fail, then prompt and login again. 

That said, this works as well.

locksmith/src/routes/v2/verifier.ts
```diff
@@ -25,4 +26,17 @@ router.delete(
   (req, res) => verifierController.removeVerifier(req, res)
 )
 
+router.get(
+  '/enabled/:network/:lockAddress/:verifierAddress',
+  userMatchAddress,
+  (req, res) => verifierController.isVerifierEnabled(req, res)
+)
+
+router.put(
```
julien51, line 35: please move to ticket controller

locksmith/src/routes/v2/verifier.ts
```diff
@@ -25,4 +26,17 @@ router.delete(
...
+  '/:network/lock/:lockAddress/key/:keyId/check',
+  lockManagerMiddleware,
```
julien51, line 37: It needs to check if the user is a verifier, not a lock manager

locksmith/src/routes/v2/verifier.ts
```diff
@@ -25,4 +26,17 @@ router.delete(
...
+  (req, res) => {
+    verifierController.markTicketAsCheckIn(req, res)
```
julien51, line 39: Move to tickets controller

locksmith/src/routes/v2/verifier.ts
julien51: Let's use a isVerifierMiddleware

locksmith/src/utils/middlewares/userMatchAddress.ts
```diff
@@ -0,0 +1,14 @@
+import { RequestHandler } from 'express'
+import Normalizer from '../../utils/normalizer'
+
+export const userMatchAddress: RequestHandler = async (req, res, next) => {
```
julien51, line 4: We don't need this

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -56,6 +57,35 @@ export default class VerifierController {
     )
   }
 
+  async #isVerifier({
```
julien51, line 60: Keep in separate PR

locksmith/src/routes/v2/verifier.ts
julien51: Put that in a dedicated PR

unlock-app/src/services/storageService.ts
```diff
@@ -564,6 +572,28 @@ export class StorageService extends EventEmitter {
     })
   }
 
+  async setKeyMetadataWithVerifier({
```
julien51, line 575: and call it `markTicketAsCheckedIn`

unlock-app/src/services/storageService.ts
```diff
@@ -677,4 +707,10 @@ export class StorageService extends EventEmitter {
       return false
     }
   }
+
+  getLockImage(lockAddress: string): string {
```
julien51, line 711: Move to its own PR?

julien51: Let's use the same approach as the keychain does

julien51: create a util if needed that both keychain + verifier UI will use

PR #8968, opened 2022-06-03
feat(unlock-app) - add more specific parameters check for verifier page

kalidiagne:
# Description
- add more specific parameters check for verifier page
Screenshot 2022-06-03 at 16 21 04

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/pages/[network]/locks/[lock]/verifiers.js
+++ b/unlock-app/src/pages/[network]/locks/[lock]/verifiers.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import { useRouter } from 'next/router'
-import VerifiersContent from '../../../components/content/VerifiersContent'
+import VerifiersContent from '../../../../components/content/VerifiersContent'
 
 const Verifiers = () => {
   const { query } = useRouter()


julien51 (changes requested):
Maybe we should use toasts so it is consistent in all of the application?

searchableguy:
There is a middleware for valid network somewhere. You can use that.

searchableguy:
> @searchableguy change verifier page URL from query to parameters as suggested, thk!

Need one more param for network in the url as well. We want to move away from current system of switching network to everything networks usable at once.

kalidiagne:
> > @searchableguy change verifier page URL from query to parameters as suggested, thk!
> 
> Need one more param for network in the url as well. We want to move away from current system of switching network to everything networks usable at once.

@searchableguy changed with this `/verifiers/[network]/[lock]` path

I'm wondering if this does not cause  issues, for example, the case when the current network in metamask doesn't match the network in the parameters

searchableguy:
It won't

kalidiagne:
> It won't

Actually sure! no issue because is done via API!
if all is okay for you please approve it so I can merge, thk!

searchableguy:
Hm, do you think using `/[network]/locks/[lock]/verifiers` as path?

kalidiagne:
> Hm, do you mind using `/[network]/locks/[lock]/verifiers` as path?

Done!

searchableguy (approved):

julien51 (approved):

PR #8969, opened 2022-06-03
feat(locksmith) - add lock manager support in verifiers endpoint

kalidiagne:
# Description
This PR adds
- add more security by checking for verifiers endpoints
  -  if logged user is `lockManager`, can use the endpoints without any un-authorized error message 
- Spinner on loading when when `/verifiers/list/[...]` is retrieving data
- endpoint to check a specific verifier for and address ( defined in `isVerifierEnabled` )

# Issues

Fixes #
Refs #

Screenshot 2022-06-03 at 19 12 13
# Checklist:


locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -82,6 +111,19 @@ export class VerifierController {
         request.user!.walletAddress!
       )
 
+      const isLockManager = await this.#isLockManager({
```
julien51, line 114: Good catch!

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -156,4 +212,37 @@ export class VerifierController {
...
+
+      const isVerifier = await Verifier.findOne({
+        where: {
+          lockAddress,
+          address,
+          network,
+        },
+      })
+
+      const isLockManager = await this.#isLockManager({
```
julien51, line 231: Maybe do that only if `isVerifier` is `false`?

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -156,4 +212,37 @@ export class VerifierController {
       })
     }
   }
+
+  // check is address is a Verifier of a specific lock
+  async isVerifierEnabled(request: Request, response: Response) {
```
julien51, line 217: This is good, and we will need to make sure that the `request.params.verifierAddress` matches the token on the API request too!

julien51 (approved):

locksmith/src/controllers/v2/verifierController.ts
searchableguy: Wasn't this handled in a middleware?

kalidiagne: oh, you're right!

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -156,4 +173,49 @@ export class VerifierController {
...
+    try {
+      const lockAddress = Normalizer.ethereumAddress(request.params.lockAddress)
+      const address = Normalizer.ethereumAddress(request.params.verifierAddress)
+      const network = Number(request.params.network)
+
+      const verifierMatchesLoggedUser = request.user?.walletAddress === address
```
julien51, line 184: That is not right (I think my comment was misleading!). We need to verify that `request.user?.walletAddress` is a lock manager on the lock...

julien51, line 184: Actually I think my comment was misleading, sorry :/

julien51 (changes requested):
I am sorry for the confusion here.
I see multiple problems. First `getVerifiersList` should return __all__ the verifiers, not just the ones matching the lockManager` (I actually think we should not use `lockManager` in that function).

Then the `list` function should check that `request.user!.walletAddress!` is indeed a `lockManager` and fail if not. (only a lock manager should be able to see the list of verifiers)
Sinilarly the functions to add/remove verifiers can only be called by a lock manager.

Finally, I am not sure what `isVerifierEnabled` is used for?

kalidiagne:
> I am sorry for the confusion here. I see multiple problems. First `getVerifiersList` should return **all** the verifiers, not just the ones matching the lockManager`(I actually think we should not use`lockManager` in that function).
> 
> Then the `list` function should check that `request.user!.walletAddress!` is indeed a `lockManager` and fail if not. (only a lock manager should be able to see the list of verifiers) Sinilarly the functions to add/remove verifiers can only be called by a lock manager.
> 
> Finally, I am not sure what `isVerifierEnabled` is used for?

@julien51  Changed `getVerifiersList ` as suggested, without the use of `lockManager`

Also
- All the routes use `lockManagerMiddleware`, so only lockManager can call it.
- `isVerifierEnabled` is needed for #8966, basically to check if the user is a verifier of a specific lock, and eventually to enable it on the verification page

locksmith/src/controllers/v2/verifierController.ts
```diff
@@ -156,4 +160,49 @@ export class VerifierController {
...
+
+      if (!verifierMatchesLoggedUser) {
+        return response.status(401).send({
+          message: `User not authozired`,
```
julien51, line 175:
```suggestion
          message: `User not authorized`,
```

julien51:
> @julien51 Changed `getVerifiersList ` as suggested, without the use of `lockManager`

Thank you!

> * All the routes use `lockManagerMiddleware`, so only lockManager can call it.

Oh wow, I had missed that! Thank you, that clarifies it...

> * `isVerifierEnabled` is needed for [feat(unlock-app) - add verifier check into verification page #8966](https://github.com/unlock-protocol/unlock/pull/8966), basically to check if the user is a verifier of a specific lock, and eventually to enable it on the verification page

Ok I misunderstood what that function was for but now I think I get it! 

Thanks! Approved!

PR #8979, opened 2022-06-06
feature(unlock-protocol-com) : adding dynamic featured users

julien51:
# Description

Adding a way to dynamically load featured users!
Ideally in a follow up PR we add transitions but if we do this we need to find a mechanism to ensure images are all of the same size.

# Checklist:


searchableguy:
You can object fit images to be same size. No need to depend on the original size. The only thing to keep in mind is similarish aspect ratio and good resolution.

searchableguy:
https://tailwindcss.com/docs/object-fit

along with width and height className and aspect/overflow for containment if too big.

searchableguy:
You forgot to add width to the image for consistency.  :)

Also I wouldn't do transition using useEffect. Would be good to use embla and Transition component from headless in later iteration.

unlock-protocol-com/src/components/pages/Home/sections/Connect.tsx
```diff
@@ -10,7 +11,34 @@ const UNLOCK_BENEFITS = [
...
+
 export function Connect() {
+  const [featured, setFeatured] = useState(1)
+
+  useEffect(() => {
+    setInterval(() => {
+      setFeatured((featured + 1) % featuredUsers.length)
+    }, 5000)
+  })
+
```
searchableguy, line 41: no clean up? This will cause problems.

PR #9002, opened 2022-06-09
feat(locksmith) - fix verifier endpoint middleware and tests

kalidiagne:
# Description
 - fix verifier endpoint middleware and tests

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/routes/v2/verifier.ts
```diff
@@ -7,18 +7,22 @@ const router = express.Router({ mergeParams: true })
 
 const verifierController = new VerifierController()
 
-router.all('/', authenticatedMiddleware, lockManagerMiddleware)
+router.use('/', authenticatedMiddleware)
```
julien51, line 10: It looks like all routes use `lockManagerMiddleware`. Why not keeping it there too?

locksmith/src/utils/lockManager.ts
```diff
@@ -5,31 +5,37 @@ import Normalizer from './normalizer'
...
-  if (!isLockManager) {
-    return res.status(401).send({
-      message: `${lockManager} is not a lock manager for ${lockAddress} on ${network}`,
+    if (!isLockManager) {
+      return res.status(401).send({
+        message: `${lockManager} is not a lock manager for ${lockAddress} on ${network}`,
+      })
+    }
+  } catch (err) {
+    return res.status(404).send({
```
julien51, line 37: Let's use `422`

julien51 (changes requested):
That's a good start but you are missing some tests for the "happy" path: when the user is indeed a lock manager

kalidiagne:
> That's a good start but you are missing some tests for the "happy" path: when the user is indeed a lock manager

Added more test

locksmith/__tests__/controllers/v2/verifierController.test.ts
```diff
@@ -1,86 +1,122 @@
...
-      walletAddress: address,
-      message,
-      signedMessage,
-    } = await getWalletInput()
-    const loginResponse = await request(app).post('/v2/auth/login').send({
-      signature: signedMessage,
-      message: message.prepareMessage(),
-    })
-    expect(loginResponse.status).toBe(200)
+jest.mock('../../../src/controllers/v2/verifierController', () => {
```
julien51, line 23: I don't think you need to mock the verifier... Since these are your actual tests!

locksmith/__tests__/controllers/v2/verifierController.test.ts
```diff
@@ -1,86 +1,122 @@
...
+
+  return jest.fn().mockImplementation(() => {
+    return mockDispatcher
+  })
+})
 
-    const addVerifierResponse = await request(app)
-      .put(`/v2/api/verifier/${network}/${lockAddress}/${address}`)
-      .set('authorization', `Bearer ${loginResponse.body.accessToken}`)
+jest.mock('../../../src/utils/lockManager', () => {
```
julien51, line 39: Same here you don't want to mock the middleware... because you want to make sure it does what you expect it to do.

locksmith/src/routes/v2/verifier.ts
searchableguy: hm, that's odd. I will check.

locksmith/__tests__/controllers/v2/verifierController.test.ts
```diff
@@ -1,86 +1,95 @@
...
 
-    const {
-      walletAddress: address,
-      message,
-      signedMessage,
-    } = await getWalletInput()
+  it('Get list items from lock with random address (not lockManager)', async () => {
+    expect.assertions(2)
+    const lock = await ethers.Wallet.createRandom().getAddress()
+    const { message, signedMessage } = await getWalletInput()
```
julien51, line 34: Ok you can use `loginRandomUser` from [test-helpers/utils.ts](https://github.com/unlock-protocol/unlock/commit/4bce4c2baf3d7f05ae873115250133eb38d587c7#diff-a56a299a69adfb724d9c98bdffe39f8a43c136d2b514d1f442ad48cc662033f6)

julien51 (approved):
Very small change to use a helper than then LG!

locksmith/__tests__/controllers/v2/verifierController.test.ts
kalidiagne: oh thk! done!

Issue #8692, opened 2022-04-29
Balances does not account for cancelled and expired keys in the same way


clemsos:
**Describe the bug**
Currently, when a key is cancelled we internally call `cancelKey` which triggers `_deleteOwnershipRecord`, which will remove ownership data and decrease the user balance. 

As a result, after a key has been cancelled, calling `balanceOf` for the owner address will return a null value (even though a cancelled key exists). When a key expires, no specific event is triggered, therefore leaving an expired key in the user balance.

Therefore, the results of `balanceOf` can be considered inconsistent.

**To Reproduce**

1. Buy a key
2. Cancel it
3. call `balanceOf`

**Expected behavior**

`balanceOf` should return the total number of keys, expired or not.

julien51:
Let's make sure that explicit cancellation (by lock owner, key manager or key owner) behaves the same as "natural" expiration.

julien51:
The only difference is that we *must* reset the `_originalPrices` in case of explicit expiration/cancellation?

clemsos:
solved by https://github.com/unlock-protocol/unlock/issues/8693

PR #9003, opened 2022-06-10
fix(smart-contracts): reset original price when a key is cancelled

clemsos:
# Description

This PR resets the `_originalPrices` in case of explicit expiration/cancellation (on top of `_originalDurations`), to prevent future key renewals

# Issues

Fixes #
Refs #8692

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinRefunds.sol
+++ b/smart-contracts/contracts/mixins/MixinRefunds.sol
@@ -108,7 +108,7 @@ contract MixinRefunds is
 
     // make future reccuring transactions impossible
     _originalDurations[_tokenId] = 0;
-    _originalPrices [_tokenId] = 0;
+    _originalPrices[_tokenId] = 0;
     
     // inform the hook if there is one registered
     if(address(onKeyCancelHook) != address(0))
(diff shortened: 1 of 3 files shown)


smart-contracts/contracts/mixins/MixinRefunds.sol
```diff
@@ -108,6 +108,7 @@ contract MixinRefunds is
 
     // make future reccuring transactions impossible
     _originalDurations[_tokenId] = 0;
+    _originalPrices[_tokenId] = 0;
```
julien51, line 111: Oh, now that I think about it I wonder what happens on renewals for a lock that has free keys?

julien51 (approved):
Please, add a test :)

smart-contracts/contracts/mixins/MixinRefunds.sol
clemsos: yes why I using with duration together is safer

clemsos:
> Please, add a test :)

actually we are already testing that the tx reverts if there is changes in the lock. The revert code is `LOCK_HAS_CHANGED` and as we check that none of duration, price and token address have changed at once, it makes it impossible to dissociate the exact reason why it reverts in test.

what we could do will be add an error code like `LOCK_HAS_CHANGED(0)` for change in duration, `LOCK_HAS_CHANGED(1)` for a change in price, and `LOCK_HAS_CHANGED(2)` is the token address changed.

@julien51 wdyt?

https://github.com/unlock-protocol/unlock/blob/d62617027b94b25914a7bc1c3822c4318882ffc7/smart-contracts/test/Lock/renewMembershipFor.js#L277-L290

julien51:
I don't think that's necessary and would add weight... so I think we're good!

PR #9007, opened 2022-06-10
feature(locksmith): Grant keys api

julien51:
# Description

Adding an API to grant keys. This has been a feature request.
There are a few more things to do:
* [x] use an `app` token (rather than a user token)
* [X] support `keyGranter` on top of `lockManager`
* [x] improve the grantKeys calls so we can take into account the durations and managers
* [x] add a limit on gas to be spent.
* [x] ensure that the purchaser has enough funds.

I ended up moving the utils middleware into their own folder for clarity because we already have a bunch of `utils`

# Checklist:


locksmith/__tests__/controllers/v2/grantKeysController.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      walletAddress: address,
+      message,
+      signedMessage,
+    } = await getWalletInput()
+    const loginResponse = await request(app).post('/v2/auth/login').send({
+      signature: signedMessage,
+      message: message.prepareMessage(),
+    })
+    expect(loginResponse.status).toBe(200)
+
```
searchableguy, line 97: Wonder if we should abstract it into a wrapper because it is used quite often.

locksmith/src/controllers/v2/grantKeysController.ts
```diff
@@ -0,0 +1,62 @@
...
+   * @returns
+   */
+  async grantKeys(request: Request, response: Response) {
+    try {
+      const lockAddress = Normalizer.ethereumAddress(request.params.lockAddress)
+      const network = Number(request.params.network)
+      const keys = request.body.keys
+
+      /** Duration and managers are ignored for now, until a later PR since dispatcher does not support them yet */
+      const recipients = keys.map((k: any) => k.recipient)
```
searchableguy, line 33: Use zod and type the keys field for proper validation of input.

locksmith/src/controllers/v2/grantKeysController.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      dispatcher.grantKeys(
+        lockAddress,
+        recipients,
+        network,
+        async (error: any, hash: string) => {
+          if (error) {
+            response.status(500).send({
+              error,
+            })
```
searchableguy, line 44: I would also attach a message field with a useful but generalised error here. It can be hard to know what error spits out.

searchableguy (changes requested):
LGTM so far.

locksmith/__tests__/controllers/v2/grantKeysController.test.ts
julien51: Good point!

julien51: Added in #661c38e09

locksmith/src/controllers/v2/grantKeysController.ts
julien51: Done :)

julien51:
@searchableguy Any pointer for how to create an app and ensure that the token is only an app token?

searchableguy:
@julien51 look into tests files for the application controller. It creates and deleted an app inside tests.

You can write a `if(request?.user?.type === "app") {}` check inside your handler.

julien51:
Alright! Ready for a final review! I am happy to split things if there is too much here :)

locksmith/package.json
```diff
@@ -10,9 +10,9 @@
     "dev": "NODE_ENV=development yarn db:migrate && ts-node-dev -r ./src/utils/envLoader.js server.js",
     "test": "yarn test:db:setup && yarn test:run",
     "test:db:setup": "yarn run test:db:drop && yarn run test:db:create && yarn run test:db:migrate",
-    "test:db:create": "NODE_ENV=test yarn sequelize db:create",
+    "test:db:create": "",
     "test:db:migrate": "NODE_ENV=test yarn sequelize db:migrate",
-    "test:db:drop": "NODE_ENV=test yarn sequelize db:drop",
```
searchableguy, line 15: Any reason?

locksmith/src/controllers/v2/grantKeysController.ts
```diff
@@ -0,0 +1,92 @@
...
+          }
+          response.status(200).send({
+            hash,
+          })
+          return
+        }
+      )
+      return
+    } catch (error) {
+      console.error(error)
```
searchableguy, line 89: Also send a 500 response?

searchableguy (approved):
LGTM

locksmith/package.json
julien51: Nah. That's a mistake :) Fixed.

locksmith/src/controllers/v2/grantKeysController.ts
julien51: Good catch! fixed.

Issue #9005, opened 2022-06-10
Disable "native" transfers for keys with a manager


julien51:
One of the audits revealed a risk for keys to be "stolen" with the following scenario:
* a malicious user purchases a key on a lock
* the set an address of theirs as the key manager
* they put that key up for sale on a platform like OpenSea
* a victim buys the key from them
* the malicious user recaptures the key they sold.

We can't expect platforms like OpenSea to surface the fact that a key manager is set and that the seller has the ability to re-transfer a token their way. 

In order to mitigate this, we should do the following
* disable the "default" `transfer` and `transferFrom` functions if a key manager is set (and/or reset the keyManager when these are called)
* add a new `lend` function that lets someone change the owner, without changing the manager.

clemsos:
the `transfer` function is actually a call to `shareKey` that will just create a new key with the specified amount of time for the recipient, so this change should only affect `transferFrom`

PR #9011, opened 2022-06-14
feat(smart-contracts): prevent `transferFrom` from being used when a key manager is set

clemsos:
# Description

This PR fix security issue that allows a key to be sold while retaining key manager privileges on it by preventing `transferFrom` from being used when a key manager is set.

# Issues

Fixes #
Refs #9005

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinKeys.sol
+++ b/smart-contracts/contracts/mixins/MixinKeys.sol
@@ -65,7 +65,7 @@ contract MixinKeys is
   // the transfer of a key to another address where their key can be transferred
   // Note: the approver may actually NOT have a key... and there can only
   // be a single approved address
-  mapping (uint => address) private approved;
+  mapping (uint => address) internal approved;
 
   // Keeping track of approved operators for a given Key manager.
   // This approves a given operator for all keys managed by the calling "keyManager"
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -112,7 +112,14 @@ contract MixinTransfer is
     }
   }
 
-
+  /** 
+  * an ERC721-like function to transfer a token from one account to another
+  * @param _from the owner of token to transfer
+  * @param _recipient the address that will receive the token
+  * @param _tokenId the id of the token
+  * @notice To prevent the key manager to retain ownership rights on the token after transfer, the 
+  * operation will fail if a key manager if set. 
+  */
   function transferFrom(
     address _from,
     address _recipient,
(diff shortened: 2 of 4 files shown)


smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -65,7 +65,7 @@ contract MixinKeys is
   // the transfer of a key to another address where their key can be transferred
   // Note: the approver may actually NOT have a key... and there can only
   // be a single approved address
-  mapping (uint => address) private approved;
+  mapping (uint => address) internal approved;
```
julien51, line 68: Any reason for the change? (curious...)

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -121,8 +128,18 @@ contract MixinTransfer is
     public
   {
     _isValidKey(_tokenId);
-    _onlyKeyManagerOrApproved(_tokenId);
-    if(ownerOf(_tokenId) != _from) {
+    if(
```
julien51, line 131: That if statement is a bit confusing. Correct me if I am wrong but the goal here is to only allow this if:
* the msg.sender is both _from AND the ownerOf(_tokenId) ANd _tokenId does not have a manager 
OR
* the msg.sender is the the keyManager of tokenId
OR
* the msg.sender has been approved for the transfer

If so, would something like this be more readable:
```
if (!
    (msg.sender == _from && ownerOf(_tokenId) == _from && approved[_tokenId] == 0x) OR
    msg.sender == approved[_tokenId] OR
    approved[_tokenId] == msg.sender OR
    isApprovedForAll(_ownerOf[_tokenId], msg.sender)
)
```

smart-contracts/contracts/mixins/MixinKeys.sol
clemsos: this allows to access it from another mixin (private limits the scope inside the file itself)

smart-contracts/contracts/mixins/MixinTransfer.sol
clemsos: oh that want what I had in mind. I thought in any case if a key manager was set then the transfer will be made impossible, regardless if the transfer was initiated by the key owner, the key manager or an approved address.

julien51 (approved):
:ship it

smart-contracts/contracts/mixins/MixinTransfer.sol
julien51: ok in that scenario, can we make sure that a key manager can "renounce" that role?

PR #9013, opened 2022-06-14
feat(smart-contracts): add `lendKey` 

### LINKED ISSUE #9005: Disable "native" transfers for keys with a manager

clemsos:
# Description

This PR adds the `lendKey` function to locks, which allows to transfer a key to an address while retaining ownership as a key manager. 

NB: needs to rebase once #9011 is merged
 

# Issues

Fixes #9005
Refs #

# Checklist:


clemsos:
Just to clarify if we expect any difference in behaviour btw a key that has been lent and a "normal" key transfer: 

- to transfer a key will stop it from being renewable. Is is the same for a key that is lent? Or should it sill auto-renew?
- can lending keys been disabled ? (by setting a high value for  `transferFeeBasisPoints`)
- does the transfer fee applied to key lending (if `transferFeeBasisPoints` is set ) ?
-

julien51:
Hey! These are good questions!

> to transfer a key will disable renewals. Is is the same for a key that is lent? Or should it sill auto-renew?

That's a good question. I think it should stop auto-renewals to be safe.

> can lending keys been disabled ? (by setting a high value for transferFeeBasisPoints)

Yes. It should behave the same as transfers.

> does the transfer fee applied to key lending (if transferFeeBasisPoints is set ) ?

Yes.

clemsos:
> Hey! These are good questions!

ok thanks ! based on the answers, then there isnt any specific change to make

smart-contracts/contracts/interfaces/IPublicLock.sol
```diff
@@ -479,15 +479,32 @@ interface IPublicLock
...
+
+  /** 
+  * Lending a key allows you to transfer the token while retaining the 
+  * ownerships right by setting yourself as a key manager first
+  * @param from the owner of token to transfer
+  * @param to the address that will receive the token
+  * @param tokenId the id of the token
+  * @notice This requires the `msg.sender` to be set as a key manager 
+  */
+
```
julien51, line 503:
```suggestion
```

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -128,10 +131,7 @@ contract MixinTransfer is
     public
   {
     _isValidKey(_tokenId);
-    if(
-      // the specified address does not own the token
-      ownerOf(_tokenId) != _from
```
julien51, line 133: Expected?

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -142,6 +142,46 @@ contract MixinTransfer is
...
+
+  function lendKey(
+    address _from,
+    address _recipient,
+    uint _tokenId
+  )
+    public
+  {
+    _isValidKey(_tokenId);
+    _onlyKeyManagerOrApproved(_tokenId);
```
julien51, line 165: Let's not use `approvals` here for simplicity? Only the key manager should be able to lend a key,

julien51: Actually let's make things simple:
* LendKey can only be called by the KeyManager, OR
* if key manager is not set, by the key owner and, in that case, it should set the keyManager to be the current owner (not the one to which the key is transferred)

Let's also make sure we have a function to *unset* the key manager (or, alternatively, change the transfer to function to allow transfers even if there is a key manager, but in that case, reset the key manager).

wdyt?

smart-contracts/contracts/mixins/MixinTransfer.sol
clemsos: hmm no. I rewrote the conditions to make them more readable

smart-contracts/contracts/mixins/MixinTransfer.sol
clemsos: yep, better to remove approvals altogether and simplify the key manager logic

clemsos:
Thinking about it, there is a flaw in the design here. The owner of a key can change the key manager, so anyone how has been lended a key can easily take full ownership of it. Steps are as follow

- x lend a key to y
- y is not owner, x remains as key manager
- y uses `setKeyManagerOf` to set itself or the zero address as key manager
- y now fully owns the key

A simple way to prevent this is to introduce a `mapping (uint => bool) isCurrentlyLent` to track if a token is effectively owned or just lent. When a key is lent we set `isCurrentlyLent[tokenId] = true` then we can use it to prevent reseting the key manager

julien51:
Hum, I think there is a simpler way: if there is a `keyManager` (explicitly set) we only allow that address to do anything, except for transfers. So a "owner" cannot set the `keyManager` to be themselves, if they are not also the `keyManager.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -128,20 +131,67 @@ contract MixinTransfer is
...
+
+  function lendKey(
+    address _from,
+    address _recipient,
+    uint _tokenId
+  )
+    public
+  {
+    // make sure caller is either owner or key manager 
+    if(msg.sender != keyManagerOf[_tokenId] && msg.sender != ownerOf(_tokenId)) {
```
julien51, line 162: What happens if the key manager is not set? Does it mean the key owner has to explicitly set themselves as key manager first and then call lendKey?

julien51: Ideally we would allow the lendKey if the caller is the ownerOf(_tokenId) AND keyManagerOf[_tokenId] is not set, and set the keyManagerOf _inside_ this function.

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -128,20 +131,67 @@ contract MixinTransfer is
...
+    
+    // store previous owner
+    address previousOwner = ownerOf(_tokenId);
+
+    // transfer key ownership to lender
+    _transferFrom(_from, _recipient, _tokenId);
+
+    // set previous owner as key manager (if keys does not already have one KM)
+    if(keyManagerOf[_tokenId] == address(0)) {
+      keyManagerOf[_tokenId] = previousOwner;
```
julien51, line 174: That's good!

julien51 (commented):
Ok, I want to do one more pass but I am worried I don't see changes in the cancellations: they should only be by `_isKeyManager()`

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -128,20 +131,67 @@ contract MixinTransfer is
...
-        && approved[_tokenId] != msg.sender
-        && !isApprovedForAll(_ownerOf[_tokenId], msg.sender)
-      )
-      || // a key manager is set
-      keyManagerOf[_tokenId] != address(0)
-    ) {
+    _onlyKeyManagerOrApproved(_tokenId);
+    
+    // revert if key manager is set
+    if(keyManagerOf[_tokenId] != address(0)) {
```
julien51, line 137: I think we don't need to revert, but just reset the keyManager to be 0x0

smart-contracts/contracts/mixins/MixinTransfer.sol
julien51:
```suggestion
    if(msg.sender != _isKeyManager(_tokenId)) {
```

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -128,20 +131,67 @@ contract MixinTransfer is
...
     }
+    
+    // store previous owner
+    address previousOwner = ownerOf(_tokenId);
+
+    // transfer key ownership to lender
+    _transferFrom(_from, _recipient, _tokenId);
+
+    // set previous owner as key manager (if keys does not already have one KM)
+    if(keyManagerOf[_tokenId] == address(0)) {
```
julien51, line 173:
```suggestion
```

smart-contracts/contracts/mixins/MixinTransfer.sol
```diff
@@ -128,20 +131,67 @@ contract MixinTransfer is
...
+      keyManagerOf[_tokenId] = previousOwner;
+    }
```
julien51, line 175:
```suggestion
   keyManagerOf[_tokenId] = msg.sender;
```

julien51: Since the function can only be called by the key manager, or the owner if no key manager is set, I think we can set the keyManagerOf to be the sender safely

smart-contracts/contracts/mixins/MixinTransfer.sol
clemsos: ok. It means the keyManager needs to be set to address zero before using `lendKey` which is fine

smart-contracts/contracts/mixins/MixinTransfer.sol
clemsos: yep makes sense, will just reset the key manager on transfer then

clemsos:
Here are the cases covered by the tests

```
Contract: Lock / lendKey
    ✓ can lend a FREE key
    failures
      ✓ should abort when there is no key to lend
      ✓ should abort if the recipient is 0x
      ✓ should only allow key manager or owner
      ✓ should prevent lending an expired key
      ✓ should fail if the sender has been approved for that key
      ✓ should fail if the sender has been approved for all owner keys
    when the sender is the key owner
      no key manager is set
        ✓ should lend ownership correctly
        ✓ update balances properly
        ✓ update key validity properly
        ✓ should set previous owner as key manager
      a key manager is set
        ✓ should prevent from lending a key
    when the sender is a key manager
      ✓ should lend ownership correctly
      ✓ update balances properly
      ✓ update key validity properly
      ✓ retains the correct key manager
    when the lock is sold out
      ✓ should still allow the lend of keys
    approvals with lent key
      ✓ can not approve another account to lend the key

      1) can not used an "approved for all" account to transfer the key
    a lent key
      ✓ has correct ownership
      ✓ can not be lent by owner
      ✓ can not be transferred by owner
      ✓ can not be burn by owner
      ✓ can not be merged by owner
      ✓ can not be cancelled by owner
```

julien51:
I think that last failing tests is faulty? It should actually transfer since the `key manager` is set to be `accounts[6]` and `accounts[6]` is the one calling `transferFrom`.

clemsos:
> I think that last failing tests is faulty? It should actually transfer since the `key manager` is set to be `accounts[6]` and `accounts[6]` is the one calling `transferFrom`.

yes... not sure why we had a condition where the transfer hook should fail if initiated by a key manager?

julien51:
That's because we blocked transfers when there was a key manager?

julien51:
In any case, lg

julien51 (approved):

PR #9017, opened 2022-06-15
feat(smart-contracts): add support for custom referrer fee

### LINKED ISSUE #9004: referrerFee for platform monetization
In its current state, the protocol does not include an explicit mechanism for locks to share revenues with the platforms that hold them. Several people in the community suggested that it might be a good addition to the core protocol. Let's work on a first iteration which we can improve later as we make progress.

For this, we suggest the addition of the concept of `referrerFee` to the lock. This fee should be customized for any referrer address. A lock manager should be able to call `setReferrerFee` on a lock and pass both a `referrer` (address) and a `fee` (uint), expressed in basis points. The fees should be stored in the lock and be "public" (a platform should be able to check if it would receive a fee or not).

We should update all the function for which there is a referrer (purchase, renewal and extension) so that if the referrer matches one of the records it receives immediately a share (corresponding to the fee) of the price paid by the user.

clemsos:
# Description

This PR adds support for adding referrer fees per address or once for the whole contract to share a percentage of the revenue generated by the lock with the referrers.

# Issues

Fixes #9004
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -232,6 +232,23 @@ interface IPublicLock
     bytes calldata _data
   ) external payable;
 
+
+  /**
+  * Returns the percentage of the keyPrice to be sent to the referrer (in basic points)
+  * @param _referrer the address of the referrer
+  */
+  function referrerFees(address _referrer) external view;
+  
+  /**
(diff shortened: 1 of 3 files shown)


julien51 (commented):
Can we add tests to check if the referrer is paid in all contexts? (purchase, extend, renewals?)

julien51 (approved):

julien51:
Very clean!

PR #9025, opened 2022-06-15
feat(locksmith) - add mark as checkin for verifiers and lock manager endpoint

kalidiagne:
# Description
- add mark as checkin for verifiers and lock manager endpoint

# Issues

Fixes #
Refs #

# Checklist:


julien51 (changes requested):
Please add tests!

locksmith/src/controllers/v2/ticketsController.ts
```diff
@@ -24,4 +24,26 @@ export class TicketsController {
     )
     response.status(200).send({ payload, signature })
   }
+
+  async markTicketAsCheckIn(request: Request, response: Response) {
+    const lockAddress = Normalizer.ethereumAddress(request.params.lockAddress)
+    const network = Number(request.params.network)
+    const metadata = request.body.message.KeyMetaData
```
julien51, line 31: Let's not actually have `KeyMetaData` be submitted since we know the format. Let's just built it here. This way we avoid malicious behaviors.

locksmith/src/controllers/v2/ticketsController.ts
```diff
@@ -24,13 +25,20 @@ export class TicketsController {
...
+   * This will mark a ticket as check-in, this operation is only allowed for a lock verifier of a lock manager
+   * @param {Request} request
+   * @param {Response} response
+   * @return
+   */
   async markTicketAsCheckIn(request: Request, response: Response) {
     const lockAddress = Normalizer.ethereumAddress(request.params.lockAddress)
     const network = Number(request.params.network)
-    const metadata = request.body.message.KeyMetaData
+    const payload = request.body
```
julien51, line 37: No need for a payload anymore

unlock-app/src/services/storageService.ts
```diff
@@ -677,4 +677,26 @@ export class StorageService extends EventEmitter {
       return false
     }
   }
+
```
julien51, line 680: Let's remove this from this PR and focus _only_ on locksmith.

locksmith/src/utils/metadata.ts
```diff
@@ -0,0 +1,51 @@
+export interface KeyMetadataPayload {
+  types: {
```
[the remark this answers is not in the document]
julien51: This is only useful if the user is signing things, which they are not.

locksmith/src/routes/v2/ticket.ts
```diff
@@ -14,4 +15,12 @@ router.get(
   ticketsController.sign
 )
 
+router.put(
+  '/:network/lock/:lockAddress/key/:keyId/check',
+  isVerifierMiddleware,
```
julien51, line 20: Do we have this in another PR?

locksmith/src/controllers/v2/ticketsController.ts
```diff
@@ -24,4 +25,33 @@ export class TicketsController {
...
+   * @param {Request} request
+   * @param {Response} response
+   * @return
+   */
+  async markTicketAsCheckIn(request: Request, response: Response) {
+    const lockAddress = Normalizer.ethereumAddress(request.params.lockAddress)
+    const network = Number(request.params.network)
+    const id = request.params.keyId.toLowerCase()
+
+    const metadata = generateKeyMetadataPayload(request.body).message
```
julien51, line 39: You don't need this. You can just create the JSON object.

julien51:
```suggestion
    const metadata = {
        checkedInAt: new Date().getTime(),
      }
```

julien51 (changes requested):
Please add tests to the controller that do the following:
* test that if called without auth it fails
* test that if called from a non-verifier, it fails
* test that if called from a verifier it succeeds and updates the metadata correctly if there is existing metadata
* * test that if called from a verifier it succeeds and sets the metadata correctly if there is no metadata before

locksmith/src/routes/v2/ticket.ts
kalidiagne: @julien51 added middleware

locksmith/__tests__/controllers/v2/ticketsController.test.ts
```diff
@@ -65,4 +66,16 @@ describe('sign endpoint', () => {
     expect(typeof response.body.signature).toBe('string')
     expect(response.status).toBe(200)
   })
+
+  it('marks ticket as check-in', async () => {
```
julien51, line 70:
```suggestion
  it('marks ticket as checked-in', async () => {
```

julien51:
@kalidiagne Please add more tests per above. It is important to avoid regression.

locksmith/__tests__/controllers/v2/ticketsController.test.ts
```diff
@@ -67,6 +69,29 @@ describe('sign endpoint', () => {
     expect(response.status).toBe(200)
   })
 
+  it('marks ticket as checked-in without auth fails', async () => {
+    expect.assertions(1)
+    const response = await request(app).put(
+      `/v2/api/ticket/${network}/lock/${lockAddress}/key/${tokenId}/check`
+    )
+
+    expect(response.status).toBe(500)
```
julien51, line 78: It should not return 500 but 401

locksmith/__tests__/controllers/v2/ticketsController.test.ts
```diff
@@ -67,6 +69,29 @@ describe('sign endpoint', () => {
...
+    const { loginResponse } = await loginRandomUser(app)
+    expect(loginResponse.status).toBe(200)
+
+    const response = await request(app)
+      .put(
+        `/v2/api/ticket/${network}/lock/${wrongLockAddress}/key/${tokenId}/check`
+      )
+      .set('authorization', `Bearer ${loginResponse.body.accessToken}`)
+
+    expect(response.status).toBe(401)
```
julien51, line 92: It should be 403

locksmith/__tests__/controllers/v2/ticketsController.test.ts
```diff
@@ -78,4 +103,44 @@ describe('sign endpoint', () => {
...
+    expect(loginResponse.status).toBe(200)
+
+    const metadata = {}
+    const response = await request(app)
+      .put(`/v2/api/ticket/${network}/lock/${lockAddress}/key/${tokenId}/check`)
+      .set('authorization', `Bearer ${loginResponse.body.accessToken}`)
+      .send(metadata)
+    expect(response.status).toBe(202)
+
+    const updateStatus = await metadataOperations.updateKeyMetadata(metadata)
```
julien51, line 119: Let's read the data rather than try to update it and check that the value is correct

locksmith/__tests__/controllers/v2/ticketsController.test.ts
```diff
@@ -78,4 +103,44 @@ describe('sign endpoint', () => {
...
+    expect(updateStatus).toBe(false)
+  })
+
+  it('correctly marks ticket as checked-in and updates metadata', async () => {
```
julien51, line 123: This test should set metadata first, then check the ticket in and read the metadata to verify it is correct

locksmith/src/routes/v2/ticket.ts
julien51, line 20: Just add the middelare for authrequired

locksmith/__tests__/controllers/v2/ticketsController.test.ts
kalidiagne: is not `401` correct? 

user is signed-in but, is not authorized to perform this specific request because is not a verifer

locksmith/__tests__/controllers/v2/ticketsController.test.ts
julien51: ha, we probably need to change this in another PR then.
Don't worry about it here.

locksmith/__tests__/controllers/v2/ticketsController.test.ts
kalidiagne: nice! changed also the middleware status code to match `403`

locksmith/__tests__/controllers/v2/ticketsController.test.ts
```diff
@@ -1,12 +1,14 @@
...
+const metadataOperations = require('../../../src/operations/metadataOperations')
 
 const app = require('../../../src/app')
 
 jest.setTimeout(600000)
 const lockAddress = '0x3F09aD349a693bB62a162ff2ff3e097bD1cE9a8C'
+const wrongLockAddress = '0x00'
 const network = 4
-const tokenId = '123'
+const tokenId = '2244'
```
julien51, line 11: Why the change?

locksmith/__tests__/controllers/v2/ticketsController.test.ts
```diff
@@ -65,4 +68,97 @@ describe('sign endpoint', () => {
...
+      .put(`/v2/api/ticket/${network}/lock/${lockAddress}/key/${tokenId}/check`)
+      .set('authorization', `Bearer ${loginResponse.body.accessToken}`)
+    expect(response.status).toBe(202)
+
+    const keyData = await metadataOperations.getKeyCentricData(
+      lockAddress,
+      tokenId
+    )
+
+    expect(keyData.metadata.value).toBe('12')
```
julien51, line 162: Can you also check that `custom_field` is persisted please?

locksmith/__tests__/controllers/v2/ticketsController.test.ts
kalidiagne: now the endpoint currently avoids overrides, the old stored value for this tokenId have a not valid structure

julien51 (approved):

PR #9026, opened 2022-06-15
refactor(locksmith): adding verifier operation

julien51:
# Description

This is a refactor that isolates the logic better between controllers, middleware and services to access the data.
This would allow us for better re-usability (we should reuse the `VerifierOperations` in more places like middleware) as well as test things in isolation.

@kalidiagne @searchableguy Let's use that approach in the future please.

It will probably fail at first since some things are still missing from other PR but I'll rebase soon.

# Checklist:

### DIFF
--- a/locksmith/src/operations/verifierOperations.ts
+++ b/locksmith/src/operations/verifierOperations.ts
@@ -1,0 +1,88 @@
+import { Verifier } from '../models/verifier'
+
+namespace VerifierOperations {
+  /**
+   * Returns true if the verifier already exists on the lock and network
+   * @param lockAddress
+   * @param address
+   * @param network
+   * @returns boolean
+   */
+  export const isVerifierAlreadyExits = async (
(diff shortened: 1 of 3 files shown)


kalidiagne (approved):

locksmith/src/operations/verifierOperations.ts
```diff
@@ -0,0 +1,88 @@
...
+    address: string,
+    network: number
+  ): Promise<boolean> => {
+    return !!(await Verifier.findOne({
+      where: {
+        address,
+        lockAddress,
+        network,
+      },
+    }))
```
searchableguy, line 22: separate into a variable and then do `!!`.

I would also just use `Boolean` which is more clear.

julien51: rahaaa just saw this. Ok I will use `Boolean` in the future.

PR #9030, opened 2022-06-16
chore(smart-contracts): remove use of web3 from tests

### LINKED ISSUE #8856: Remove web3js dependencies
Our smart contract code is tested with 2 different framework : ethers (through HardHat) and web3js. Most of the `web3js` tests are for older versions of Unlock which are 1) not used anymore and 2) deprecated.
We should remove the `web3js` dependencies, starting by removing support for old versions of Unlock (only >= 4 are supported right now).
Then, we would need to rewrite remaining tests using hardhat/ethers.

PS: we could consider also deprecating v4 or v5 depending on how many active (has active memberships and sold/minted >1 in the last 6 months)

clemsos:
# Description

This removes all mentions of web3 from the contracts tests (expect for hardly-difficult package which will be addressed in another PR)

# Issues

Fixes #8856 
Refs #

# Checklist:


julien51 (approved):
There are a bunch of files in which you just added `const { ethers } = require('hardhat')` without actually using `ethers`... (or at least not adding new use of it).

Appart from this, it looks good.

clemsos:
> There are a bunch of files in which you just added `const { ethers } = require('hardhat')` without actually using `ethers`... (or at least not adding new use of it).

Oh... added back the lint `no-unused-var` rule in #9034 so we wont end up with this sort of issues

julien51:
ooof, conflicts :(

julien51 (approved):
These are hard to review... So I'll assume you have not removed tests and that or coverage is good...

PR #9050, opened 2022-06-17
feat(unlock-app): Add close button and emit transaction info in new checkout

searchableguy:
# Description

This adds onClose function which would be invoked on using the close button or clicking the return to site after finishing mint. 

This emits the transaction info on a successful mint and respect optimistic checkout setting in paywallConfig.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/alpha/Checkout/index.tsx
+++ b/unlock-app/src/components/interface/checkout/alpha/Checkout/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import type { PaywallConfig } from '~/unlockTypes'
 import { useCheckoutCommunication } from '~/hooks/useCheckoutCommunication'
 import { checkoutMachine, CheckoutPage } from './checkoutMachine'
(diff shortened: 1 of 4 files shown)


unlock-app/src/components/interface/checkout/alpha/Checkout/index.tsx
```diff
@@ -30,9 +36,42 @@ export function Checkout({ paywallConfig, injectedProvider }: Props) {
...
+  const { messageToSign, mint, lock } = state.context
 
-  const onClose = () => {
-    // TODO
+  useEffect(() => {
+    if (
+      mint &&
+      mint.transactionHash &&
+      mint.status === 'FINISHED' &&
+      communication
```
kalidiagne, line 46: can we change it with a more "speaking" var?

searchableguy: 🤔 I don't understand what that means. 

It is meant to be string enum of various states like `ERROR`, `PROCESSING`, `FINISHED`.

searchableguy: will do

kalidiagne (approved):

julien51 (approved):

PR #9076, opened 2022-06-22
feat(unlock-app) - ability to filter items on members page

kalidiagne:
# Description
- Ability to filter items on members page, this includes also metadata
  - user can select to filter by `keyId` or `owner adress`
  - search will done with a debounce of 500ms

# Issues

Fixes #
Refs #
### filter by keyId
Screenshot 2022-06-30 at 22 26 57

### filter by address
Screenshot 2022-06-30 at 22 26 05

### loading status on search 
Screenshot 2022-06-30 at 22 26 35

# Checklist:


unlock-app/src/utils/search.ts
```diff
@@ -0,0 +1,17 @@
...
+  const options = {
+    keys,
+    // because we deal with wallet address let's keep threshold to 0 to have perfect match on search
+    threshold: 0,
+  }
+
+  const fuse = new Fuse(items, options)
+
+  return [...fuse.search(searchTerm).map(({ item }) => item)]
+}
```
searchableguy, line 17: I would not abstract it. Use memo hook to load the fuse inside component with data once and then use it directly.

1. You are creating new index every time you use this method which will impact performance with larger set. 

2. You want to optimize the options for different type of search. So this won't be reusable elsewhere.

unlock-app/src/components/content/MembersContent.tsx
```diff
@@ -165,6 +166,24 @@ const MetadataTableWrapper = ({
...
+    const searchTermEmpty = search.length === 0
+
+    if (searchTermEmpty) {
+      setFiltredItems(list)
+    } else {
+      const filtredResults = searchFromList(list, search, columns)
+      setFiltredItems(filtredResults)
+    }
+  }
+
```
searchableguy, line 186: Instead of storing filtered items. 

I would only store the query and do search on each render pass using fuse.

```ts
const [query, setQuery] = useState('')
const fuseSearch = useMemo(() => { return new Fuse(items, {})}, [items])
const items = fuse.search(query) // any fallbacks can be configured in options or via ternary here
// render
return items.map(item => (<div> {item.value} </div> ))
```

unlock-app/src/utils/search.ts
kalidiagne: agree! but i think we need to define a pattern for search. let's talk about it!

julien51:
Ok, so unfortunately I am not sure this is very useful because if there are multiple pages it only filters on the current page... which is not very useful.
We need to filter at the graphQL query level to go across all of the records...

searchableguy:
That would require a separate search index if we want to cover metadata for filtering as well though.

If not, query to subgraph graphql API should suffice for the current purpose.

julien51:
> That would require a separate search index on the backend if we want to cover metadata for filtering as well though. I thought an easy solution would be to load all data at once on the client side and split into pages there instead of at API level.

Yeah, that could be possible even though hard to scale beyond a few hundred members...

> If not, query to subgraph graphql API should suffice for the current purpose.

I think that's the right approach for now. lookup by token id and/or owner address

kalidiagne:
Update GraphQL query to filter items with the `search` parameter, before merging this PR we need to have 

https://github.com/unlock-protocol/unlock/pull/9102 to avoid multiple requests for sign of message every time a new search query is passed

unlock-app/src/queries/keyholdersByLock.js
```diff
@@ -7,10 +7,11 @@ export default function keyholdersByLockQuery() {
       $expiresAfter: BigInt! = 0
       $first: Int! = 100
       $skip: Int! = 0
+      $owner: String = "0x4"
```
julien51, line 10: I am not sure I understand this?

kalidiagne: fixed!

julien51 (approved):

julien51:
Looking good!

julien51 (approved):

julien51:
@searchableguy Do you mind looking at it as well please?

searchableguy (approved):

PR #9079, opened 2022-06-23
feat(unlock-app) - verifiers should see metadata when scanning qr code

kalidiagne:
# Description
On the verification page, also users with verifiers status would be able to see metadata.
To solve this issue and also avoid the double sign-message issue, this PR includes: 

- [x] use new endpoint to get metadata and avoid double message sign-in
- [x] change endpoint to include metadata also for verifiers 
- [x]  remove old way to get user metadata

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/routes/metadata.ts
```diff
@@ -71,7 +64,6 @@ router.get(
 )
 
 router.get('/:address/keyHolderMetadata', MetadataController.keyHolderMetadata)
-router.get('/:address/:keyId', MetadataController.data)
```
searchableguy, line 74: ditto as above.

locksmith/src/controllers/v2/metadataController.ts
```diff
@@ -41,7 +42,7 @@ export class MetadataController {
     this.web3Service = web3Service
   }
 
-  async #isKeyOrLockOwner({
+  async #isKeyOwnerOrLockVerifier({
```
julien51, line 45: Let's move that to a metadataOperations file so we can re-use?

julien51 (commented):
One quick comment about making code more re-usable and keeping things outside of the controller :)

julien51 (approved):

locksmith/src/operations/metadataOperations.ts
```diff
@@ -138,3 +149,32 @@ const defaultMappings = (address: string, host: string, keyId: string) => {
   defaultResponse.description = `${defaultResponse.description} Unlock is a protocol for memberships. https://unlock-protocol.com/`
   return defaultResponse
 }
+
+export const isKeyOwnerOrLockVerifier = async ({
```
searchableguy, line 153: Should this not take into account lock manager as well like the original API?

Afaik, we don't automatically create lock manager as verifier on the backend.

searchableguy (changes requested):

locksmith/src/operations/metadataOperations.ts
searchableguy: I'm saying it should also allow loco manager to access the data.

kalidiagne: Oh, get it! 
added also check for lock manager, thk!

searchableguy:
Need to remove changes from another PR.

searchableguy (approved):

Issue #8853, opened 2022-05-23
Use the `fetch` package only


julien51:
In our codebase, we currently use multiple HTTP client packages: `fetch`, `node-fetch`, `cross-fetch` and `axios`...
This is adding unnecessary complexity (cognitive load as these API differ slightly), extra build time, heavier bundles... etc.

We should normalize to _only_ use `fetch` is nos supported by all major browsers and node.js natively (as of 18... which we don't support yet but would soon).

jordang7:
I will look into working on this! 

Curious is there a set amount for the bounty? I can only see the bounty tag on the issue but no other details

julien51:
@jordang7 Thanks for the follow up! Let's say 500 DAI!

julien51:
@jordang7 any progress?

jordang7:
Yes, sorry was a little more busy than expected and had some trouble getting setup. I've replaced all HTTP client packages with fetch currently. Having some trouble transitioning the tests to mock fetch instead of axios, specifically the wedlockService test. Having 4 tests fail there because of differing return structures. Haven't found a fix  for this yet

![wedlocks-error](https://user-images.githubusercontent.com/32372824/175119304-e825efa8-222b-431f-8cb5-f91af800c619.png)

PR #9080, opened 2022-06-23
refactor(all): standardize use of single fetch package

jordang7:
# Description

Replaced all instances of other HTTP client packages (axios,cross-fetch,node-fetch) with fetch. Which also included updating wedlock and useMetadata tests to mock fetch. 

@julien51 let me know if this is what you were expecting/wanted

# Issues
#8853 

Fixes #
Refs #

# Checklist:


searchableguy (changes requested):
This removes cross-fetch and other fetch imports but does not replace them with `node-fetch`. Please add the relevant import for fetch usage using `node-fetch`. 

Note: Native fetch support in only available in node v18 as an experimental feature. We are on node 16 so we cannot depend on that for now.

jordang7:
> This removes cross-fetch and other fetch imports but does not replace them with `node-fetch`. Please add the relevant import for fetch usage using `node-fetch`.
> 
> Note: Native fetch support in only available in node v18 as an experimental feature. We are on node 16 so we cannot depend on that for now.

Understood, just pushed a commit reflecting this.

packages/unlock-js/src/locksmithService.ts
```diff
@@ -1,4 +1,4 @@
-import fetch, { Headers } from 'cross-fetch'
+import fetch from 'node-fetch'
```
searchableguy, line 1: Keep `cross-fetch` import with Headers here since this is a library for both backend and frontend. We will remove this cross-fetch package when upgrading to node v18.

searchableguy:
LGTM so far. I will take another look after all tests are passed.

unlock-app/src/__tests__/hooks/useMetadata.test.js
```diff
@@ -1,17 +1,18 @@
...
 const metadata = {
   image: 'https://...',
 }
 
 describe('useMetadata', () => {
   beforeEach(() => {
-    jest.clearAllMocks()
-    axios.get = jest.fn(() => {
+    fetchMock.resetMocks()
+    global.fetch = jest.fn(() => {
```
julien51, line 15: I wonder if we should make sure we clean that up after each test?

jordang7:
> Very cool @jordang7 ! Let's see why the tests don't pass, shall we?
> 
> Thank you!

Looks like it is having trouble importing node-fetch. I will look into it

node-fetch-error

searchableguy:
Yeah, that error is due to ESM requirement of node-fetch v3. 

Shouldn't be a problem to downgrade as they will backport all fixes.

julien51:
Thanks a lot @jordang7 ! Any progress so we can push this over the finish line?

jordang7:
> Thanks a lot @jordang7 ! Any progress so we can push this over the finish line?

Sorry for the delay, went on vacation for a bit. 

I don't have much experience with jest, but I am looking into making the tests compatible with fetch now.

jordang7:
@julien51 @searchableguy Any idea why [smart-contract-extensions](https://github.com/unlock-protocol/unlock/runs/7219223234?check_suite_focus=true#logs) is failing now? Don't believe any of the changes in the last commit should be affecting smart-contract-extensions.

searchableguy:
@jordang7 You may want to rebase or merge master because something which broke the CI was merged.

jordang7:
> @jordang7 You may want to rebase or merge against master again because something which broke the CI was merged earlier.

Ah gotcha. Should be good to go now.

searchableguy:
thanks @jordang7 will merge soon :D

julien51:
Looking good @jordang7 !

searchableguy (approved):
Other than removing unused deps, looks good to me.

julien51:
@jordang7 Want to do a final rebase please?

unlock-app/src/services/storageService.ts
```diff
@@ -225,8 +235,13 @@ export class StorageService extends EventEmitter {
...
   async createUser(user: any) {
-    const opts = {}
-    return axios.post(`${this.host}/users/`, user, opts)
+    return fetch(`${this.host}/users/`, {
+      method: 'POST',
+      body: JSON.stringify(user),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    })
```
searchableguy, line 244: Can you check this doesn't actually change the assumption behind the function?

The fetch version will return response while the post version should return the json/text body, I believe.

searchableguy (changes requested):
One more thing can you re-check?

unlock-app/src/services/storageService.ts
jordang7: This function is used in 'createUserAccount` in useAccount.ts. I modified that function get the response data using  `response.json()` , so that should be good. I will check the rest to make sure. 
cua

searchableguy: Is that the only place where it's used?

searchableguy (approved):

unlock-app/src/services/storageService.ts
jordang7: Yes I believe that's the only place it's used!

kalidiagne (approved):

PR #9085, opened 2022-06-27
feat(unlock-app): Add progress indicator on the checkout head

searchableguy:
# Description

image

# Issues

Fixes #
Refs #

# Checklist:


packages/unlock-assets/dist/index.js
```diff
@@ -40,232 +40,349 @@ function createCommonjsModule(fn, module) {
 	return module = { exports: {} }, fn(module, module.exports), module.exports;
 }
 
-/** @license React v16.13.1
```
julien51, line 43: I wonder if we need these files in git?

searchableguy: We don't IMO. We should fix this along with adding typescript but that would require some changes in the scripts and CI.

PR #9087, opened 2022-06-27
feat: remove eth-sig-util and ethereumjs-util from unlock-app

### LINKED ISSUE #9083: Removing `eth-sig-util` and `ethereumjs-util` from unlock-app
We removed these dependencies in `locksmith` and we also need to remove them from `unlock-app` and replace with ethers.

Attaching a 200 DAI bounty to that one as this should be much easier than the locksmith changes.

m00n620:
This should close #9083

Removed eth-sig-util and ethereumjs-util from unlock-app


m00n620:
@julien51 , please take a look

julien51 (commented):
ok so now I am worried that we need to test thoroughly that this still works with the backend because the messages signed are different...

julien51 (approved):

julien51:
Thank you! I will do dinner thorough testing after this had been deployed!

PR #9171, opened 2022-07-04
feat(unlock-js): add support for PublicLock v11

clemsos:
# Description

This adds support for PublicLock V11  in unlock-js

Needs to be rebase once #9229 is merged

# Issues

Fixes #
Refs #

# Checklist:


packages/unlock-js/src/PublicLock/v11/index.js
```diff
@@ -0,0 +1,8 @@
+import v10 from '../v10'
+import abis from '../../abis'
+
+export default {
+  version: 'v11',
+  PublicLock: abis.PublicLock.v11,
```
julien51, line 6: Wow, all functions identical? That's convenient :D

julien51 (commented):
Do you think it would make sense to pass an expiration to the grantKeyExtension method to match grant keys?

clemsos:
> Do you think it would make sense to pass an expiration to the grantKeyExtension method to match grant keys?

I quite like the simplicity of the function that way but that's just a code thing :) Not sure how passing a custom duration will impact things like renewal by cc.

clemsos:
I thought the goal of `grantKeyExtension` was to mimic `extend` - without requiring a direct payment. The `extend` itself will extend by the amount of time specified in the lock. Internally they both use `_extendKey`, which just requires a tokenId. If we really need to pass the duration as a param to `grantKeyExtension` then it will require substantial changes to the contract under the hood. 

Also seems a good practice to have correspondance as  `grantKeyExtension` =  `extend` for key granters.

Then maybe there is a direct use case I dont know about that will require to extend a key by a given amount that differs from the lock duration.

julien51:
Well, technically `grantKeyExtension` should be to `extend` what `grantKeys` is to `purchase`, so it should mimick _both_ `extend` and `grantKeys`.

I understand about the larger changes, so let's maybe hold on that and wait a bit.

julien51 (commented):
Ca m'a l'air bien mais il y a un truc qui plante :/

clemsos:
> Ca m'a l'air bien mais il y a un truc qui plante :/

Yes, keep hitting some sort of race conditions in tests. I fixed several but still happening for some reason

clemsos:
> Any idea why this race condition only happens with v11?

We were reusing an existing account that already has a key (`accounts[11]` ) so we were reaching the max keys per address (added a condition to set more when v11). I tried similar tests directly in smart-contracts and it all works smoothly.

What is weird is that we weren't catching the custom error (we moved from `require` statements to custom errors in ) 

I bumped hardhat version, but it seems the errors arent exported in the abi so we have a 'VM Exception while processing transaction: reverted with an unrecognized custom error' instead of showing the actual error (should be `MAX_KEYS_REACHED`).

I added the custom errors to the lock ABI (following [ethers guidelines ](https://docs.ethers.io/v5/api/utils/abi/interface/#Interface--creating) but still not showing up :/

packages/unlock-js/src/abis.ts
```diff
@@ -600,6 +600,39 @@ const abis = {
     "v11": {
       "contractName": "PublicLock",
       "abi": [
+        "error CANNOT_APPROVE_SELF()",
```
julien51, line 603: That's cool!

julien51 (approved):

PR #9174, opened 2022-07-04
feat(unlock-app) - Optimize members page checkin

kalidiagne:
# Description
This PR optimizes the member's page for the eventual undefined values on members' object.
From the members page, a ticket can be checked-in without extra metadata, only checkInTime will be returned 

# Issues

Fixes #
Refs #

https://user-images.githubusercontent.com/20865711/177193868-def2fcd6-8855-499f-8495-168ec92716e1.mov

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [x] I have performed a self-review of my own code
- [x] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #9175, opened 2022-07-04
feat(locksmith) - Bulk metadata optimize

kalidiagne:
# Description
This PR optimize `getBulkKeysMetadata` controller, related to https://github.com/unlock-protocol/unlock/pull/9174 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/metadataController.ts
+++ b/locksmith/src/controllers/v2/metadataController.ts
@@ -416,30 +416,39 @@ export class MetadataController {
         return await metadataOperations.getKeyCentricData(lockAddress, keyId)
       })
 
+      // get owners address from `ownerOf`, because not every item can include the user address, and void undefined values
+      const ownersAddressPromise = keyIds.map(async (keyId) => {
+        return await this.web3Service.ownerOf(lockAddress, keyId, network)
+      })
+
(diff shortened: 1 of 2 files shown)


locksmith/src/controllers/v2/metadataController.ts
```diff
@@ -416,30 +416,39 @@ export class MetadataController {
         return await metadataOperations.getKeyCentricData(lockAddress, keyId)
       })
 
+      // get owners address from `ownerOf`, because not every item can include the user address, and void undefined values
+      const ownersAddressPromise = keyIds.map(async (keyId) => {
```
julien51, line 420: I think this is going to make the request timeout because looking up the owner for each of them will be a problem.

julien51 (changes requested):
We need to make sure we have the owner in the request coming from the front-end or this will timeout as looking up each on chain will be take too much time

locksmith/src/controllers/v2/metadataController.ts
kalidiagne: refactored to avoid this issue

kalidiagne:
Removed use of `ownerOf` to avoid request timeout and better performance

julien51 (approved):

PR #9192, opened 2022-07-06
feat(unlock-app): Stop asking user to sign in repeatedly

searchableguy:
# Description

This fixes an issue where we repeatedly ask user to sign in. It's a temporary fix which I will remove in rewrite. 

Please look carefully if this breaks any of your code (I've minimized the change so it shouldn't) @kalidiagne 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useProvider.ts
+++ b/unlock-app/src/hooks/useProvider.ts
@@ -119,6 +119,7 @@ export const useProvider = (config: any) => {
 
       // remove access token when account wallet changes
       const removeAccessToken = () => {
+        localStorage.removeItem(`locksmith-access-token`)
         localStorage.removeItem(`locksmith-refresh-token`)
       }
 
--- a/unlock-app/src/hooks/useProvider.ts
+++ b/unlock-app/src/hooks/useProvider.ts
@@ -121,12 +121,12 @@ export const useProvider = (config: any) => {
 
       if (provider.on) {
         provider.on('accountsChanged', async () => {
-          await storageService.signout()
+          await storageService.signOut()
           resetProvider(new ethers.providers.Web3Provider(provider))
         })
 
         provider.on('chainChanged', async () => {
-          await storageService.signout()
+          await storageService.signOut()
           resetProvider(new ethers.providers.Web3Provider(provider))
         })
       }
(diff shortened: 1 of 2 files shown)


julien51 (approved):
Works for me, but please @kalidiagne let us know if it works for you too!

kalidiagne:
@searchableguy were to do you see repeated sign-in popup? in the current version sign-in will be shown only when wallet account is changed or the token expired, is it not? 

Testing with this branch, i see an issue for example in `verifiers page`, when i look at the verifiers page and change account from Metamask I'm able to see the list even if I'm not the lock manager because the token is still the same and a new pop-up request to sign-in does not show when I reload the page.

kalidiagne:
@searchableguy added a commit that solve the issue described before, we need to remove access and refresh token when  wallet account changes

unlock-app/src/hooks/useProvider.ts
```diff
@@ -117,8 +117,10 @@ export const useProvider = (config: any) => {
       }
       const ethersProvider = new ethers.providers.Web3Provider(provider)
 
+      // remove access token when account wallet changes
       const removeAccessToken = () => {
-        localStorage.removeItem(`${APP_NAME}.token`)
+        localStorage.removeItem(`locksmith-access-token`)
```
kalidiagne, line 122: @searchableguy maybe store this name as done before?

searchableguy:
That case should be handled by logging out the user when they disconnect an account. You would need to call the logout endpoint and remove the tokens. I've not seen anything implemented to that extent already.

Want me to do it? @kalidiagne

searchableguy:
Logging out wipes out the token from backend so it's an extra precaution in case of any possible leaks.

searchableguy:
@kalidiagne can you re-test and see if all issues are resolved?

I will add the logout call later when I refactor it. For now, should be okay.

searchableguy:
@kalidiagne re-test and see if there are any problems?

unlock-app/src/services/storageService.ts
```diff
@@ -71,13 +70,27 @@ export class StorageService extends EventEmitter {
     return this.locksmith.login(message, signature)
   }
 
+  async signOut() {
+    const endpoint = `${this.host}/v2/auth/revoke`
+    const response = await fetch(endpoint, {
```
julien51, line 75: Let's make this fail silently (try/catch) since it should not affect the rest of the app if it does.

kalidiagne:
@searchableguy tested and all fine!

kalidiagne (approved):

PR #9205, opened 2022-07-07
feat(unlock-app) - Send QR-code to email metadata

kalidiagne:
# Description
- [x] Show `send QR code by email` button in members page, that's shows only when email metadata is present
- [x] Genate and and QR-code as attachments to sendEmail function 
- [x] add keyGranter on check for signature in verification page

# Issues

Fixes #
Refs #
Screenshot 2022-07-07 at 11 11 22

# Checklist:


locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -78,7 +89,7 @@ export const notifyNewKeysToWedlocks = async (
  * and email based on the lock's template if applicable
  * @param key
  */
-export const notifyNewKeyToWedlocks = async (key: any, network?: number) => {
+export const notifyNewKeyToWedlocks = async (key: Key, network?: number) => {
```
julien51, line 92: Let's make sending the QR code optional here so that the background job _does not_ send the QR code, but the explicit API call does. Does it make sense?

locksmith/src/controllers/v2/ticketsController.ts
```diff
@@ -98,7 +98,8 @@ export class TicketsController {
             address: keyOwner,
           },
         },
-        network
+        network,
+        true
```
julien51, line 102:
```suggestion
        true /* includeQrCode */
```

julien51 (approved):

PR #9209, opened 2022-07-07
feat(wedlocks) - nft email confirmation template

kalidiagne:
# Description
Change email confirmation template with more details, tx URL and open sea URL are optional, so in that case the value will be print only if passed as parameter

- Add `opensea.tokenUrl` and `isTestNetwork` parameter is networks

# Issues

Fixes #
Refs #

# Checklist:


wedlocks/src/templates/keyMined.js
```diff
@@ -1,12 +1,28 @@
+const links = (params) => {
+  const hasTxUrl = params?.txUrl?.length > 0
```
julien51, line 2: I don't think txUrl has been defined

kalidiagne: added to parameters as optional

locksmith/src/utils/openSea.ts
```diff
@@ -0,0 +1,21 @@
...
+  tokenId,
+  network,
+}: {
+  lockAddress: string
+  tokenId: string
+  network: any
+}): string | undefined => {
+  const isTestNetwork = [4, 31337, 80001].includes(network)
+
+  const baseUrl = isTestNetwork
```
julien51, line 14: Can you check the network package and put that logic there please do it's reusable?

Thank you!

julien51 (changes requested):

julien51:
Also, now that I think about it the OpenSea URL should also be in the network packages (like we do for the block explorer!)

julien51 (approved):

locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -87,10 +86,14 @@ export const notifyNewKeysToWedlocks = async (
  * @param key
  */
 export const notifyNewKeyToWedlocks = async (key: any, network?: number) => {
+  const lockAddress = key.lock.address
+  const ownerAddress = key.owner.address
+  const tokenId = key?.keyId
```
julien51, line 91: Not sure we need `?` here, or we need it everywhere above as well

kalidiagne: `keyId` is not always passed as parameter, i update the `key: any` type definition to: 

```
 key: {
    lock: {
      address: string
      name?: string
    }
    owner: {
      address: string
    }
    keyId?: string
  },
  ```

locksmith/src/utils/openSea.ts
kalidiagne: added!

PR #9218, opened 2022-07-09
feat(unlock-app): Redesign ticket membership card for verification page

searchableguy:
# Description

This redesigns the ticket membership card for use in scanner. I've split it from the original PR. 

image
image

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Ok this is looks good, but I am really worried about making sure we do not break the (fragile) flow that we have right now.
We need to make sure the following flows work

With a QR code, either user-signed (thru keychain) or locksmith (thru email...), it can be scanned by a regular camera and prompts the user to open a URL.
This URL _can_ be loaded in any web browser, even if the user does not have a proper crypto wallet.
When loaded, it shows if the ticket is valid. 
The ticket is valid only in the following:
- signature matches the owner of the token/key OR is the locksmith purchaser's address
- key is not expired
- the current owner of the token is the same as in the signed blob

If the ticket was previously checked-in, the ticket is valid but a warning is shown indicating that it has already been checked in.

Additionally, there is a button to log in/connect which should support wallet and Unlock accounts.
When the user is logged in, if they are a verifier, the UI shows the metadata associated with the key/ticket and a button to "mark as checked-in" is visible. If the user clicks on the button, the ticket is immediately marked as checked-in and when scanned again, it cannot be marked as checked-in again.

Can you please confirm that your changes do not break any of these?

unlock-app/src/components/interface/Account.tsx
```diff
@@ -62,7 +64,10 @@ export function Account() {
                 <button
                   className="px-2 py-1 text-gray-900 bg-gray-200 rounded"
                   type="button"
-                  onClick={deAuthenticate}
+                  onClick={() => {
+                    deAuthenticate()
+                    storageService.signOut()
```
kalidiagne, line 69: can we move`signOut` inside `deAuthneticate`?

unlock-app/src/components/interface/verification/MembershipCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+  RiCheckboxCircleFill as ValidIcon,
+} from 'react-icons/ri'
+import { ReactNode } from 'react'
+
+dayjs.extend(relativeTimePlugin)
+
+interface Props {
+  timestamp: number
+  lock: Lock
+  membershipData: any
```
kalidiagne, line 22:
```suggestion
  membershipData: {
  expiration: number
    image?: string
    keyId?: string | number
    owner?: string
    userMetadata?: {
      public?: any
      protected?: any
    }
  }
```

unlock-app/src/components/interface/Account.tsx
searchableguy: We can but I didn't want to do it in this PR be because might change the behavior in multiple places instead being isolated.

unlock-app/src/components/interface/verification/MembershipCard.tsx
searchableguy: Will add this as a separate type so it's reusable. Thanks.

unlock-app/src/components/interface/VerificationStatus.tsx
```diff
@@ -1,118 +1,179 @@
...
+
+  const checkedInAt = membershipData?.metadata?.checkedInAt
+
+  const CardActions = () => (
+    <div className="grid w-full">
+      {viewer ? (
+        <Button
+          loading={isCheckingIn}
+          disabled={!isVerifier || isCheckingIn || !!invalid || checkedInAt}
+          onClick={async (event) => {
```
julien51, line 136: I think it's better to move to its own function for clarity?

julien51 (approved):
Some minor changes and then LG. Please do sort the small issues and then merge!

searchableguy:
They do not break any of it. You can see how it would look for checked in users already above. It will disable the button with checked in message and show checked in at date in the fields.

kalidiagne (approved):

julien51 (approved):

PR #9225, opened 2022-07-11
fix(locksmith): Return correct key expiration date from KeyData

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/utils/keyData.ts
+++ b/locksmith/src/utils/keyData.ts
@@ -17,13 +17,21 @@ export default class KeyData {
       if (!owner) {
         return {}
       }
+      const lock = await web3Service.getLock(lockAddress, network)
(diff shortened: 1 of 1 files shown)


locksmith/src/utils/keyData.ts
```diff
@@ -17,13 +17,21 @@ export default class KeyData {
...
+          network
+        )
+      } else {
+        key.expiration = await web3Service.getKeyExpirationByLockForOwner(
+          lockAddress,
+          owner,
+          network
+        )
+      }
       key.owner = owner
```
julien51, line 34: Let's move this _above_ the expiration so that it's set even when the retrieval of expiration fails please.

julien51 (approved):

PR #9183, opened 2022-07-05
feat(unlock): Add network description to network package and use those in checkout

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:

- [ ] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #9231, opened 2022-07-12
doc(networks): scripts to generate the docs for the networks package

### LINKED ISSUE #9081: Generate contract address docs automatically from network package
Our docs include a list of all addresses for all networks. We currently maintain these list manually in the docs. Instead we should create a script that retrieves these automatically.

clemsos:
# Description

This PR adds a script that will generate the docs about the existing networks, and opens a PR (or commits to an existing one) in the docs repo.

Usage:

```
sh scripts/docs-networks-pr.sh
```

# Issues

Fixes #9081
Refs #

# Checklist:

### DIFF
--- a/packages/networks/.eslintrc.js
+++ b/packages/networks/.eslintrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+  extends: ['@unlock-protocol/eslint-config'],
+}
--- a/packages/networks/bin/doc.js
+++ b/packages/networks/bin/doc.js
@@ -1,8 +1,17 @@
 const { networks } = require('../dist')
 
-const parseNetwork = ({ name, id, unlockAddress, explorer, nativeCurrency }) =>
+const parseNetwork = ({
+  name,
+  id,
+  unlockAddress,
+  explorer,
+  nativeCurrency,
+  description,
+}) =>
   `
 ## ${name}
+
+${description || ''}
 
 - chainId: ${id}
 - native currency: ${nativeCurrency.name} (${nativeCurrency.symbol})
--- a/packages/networks/bin/doc.js
+++ b/packages/networks/bin/doc.js
@@ -1,0 +1,24 @@
+const { networks } = require('../dist')
+
+const parseNetwork = ({ name, id, unlockAddress, explorer, nativeCurrency }) =>
+  `
+## ${name}
+
+- chainId: ${id}
+- native currency: ${nativeCurrency.name} (${nativeCurrency.symbol})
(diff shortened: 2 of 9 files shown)


packages/networks/.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  extends: ['@unlock-protocol/eslint-config'],
+}
```
clemsos, line 3: had to add this to make our pre-commit checks pass

packages/networks/bin/doc.js
```diff
@@ -0,0 +1,24 @@
+const { networks } = require('../dist')
+
+const parseNetwork = ({ name, id, unlockAddress, explorer, nativeCurrency }) =>
+  `
+## ${name}
+
+- chainId: ${id}
+- native currency: ${nativeCurrency.name} (${nativeCurrency.symbol})
+- unlockAddress: \`[${unlockAddress}](${explorer.urls.address(unlockAddress)})\`
```
searchableguy, line 9: There is also description added in https://github.com/unlock-protocol/unlock/pull/9183 

Let me merge it and then you can rebase?

searchableguy: You can rebase now :D

clemsos:
Currently there is a typo in the way URL are parsed in https://docs.unlock-protocol.com/core-protocol/unlock/networks

clemsos:
Looks like rebase works. I didn’t push a fix for the urls so far.

about public lock addresses, it’s possible but it will require to change the place in docs as the networks file currently lives in the Unlock folder. Also maybe PublicLock addresses are more internal and don’t really needs to be exposed

clemsos:
updated PR https://github.com/unlock-protocol/docs/pull/68

julien51:
> about public lock addresses, it’s possible but it will require to change the place in docs as the networks file currently lives in the Unlock folder. Also maybe PublicLock addresses are more internal and don’t really needs to be exposed

Ok, let's do that in a later PR then and indeed, we might move the addresses to be in the same place.
It's still important to show them so people can "inspect" them?

searchableguy (approved):

julien51:
One final thought: do you think we could run that script automatically (same as the smart contract doc generation one) when master gets a new commit?

clemsos:
> One final thought: do you think we could run that script automatically (same as the smart contract doc generation one) when master gets a new commit?

yes, we can have them run on CI. I filled an issue there to track that #9270

PR #9238, opened 2022-07-12
feat(unlock-app): Use different qr scanner library for verification page

searchableguy:
# Description

The underlying library behind react-qr-reader (zxing) doesn't work properly with some of our generated QR code. I've replaced it with qr-scanner package which uses jsqr library for qrcode scanning which seems to have better success rate in my testing.

# Issues

Fixes #
Refs #

# Checklist:


unlock-app/src/utils/verification.ts
```diff
@@ -3,7 +3,9 @@ import * as z from 'zod'
 export const MembershipVerificationData = z.object({
   account: z.string(),
   timestamp: z.number(),
-  tokenId: z.string(),
+  tokenId: z
+    .union([z.string(), z.number()])
+    .transform((value) => value.toString()),
```
searchableguy, line 8: Different data type in QR codes.

unlock-app/src/components/interface/VerificationStatus.tsx
```diff
@@ -127,7 +143,9 @@ export const VerificationStatus = ({ config }: Props) => {
   )
 
   const invalid = invalidMembership({
-    membershipData: membershipData!,
+    keyId: key!.tokenId.toString(),
+    owner: key!.owner,
+    expiration: key!.expiration,
```
searchableguy, line 148: Same here. We don't consistent data so I'm directly passing the props after converting.

julien51:
This needs very thorough testing. Have you been able to do some of this?

searchableguy:
Tests done

[ ] Disconnect and scan - make sure it works without logging in
[ ] Able to check in ticket as a verifier
[ ] Able to check in ticket as a lock manager
[ ] Checked in ticket shows as yellow with correct time
[ ] Test with v10 and v9 locks respectively.
[ ] Use the connect button to login

kalidiagne (approved):

julien51 (approved):

PR #9259, opened 2022-07-14
feat(unlock-app) - fix invalid keyId value in query

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/services/graphService.ts
+++ b/unlock-app/src/services/graphService.ts
@@ -4,7 +4,7 @@ import locksByManager from '../queries/locksByManager'
 import keyHoldersByLocks from '../queries/keyholdersByLock'
 import { ToastHelper } from '../components/helpers/toast.helper'
 import keyholdersByKeyIdQuery from '../queries/keyholdersByKeyId'
-
+import { getValidNumber } from '~/utils/strings'
 export class GraphService {
   public client: any
 
@@ -55,9 +55,7 @@ export class GraphService {
     const query =
       filterKey === 'owner' ? keyHoldersByLocks() : keyholdersByKeyIdQuery()
 
-    const keyId = !isNaN(parseInt(`${search}`))
-      ? parseInt(`${search}`)
-      : undefined
+    const keyId = getValidNumber(search)
 
     const result = await this.client.query({
       query,
--- a/unlock-app/src/__tests__/utils/strings.test.js
+++ b/unlock-app/src/__tests__/utils/strings.test.js
@@ -1,8 +1,26 @@
-import { camelCaseToTitle } from '../../utils/strings'
+import { camelCaseToTitle, getValidNumber } from '../../utils/strings'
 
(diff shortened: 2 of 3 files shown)


unlock-app/src/services/graphService.ts
```diff
@@ -55,9 +55,7 @@ export class GraphService {
     const query =
       filterKey === 'owner' ? keyHoldersByLocks() : keyholdersByKeyIdQuery()
 
-    const keyId = !isNaN(parseInt(`${search}`))
```
julien51, line 58: We need to show an indication to the user when the number is not valid, otherwise the user does not know what is going on.
Maybe add validation to the form field rather than here?

kalidiagne: Need it also here, but fine to add also a visible error.
Otherwise we have to add a `Confirm/Search` button that's run the query when pressed only when validation passes

unlock-app/src/components/content/MembersContent.tsx
```diff
@@ -219,7 +230,7 @@ const MetadataTableWrapper = ({
           </select>
         </span>
         <Input
-          type="text"
+          type={currentFilter?.type ?? 'text'}
```
kalidiagne, line 233: In case of `keyId` type is `number`, so this prevents invalid entries

unlock-app/src/components/content/MembersContent.tsx
```diff
@@ -166,12 +167,15 @@ const MetadataTableWrapper = ({
...
   const [query, setQuery] = useState<string>('')
   const [filterKey, setFilteKey] = useState<string>('owner')
+  const [currentFilter, setCurrentFilter] = useState<Filter>()
   const queryValue = useDebounce<string>(query)
 
-  const filters: Filter[] = [
-    { key: 'owner', label: 'Owner' },
-    { key: 'keyId', label: 'Token id' },
-  ]
+  const filters: Filter[] = useMemo(() => {
```
julien51, line 173: Why `useMemo` ?

unlock-app/src/components/content/MembersContent.tsx
julien51: Ok good!

julien51 (approved):

unlock-app/src/components/content/MembersContent.tsx
searchableguy: same ^ 

it's unnecessary given the negligible cost of creating an object in js. In fact, it might be slower due to the fact it does a lot in the background. Just put this static variable outside the component if you don't want to recreate it on component mount/unmount.

unlock-app/src/components/content/MembersContent.tsx
searchableguy: I would actually keep it as text and parse it to number. Due to how react-controlled input works, it's better that way to avoid a situation where you end up with 0 in value.

PR #9275, opened 2022-07-18
feat(unlock-app)- query for all keys (expired + active)

kalidiagne:
# Description
Add query foll all keys (expired + active)  

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/content/MembersContent.tsx
+++ b/unlock-app/src/components/content/MembersContent.tsx
@@ -7,7 +7,7 @@ import BrowserOnly from '../helpers/BrowserOnly'
 import Layout from '../interface/Layout'
 import Account from '../interface/Account'
 import { pageTitle } from '../../constants'
-import { MemberFilters } from '../../unlockTypes'
+import { MemberFilter } from '../../unlockTypes'
 import { MetadataTable } from '../interface/MetadataTable'
 import useMembers from '../../hooks/useMembers'
 import LoginPrompt from '../interface/LoginPrompt'
@@ -140,7 +140,7 @@ const filters: Filter[] = [
   {
     key: 'expiration',
     label: 'Expiration',
-    options: [MemberFilters.ACTIVE, MemberFilters.EXPIRED, MemberFilters.ALL],
+    options: ['active', 'expired', 'all'],
   },
 ]
 /**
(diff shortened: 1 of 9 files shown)


unlock-app/src/queries/keyholdersByLock.ts
```diff
@@ -58,6 +87,20 @@ const ExpiredKeys = gql`
   }
 `
 
-export default function keyholdersByLockQuery(showActive = true) {
-  return showActive ? ActiveKeys : ExpiredKeys
+const QUERY_BY_TYPE = {
+  [MemberFilters.ACTIVE]: ActiveKeys,
+  [MemberFilters.EXPIRED]: ExpiredKeys,
+  [MemberFilters.ALL]: AllKeys,
+}
```
searchableguy, line 94: Not a big deal for this but we should deprecate use of enum (they are not recommended to be used by the typescript team)

unlock-app/src/services/graphService.ts
```diff
@@ -62,11 +62,10 @@ export class GraphService {
     search: string | number
     filterKey: string
   }) => {
-    const showActive = expiration === MemberFilters.ACTIVE
     const query =
       filterKey === 'keyId'
         ? keyholdersByKeyIdQuery()
-        : keyHoldersByLocks(showActive)
+        : keyHoldersByLocks(expiration)
```
searchableguy, line 68: Some explanation would be helpful.

unlock-app/src/queries/keyholdersByLock.ts
kalidiagne: remove the enum with `type` definition

unlock-app/src/queries/keyholdersByLock.ts
```diff
@@ -1,4 +1,33 @@
-import { gql } from 'apollo-boost'
+import { DocumentNode, gql } from 'apollo-boost'
+import { MemberFilter } from '~/unlockTypes'
+
+const AllKeys = gql`
```
julien51, line 4: Let's move to another file to keep the name consistent

julien51 (commented):
Please check tests

julien51 (approved):

unlock-app/src/components/content/MembersContent.tsx
```diff
@@ -140,7 +140,7 @@ const filters: Filter[] = [
   {
     key: 'expiration',
     label: 'Expiration',
-    options: [MemberFilters.ACTIVE, MemberFilters.EXPIRED, MemberFilters.ALL],
+    options: ['active', 'expired', 'all'],
```
searchableguy, line 143: you can actually type this btw. 

```ts
type Filters = "Active" | "Expired" | "All"

// use it elsewhere

options: Filters[]
option: Filter

```

kalidiagne: oh, i do it by purpose, but eventually can be changed when we need different filters!

PR #9276, opened 2022-07-19
fix(unlock-app): Handle an edge case where QR code urls have double encoded the data and missing tokenId

searchableguy:
This is a fix for QR codes generated before june.

1. There was a bug in the past which caused data to be double encoded in the URI. We fixed this sometime later but since some QR codes were generated through that bug, this fixes it by double decoding which shouldn't cause problem even if URI was only once encoded.
2. Similar to this, our old QR codes didn't include tokenId. I've made it optional.

# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/VerificationStatus.tsx
+++ b/unlock-app/src/components/interface/VerificationStatus.tsx
@@ -61,6 +61,7 @@ export const VerificationStatus = ({ config, onVerified, onClose }: Props) => {
   const { isLoading: isKeyLoading, data: key } = useQuery(
     [network, tokenId, lockAddress],
     async () => {
+      // Some older QR codes might have been generated without a tokenId in the payload. Clean up after January 2023
       if (lockVersion && lockVersion >= 10 && tokenId) {
         return web3Service.getKeyByTokenId(lockAddress, tokenId, network)
       } else {
--- a/unlock-app/src/components/interface/VerificationStatus.tsx
+++ b/unlock-app/src/components/interface/VerificationStatus.tsx
@@ -61,7 +61,7 @@ export const VerificationStatus = ({ config, onVerified, onClose }: Props) => {
   const { isLoading: isKeyLoading, data: key } = useQuery(
     [network, tokenId, lockAddress],
     async () => {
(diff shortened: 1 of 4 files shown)


searchableguy:
We can either merge this or don't handle the problematic past QR codes.

searchableguy:
Since we do warn people to generate QR codes if it's been a while.

julien51:
That's great. I'll approve but we should probably add comments to make sure we can clean that up in a few months.

unlock-app/src/components/interface/VerificationStatus.tsx
```diff
@@ -61,7 +61,7 @@ export const VerificationStatus = ({ config, onVerified, onClose }: Props) => {
   const { isLoading: isKeyLoading, data: key } = useQuery(
     [network, tokenId, lockAddress],
     async () => {
-      if (lockVersion && lockVersion >= 10) {
+      if (lockVersion && lockVersion >= 10 && tokenId) {
```
julien51, line 64: Some older QR codes might have been
```suggestion
      // Some older QR codes might have been generated without a tokenId in the payload. Clean up after January 2023
      if (lockVersion && lockVersion >= 10 && tokenId) {
```

unlock-app/src/utils/verification.ts
```diff
@@ -24,7 +24,7 @@ interface Options {
 export function getMembershipVerificationConfig({ data, sig }: Options) {
   try {
     if (sig && data) {
-      const raw = decodeURIComponent(data)
+      const raw = decodeURIComponent(decodeURIComponent(data))
```
julien51, line 27: Will that create issues if the payload is not encoded twice?

julien51 (approved):

julien51:
Lgtm

searchableguy:
@kalidiagne please test on your end as well and report if it doesn't break anything.

kalidiagne (approved):

kalidiagne:
> @kalidiagne please test on your end as well and report if it doesn't break anything.

Works fine!

PR #9295, opened 2022-07-22
refactor(locksmith): update sequelize, sequelize-typescript version

### LINKED ISSUE #9271: Removing `sequelize-typescript`
When we started working on locksmith, Sequelize did not have support for TypeScript so we started to use `sequelize-typescript`.
That said, sequelize now has full native support for typescript!
We should remove `sequelize-typescript` and make sure that our Sequelize setup is fully typescript ready.

As part of this work let's make sure we move to the latest version of TypeScript: https://github.com/unlock-protocol/unlock/pull/7977

m00n620:
This will close #9271


locksmith/package.json
```diff
@@ -66,19 +66,19 @@
...
-    "reflect-metadata": "0.1.13",
+    "reflect-metadata": "^0.1.13",
     "request": "2.88.2",
     "request-promise-native": "1.0.9",
-    "sequelize": "5.22.5",
-    "sequelize-cli": "6.4.1",
-    "sequelize-typescript": "1.1.0",
+    "sequelize": "^6.21.3",
+    "sequelize-cli": "^6.4.1",
+    "sequelize-typescript": "^2.1.3",
```
searchableguy, line 81: Pin versions?

locksmith/package.json
```diff
@@ -100,7 +100,7 @@
     "@types/node": "16.11.21",
     "@types/node-cron": "3.0.2",
     "@types/request-promise-native": "1.0.18",
-    "@types/sequelize": "4.28.13",
+    "@types/sequelize": "^4.28.14",
```
searchableguy, line 103: Is this needed? I thought sequelize packages types along with itself.

searchableguy (changes requested):
One question:

Why not use the infer utilities for generating type from class declaration itself instead of creating separate typescript types?

They don't seem to provide much value when both parts need to be explicitly declared by us anyway. 

https://sequelize.org/docs/v6/other-topics/typescript/#usage

locksmith/__tests__/controllers/metadataController/requestingKeyholderMetadata.test.ts
```diff
@@ -127,7 +127,7 @@ describe('Metadata Controller', () => {
         )
 
         const response = await request(app)
-          .get(`/api/key/${lockAddress}/keyHolderMetadata`)
+          .get(`/api/key/${lockAddress}/keyHolderMetadata?chain=${chain}`)
```
julien51, line 130: Is this relevant to this PR?

locksmith/migrations/20220722011748-lockConstraints.js
```diff
@@ -0,0 +1,20 @@
...
+    })
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.removeContraint(
+      table,
+      'lock_address_unique_constraint'
+    )
+  },
+}
```
searchableguy, line 20: Maybe better to be in a separate PR as this  does seem related to upgrading the package. Same with other constraint added. 

This is adding another constraint at database level.

locksmith/__tests__/controllers/metadataController/tokenData.test.ts
```diff
@@ -8,6 +8,8 @@ import { lockTypedData } from '../../test-helpers/typeDataGenerators'
 import app = require('../../../src/app')
 import Base64 = require('../../../src/utils/base64')
 
+jest.setTimeout(60000)
```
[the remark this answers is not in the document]
m00n620: Timeout happend on local

locksmith/migrations/20191001184934-usermetadataConstraint.js
```diff
@@ -1,16 +1,9 @@
...
+'use strict'
 
-const table ='UserTokenMetadata'
+const table = 'UserTokenMetadata'
 
 module.exports = {
-  up: async (queryInterface, Sequelize) => {
-    return queryInterface.addConstraint(table, ['tokenAddress', 'userAddress'],{
-      type: 'unique',
-      name : 'token_user_address_unique_constraint'
```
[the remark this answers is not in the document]
m00n620: these constraints break upsert on sequelize v6

locksmith/__tests__/controllers/metadataController/requestingKeyholderMetadata.test.ts
m00n620: yes, test will fail, because req.chain should be 31337

m00n620:
@searchableguy , It will be better to create types instead of using infer attributes. We need to handle strict types.

m00n620:
> I think you have changed some of the constrains. Would you be able to make sure this does not change?

@julien51 , to use sequelize v6 upsert, we need unique constraints

searchableguy:
Hm, I'm not sure they add more value here since you will be declaring types again on the class in the same file anyway. Inferring from that allow us to have single source of truth and sequelize typescript page has explicit utilities so that's a supported path.

locksmith/migrations/20191001184934-usermetadataConstraint.js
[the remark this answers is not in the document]
searchableguy: Please do it in another migration instead of modifying existing one because they won't be applied properly and re-check.

locksmith/package.json
julien51: +1

locksmith/__tests__/controllers/metadataController/requestingKeyholderMetadata.test.ts
julien51: ok

locksmith/migrations/20191001184934-usermetadataConstraint.js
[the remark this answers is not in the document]
julien51: @m00n620 I think @searchableguy is right. We should not modify older migrations but add new ones ...

locksmith/__tests__/controllers/metadataController/tokenData.test.ts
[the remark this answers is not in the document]
julien51: Let's remove for now though.

julien51, line 11:
```suggestion
```

locksmith/migrations/20191001184934-usermetadataConstraint.js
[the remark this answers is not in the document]
searchableguy: @m00n620

Sequelize moved away from using temporary replace function in v5 to `INSERT ON CONFLICT DO UPDATE` on postgres in v6. This requires having a unique constraint or a primary key.

We need the composite constraint because token address or user address on their own are not unique. A user address can have data on multiple locks so we require a compose constraint here. Anywhere we use upsert with usermetadata, we pass both token address as well as user address so that should work, right?

Could you please tell me where it breaks if you know or remember?

locksmith/src/controllers/v2/metadataController.ts
```diff
@@ -139,8 +139,7 @@ export class MetadataController {
           returning: true,
         }
       )
-      const statusCode = created ? 201 : 204
```
searchableguy, line 142: does not return created on sequelize v6 so can't differentiate.

locksmith/migrations/20191001184934-usermetadataConstraint.js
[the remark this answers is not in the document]
searchableguy: @m00n620 Just to confirm, this was because upsert had a validation error due to the unique constraints. 

Did you check if it was an existing bug or issue? I found https://github.com/sequelize/sequelize/issues/13734

after debugging the raw queries. Those were all using primary key on ON_CONFLICT regardless if you passed a primary or not in upsert. So I'm passing conflict field manually now. 

Removing the constraints here would lead to inserting new data instead of updating old data since the id is not passed and we depend on userAddress and tokenAddress as the unique key for conflict in our codebase. 

Let me know if I'm correct.

locksmith/src/controllers/v2/metadataController.ts
julien51: ha! that's ok

locksmith/src/models/authorizedLock.ts
```diff
@@ -1,7 +1,14 @@
 import { Table, Column, Model } from 'sequelize-typescript'
 
+interface AuthorizedLockAttributes {
+  address: string
```
julien51, line 4: It's a pity that we have to replicate these but I guess that's the only way?

locksmith/src/models/hookEvent.ts
```diff
@@ -1,7 +1,22 @@
...
+  lock?: string
+  key?: string
+  hookId: number
+  topic: string
+  body: any
+  state: string
+  lastError?: string
+  attempts: number
+  createdAt?: Date
+  updatedAt?: Date
```
julien51, line 15: We have not added them to _all_. I guess it's only needed if we access these?

locksmith/src/models/lockIcons.ts
```diff
@@ -1,7 +1,13 @@
 import { Table, Column, Model } from 'sequelize-typescript'
 
+interface LockIconsAttributes {
+  lock: string
+  chain: string
```
julien51, line 5:
```suggestion
  chain: number
```

julien51 (approved):
I think there is one type issue and then LG!

locksmith/src/models/hookEvent.ts
searchableguy: it's not really needed. Nice to have for stricter typing. m00n added it so I'm keeping it because I'm going to remove the typescript package later.

searchableguy (approved):

PR #9302, opened 2022-07-22
feat(unlock-app)- add or update email from members page

kalidiagne:
# Description
Add the ability to save/update the email address on the members page for any saved email address. Only a lock manager should be able to do it.

# Issues

Fixes #
Refs #

# Recordings 
### add/update email and keep old metadata if present 

https://user-images.githubusercontent.com/20865711/180753197-2bdf585d-a755-4d33-8ba0-396587795900.mov

### Add email flow 

https://user-images.githubusercontent.com/20865711/180740294-3b10208c-4f92-4b90-841c-ad2c637ba604.mov

### Edit email flow 
https://user-images.githubusercontent.com/20865711/180740041-8d02f65f-2dc1-4464-ad86-ee4c709d934a.mov

# Edit email screens 

- [x] show edit email button when email is present  
Screenshot 2022-07-22 at 19 10 12

# Add email screens

 - [x] show add email button when is not present 
Screenshot 2022-07-22 at 18 43 22

- [x] add email modal 
Screenshot 2022-07-22 at 18 43 22

# Checklist:


julien51:
ok that's looking good but why do we refresh the whole page when saving?

kalidiagne:
> ok that's looking good but why do we refresh the whole page when saving?

The list of members and metadata are taken from a single endpoint, not one by one for each card.
To see the value update we need to recall that endpoint

julien51:
> The list of members and metadata are taken from a single endpoint, not one by one for each card.
> To see the value update we need to recall that endpoint

But we know what we are updating, right? So there is no need to refetch the data... we can just update what we changed (as long as the API call to update was successful!

kalidiagne:
> > The list of members and metadata are taken from a single endpoint, not one by one for each card.
> > To see the value update we need to recall that endpoint
> 
> But we know what we are updating, right? So there is no need to refetch the data... we can just update what we changed (as long as the API call to update was successful!

Yep, double! just updated the PR

julien51 (approved):
Looking good! Please do a solid round of testing on this and let;s wait for @searchableguy 's feedback!

PR #9313, opened 2022-07-25
feat(unlock-app): Add payment screen and remove payment buttons from quantity page

searchableguy:
# Description

image

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/alpha/Checkout/Select.tsx
+++ b/unlock-app/src/components/interface/checkout/alpha/Checkout/Select.tsx
@@ -85,24 +85,11 @@ export function Select({ checkoutService, injectedProvider, onClose }: Props) {
                       })
                       if (account && lock) {
                         setIsLockLoading(lock.address)
-                        let existingMember = false
-                        if (
-                          lock?.publicLockVersion &&
-                          lock.publicLockVersion >= 10
-                        ) {
-                          existingMember = await web3Service?.getHasValidKey(
-                            lock.address,
-                            account,
-                            lock.network
-                          )
-                        } else {
-                          const key = await web3Service.getKeyByLockForOwner(
-                            lock.address,
-                            account,
-                            lock.network
-                          )
-                          existingMember = !!key
-                        }
+                        const existingMember = await web3Service.getHasValidKey(
+                          lock.address,
+                          account,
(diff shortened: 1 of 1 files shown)


unlock-app/package.json
```diff
@@ -66,6 +66,7 @@
     "qr-scanner": "1.4.1",
     "qrcode.react": "3.0.2",
     "react": "17.0.2",
+    "react-crypto-icons": "^1.0.5",
```
julien51, line 69:
```suggestion
    "react-crypto-icons": "1.0.5",
```

julien51 (commented):
Do you think there would be a way to re-use the close logic across all screens?

unlock-app/src/components/interface/checkout/alpha/Connect/UnlockAccountSignIn.tsx
```diff
@@ -1,5 +1,5 @@
 import { useActor } from '@xstate/react'
-import { Shell } from '../Shell'
```
julien51, line 2: Can we split changes in this file into another PR for clarity?

julien51 (commented):
There is a lot going on there. I would love to break out small pieces away into smaller PR so we only keep the payment logic to review. The rest seems straightforward.

unlock-app/src/components/interface/checkout/alpha/Checkout/Select.tsx
```diff
@@ -82,15 +83,26 @@ export function Select({ checkoutService, injectedProvider, onClose }: Props) {
                         type: 'SELECT_LOCK',
                         lock,
                       })
-                      if (account) {
+                      if (account && lock) {
                         setIsLockLoading(lock.address)
-                        const existingMember =
-                          await web3Service?.getHasValidKey(
+                        let existingMember = false
```
julien51, line 88: Do you think we should move this logic inside unlock js?

julien51: Actually i think this works for older versions already

julien51 (changes requested):
One small refactor and then lg

julien51 (approved):

PR #9325, opened 2022-07-27
feat(unlock-app) - get totals members from contract

kalidiagne:
# Description
Get total of members from contract due to graphQL limitations on query total.

# Issues

Fixes #
Refs #
Screenshot 2022-07-27 at 12 46 03
Screenshot 2022-07-27 at 12 45 40
# Checklist:

### DIFF
--- a/packages/unlock-js/src/web3Service.ts
+++ b/packages/unlock-js/src/web3Service.ts
@@ -412,6 +412,6 @@ export default class Web3Service extends UnlockService {
       lockAddress,
       this.providerForNetwork(network)
     )
-    return lockContract.numberOfOwners()
+    return ethers.BigNumber.from(await lockContract.numberOfOwners()).toNumber()
   }
 }
--- a/packages/unlock-js/src/web3Service.ts
+++ b/packages/unlock-js/src/web3Service.ts
@@ -401,4 +401,17 @@ export default class Web3Service extends UnlockService {
       this.providerForNetwork(network)
     )
   }
+
+  /**
+   * Returns numbers of owners for a specific lock
+   * @param {*} lockAddress
+   * @param {*} network
+   */
+  async numberOfOwners(lockAddress: string, network: number) {
+    const lockContract = await this.getLockContract(
+      lockAddress,
+      this.providerForNetwork(network)
+    )
+    return lockContract.numberOfOwners()
+  }
 }
--- a/unlock-app/src/hooks/useMembers.ts
+++ b/unlock-app/src/hooks/useMembers.ts
@@ -208,7 +208,7 @@ export const useMembers = ({
       data: { activeKeys },
     } = await graphService.keysCount(lockAddresses)
 
-    const totalsPromise = await Promise.all(
+    const locksTotalList = await Promise.all(
       lockAddresses.map((lockAddress) =>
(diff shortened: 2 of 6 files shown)


unlock-app/src/queries/keysCount.ts
```diff
@@ -13,7 +13,7 @@ export default function keysCount() {
         }
       }
       activeKeys: locks(where: { address_in: $addresses }) {
-        keys(where: { expiration_gt: $timestamp }) {
+        keys(where: { expiration_gt: $timestamp }, first: 1000) {
```
kalidiagne, line 16: Set max value for results entries

unlock-app/src/queries/keysCount.ts
```diff
@@ -4,7 +4,7 @@ export default function keysCount() {
   return gql`
     query keysCount($addresses: [String!], $timestamp: BigInt! = 0) {
       totalKeys: locks(where: { address_in: $addresses }) {
-        keys(where: { expiration_gt: 0 }) {
+        keys(where: { expiration_gt: 0 }, first: 1000) {
```
kalidiagne, line 7: Set max value for results entries

packages/unlock-js/src/web3Service.ts
```diff
@@ -401,4 +401,17 @@ export default class Web3Service extends UnlockService {
...
+   * Returns numbers of owners for a specific lock
+   * @param {*} lockAddress
+   * @param {*} network
+   */
+  async numberOfOwners(lockAddress: string, network: number) {
+    const lockContract = await this.getLockContract(
+      lockAddress,
+      this.providerForNetwork(network)
+    )
+    return lockContract.numberOfOwners()
```
kalidiagne, line 415: We also need from the contacts a way to get total of active keys

julien51: This might differ based on the lock version. Please test if it works with older versions.
Also when changing unlockJS, please add to the changelog and make sure you update the version in the package file!

julien51: Also, make sure you convert to a `Number` here so that applications don't have to do it themselves.

unlock-app/src/hooks/useMembers.ts
```diff
@@ -204,17 +205,22 @@ export const useMembers = ({
...
+      )
+    )
+
     // get total for every locks
     const locksActiveList: number[] = activeKeys.map(
       (lock: any) => lock?.keys?.length
     )
-    const locksTotalList: number[] = totalKeys.map(
-      (lock: any) => lock?.keys?.length
+    const locksTotalList: number[] = totalsPromise.map((total: BigNumber) =>
```
julien51, line 221: Conversion should happen in UnlockJS

unlock-app/src/queries/keysCount.ts
julien51: Is that going to break components who are using this query?

kalidiagne: Query used to retrieve active/total members, used only in this section.
Now the total members is fetched from the contract, and the active members is get from graphQL, value is set to 1000 (max) to prevent the same issue for the total (need to replace if later and get active members from contract)

julien51 (approved):

PR #9326, opened 2022-07-27
feat(unlock-app) - Update metadata keychain

kalidiagne:
# Description
Give the ability to metadata to key owner from keychain.

# Issues

Fixes #
Refs # 

** flow ** 

https://user-images.githubusercontent.com/20865711/181249418-17ed304d-8d41-4d9f-9f11-6d920c1e7d0d.mov

**Metadata form with inputs**

- Update button is disabled because no data is changed from the form 
Screenshot 2022-07-27 at 14 24 46

**Metadata form with success state on update**

Screenshot 2022-07-27 at 14 25 03

 **Metadata form with loading state on update**

Screenshot 2022-07-27 at 14 31 34

 **Metadata form with validation if field is empty**

Screenshot 2022-07-27 at 14 32 10

# Checklist:


unlock-app/src/components/interface/keychain/KeyMetadataModal.tsx
```diff
@@ -92,31 +133,52 @@ export const KeyMetadataModal: React.FC<MetadataModalProps> = ({
...
-                    })}
-                  />
+                  <div key={index}>
+                    <Input
+                      label={key}
+                      key={index}
+                      disabled={updating}
+                      {...register(key, {
+                        value,
+                        required: true,
```
kalidiagne, line 149: Set all fields as required, to avoid users emptying some fields.
There is a way to get the required fields? 

actually is only set in checkout flow with `metadataInputs`? 
@julien51  @searchableguy

julien51:
Let's this one for for now as I think we should keep the "type" information and we don't currently store this.

PR #9349, opened 2022-07-29
press kit set

ChialiTsai:
updating logo along with press kit.
this is to help contributors, partners to easily download our logos and tips on how to use them.


julien51 (approved):
💯

PR #9420, opened 2022-08-03
feat(unlock-app): Check all available memberships for the user by default

searchableguy:
…

# Description

We check for all possible memberships now instead of one user clicks. This does result in a slight lock reload when you login but it works fine, I think.

# Issues

Fixes #
Refs #

# Checklist:


julien51 (approved):

julien51:
I think it is ok. In the _vast_ majority of cases there won't be more than 1 or 2 locks anyway.

PR #9474, opened 2022-08-19
feat(unlock-app) -  keys controller

kalidiagne:
# Description
This PR adds a controller for a new endpoint to list of keys based on a specific query key (`keyId`, `email`, `owner`) 

In order to simplify and to get metadata and keys information in different data structures, so to achieve the result all data for the specific`lockAddress` will be retrieved and merged with the respective metadata. For that reason, as a side effect were are going to lose pagination because we get all keys at one or based on the query. discussed that with @searchableguy 

# Issues

Fixes #
Refs #

# Checklist:


kalidiagne:
> You can still add pagination. There is no limitation on that even if we fetch all keys at once but can provide artificial pagination to end users in any increments.
> 
> I would recommend it in case we need to optimize endpoints in future.

Oh yes, this is also an alternative, double directly from the FE.

kalidiagne:
> You can still add pagination. There is no limitation on that even if we fetch all keys at once but can provide artificial pagination to end users in any increments.
> 
> I would recommend it in case we need to optimize endpoints in future.

Opened specific PR with artificial pagination: https://github.com/unlock-protocol/unlock/pull/9488

julien51 (approved):

PR #9481, opened 2022-08-19
fix(unlock-app): Use correct currency and only show refund on connected network keys

searchableguy:
# Description

Key cancel and refund wasn't using correct currency if the lock was denominated in anything other than base currency. I've changed it to show the correct currency. 

Keychain does not handle lock changes properly. I've disabled the button if the network does not match the connected network.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/utils/keyPricer.ts
+++ b/locksmith/src/utils/keyPricer.ts
@@ -29,6 +29,13 @@ export default class KeyPricer {
       case 100:
         // we max at $1
         return gasCost < 100
+
+      case 137:
+        return gasCost < 100
+
+      case 10:
+        return gasCost < 100
+
       default:
         // We max at 1 cent
         return gasCost < 1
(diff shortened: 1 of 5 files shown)


kalidiagne (approved):

unlock-app/src/components/interface/keychain/Key.tsx
```diff
@@ -229,8 +239,15 @@ const Key = ({ ownedKey, account, network }: Props) => {
...
   const isAvailableOnOpenSea = [1, 4, 137].indexOf(network) > -1
-  const baseCurrencySymbol =
-    walletService.networks[network].baseCurrencySymbol ?? ''
+  const baseSymbol = walletService.networks[network].baseCurrencySymbol!
+  const symbol = isLockDataLoading
+    ? baseSymbol
+    : lockTickerSymbol(lockData, baseSymbol)
+
+  const isRefundable =
+    !isLockDataLoading && !isKeyExpired && lockData.publicLockVersion >= 10
```
julien51, line 248: Refunds have been availbale for longer than version 10...

julien51 (commented):
There is also an issue where the amount displayed has way too many decimals. Do you mind fixing it at the same time?

PR #9497, opened 2022-08-23
fix(unlock-app): Use lock specific constraints and hide captcha step when using credit card

searchableguy:
# Description

Currently, we show captcha step when person is using credit card in the progress bar. It is skipped but I think it should be hidden.

Refactoring in the quantity page to use lock specific constraints over global. 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/alpha/Checkout/Captcha.tsx
+++ b/unlock-app/src/components/interface/checkout/alpha/Checkout/Captcha.tsx
@@ -20,7 +20,7 @@ export function Captcha({ injectedProvider, checkoutService }: Props) {
   const config = useConfig()
   const storage = useStorageService()
   const [recaptchaValue, setRecaptchaValue] = useState<string | null>(null)
-  const { recipients } = state.context
+  const { recipients, payment } = state.context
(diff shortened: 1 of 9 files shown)


unlock-app/src/components/interface/checkout/alpha/Checkout/Captcha.tsx
```diff
@@ -83,7 +83,8 @@ export function Captcha({ injectedProvider, checkoutService }: Props) {
       id: 6,
       name: 'Solve captcha',
       to: 'CAPTCHA',
-      skip: !paywallConfig.captcha,
+      skip:
+        !paywallConfig.captcha || ['card', 'claim'].includes(payment.method),
```
kalidiagne, line 87: is not needed to show the captcha for credit card? what if `captcha=true` and payment method is `card` or 'claim'? it this case the captcha will not be visible

kalidiagne: Awesome! so LGTM!

kalidiagne (approved):

PR #9498, opened 2022-08-23
Ccarfi update checkout blog post for launch

ccarfi:
a number of updates to searchableguys's blog post

- updated to be a "launch" post instead of "coming soon"
- rephrased a number of things to connect with "we've done this to improve user experience"
- removed copy around the /alpha/ path
- added copy around the /legacy/ path
- updated share image to be a GIF


ccarfi:
added newest screenshots from @searchableguy's post

julien51 (commented):
I wonder if we should rather publish a whole new post?

ccarfi:
i made the decision to update the old post for two reasons:

1) anyone finding the old post would be finding out of date information (we're no longer in alpha)
2) it's better to have a single post about a thing in most cases, rather than split google juice.

julien51:
Ok, works for me!

ccarfi:
@julien51 or @searchableguy can you please review/approve so this can be merged? thank you!

ccarfi:
>Shouldn't we wait until tomorrow?

once it's reviewed and approved, we can merge any time we want. 

(and it already is "tomorrow" in a number of time zones. :) )

PR #9506, opened 2022-08-24
fix(unlock-app): Hide credit card and superfluid option when recurring payment is enabled

searchableguy:
# Description

I think we should hide the option since even if users confirm, they won't be charged automatically for subsequent purchase. 

I will enable it once we have support for stripe subscriptions. 

# Issues

Fixes #
Refs #

# Checklist:


julien51 (commented):
I am not sure we should hide them. It's ok to have them as is right now... I think users would be more confused to _not_ see these options.

julien51:
> Then maybe hide the recurring part on confirmation screen since the payment is not actually recurring.

Do we show a message about it being recurring when using CC or Superfluid?

> It will still go through but user will be shown normal confirmation page for singe purchase.

I think that's the correct behavior.

julien51 (approved):

Issue #8851, opened 2022-05-23
Convert unlock-asset package to typescript


searchableguy:
**Is your feature request related to a problem? Please describe.**
Types are missing from unlock-assets package because we haven't set up svgr to generate ts files instead of js for svg components. 

**Describe the solution you'd like**
Add typescript support by modifying the template and package.

julien51:
Maybe @ m00n620 might be interested :)

jordang7:
Hey, I was playing around with this a little and it seems the SVGR CLI tool has some weird compatibility issues with typescript. I am only able to get the generated components to be typescript files. Couldn't find a solution to convert the actual template file and the rest into typescript.

Let me know if just generating the ts files has some value, or if the whole package needs to be typescript as well.

PR #9515, opened 2022-08-25
refactor(unlock-assets): Generate typescript components

jordang7:
# Description

Modified unlock-assets package to have svgr generate typescript components
- Removed js files in build/svg-components and replaced with typescript
- Added interfaces variable to template.js to allow for typescript and --typescript flag added to svg-2-components script

# Issues

#8851 

Fixes #
Refs #

# Checklist:


jordang7:
> Can you run `unlock-app` after rebuilding everything locally and see if the icons are displayed properly on checkout, dashboard, etc?

Yes just did a packages:clean and build and the icons look to be rendering properly. 
Screenshot from 2022-08-29 10-49-17

jordang7:
Secondly, while we are generating typescript components for build folder. We aren't actually outputing them in dist when we use bundler to export them. Please follow the unlock-js build command and replace micro bundler here with what we are using there to output the typescript definition as well.

Just replaced microbundler with tsup. However I can't seem to replicate the watermark not displaying. It's showing on the demo page correctly for me. 
unl

Could you please send me the steps you took to produce the bug? Thanks!

packages/unlock-assets/package.json
```diff
@@ -5,11 +5,11 @@
   "author": "unlock inc",
   "license": "MIT",
   "main": "dist/index.js",
-  "module": "dist/index.modern.js",
-  "source": "src/index.js",
+  "module": "dist/index.mjs",
+  "source": "src/index.tsx",
```
searchableguy, line 9: add `typings` property referencing the ts definition file.

packages/unlock-assets/package.json
```diff
@@ -5,11 +5,11 @@
   "author": "unlock inc",
   "license": "MIT",
   "main": "dist/index.js",
-  "module": "dist/index.modern.js",
-  "source": "src/index.js",
+  "module": "dist/index.mjs",
```
searchableguy, line 8: Use `exports` instead of module defined on the node official documentation site.

searchableguy:
Reproduction steps:

```shell
git pull jordan branch
cd packages/unlock-assets
yarn install
yarn svg-2-components
yarn build
cd ../../unlock-app
yarn dev
```

The icon does not show up for me. Some are missing.

jordang7:
> CleanShot 2022-09-10 at 02 59 52@2x
> 
> There are some icons which have changed and aren't appearing. This is from the checkout bottom - unlock word mark.
> 
> Can you check and fix what caused this difference after rebuilding?

Did a total clean/rebuild and I still haven't been able to replicate this bug on my end so haven't been able to find a fix. This screenshot is from the demo feature on the unlock-app dashboard, correct? Or is there another way you are reaching this checkout page? Also are there any console errors when seeing this bug?

searchableguy:
Hm, that's very odd. Can you tell me your reproduction steps, browser environment, and OS?

jordang7:
Ah yes that was it!

I can see the bug now, thanks. Will take a look at figure out what's happening.

searchableguy:
is this fixed?

jordang7:
Not fully. I fixed the way it displays on the checkout page but I realize it broke the way it displays on the dashboard if forContent is true (shows black instead of brand color). Need to look into it a little more

jordang7:
Yes should be fixed now

searchableguy (approved):
Approving. I've made some changes to the template. 

Though type checking isn't available in the unlock-app package right now due to vendor.d.ts and there are some changes required there but that's out of scope here.

PR #9517, opened 2022-08-26
feat(unlock-app) - check key validation with label to enable actions

kalidiagne:
# Description
fix and simplify how key statatus is checked to enable/disable `Expire and Refund` button.
to do this `expirationAsDate` fn is used, because have one of this states `never` | `expired` | `${VALID_DATE}`

# Issues

Fixes #
Refs #

**Key not expired** 
Screenshot 2022-08-26 at 11 45 42

**Expired keys** 
Screenshot 2022-08-26 at 11 45 58

**Key with an expiration date** 
Screenshot 2022-08-26 at 11 53 53

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/MetadataTable.tsx
+++ b/unlock-app/src/components/interface/MetadataTable.tsx
@@ -131,8 +131,6 @@ export const MetadataTable: React.FC<MetadataTableProps> = ({
   }
 
   const isKeyValid = (metadata: KeyMetadata) => {
-    if (!metadata?.expiration) return // expiration not present
-
     const expiration = expirationAsDate(metadata?.expiration)
     if (expiration.toLowerCase() === 'never') return true
     if (expiration.toLowerCase() === 'expired') return false
--- a/unlock-app/src/components/interface/MetadataTable.tsx
+++ b/unlock-app/src/components/interface/MetadataTable.tsx
@@ -8,7 +8,6 @@ import {
   MemberCardPlaceholder,
 } from '../interface/members/MemberCard'
 import { Button } from '@unlock-protocol/ui'
-import { expirationAsDate } from '~/utils/durations'
 interface KeyMetadata {
   // These 3 properties are always present -- they come down from the graph as
   // strings
--- a/unlock-app/src/components/interface/MetadataTable.tsx
+++ b/unlock-app/src/components/interface/MetadataTable.tsx
@@ -8,6 +8,7 @@ import {
   MemberCardPlaceholder,
 } from '../interface/members/MemberCard'
 import { Button } from '@unlock-protocol/ui'
+import { expirationAsDate } from '~/utils/durations'
 interface KeyMetadata {
   // These 3 properties are always present -- they come down from the graph as
   // strings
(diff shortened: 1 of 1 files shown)


unlock-app/src/components/interface/MetadataTable.tsx
```diff
@@ -130,11 +131,12 @@ export const MetadataTable: React.FC<MetadataTableProps> = ({
   }
 
   const isKeyValid = (metadata: KeyMetadata) => {
-    if (!metadata?.expiration) return false
-    if (metadata?.expiration.toLowerCase() === 'never') return true
-    const now = new Date().getTime()
-    const expiration = new Date(metadata?.expiration).getTime()
-    return expiration > now
+    if (!metadata?.expiration) return // expiration not present
```
julien51, line 134: I am confused about _why_ the expiration would be missing?

unlock-app/src/components/interface/MetadataTable.tsx
```diff
@@ -130,11 +131,12 @@ export const MetadataTable: React.FC<MetadataTableProps> = ({
...
+
+    const expiration = expirationAsDate(metadata?.expiration)
```
julien51, line 136: I don't think it is the rigth approach.
I assume `metadata?.expiration` is a number?

If the number is smaller than the timestamp for "now" then it is expired.
Otherwise it is not.

unlock-app/src/components/interface/MetadataTable.tsx
kalidiagne: Non necessary, now the function recive directly `expiration` as parameter

unlock-app/src/components/interface/MetadataTable.tsx
kalidiagne: this is because `metadata.expiration` is a string

```
interface KeyMetadata {
  // These 3 properties are always present -- they come down from the graph as
  // strings
  lockName: string
  expiration: string
  keyholderAddress: string
  // Can have any other arbitrary properies, as long as the values are strings.
  [key: string]: string
}
```

kalidiagne: `expiration` parsed as numbed to check as suggested

unlock-app/src/components/interface/MetadataTable.tsx
```diff
@@ -129,19 +129,16 @@ export const MetadataTable: React.FC<MetadataTableProps> = ({
...
 
-  const isKeyValid = (metadata: KeyMetadata) => {
-    if (!metadata?.expiration) return false
-    if (metadata?.expiration.toLowerCase() === 'never') return true
-    const now = new Date().getTime()
-    const expiration = new Date(metadata?.expiration).getTime()
-    return expiration > now
+  const isKeyValid = (timestamp: KeyMetadata['expiration']) => {
+    const now = new Date().getTime() / 1000
+    return parseInt(timestamp) > now
```
julien51, line 134: Unfortunately you need to handle the case where `timestamp` is MAX UINT (unlimited keys!) because parseInt will fail..

kalidiagne: should work also with `parseInt`, added check for `MAX_UINT` value

julien51 (approved):
Did you test it with a few locks on Goerli or other test nets?

kalidiagne:
> Did you test it with a few locks on Goerli or other test nets?

Tested with multiple locks and lock versions

Issue #9451, opened 2022-08-16
Check if key can be granted


julien51:
Our UI to airdrop keys does not check if a key can effectively be granted.
We should do that and show an error message to the user when it cannot.
To check if a key can be granted, we need to retrieve the "max key per address" setting on the lock, and see if a given addresss already has one (or more) keys.

kalidiagne:
Solved by https://github.com/unlock-protocol/unlock/pull/9527

kalidiagne:
Solved by https://github.com/unlock-protocol/unlock/pull/9527

PR #9527, opened 2022-08-29
check if key can be granted

kalidiagne:
# Description
Check if key can be granted based on `maxKeysPerAddress`.
If the limit is reached (total keys owned + total keys to aidrop) the error message will show to the user to inform that the limit is reached.

**extra:**
- Also added address minify for better reading to user

# Issues 

Fixes # https://github.com/unlock-protocol/unlock/issues/9451 
Refs #

Screenshot 2022-08-29 at 22 00 39

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useMultipleRecipient.ts
+++ b/unlock-app/src/hooks/useMultipleRecipient.ts
@@ -117,7 +117,7 @@ export const useMultipleRecipient = (
     return recipients.size < maxRecipients
   }
 
-  const isKeyCanBeGranted = async ({
+  const canKeyBeGranted = async ({
     lockAddress,
     owner,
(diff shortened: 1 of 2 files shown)


unlock-app/src/hooks/useMultipleRecipient.ts
```diff
@@ -115,30 +117,48 @@ export const useMultipleRecipient = (
     return recipients.size < maxRecipients
   }
 
+  const isKeyCanBeGranted = async ({
```
julien51, line 120:
```suggestion
  const canKeyBeGranted = async ({
```

julien51 (approved):
Let's use a better name for the function and then LG!

PR #9529, opened 2022-08-30
feat(unlock-app) - ability to extend key from frontend

kalidiagne:
# Description
This PR adds the ability to extend a key from frontend. 

For every key, if the is expired or have one expiration day, a new button `Extend key` will shown (Keys with unlimited duration are already fine, and dont need to extend)

**Loading state**
Screenshot 2022-08-31 at 16 09 35

**Keys with a end date have 'Extend key' button**
Screenshot 2022-08-31 at 10 18 59

**Expired keys with 'Extend key' button**
Screenshot 2022-08-31 at 10 19 16

**Expired by date*
Screenshot 2022-08-31 at 14 31 17

**Set key never expires**
Screenshot 2022-08-31 at 14 29 05

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Ok first thing first: the "extend key" button should be displayed *inside* the member item (close to the "expire" button).

kalidiagne:
> Ok first thing first: the "extend key" button should be displayed _inside_ the member item (close to the "expire" button).

I think we dont simplity things with this aproach, the lockManager have to do a lot of steps to extend more than one key 
- Filter by 'expiration' -> Filter by 'expired' ->  search key to expire -> Repet for one by one for all of the keys he want to extend

instead, he can select more that one keys at once and define the duration extend time, the dispayed keys are only the expired ones.

julien51:
> I think we dont simplity things with this aproach, the lockManager have to do a lot of steps to extend more than one key

The case where the lock manager extends more than one key is pretty rare. 
We should optimise for simplicity in the most common use case: the lock manager extends a single key. (it's very similar to how the lock manager can cancel a key!)

julien51 (approved):

julien51:
Hum, what is this page reload???

unlock-app/src/components/creator/members/ExtendKeysDrawer.tsx
```diff
@@ -0,0 +1,240 @@
...
+    expiration: currentExpiration,
+  } = selectedKey ?? {}
+
+  if (!lockAddress && !tokenId) return null
+
+  const onComplete = async () => {
+    setIsOpen(false)
+    if (typeof refetch === 'function') {
+      await refetch()
+      // force reload to show keys in defaut list
```
julien51, line 201: No... I have told you in the past, a full page reload is *never* the right solution,.

julien51: Why do you need to reload? If it is just to update the value of the expiration, just update it from what the user entered?

kalidiagne: When we extend expired key, generally we are filtering by `Expired -> expired`, need to go back to the previous list to see the key, or change filters because it will no be present with the actual filter

kalidiagne: Ill change it with a nicer way

kalidiagne: @julien51 updated!

unlock-app/src/components/creator/members/ExtendKeysDrawer.tsx
```diff
@@ -196,12 +196,9 @@ export const ExtendKeysDrawer = ({
...
   const onComplete = async () => {
     setIsOpen(false)
-    if (typeof refetch === 'function') {
-      await refetch()
-      // force reload to show keys in defaut list
-      setTimeout(() => {
-        window.location.reload()
-      }, 2000)
+    if (typeof resetSearchFilters === 'function') {
+      console.log('reset filters')
```
julien51, line 200: Let's remove this please :)

julien51 (approved):
ok that's better!

PR #9535, opened 2022-08-30
fix(locksmith): Handle coinbase not returning key price

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/utils/priceConversion.ts
```diff
@@ -28,7 +28,12 @@ export default class PriceConversion {
...
       )
 
+      if (!response.ok) {
+        return undefined
+      }
+
       const { data } = await response.json()
+
       if (!data?.amount) {
         return 0
```
julien51, line 38: Maybe we also handle things here... if there is no `amount` I think we can assume it's not "convertible"

julien51 (approved):

PR #9540, opened 2022-08-31
feat(unlock-app): Hide payment option based on eligibility

searchableguy:
# Description

Hide crypto payment option like others when balance is low or unlock account is used to discourage user funding it.

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Thanks for this!
I think the rules should be these:
* if there are *no* other option, show the "crypto" checkout as the only option...
* if claim is available, only show it for users who don't have tokens to pay for gas

unlock-app/src/components/interface/checkout/alpha/Checkout/Payment.tsx
```diff
@@ -231,6 +244,15 @@ export function Payment({ injectedProvider, checkoutService }: Props) {
                 </div>
               </button>
             )}
+            {allDisabled && (
+              <div>
+                <p className="text-sm">
```
julien51, line 249: I like that!

julien51 (approved):

PR #9550, opened 2022-09-02
fix(hardhat-plugin): hardhat plugin improvements and fixes

### LINKED ISSUE #9525: Hardhat plugin improvements
As we are working on integration testing, it will be great to use our hardhat plugin internally to deploy various versions and test the subgraph against it. Here are a list of some fixes / improvements that could remove extra steps and make the plugin MUCH easier to use.

### TS support

Currently, the way types are defined are not very logical and doesnt work properly when importing `import { unlock } from 'hardhat'` in a `.ts` file

### Explicit contracts in `/contracts` folder

For instance, it will be much easier to have all contracts explicitly as a simple import in the `contracts` folder -- `contracts/UnlockV6`, `contracts/UnlockV7`, etc (like [Uniswap](https://github.com/Uniswap/hardhat-v3-deploy/tree/master/contracts) plugin). We can still generate this from our contracts package source before deploying, but better to NOT hide this logic so its easy to check and modify and don't create overhead at runtime.

### Even better : deploy from bytecode

Currently we are using OZ upgrades plugin so we need to import the plugin, specify all solc versions, build the contracts, etc. Thats a lot of overhead and we should be able to just deploy a basic transparent proxy (from bytecode) and deploy the Unlock contract directly using it.

These are the most important ones that will allow to use it easily across our own code - and makes things easier for people to test hooks etc on locks locally before going live.

clemsos:
# Description

This removes several hardhat dependencies (esp. OZ upgrades plugin) and simplifies the way deployments for testing are handled. This also adds a bunch of typescript fixes

# Issues

Fixes #9525 
Refs #

# Checklist:


julien51 (commented):
This is looking very good! 

Please add multiple example to the README and/or an `example` folder and let's make sure we add this to the `tools` section of our docs too!

clemsos:
ok so that all looks fine now, much cleaner and simpler imo. 

I created a [ts example](https://github.com/unlock-protocol/hardhat-plugin-ts-example) and updated the existing [js example repo](https://github.com/unlock-protocol/hardhat-plugin-example). Not sure both are needed but it was to make sure everything was working properly w Typescript

julien51:
ok we need to move some of the docs to the docs repo (or create a new guide?) 
What do you think @wonderwomancode ?

julien51 (approved):
LG

Issue #9458, opened 2022-08-18
Use Local Memory Type Variable Instead of Global Storage Type Variable in Event to Save Gas


Ether1oop:
Hi, we recently have conducted a systematic study about Solidity event usage, evolution, and impact, and we are attempting to build a tool to improve the practice of Solidity event use based on our findings. We have tried our prototype tool on some of the most popular GitHub Solidity repositories, and for your repository, we find a potential optimization of gas consumption arisen from event use.  

The point is that when we use emit operation to store the value of a certain variable, local memory type variable would be preferable to global storage type (state) variable if they hold the same value. The reason is that an extra SLOAD operation would be needed to access the variable if it is storage type, and the SLOAD operation costs 800 gas.

For your repository, we find that the following event use can be improved:

- [PublicLockV4.sol](https://github.com/unlock-protocol/unlock/blob/master/packages/contracts/src/contracts/PublicLock/PublicLockV4.sol)
**function name:**&emsp;*initialize*
**event name:**&emsp;&emsp;*OwnershipTransferred*
**variable:**&emsp;&emsp;&emsp;&emsp;*_owner->sender*

```Solidity
    function initialize(address sender) public initializer {
        _owner = sender;
        emit OwnershipTransferred(address(0), _owner);
    }
```
&emsp;&emsp;**function name:**&emsp;*updateKeyPrice*
&emsp;&emsp;**event name:**&emsp;&emsp;*PriceChanged*
&emsp;&emsp;**variable:**&emsp;&emsp;&emsp;&emsp;*keyPrice->_keyPrice*
```Solidity
  function updateKeyPrice(
    uint _keyPrice
  )
    external
    onlyOwner
    onlyIfAlive
  {
    uint oldKeyPrice = keyPrice;
    keyPrice = _keyPrice;
    emit PriceChanged(oldKeyPrice, keyPrice);
  }
```

- [PublicLockV0.sol](https://github.com/unlock-protocol/unlock/blob/master/packages/contracts/src/contracts/PublicLock/PublicLockV0.sol)
**function name:**&emsp;*updateKeyPrice*
**event name:**&emsp;&emsp;*PriceChanged*
**variable:**&emsp;&emsp;&emsp;&emsp;*keyPrice->_keyPrice*

```Solidity
  function updateKeyPrice(
    uint _keyPrice
  )
    external
    onlyOwner
  {
    uint oldKeyPrice = keyPrice;
    keyPrice = _keyPrice;
[...]

PR #9565, opened 2022-09-07
[fix]Use Local Memory Type Variable Instead of Global Storage Type Variable in Event to Save Gas

Ether1oop:
# Description
Hi, we are a research group on programming languages and software engineering. We recently have conducted a systematic study about Solidity event usage, evolution, and impact, and we are attempting to build a tool to improve the practice of Solidity event use based on our findings. We have tried our prototype tool on some of the most popular GitHub Solidity repositories, and for your repository, we find a potential optimization of gas consumption arisen from event use.  

The point is that when we use emit operation to store the value of a certain variable, `local memory type variable` would be preferable to `storage type (state) variable` if they hold the same value. The reason is that an extra SLOAD operation would be needed to access the variable if it is storage type, and the SLOAD operation costs 800 gas.

For your repository, we find that several event uses can be improved.

# Issues

[#9458](https://github.com/unlock-protocol/unlock/issues/9458)

# Checklist:

### DIFF
--- a/packages/contracts/src/contracts/PublicLock/PublicLockV0.sol
+++ b/packages/contracts/src/contracts/PublicLock/PublicLockV0.sol
@@ -976,7 +976,7 @@ contract PublicLock is ILockCore, ERC165, IERC721, IERC721Receiver, Ownable {
   {
     uint oldKeyPrice = keyPrice;
     keyPrice = _keyPrice;
-    emit PriceChanged(oldKeyPrice, keyPrice);
+    emit PriceChanged(oldKeyPrice, _keyPrice);
   }
 
   /**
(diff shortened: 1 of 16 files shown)


julien51:
@Ether1oop Thanks a lot! would you be able to post about the savings here? How much exactly are they?

wonderwomancode:
@Ether1oop I'm Angela Steffens and I handle developer relations and community here at Unlock Protocol. I'd love to nominate you for our retroactive funding for your contribution here. I just need your wallet address so we know where to send your UDT reward if you receive the votes. It's also not against the rules to nominate yourself if you'd like to do that. https://unlock.community/t/retroactive-funding-september-2022/347

PR #9569, opened 2022-09-07
feat(locksmith): Adding a script for dappcon so we can resend QR codes 

julien51:
# Description

When dappcon went on sale we did not send QR codes as part of the confirmation email.... so we need to send another email with the QR code to all the attendees to make sure everybody has it.

# Checklist:


locksmith/dappcon-mailer.js
```diff
@@ -0,0 +1,52 @@
...
+    page,
+    filterKey: '',
+    expiration: 'active',
+  }
+
+  const keys = await keysOperations.getKeysWithMetadata({
+    network,
+    lockAddress,
+    filters,
+    loggedInUserAddress: '0x2F3c9b71f256E7450c47872aE00389C616B08646',
```
kalidiagne, line 21: @julien51 this is not the lockAddress?

julien51: Ha it is but the address does not matter here :)

kalidiagne (approved):

PR #9580, opened 2022-09-08
feat(unlock-app) - ignore recipient key

kalidiagne:
# Description
Ignore recipient key in member card

**Issue**
Screenshot 2022-09-08 at 13 09 18

**Fix**
Screenshot 2022-09-08 at 13 09 27

# Issues

Fixes #
Refs #

# Checklist:


julien51:
Hum. I don't think this is enough. If I understand correctly this, won't show it but what we need is to not *save* it.

julien51 (changes requested):
@kalidiagne This is not enough. It's not about not showing but also not saving that data!

PR #9590, opened 2022-09-12
fix(locksmith): Don't round decimal prices in fiat pricing API

searchableguy:
# Description

The code was rounding 0.x decimal values to 0 due to `parseInt`. I've removed the rounding and used big number js here for precision. It will remove trailing 0s or any floating point oddities. 

We can't use bigNumber provided by ethers since that only deals with integers, not decimals.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/packages/networks/src/networks/avalanche.ts
+++ b/packages/networks/src/networks/avalanche.ts
@@ -36,6 +36,38 @@ export const avalanche: NetworkConfig = {
   isTestNetwork: false,
   description:
     'Avalanche is an open, programmable smart contracts platform for decentralized applications.',
+  tokens: [
+    {
+      name: 'Ethereum',
+      symbol: 'WETH',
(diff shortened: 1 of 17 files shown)


locksmith/package.json
```diff
@@ -50,6 +50,7 @@
     "apollo-link-http": "1.5.17",
     "apollo-server": "3.6.7",
     "apollo-server-express": "2.25.4",
+    "bignumber.js": "9.1.0",
```
julien51, line 53: Can we use the Ethers version please? These are different libs and I really want to avoid using multiple libs for big numbers!

julien51 (changes requested):
Approach is good but please let's use `ethers.BigNumber` to reduce number of dependencies!

julien51 (approved):

PR #9591, opened 2022-09-12
feat(unlock-app) - add tokens list to network

kalidiagne:
# Description
On our new create lock page, we need to give to the user the ability to select a specific token from the selected network, to provide this feature, for every network we have in unlock we need to have the list of tokens.

# Issues

Fixes #
Refs #

# Checklist:


julien51 (commented):
This is a great idea! Can we actually add that inside the `networks` package to make things more reusable?

kalidiagne:
> Where is this copied from?
> 
>     1. It introduces third party dependency. All those urls pointing to assets might not work (raw github paths can change) or need to be updated constantly.
> 
>     2. We can't add all erc-20s as suggestions. That would be tough to maintain.
> 
>     3. Have you looked to see if there is a standard API or way to get these? So we don't have to deal with 1 and 2.

Discussed with Julien last week, which suggested to use list from https://tokenlists.org/, but I see only Ethereum network are present, so the alternative is to use an internal list managed by us. 

The list is from different resources from github, i think we can keep the list and ignore `logoURI` to use `react-crypto-icons` instead for the currency logo

packages/networks/src/networks/mainnet.ts
```diff
@@ -38,6 +38,1028 @@ export const mainnet: NetworkConfig = {
   description: 'The most popular network',
   isTestNetwork: false,
   teamMultisig: '0xa39b44c4AFfbb56b76a1BF1d19Eb93a5DfC2EBA9',
+  tokens: [
+    {
```
julien51, line 42: I think there is too many here... Let's keep the following: 
WETH, DAI, USDC, WBTC, MATIC, BAT and that;s it for now.

packages/networks/src/networks/polygon.ts
```diff
@@ -43,6 +43,224 @@ export const polygon: NetworkConfig = {
   description: 'Popular side chain network. Cheaper transaction cost.',
   isTestNetwork: false,
   teamMultisig: '0x479f3830fbd715342868BA95E438609BCe443DFB',
+  tokens: [
+    {
```
julien51, line 47: Same here... we do not need all of these tokens by default.

julien51 (approved):

packages/networks/src/networks/mainnet.ts
kalidiagne: also keep `USDT` in list

PR #9600, opened 2022-09-13
feature(unlock-js, locksmith): using wallet service to renewMembershipFor so we can bump gas

julien51:
# Description

I noticed that some renewal transactions are running out of gas... it is because the estimate is sometimes not accurate.
Moving the logic to unlock-js allows us to also add a way to bump gas limit!

# Checklist:


clemsos (approved):
LGTM! 

Side note: why are we passing the chainId as string in unlock js? Seems like we may want to change that no?

PR #9607, opened 2022-09-14
feat(ui) - add collapse component

kalidiagne:
# Description
Add `Collapse`  component to UI library

https://user-images.githubusercontent.com/20865711/190169260-fe85089c-eb08-40b4-9024-8a9befe98737.mov

# Issues

Fixes #
Refs #

# Checklist:


packages/ui/lib/components/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,52 @@
...
+  collapsed = false,
+}: CollapseProps) => {
+  const [isOpen, setIsOpen] = useState(false)
+
+  const onCollapse = () => {
+    if (disabled) return
+    setIsOpen(!isOpen)
+  }
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
kalidiagne: i think the button should be this one, and not customizable, the structure will be the same and allow to customize content inside. 

what if the custom icon is not an arrow? for what we actually need I think we can keep it as is.
ref: https://www.figma.com/file/pNN7ER1F4mdToFfk1bB0gu/App-v2?node-id=1432%3A21155

packages/ui/lib/components/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,52 @@
...
+    setIsOpen(collapsed)
```
kalidiagne, line 25: @searchableguy the idea behind is to enable the ability to collapse/expand for a different source.

So that is why the component has `collapsed` props that change the internal state.
Any suggestion to do it better? member card for example can be expanded/closed from the filter bar

https://www.figma.com/file/pNN7ER1F4mdToFfk1bB0gu/App-v2?node-id=1587%3A28822

searchableguy: Do you need `useEffect` here to change the state to default?

You can pass default value in the useState directly. Correct me if I'm missing something.

Since useState is a single line use. A user can use it outside the pure view component that this is and pass function and state down.

PR #9608, opened 2022-09-14
feat(unlock-app): Add support for payment element

searchableguy:
# Description

CleanShot 2022-09-15 at 10 33 46@2x

CleanShot 2022-09-15 at 14 39 10@2x

Stripe payment element require setupIntent to save payment details for future usage instead of tokens which was the case with cards element. 

The token & source API has been superseded by the payment methods API. So I've made a minor change in the payment intent to clone payment methods on connected account and use it directly. 

This flow *should* support enabling apple and google pay from the stripe dashboard but testing them require a https environment and merchant verification (apple pay). 

As part of the clean up, I've also moved a few API endpoints to using SIWE auth. 

https://stripe.com/docs/payments/payment-methods/transitioning

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/controllers/v2/purchaseController.ts
```diff
@@ -0,0 +1,60 @@
...
+import PaymentProcessor from '../../payment/paymentProcessor'
+import * as Normalizer from '../../utils/normalizer'
+import logger from '../../logger'
+const config = require('../../../config/config')
+
+export class PurchaseController {
+  processor = new PaymentProcessor(config.stripeSecret)
+  async createSetupIntent(request: Request, response: Response) {
+    try {
+      const userAddress = Normalizer.ethereumAddress(
```
julien51, line 15: I wonder if we should start doing this in the middleware..

unlock-app/public/.well-known/apple-developer-merchantid-domain-association
```diff
@@ -0,0 +1 @@
...33303131303630333535303430613063306134313730373036633635323034393665363332653331306233303039303630333535303430363133303235353533303230383463333034313439353139643534333633303064303630393630383634383031363530333034303230313035303061303831393533303138303630393261383634383836663730643031303930333331306230363039326138363438383666373064303130373031333031633036303932613836343838366637306430313039303533313066313730643331333933303338333133393331333733313332333333303561333032613036303932613836343838366637306430313039333433313164333031623330306430363039363038363438303136353033303430323031303530306131306130363038326138363438636533643034303330323330326630363039326138363438383666373064303130393034333132323034323062303731303365313430613462386231376262613230316130336163643036396234653431366232613263383066383661383338313435633239373566633131333030613036303832613836343863653364303430333032303434363330343430323230343639306264636637626461663833636466343934396534633035313039656463663334373665303564373261313264376335666538633033303033343464663032323032363764353863393365626233353031333836363062353730373938613064643731313734316262353864626436613138363633353038353431656565393035303030303030303030303030227D
```
julien51: Let's isolate that in its own PR please with explanation of what that is and where to get it if it needs to be updated!

unlock-app/src/components/interface/checkout/alpha/Checkout/CardPayment.tsx
```diff
@@ -106,82 +123,117 @@ export function CardPayment({ checkoutService, injectedProvider }: Props) {
...
 
-  let errorMessage = ''
-  if (errors?.name?.message) {
-    errorMessage = errors.name.message as unknown as string
-  }
+    onSubmit()
+
+    const { error, setupIntent } = await stripe.confirmSetup({
+      elements,
+      redirect: 'if_required',
```
julien51, line 202: Did you solve the redirection issue you were asking for?

julien51 (approved):
A few small comments

unlock-app/src/components/interface/checkout/alpha/Checkout/CardPayment.tsx
searchableguy: So re-direct isn't mandatory for card & google pay/apple pay. This won't end up redirecting unless we enable more payment option.

I tried a hack to solve redirect by storing the state in local storage and adding a flag to reload from it but I think it's unnecessary for now so instead of hack, we will come back when we store information about the flow in the routes properly to load them up without local storage.

locksmith/src/controllers/v2/purchaseController.ts
searchableguy: hm, maybe. Do we have any case where we do want the normalized address?

PR #9613, opened 2022-09-15
feat(unlock-app) - add support for totalSupply

kalidiagne:
# Description
Add support for totalSupply

# Issues

Fixes #
Refs #

# Checklist:


searchableguy (approved):

packages/unlock-js/src/web3Service.ts
```diff
@@ -509,4 +509,18 @@ export default class Web3Service extends UnlockService {
...
     return id.toNumber()
   }
+
+  /**
+   * Returns the number of keys already sold
+   * @param lockAddress
+   * @param network
+   * @returns
+   */
+  async totalSupply(lockAddress: string, network: number) {
```
julien51, line 519: I think we have that thru `getLock` already... but maybe it is worth adding independently?

kalidiagne: Yes, I see we have it also inside `getLock`, there is some place we need only this is information, we can avoid to use `getLock` in those scenarios

PR #9632, opened 2022-09-17
refactor(unlock-app): Move away from nextjs runtime behavior for loading configuration

searchableguy:
# Description

This PR removes the nextjs runtime behavior for loading configuration. It is not recommended by nextjs anymore since they introduced NEXT_PUBLIC.

This is required to change the router in later PRs. 

# Issues

Fixes #
Refs #

# Checklist:


julien51 (commented):
Nice cleanup!
That said, we have been bitten many times in the past ny missing environment variables. 
Can we have keep the part where the build fails when it is missing some variables please?

unlock-app/next.config.js
```diff
@@ -1,90 +1 @@
...
-  if (!requiredConfigVariables[configVariableName]) {
-    if (
-      ['dev', 'dev-kovan', 'test'].indexOf(requiredConfigVariables.unlockEnv) >
-      -1
-    ) {
-      return console.error(
-        `The configuration variable ${configVariableName} is falsy.`
-      )
-    }
-    throw new Error(
```
julien51, line 75: That part is important to make sure we do not build/start the app in production when required env vars are missing.

searchableguy: Yeah, so nextjs doesn't have a straight forward way to do this that they recommend. 

https://nextjs.org/docs/basic-features/environment-variables

I will add a script check before build because I'm gonna tweak/remove some parts in next PRs.

.github/workflows/_netlify.yml
```diff
@@ -29,11 +29,11 @@ on:
       NEXT_PUBLIC_UNLOCK_ENV:
         required: false
       # unlock-app only
-      BASE64_WEDLOCKS_PUBLIC_KEY:
+      NEXT_PUBLIC_BASE64_WEDLOCKS_PUBLIC_KEY:
         required: false
```
julien51, line 33: What if we make required to true here?

julien51 (approved):

.github/workflows/_netlify.yml
searchableguy: wouldn't work because we reuse this action for multiple apps. 

So we can't make any of the variable required unless it's needed for all.

searchableguy:
@clemsos 🤔 CI stuck here. Any reason?

clemsos:
Hmm not sure why. Let me check

.github/workflows/production.yml
```diff
@@ -29,14 +29,14 @@ jobs:
       bypass_diff_check: bypass
       service: locksmith-websub
       build-dir: locksmith
-      command: "yarn websub:prod"
+      command: 'yarn websub:prod'
```
clemsos, line 32:
```suggestion
      command: "yarn websub:prod"
```

.github/workflows/production.yml
```diff
@@ -1,4 +1,4 @@
-name: "Production Branch"
+name: 'Production Branch'
```
clemsos, line 1:
```suggestion
name: "Production Branch"
```

PR #9642, opened 2022-09-19
feat(unlock-js, locksmith): updated UnlockJS to support transaction options

julien51:
# Description

I am adding a new critical feature to the unlock-js library: the ability to pass transaction options.

This will be used to run "simulations" that can then help us do 2 things:
* account for real gas costs on renewals or credit card purchases (rather than hardcoded value)
* not showing the "claim" button when the key could not be airdropped because the lock uses a hook

# Checklist:


packages/unlock-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@unlock-protocol/unlock-js",
-  "version": "0.26.8",
+  "version": "0.27.0",
```
julien51, line 3: Since it changes the API, I am bumping minor

julien51:
@clemsos what's the way to run all tests again?

clemsos:
> /run-all-tests

so this works but the results aren't shown here :/ 

you have to dig into the "actions" tab to find it > https://github.com/unlock-protocol/unlock/actions/runs/3086054579

there is an open issue for this already but no fix in mind so far #9306
should be a way with existing gh actions

julien51:
HA! Thank you for this. I am _very_ surprised that tests are passing here because I changed the API for `walletService` on many calls and I would expect them to fail as a consequence :(

searchableguy:
The code itself looks good (It does change the grantKeyExtension and grantKeys in dispatcher we depend on). I tested bunch of functions on the dashboard too.

searchableguy (approved):
I'm guessing the transaction object isn't used intentionally in most cases in this PR.

PR #9651, opened 2022-09-21
feat(unlock-app) - fix members export truncates export

kalidiagne:
# Description
Fix truncate export by continuing fetching the next page on graphql query if the 1000 member limit is reached on the result

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/graphql/datasource/keysByQuery.ts
```diff
@@ -153,27 +153,53 @@ export class keysByQuery extends GraphQLDataSource {
...
+          },
+        })
+      }
+      const {
+        data: { locks },
+      } = (await getData()) ?? {}
+
+      const keysList = locks[0]?.keys ?? []
+
+      let getForNextPage = keysList?.length === first
```
searchableguy, line 185: it's unfortunate that total isn't working on current subgraph otherwise we could use that for detecting whether there are more keys. Maybe add a comment here for future that we have no way to detect how many keys there are left to query on this subgraph query.

There should be cursor on the graphql side.

julien51 (approved):

PR #9674, opened 2022-09-26
feat(locksmith) - show expired and now expired key when search for keyId

kalidiagne:
# Description

This PR allow for a user when search by `keyId`, to see both active and expired memberships.
Actually, only active keys are returned 

# Issues

Fixes #
Refs #

# Checklist:


searchableguy (approved):

locksmith/src/graphql/datasource/keysByQuery.ts
```diff
@@ -8,14 +8,13 @@ export type KeyFilter = 'all' | 'active' | 'expired' | 'keyId'
...
     $addresses: [String!]
-    $expireTimestamp: BigInt! = 0
     $first: Int! = 100
     $skip: Int! = 0
     $keyId: BigInt
   ) {
     locks(where: { address_in: $addresses }) {
       keys(
-        where: { expiration_gt: $expireTimestamp, keyId: $keyId }
+        where: { expiration_gt: 0, keyId: $keyId }
```
searchableguy, line 17: though, I'm wondering why the default value isn't working here.

kalidiagne: because tha value is always passed
1: https://github.com/unlock-protocol/unlock/blob/8b18db8ed8e3ba03ba4a737eecf333b774c8eb25/locksmith/src/graphql/datasource/keysByQuery.ts#L156
2: https://github.com/unlock-protocol/unlock/blob/8b18db8ed8e3ba03ba4a737eecf333b774c8eb25/locksmith/src/graphql/datasource/keysByQuery.ts#L174

searchableguy: You can just stop passing the value too in the code, right? I would remove that part because it's confusing to pass variables when they aren't respected in the query. 

Also, Are you sure this doesn't actually break any behaviour?

kalidiagne: need to pass the value because we have different query where the value is needed, best thing here I think to remove it in the specific query

kalidiagne: let meet update it in a better way

kalidiagne:
Searching by `keyId` will search for all the keys, this only returns a single value, so we also hide the extra `expiration` filter to avoid adding more complexity in locksmith

Screenshot 2022-10-11 at 15 19 14

Screenshot 2022-10-11 at 15 19 28

searchableguy (approved):

PR #9688, opened 2022-09-27
feat(unlock-app): Add redesigned airdrop drawer for members page

searchableguy:
# Description

Work in progress.

CleanShot 2022-09-27 at 22 32 34@2x

CleanShot 2022-09-27 at 22 32 50@2x

CleanShot 2022-09-27 at 22 38 48@2x

# Issues

Fixes #
Refs #

# Checklist:


packages/contracts/package.json
```diff
@@ -19,11 +19,11 @@
     "publish:npm": "yarn prepublish && npm publish"
   },
   "devDependencies": {
+    "@nomiclabs/hardhat-etherscan": "^3.1.0",
```
julien51, line 22:
```suggestion
    "@nomiclabs/hardhat-etherscan": "3.1.0",
```

I think renovate will complain about the `^`

unlock-app/package.json
```diff
@@ -64,6 +65,7 @@
     "react": "17.0.2",
     "react-crypto-icons": "1.0.5",
     "react-dom": "17.0.2",
+    "react-dropzone": "^14.2.2",
```
julien51, line 68: And here

unlock-app/package.json
```diff
@@ -73,6 +75,7 @@
     "react-jwt": "1.1.6",
     "react-query": "3.39.2",
     "react-test-renderer": "17.0.2",
+    "react-use": "^17.4.0",
```
julien51, line 78: And here!

unlock-app/public/templates/airdrop.csv
```diff
@@ -0,0 +1,2 @@
+recipient,count,email,expiration,expire,manager
+julien51.eth,1,julien@unlock-protocol.com,0,true,julien51.eth
```
julien51, line 2: hehe :D

julien51: I assume `0` for expiration means `never` ? We should probably document in the guide on how to airdrop memberships.

unlock-app/src/components/interface/locks/Manage/index.tsx
```diff
@@ -7,27 +7,29 @@ import { LockDetailCard } from './elements/LockDetailCard'
...
 
   return (
     <>
-      <GrantKeysDrawer
+      <AirdropKeysDrawer
         isOpen={isOpen}
         setIsOpen={setIsOpen}
-        lockAddresses={[lockAddress]}
+        lockAddress={lockAddress}
+        network={network!}
```
julien51, line 32: The `network` should be the network for which `lockAddress` makes sense not (necessarily) the one the user is connected to. We should only prompt the user to switch network when they click the button _and_ they are on the wrong network.

unlock-app/src/components/interface/members/airdrop/AirdropBulkForm.tsx
```diff
@@ -0,0 +1,139 @@
...
+              clear()
+            }}
+          >
+            Confirm aidrop
+          </Button>
+        </div>
+      ) : (
+        <div className="space-y-6">
+          <p>
+            Once you upload the csv, you can see all the members. Lock&apos;s
```
julien51, line 123:
```suggestion
            Once you upload the csv, you can see all the list of memberships to be granted. Lock&apos;s
```

julien51 (approved):
This is looking very GOOD!

unlock-app/src/components/interface/locks/Manage/index.tsx
searchableguy: Yeah, I know. this is actually to not break existing code. 

The current keydrop drawer assumed network of existing user so it doesn't check for network in the URL in two places so that would need to be changed so I have a placeholder network here for now.

unlock-app/public/templates/airdrop.csv
searchableguy: 0 would be default. I think that's the convention we use in extend and other APIs on unlockjs. 

for never expiring, there is `expire` which is more explicit since we cannot assume lack of expiration as user not meaning to expire something.

julien51: Hum, ok if no `expire` let's also use the lock's duration (as for 0).

searchableguy: if no expire and expiration provided, we use the default yes.

julien51 (approved):

julien51 (commented):
:shipit:

kalidiagne (approved):
as discussed LGTM!

PR #9695, opened 2022-09-28
fix(packages): add @unlock-protocol/hardhat-plugin to packages build

clemsos:
# Description

Add missing ref to hardhat-plugin in clean/build workflow for packages 

# Issues

Fixes #
Refs #

# Checklist:


clemsos:
btw is that normal that we are checking the `dist`folder in unlock-assets ?
 https://github.com/unlock-protocol/unlock/tree/master/packages/unlock-assets/dist

searchableguy:
@clemsos It's been there for a while but we should remove these generated files (same for build folder in the same package).

packages/contracts/package.json
```diff
@@ -19,11 +19,11 @@
     "publish:npm": "yarn prepublish && npm publish"
   },
   "devDependencies": {
+    "@nomiclabs/hardhat-etherscan": "^3.1.0",
     "@primitivefi/hardhat-dodoc": "0.2.3",
     "@unlock-protocol/hardhat-helpers": "workspace:^",
     "@unlock-protocol/tsconfig": "workspace:./packages/tsconfig",
     "@unlock-protocol/types": "workspace:^",
-    "@nomiclabs/hardhat-etherscan": "^3.1.0",
```
searchableguy, line 26: Been wondering why this dependency keep changing place in the hierarchy 🤔

searchableguy (approved):

packages/contracts/package.json
clemsos: yarn makes deps follow alphabetical order - I added it manually. Which makes me think I need to remove that caret in the version.

PR #9702, opened 2022-09-28
feat(unlock-app) - pass lock manager down to components

kalidiagne:
# Description
Pass lock manager status to components 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/locks/Manage/elements/Members.tsx
+++ b/unlock-app/src/components/interface/locks/Manage/elements/Members.tsx
@@ -15,6 +15,7 @@ import { useState } from 'react'
 interface MembersProps {
   lockAddress: string
   network: number
+  isLockManager: boolean
   filters?: {
     [key: string]: any
   }
@@ -48,6 +49,7 @@ const MembersPlaceholder = () => {
 export const Members = ({
   lockAddress,
   network,
+  isLockManager,
   filters = {
     query: '',
     filterKey: 'owner',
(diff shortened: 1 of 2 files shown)


searchableguy:
Do you think a hook for this would be better?

React query can be used to cache the result for a long time so we can have that instead of prop drilling here because we will need this status in multiple places aside from members page.

```ts
// isManager could check if currently authenticated user is manager and manager could give us the manager object for that lock if we need it.


const { isMananger, manager } = useLockManager({ lockAddress: "", network: 5 })
```

WDYT?

kalidiagne:
> Do you think a hook for this would be better?
> 
> React query can be used to cache the result for a long time so we can have that instead of prop drilling here because we will need this status in multiple places aside from members page.
> 
> ```ts
> // isManager could check if currently authenticated user is manager and manager could give us the manager object for that lock if we need it.
> 
> 
> const { isMananger, manager } = useLockManager({ lockAddress: "", network: 5 })
> ```
> 
> WDYT?

Great! love this approach!

kalidiagne:
@searchableguy updated with the hook, wdyt?

searchableguy:
is there a method to get the manager address? 

if yes, can you use that. Because that way, we can also pass the actual manage and as well as isManager.

kalidiagne:
> is there a method to get the manager address?
> 
> if yes, can you use that. Because that way, we can also pass the actual manage and as well as isManager.

I think we don't have it, we can have multiple manager for a lock and I don't see any function that's give us the list of all the managers or I am missing something?

julien51:
> I think we don't have it, we can have multiple manager for a lock and I don't see any function that's give us the list of all the managers or I am missing something?

You are correct. We don't have such a function to save on gas.
We could use the new subgraph for this eventually but I think for now your approach is correct

PR #9712, opened 2022-09-29
feat(unlock-app): Reduce the amount of text on the airdrop

searchableguy:
# Description

CleanShot 2022-09-29 at 21 29 17@2x

CleanShot 2022-09-29 at 21 32 03@2x

Reduced the text and moved it from top to have more space in both to see members. 

I think the intent of the page is clear without explicit description but I can add text too if required. 

The email part is mentioned on the input already. 

# Checklist:


julien51:
> I think the intent of the page is clear without explicit description but I can add text too if required.


Agreed!

julien51 (approved):

PR #9724, opened 2022-09-30
feat(subgraph): add total keys property

clemsos:
# Description

Keep track `totalKeys` in the `Lock` object of the subgraph by watching the minting and burning of keys

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/docker/development/subgraph/startup.sh
+++ b/docker/development/subgraph/startup.sh
@@ -9,7 +9,7 @@ cat /home/unlock/subgraph/networks.json
 yarn workspace @unlock-protocol/subgraph codegen
 
 # build the subgraph files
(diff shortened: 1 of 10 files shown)


julien51 (approved):
This is good but we need to add the handling of a "missing" event :) Can we actually add the event in the subgraph now (even if of course it will not be triggered until we actually implement it)?

clemsos:
> This is good but we need to add the handling of a "missing" event :) Can we actually add the event in the subgraph now (even if of course it will not be triggered until we actually implement it)?

Not sure we want to add things that does not exist yet. We can track missing events in issues and add it when they are implemented so the code stay lean. 

Also they are no missing events here for `totalKeys`. We just use mint/transfer and burn to keep track of the overall number of keys.

PR #9727, opened 2022-09-30
feat(unlock-app) - switch network when does not match the connected wallet

kalidiagne:
# Description
fix: 
`
 If the user is connected on the wrong network when on the manage lock page, and they click on any action that requires to sign a transaction, let's first prompt the user to connect to the right network.
`

# Issues

Fixes #
Refs #

# Checklist:


unlock-app/src/hooks/useNetwork.ts
```diff
@@ -0,0 +1,32 @@
...
+      })
+      await callback()
+    } else {
+      await callback()
+    }
+  }
+
+  return {
+    switchToNetworkBeforeAction,
+  }
```
searchableguy, line 31: As discussed, let's not add hook for this and re-use existing hook since it's a minor change in function.

julien51 (approved):

unlock-app/src/hooks/useProvider.ts
```diff
@@ -167,32 +168,44 @@ export const useProvider = (config: any) => {
...
             },
           ],
           account
         )
         await ToastHelper.promise(changeNetworkRequest, {
-          loading: `Changing network to ${network.name}. Please Approve on your wallet.`,
-          error: `Error in changing network to ${network.name}`,
-          success: `Successfully changed network to ${network.name}`,
+          loading: `Changing network to ${name}. Please Approve on your wallet.`,
+          error: `Error in changing network to ${name}`,
```
julien51, line 200: For another PR, I think we should _not_ show this error when the errorCode is `4902` because right now we shouw the error and immediately prompt the user to add the network.

kalidiagne:
Close for this temp solution https://github.com/unlock-protocol/unlock/pull/9812

PR #9728, opened 2022-09-30
feat(unlock-app): Handle wallet icon more inline with the injected provider

searchableguy:
# Description

CleanShot 2022-10-01 at 00 47 55@2x

A slight difference from the original: We use wallet connect on both desktop and mobile when none provided since many desktop users have mobile only wallets like rainbow which they may want to use. 

So I think it makes sense to use wallet connect on both mobile and desktop when no injected provider is there.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useAuthenticate.ts
+++ b/unlock-app/src/hooks/useAuthenticate.ts
@@ -6,7 +6,6 @@ import { useAppStorage } from './useAppStorage'
 
 export interface EthereumWindow extends Window {
   ethereum?: any
-  web3?: any
 }
 
 interface RpcType {
@@ -33,11 +32,6 @@ export const selectProvider = (config: any) => {
     provider = `http://${config.httpProvider}:8545`
   } else if (ethereumWindow && ethereumWindow.ethereum) {
     provider = ethereumWindow.ethereum
-  } else if (ethereumWindow.web3) {
-    // Legacy web3 wallet/browser (should we keep supporting?)
-    provider = ethereumWindow.web3.currentProvider
-  } else {
-    // TODO: Let's let the user pick one up from the UI (including the unlock provider!)
   }
   return provider
 }
--- a/unlock-app/src/components/interface/LoginPrompt.tsx
+++ b/unlock-app/src/components/interface/LoginPrompt.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react'
 import styled from 'styled-components'
 import SvgComponents from './svg'
-
+import { RiWalletFill as WalletIcon } from 'react-icons/ri'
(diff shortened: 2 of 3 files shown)


unlock-app/src/hooks/useAuthenticate.ts
```diff
@@ -33,11 +32,6 @@ export const selectProvider = (config: any) => {
     provider = `http://${config.httpProvider}:8545`
   } else if (ethereumWindow && ethereumWindow.ethereum) {
     provider = ethereumWindow.ethereum
-  } else if (ethereumWindow.web3) {
-    // Legacy web3 wallet/browser (should we keep supporting?)
```
searchableguy, line 37: Removed since that was deprecated in 2020.

julien51:
> We use wallet connect on both desktop and mobile when none provided since many desktop users have mobile only wallets like rainbow which they may want to use.

But in that case they an just use the wallet connect icon???

Also you are not handling the specific icon per wallet : 

> a. If there is an injected wallet, try to find its flavor and show a relevant icon (Brave as isBraveWallet and MM has isMetaMask for example).
i. If an wallet can be "guessed", show the correct icon...
ii. if not, show the MetaMask one

searchableguy:
> Also you are not handling the specific icon per wallet :

Yeah, will add. Trying to find icons for the popular ones since they don't provide under `window.ethereum`.


> But in that case they can just use the wallet connect icon???

It's a generic fallback same as mobile for people who don't know wallet connect. Wallet connect does tell you to install wallet yourself but I will add a popup here with "Download metamask" as an option.

searchableguy:
CleanShot 2022-10-01 at 02 58 51@2x


What other options do I add?

julien51:
> b. if no injected wallet exists, show a greyed "generic" wallet icon that, when clicked opens a simple modal that has a simple text like "Wallets give access to your funds and Ethereum applications. Only you should have access to your wallet." and then links to https://ethereum.org/en/wallets/find-wallet/

> Also note: UA detection is not the most accurate and other ways to detect mobile may not work either (tablets pretend to be desktop for example).
Yeah, I know... but I think it's ok. Worst case, we show the link to download wallets which should maybe give a hint to the user who _knows_ they have a wallet that they need to use it!

> For wallet detection, I'm adding ones which are actually detectable (brave is).
Let's try with [0xFrame](https://frame.sh/), [Status](https://status.im/), Argent.

searchableguy:
CleanShot 2022-10-04 at 22 54 02@2x

searchableguy:
For argent, I don't think they actually have isArgent in injected provider. They mostly use wallet connect as main connector.

julien51 (approved):
LGTM!

PR #9730, opened 2022-09-30
feat(checkout): Add a way for users to attach data from their backend for contract purchase

searchableguy:
# Description

Adds databuilder argument which we will call to get data for use in-purchase hook. 

# Issues

Fixes #
Refs #

# Checklist:


julien51:
> For CORS, I think we can expect users to add proper origin headers since they are building the endpoint itself.

I am actually not so sure about that. It might not be easy to provide these depending on _where_ they create the script to generate that `data` field. I think we need to offer something that is _as generic_ as possible.

unlock-app/src/components/interface/checkout/alpha/Checkout/Metadata.tsx
```diff
@@ -142,10 +143,24 @@ export function Metadata({ checkoutService, injectedProvider }: Props) {
         })
       )
 
-      await storage.submitMetadata(users, lock!.network)
-
       const recipients = users.map((item) => item.userAddress)
 
+      const generatedData = paywallConfig.dataURL
+        ? await fetchData(paywallConfig.dataURL, {
```
julien51, line 149: This has to be a per lock specific setting please (not global)

unlock-app/src/components/interface/checkout/alpha/Checkout/Metadata.tsx
```diff
@@ -142,10 +143,24 @@ export function Metadata({ checkoutService, injectedProvider }: Props) {
...
+            recipients,
```
julien51, line 150: And let's make one call per recipient too, because the hook is per recipient

julien51 (changes requested):
Let's make the setting a lock-level setting (not global)

unlock-app/src/components/interface/checkout/alpha/Checkout/Metadata.tsx
searchableguy: We will need to multiple requests then for multiple recipients though.

julien51: > We will need to make multiple requests then for multiple recipients though.
Yes that's the correct approach because that's how the hook will work

unlock-app/src/components/interface/checkout/alpha/Checkout/Confirm.tsx
```diff
@@ -194,12 +195,28 @@ export function Confirm({
...
         ? new Array(recipients!.length).fill(paywallConfig.referrer)
         : undefined
+
+      let data = password || captcha || undefined
+
+      const dataBuilder =
+        paywallConfig.locks[lock!.address].dataBuilder ||
+        paywallConfig.dataBuilder
+
+      // if Data builder url is present, prioritize that above rest.
```
julien51, line 205: +1

unlock-app/src/components/interface/checkout/alpha/Checkout/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+      endpoint.searchParams.append('lockAddress', lockAddress)
+      endpoint.searchParams.append('recipient', recipient)
+
+      const abortController = new AbortController()
+
+      const timer = setTimeout(() => {
+        abortController.abort()
+      }, 5000)
+
+      const response = await fetch(endpoint, {
```
julien51, line 26: Is this going to work in a context where the 3rd party site is on _any_ domain? and maybe even HTTP (no s)?

julien51 (approved):
LGTM!

unlock-app/src/components/interface/checkout/alpha/Checkout/utils.ts
searchableguy: proxied using our RPC endpoint so it should work in almost all cases.

PR #9797, opened 2022-10-03
fix(smart-contracts): replace use of `balanceOf` by `totalKeys` internally

### LINKED ISSUE #9593: Onwership list record issue
We just identified 2 bugs in our current public lock contract (non critical but really annoying).

* `_createOwnershipRecord` uses `balanceOf` instead of `totalKeys` which means it might overwrite other owned keys in that list. Let's make sure we check that we have no other similar issues.
* `_cancelKey` should _not_ delete (or change) ownership records. It should only update the time (expire it by setting the timestamp value to 0)

clemsos:
# Description

As described in #9593, there is an issue with the ownership record list. We have separated `balanceOf` (valid keys) from `totalKeys` (all keys) on v11, but still uses `balanceOf` internally which leads to a broken ownership record when we cancel keys on v11

# Issues

Fixes #9593 
Refs #

# Checklist:


julien51:
Can you triple check that we ever use _balanceOf_ internally?

julien51 (approved):

PR #9799, opened 2022-10-03
fix(smart-contracts): pass `tokenId` as param to the `onPurchase` hook

### LINKED ISSUE #9588: Include tokenId in onKeyPurchaseHook
We should include the `tokenId` when calling `onKeyPurchaseHook`.

clemsos:
# Description

Pass the created `tokenId` of the key as param to the `onPurchase` hook

# Issues

Fixes #9588
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/interfaces/hooks/ILockKeyPurchaseHook.sol
+++ b/smart-contracts/contracts/interfaces/hooks/ILockKeyPurchaseHook.sol
@@ -31,6 +31,7 @@ interface ILockKeyPurchaseHook
   /**
    * @notice If the lock owner has registered an implementer then this hook
    * is called with every key sold.
+   * @param the tokenId of the purchased key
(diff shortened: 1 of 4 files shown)


julien51 (commented):
I am trying to think about what happens for existing hooks... and if there is a way to mitigate breakage

clemsos:
> I am trying to think about what happens for existing hooks... and if there is a way to mitigate breakage

yes thats a breaking change from previous versions. We could do a try/catch with the earlier signature (without `tokenId`) to ensure backward compat

julien51 (approved):

clemsos:
We also need to release the new hooks abi in the contracts package

clemsos:
maybe we can version the hooks somehow, so we get a version check when upgrading/trying to set an old one?

PR #9802, opened 2022-10-03
feat(unlock-app) - update links for new dashboard

kalidiagne:
# Description
- Update links to points to the new dashboard 

**Let's keep it as draft, this is one of the last PR we need to merge**

# Issues

Fixes #
Refs #

# Checklist:


julien51 (approved):
LGTM!
Let's merge when we're ready to ship!

unlock-app/src/components/content/DashboardContent.jsx
```diff
@@ -1,139 +1,16 @@
...
-          <Phone>
-            <Warning>
-              The Dashboard is currently not optimized for a mobile experience.
-              To create locks, please use a desktop computer.
-            </Warning>
-          </Phone>
+  useEffect(() => {
+    // force redirect to the new dashboard
+    router.push('/locks')
+  }, [router])
```
searchableguy, line 11: Is there any need for re-direct? let's remove it and make locks the index path as well?

kalidiagne: for any eventual external link

searchableguy: I think you misunderstood me here. 

I was saying we should simply make /locks available at /dashboard and remove the dashboard content.

kalidiagne: so keep `/dashboard` instead of `/locks`? 

i think the `/locks` is better here

searchableguy: You can keep both but since the locks page was dashboard, it makes sense to put it there instead of a redirect which adds friction to user experience. 

CC @julien51

julien51: I like the idea of using a more explicit `/locks` vs `/dashboard`... and I think it's better to have a single path.
That said I want to understand more about the friction? I would assume redirects are pretty painless, but maybe I am missing something?

searchableguy: redirects add delay to user experience. If we can do without them, /locks or /dashboard either would be fine.

Since `/dashboard` was in use before and have linked elsewhere, I proposed replacing it.

kalidiagne: removed the redirect and replaced the content inside `DashboardContent` page

julien51 (approved):

PR #9826, opened 2022-10-04
chore(*): removing support for Rinkeby!

julien51:
It is time to remove support for Rinkeby now that the network has been deprecated...
This PR removes it in many places.


julien51:
Also removed from the docs: https://github.com/unlock-protocol/docs/pull/173

kalidiagne:
> It is time to remove support for Rinkeby now that the network has been deprecated... This PR removes it in many places.

I see we have a lot of `rinkeby` ref inside `docker` folder, we need to remove those as well? 
for example `LOCKSMITH_UNLOCK_CONTRACT_ADDRESS`

searchableguy (approved):

julien51:
> I see we have a lot of rinkeby ref inside docker folder, we need to remove those as well?

Let me look

searchableguy:
Those variables mentioning rinkeby are not used anywhere as far I know. 

So good to clean up.

searchableguy:
not the files :p 

The rinkeby.locksmith and other related variables.

PR #9854, opened 2022-10-06
feat(provider): Add proxy path for proxifying request

searchableguy:
# Description

Adds proxy path for use in the checkout. 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/provider/src/index.ts
+++ b/provider/src/index.ts
@@ -45,25 +45,40 @@ export default {
 
     const url = new URL(request.url)
     const { pathname } = url
-    const proxyURL = url.searchParams.get('url')
+    const dataURL = url.searchParams.get('url')
 
-    if (pathname === '/proxy' && proxyURL) {
-      const endpoint = new URL(proxyURL)
+    if (pathname === '/data' && dataURL) {
+      const endpoint = new URL(dataURL)
       // Proxy the request
       const response = await fetch(endpoint.toString(), {
         method: 'GET',
         body: request.body,
         headers: new Headers({
           Accept: '*/*',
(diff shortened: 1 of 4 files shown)


provider/src/index.ts
```diff
@@ -36,56 +36,94 @@ export default {
...
     }
 
+    const url = new URL(request.url)
+    const { pathname } = url
+    const proxyURL = url.searchParams.get('url')
+
+    if (pathname === '/proxy' && proxyURL) {
+      const endpoint = new URL(proxyURL)
+      // Proxy the request
+      const response = await fetch(endpoint.toString(), {
```
julien51, line 53: I am a bit worried this might be abused because right now it allows for _any_ kind of URL.

provider/src/index.ts
```diff
@@ -36,56 +36,94 @@ export default {
...
+        method: 'GET',
+        body: request.body,
+        headers: new Headers({
+          Accept: '*/*',
+          Origin: 'https://rpc.unlock-protocol.com/',
+          'Content-type': request.headers.get('content-type')!,
+        }),
+      })
+
+      const text = await response.text()
```
julien51, line 63: What if we fail if the response is not a parse-able JSON with a `data` object in it?
This way we can make sure that it does not do anything useful for people who are not using it "correctly" ?

searchableguy: Done.

julien51 (approved):
We can't limit the URL used, but as you see we can limit what we return which means it's probably useless to anyone but people who expected a json object with a `data` property :)

LGTM!

PR #9858, opened 2022-10-06
feat(unlock-js): Add subgraph service

searchableguy:
# Description

Adds fully typed subgraph service from our v2 subgraph endpoints. 

# Issues

Fixes #
Refs #

# Checklist:


searchableguy:
Should we cover both a react-apollo-client and a generic typescript client or a single one?

julien51:
> Should we cover both a react-apollo-client and a generic typescript client or a single one?

Ha! You have reach the limits of knowledge about this GraphQL space :D 

What's the most frequent thing to do?
What's the most useful thing?
If these 2 are not aligned, what are the pros/cons of picking one vs the other?

searchableguy:
For context, I've tried the original `graphclient cli` build tool for generating cross-chain sdk. You can find it in the commit. It is a wrapper around graph mesh generator which is a standard tool by the guild so it is not actually super custom. 

They provide a few plugins on top but none of them work properly with cross-chain support. They have open issues for it so I ended up replacing it with the original graphql-codegen tool which provides more flexibility and is more robust as it extends beyond crypto. 


For the API, 

```ts
 // Query two networks by ID. To query all, don't pass any custom networks.
 const locks = await service.locks(
    {
      first: 100,
      skip: 100,
    },
    {
      networks: [1, 5],
    }
  )
  ```
  
 We allow networks to be configured when initialising the class similar to other services we provide.

File: packages/unlock-js/src/subgraph/index.ts [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+
+  createSdk(networkId = 1) {
+    const network = this.networks[networkId]
+
+    const client = new GraphQLClient(network.subgraph.endpointV2!)
+    const sdk = getSdk(client)
+    return sdk
+  }
+
+  async locks(variables: AllLocksQueryVariables, options?: QueryOptions) {
```
julien51, line 28: Let's add docstrings please!

julien51 (commented):
Please add docstrings and add to changelog for unlockjs + bump version!

packages/unlock-js/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changes
 
+# 0.29.4
```
julien51, line 3: I think it can be a 0.30.0!

searchableguy, line 3:
```suggestion
# 0.30.0
```

packages/unlock-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@unlock-protocol/unlock-js",
-  "version": "0.29.3",
+  "version": "0.29.4",
```
searchableguy, line 3:
```suggestion
  "version": "0.30.0",
```

julien51 (approved):
LGTM!

PR #9870, opened 2022-10-07
feat(smart-contracts): introduce `execTransaction`

### LINKED ISSUE #9673: Support for execTransactions
It would be useful to add a function that only lock managers can call to execute a transaction on behalf of the lock.
This can let us remove a few functions like the ones to withdraw or to approve a beneficiary... etc.

More context https://twitter.com/w1nt3r_eth/status/1572633281094647809

clemsos:
# Description

This introduces an `execTransaction` pattern that can be used to execute contract calls on behalf of the lokc , e.g. ERC20 transfer and approval.

# Issues

Resolves #9673 
Refs #

# Checklist:


smart-contracts/test/Lock/execTransaction.js
```diff
@@ -0,0 +1,79 @@
...
+const { ethers } = require('hardhat')
+
+let unlock
+let lock
+
+contract('Lock / execTransaction', (accounts) => {
+  const [deployer, lockOwner, keyOwner, spender, other] = accounts
+  let token
+
+  before(async () => {
```
julien51, line 17: Can you add test to make sure the lock can "spend" its funds (both Eth and ERC20)?

julien51 (changes requested):

smart-contracts/test/Lock/execTransaction.js
julien51: cc @clemsos ^^

clemsos: yes, added tests for ERC20 spending. As we still have `withdraw`, I dont think we should make this function support ETH transfer as it will add complexity to it for duplicating what already supported

julien51: It's just a test! We should be much better at identifying regressions. 
We meed a test to make sure that "spending" eth works with `execTransaction`. Right now we don;t have any!

julien51:
Hey, I am really worried about the security implications here. For example, I think there is an issue here with Unlock.
Someone could arbitrarily call Unlock `recordKeyPurchase` function for example... and I think it may break things.

I think we need to think about this a lot more. Waht do you think?

clemsos:
Yes agree. Having the ability to execute random calls does not really seem like a good security practice, esp with calls to Unlock.

julien51:
Ok so let's close this for now. Since we kept the withdraw function, we don't need this... agreed?

PR #9880, opened 2022-10-07
feat(ui): Add new ticket UI component

searchableguy:
# Description

CleanShot 2022-10-08 at 03 15 27@2x

# Issues

Fixes #
Refs #

# Checklist:


julien51:
I just realized we're missing the lock address on Thurs design... Do you mind adding it?

kalidiagne:
Do we also need to show metadata in this component?

julien51:
> Do we also need to show metadata in this component?

That's a good idea but we probably need to ask @ChialiTsai about it because that's going to make things a bit murkier

searchableguy:
CleanShot 2022-10-10 at 23 54 28@2x


Added the changes. 

For metadata, we can simply pass `TicketItem` to items props if we need to for stuff like `email` so this component should be adequate (see how location, date, etc works).

searchableguy:
CleanShot 2022-10-11 at 00 00 48@2x

searchableguy:
CleanShot 2022-10-11 at 00 00 48@2x
 
 @julien51

ChialiTsai:
image


@searchableguy could we move the lock address and network info at bottom instead?
the top section was meant to keep most essential info for the participant when they go to the event.

searchableguy:
CleanShot 2022-10-11 at 17 38 30@2x

@ChialiTsai @julien51

packages/ui/lib/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { ethers } from 'ethers'
+
+export const minifyAddress = (address: string) => {
+  const checked = ethers.utils.isAddress(address)
+  return checked
+    ? `${address.slice(0, 5)}...${address.slice(address.length - 5)}`
```
kalidiagne, line 6:
```suggestion
  return checked && address.length > 5
    ? `${address.slice(0, 5)}...${address.slice(address.length - 5)}`
```

julien51:
@searchableguy I think we should avoid duplicating stuff... and we need to be more flexible with the ower data. Some event will have the attendee name, email, ... etc

packages/ui/lib/components/Ticket/Ticket.tsx
```diff
@@ -0,0 +1,115 @@
...
+  recipient,
+  id,
+  QRCodeURL,
+  date,
+  time,
+  location,
+  lockAddress,
+  network,
+}: Props) {
+  const networkConfig = networks[network]
```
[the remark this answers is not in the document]
kalidiagne: maybe we can show an error message if the network passed is not correct

packages/ui/lib/utils.ts
searchableguy: No, this is incorrect because this is only for letting ENS name pass through as is instead of converting to address. This will stop any ENS less than 5 in length to be passed.

searchableguy:
okay, changed the design.

CleanShot 2022-10-11 at 18 41 30@2x

CleanShot 2022-10-11 at 18 42 06@2x

julien51 (approved):
LGTM

PR #9883, opened 2022-10-10
refactor(unlock-js): Rename export explicitly to avoid confusion

searchableguy:
# Description

Renames the exports explicitly to avoid confusion.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/packages/unlock-js/package.json
+++ b/packages/unlock-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@unlock-protocol/unlock-js",
-  "version": "0.28.2",
+  "version": "0.29.2",
   "description": "This module provides libraries to include Unlock APIs inside a Javascript application.",
   "main": "dist/index.js",
   "module": "dist/index.mjs",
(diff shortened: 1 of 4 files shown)


packages/unlock-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@unlock-protocol/unlock-js",
-  "version": "0.28.2",
```
julien51, line 3: Since this does not change the API I would only do a patch

packages/unlock-js/package.json
```diff
@@ -1,6 +1,6 @@
...
+  "version": "0.29.2",
```
julien51, line 3:
```suggestion
  "version": "0.28.3",
```

julien51 (approved):

packages/unlock-js/package.json
searchableguy: It does because it changes the export name.

PR #9886, opened 2022-10-10
feat(contracts): Merge metadata setters into a single `setLockMetadata`

clemsos:
# Description

Having a single `setLockMetadata` allows to emit an event we can track on the graph for all updates. 

Having a single function is part of an attempt to make the contract code leaner and reduce its size BUT it seems to actually increase its size. We are getting `Error: Transaction reverted: trying to deploy a contract whose code is too large`

```
 |  PublicLock               ·     24.616  ·       +0.205  │
```

# Issues

Fixes #
Refs #

# Checklist:


julien51 (commented):
Let's not use `setLockMetadata` because I worry this will be really confusing to people ...
What about  setLockAttributes` ?

clemsos:
> Also what? Why does it end up bigger????

it seems like adding an event makes it bigger...

PR #9887, opened 2022-10-10
fix(smart-contracts): remove events when setting roles

clemsos:
# Description

Here we remove the custom Lock events that are fired when setting roles (mostly to save on contract size). We will rely on the [`RoleGranted` event](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3dac7bbed7b4c0dbf504180c33e8ed8e350b93eb/contracts/access/AccessControl.sol#L230) fired by OpenZeppelin AccessControl contract

Influence on contract size:

```
 |  PublicLock               ·     24.201  ·       -0.210  │
```

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinRoles.sol
+++ b/smart-contracts/contracts/mixins/MixinRoles.sol
@@ -13,12 +13,6 @@ contract MixinRoles is AccessControlUpgradeable, MixinErrors {
   bytes32 public constant LOCK_MANAGER_ROLE = keccak256("LOCK_MANAGER");
   bytes32 public constant KEY_GRANTER_ROLE = keccak256("KEY_GRANTER");
 
-  // events
-  event LockManagerAdded(address indexed account);
-  event LockManagerRemoved(address indexed account);
-  event KeyGranterAdded(address indexed account);
-  event KeyGranterRemoved(address indexed account);
-
   // initializer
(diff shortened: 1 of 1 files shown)


clemsos:
Note that I also tried to remove strings and replace by literal constants but that does NOT affect the contract size - the compiler does this already I guess. So keeping the strings for more readability, and adding this here for reference:

```
  bytes32 public constant LOCK_MANAGER_ROLE_ENCODED = 0xb89cdd26cddd51301940bf2715f765b626b8a5a9e2681ac62dc83cc2db2530c0;
  bytes32 public constant KEY_GRANTER_ROLE_ENCODED = 0xb309c40027c81d382c3b58d8de24207a34b27e1db369b1434e4a11311f154b5e;
```

clemsos:
Note that the subgraph V2 will actually fail here

clemsos:
> > Note that the subgraph V2 will actually fail here
> 
> Dang. Ok let's make sure we add a fix there as part of this PR? Wdyt?

yes, update the graph logic and tests directly in that PR is the best maybe. Shouldn't be too crazy really, just parse the keccak et voila :P

julien51 (approved):

clemsos:
We dont need this optimization for that version. Closing for now

PR #9889, opened 2022-10-10
fix(smart-contracts): remove duplicate `safeTransferFrom`

clemsos:
# Description

This removes an additional `safeTransferFrom` signature that was just a shortcut to pass empty bytes data and can be easily reproduce by explicitly passing it to the other sig

Influence on contract size

```
 |  PublicLock               ·     24.325  ·       -0.086  │
```

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/mixins/MixinTransfer.sol
+++ b/smart-contracts/contracts/mixins/MixinTransfer.sol
@@ -275,24 +275,6 @@ contract MixinTransfer is
     return true;
   }
 
-  /**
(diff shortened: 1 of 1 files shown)


julien51 (changes requested):
Unfortunately I am not sure we can because this breaks compatibility with ERC721... platforms like OpenSea probably expect this function to exist and will call it :/

clemsos:
> ERC721

oh ok ~ I suspected it was the reason why it was there. Let's close this then no pb

PR #9895, opened 2022-10-10
fix(subgraph): don't index locks below v5

clemsos:
# Description

Don't index locks below v5 on our subgraph v2 (they are missing `name` and `tokenAddress`)

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/subgraph/src/unlock.ts
+++ b/subgraph/src/unlock.ts
@@ -21,6 +21,9 @@ export function handleNewLock(event: NewLock): void {
     version = BigInt.fromI32(publicLockVersion.value.toI32())
   }
 
+  // dont index locks with versions lower than 5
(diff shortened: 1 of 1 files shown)


clemsos:
the log just gives a block number (7120924) where we have an [unlock tx](https://etherscan.io/tx/0x460ac4569ffb6d8e602f9665969a0aae94e882635cc9490d9fc4657e0453aab7) where the lock address is  https://etherscan.io/address/0x522f8e92d8D3D11E67C3D1dd175ca4656c82a6F6

code isn't verify so I am not sure which version it is 

but on the docs, v4 has no `tokenAddress()` sig which was causing the failure

julien51:
Ok good. I was worried that was a *recent* event...

PR #9928, opened 2022-10-12
feat(unlock-app) - fix set default currency

kalidiagne:
# Description
Currency change is not working for the default one, to fix `ZERO` address is set by default in order to avoid fallback to the current currency. Also `getValues` is removed to use `fields` value passed directly on `handleSubmit`

# Issues

Fixes #
Refs #

# Checklist:


julien51 (approved):
Have you very thoroughly tested this?

kalidiagne:
> Have you very thoroughly tested this?

tested an working, setting default network from a custom one is now working.
going back to a custom one still working as before 

https://user-images.githubusercontent.com/20865711/195543550-3a4b3bf9-bd4e-4ff4-b2f4-8154cd55b2d9.mov

Issue #9867, opened 2022-10-06
Allow recurring "forever"


julien51:
In our current checkout config, the number of recurring purchases is "fixed":

> `recurringPayments`: The number of time a membership should be renewed automatically.

We should enable "forever". In that context, the membership will stop only if the user stops it explicitly and/or they run out of funds. 

When the membership is forever, the ERC20 approval should be MAX_UINT and the UI should say “Pay every X days until canceled”.

searchableguy:
Before enabling this option, I think we should have an option to cancel the subscription from keychain.

julien51:
Well we do! If you cancel (trashcan icon) it will cancel the subscription =)

PR #9935, opened 2022-10-13
feat(unlock-app): Support forever in recurring option

searchableguy:
# Description

fixes https://github.com/unlock-protocol/unlock/issues/9867

Introduces a totalApproval field in the purchaseKeys which sets the total approval instead of using calculated total based on price and renewal times. 

CleanShot 2022-10-13 at 23 49 04@2x
CleanShot 2022-10-13 at 23 51 14@2x

# Issues

Fixes #
Refs #

# Checklist:


kalidiagne:
need to add changelog and bump version for `unlock-js`

kalidiagne (approved):

julien51:
Let's get @ChialiTsai 's feedback I'd design here please. I am not sure this yes the right approach.

unlock-app/src/components/interface/checkout/alpha/Checkout/Select.tsx
```diff
@@ -254,37 +255,56 @@ export function Select({ checkoutService, injectedProvider }: Props) {
...
+                                      Renew{' '}
+                                      {typeof item.recurringPayments ===
+                                      'number'
+                                        ? `${item.recurringPayments} times`
+                                        : item.recurringPayments}
+                                    </Badge>
+                                  )}
+                                  {isMember && (
+                                    <Badge size="tiny" variant="primary">
+                                      Owned
```
julien51, line 272: I think it is a great idea to show to the user when they already have a membership to a lock, but I am not sure this UI is the right one :/

unlock-app/src/components/interface/checkout/alpha/Checkout/Select.tsx
```diff
@@ -254,37 +255,56 @@ export function Select({ checkoutService, injectedProvider }: Props) {
...
-                                      formattedData.isSoldOut
-                                        ? 'Sold out'
-                                        : formattedData.formattedKeysAvailable
-                                    }
-                                  />
+
+                              <div className="w-full space-y-2">
+                                <div className="flex items-center w-full grid-cols-2 gap-2">
+                                  {item.recurringPayments && (
+                                    <Badge variant="primary" size="tiny">
```
julien51, line 262: I am not sure this is the right place to show this... and even if it is I don;t think this is the right UI.

I think confirmation is a much better place.

julien51 (changes requested):
I would love @ChialiTsai 's comments on the best way to show these 2 elements:
* user already has a membership to the lock
* membership will renew.

searchableguy:
As mentioned - this is what we have in the figma design. Happy to make the changes once I get new design.

ChialiTsai:
image


May I propose the UI change to this? @searchableguy

searchableguy:
@julien51 is above good to implement?

julien51:
Yes! That's much better imo

searchableguy:
CleanShot 2022-10-18 at 00 56 15@2x

julien51 (approved):

Issue #9850, opened 2022-10-05
Tickets QR codes


julien51:
Our QR codes are fairly basic for now!
We need to implement a better "ticket" UI that does include the QR code but also a lot of information!

![image](https://user-images.githubusercontent.com/17735/194121906-44825ab4-416e-4a71-ab4e-b866bd5fe490.png)

Since the QR codes are generated both on the front-end and the backend and to avoid replicating this, we should create an API endpoint that, given the signature and the signed details will generate a PDF. This API will be used on the front-end from the keychain so it can be sent by email... as well as a from the websub worker directly.

The ticket should be in the PDF format and should include the following:
- the NFT image
- the lock name
- the token id
- the owner address

Once we have more specific metadata for events we will add them to the tickets (location, date... etc)

searchableguy:
Any reason to switch to PDF from PNG currently for QRcode?

PDF file won't be shown by default in gmail and other clients unlike PNGs. They also cannot be displayed on the frontend without pulling in dependencies so we will need to the actual UI or PNG.

julien51:
We don't expect to show the PDF on the front-end. It can be downloaded from there though...
And we will send it as an attachment by email.

PR #9941, opened 2022-10-14
feat(locksmith): Add endpoint for ticket generation

searchableguy:
# Description

Moved from using tailwind classes due to experimental support and few errors in satori but design is unchanged. 

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/controllers/v2/ticketsController.tsx
```diff
@@ -168,4 +173,55 @@ export class TicketsController {
...
+        title={lock.name}
+        id={tokenId}
+        recipient={userAddress}
+        lockAddress={lockAddress}
+        network={network}
+        // time="10:00 - 12:00"
+        // date="10/10/2022 - 10/15/2022"
+        // location="190 Bowery, Soho, New York"
```
searchableguy, line 202: what standard field names should we have for lock metadata here?

julien51:
The relates to https://github.com/unlock-protocol/unlock/issues/9850

julien51:
This is blocked because we need to use a library that requires us to use ESM.

We need to upgrade `jest` because the version we use 27 does not ESM. Unfortunately, we can't upgrade jest because of memory like, so we need to move to `ytest` instead of `jest`.

julien51:
Now that have ESM we should be able to upgrade the dependencies and get that out of the door! 🎉

searchableguy:
You should receive a ticket in your email `julien@unlock-protocol.com`. Let me know if any changes are required. 

@julien51

.yarn/patches/yoga-layout-prebuilt-npm-1.10.0-855b15449f.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/yoga-layout/build/Release/nbind.js b/yoga-layout/build/Release/nbind.js
```
julien51, line 1: Any context on this file?

julien51 (approved):
Could you write a quick README on how this all works please?

.yarn/patches/yoga-layout-prebuilt-npm-1.10.0-855b15449f.patch
searchableguy: It's a small one line patch for yoga dependency. It will be automatically applied by yarn on `yarn install`.

PR #10047, opened 2022-10-26
feat(unlock-app) - Encourage creators for next steps once lock is deployed

kalidiagne:
# Description

https://user-images.githubusercontent.com/20865711/197978221-873f3fb2-62f1-49a9-914e-278e726dee58.mov

Screenshot 2022-10-26 at 10 17 51

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/locks/Manage/elements/Members.tsx
+++ b/unlock-app/src/components/interface/locks/Manage/elements/Members.tsx
@@ -10,6 +10,7 @@ import { paginate } from '~/utils/pagination'
 import { PaginationBar } from './PaginationBar'
 import React from 'react'
 import { ExpirationStatus } from './FilterBar'
+import Link from 'next/link'
 
 interface MembersProps {
   lockAddress: string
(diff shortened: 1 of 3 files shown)


unlock-app/src/components/interface/locks/Manage/elements/Members.tsx
```diff
@@ -120,12 +123,31 @@ export const Members = ({
     )
   }
 
+  const checkoutLink = `/locks/checkout-url?lock=${lockAddress}&network=${network}`
+
   if (noItems && (hasSearch || hasActiveFilter)) {
```
julien51, line 128: I think we should only show the buttons to airdrop and/or generate purchase URL if there is no active filter (i.e the fact that there are no results is because there are no keys, not because none was "found")

julien51 (changes requested):
Please make sure this is only displayed when there are no keys _at all_, not just when none was found thru filters.

kalidiagne:
> Please make sure this is only displayed when there are no keys _at all_, not just when none was found thru filters.

**No keys scenario without active filter**

Screenshot 2022-10-26 at 18 37 56


**No keys scenario with filter**

Screenshot 2022-10-26 at 18 38 06

Screenshot 2022-10-26 at 18 38 22


Screenshot 2022-10-26 at 18 38 13

julien51 (approved):
LGTM!

PR #10048, opened 2022-10-26
feat(unlock-app) - only use subgraph to load locks

kalidiagne:
# Description

Get locks from `Subgraph`

https://user-images.githubusercontent.com/20865711/198006122-b48dae1c-8cc3-47c3-824b-4e1bb9a90e23.mov

# Issues

Fixes #
Refs #

# Checklist:


unlock-app/src/components/interface/locks/List/elements/LockCard.tsx
```diff
@@ -140,33 +174,42 @@ const LockIcon = ({ lock }: LockIconProps) => {
...
+    return await web3service.getLock(lock.address, network)
+  }
+
+  const {
+    isLoading: isLoadingInfo,
+    data: { keyPrice, balance, outstandingKeys, expirationDuration } = {},
+  } = useQuery(['lockDetail', network, lock?.address], async () =>
+    getLockDetail()
+  )
+
```
kalidiagne, line 200: we actually have `balance` and `outstandingKeys ` values from the subgraph

kalidiagne: Until the value is not present we show a placeholder 

Screenshot 2022-10-26 at 12 44 11

julien51: Ok lets actually make sure we add these to the subgraphs before merging.
@kalidiagne I think that could be a great occasion to learn about how to update the subgraph.
Maybe you can schedule a sessions with @clemsos so he can show you?

kalidiagne: Sure ill do it! i take first look at the example @clemsos already provided, and the talk also with him!

julien51: Since we dont have the `balance` in the graph and it's not easy to add, just use the `getBalance` function to get it...

kalidiagne:
`
utils . formatEther ( wei )   =>   string
Format an amount of wei into a decimal string representing the amount of ether. The output will always include at least one whole number and at least one decimal place, otherwise leading and trailing 0’s will be trimmed.
`
**Before**
Screenshot 2022-10-31 at 12 16 31

**After**
Screenshot 2022-10-31 at 12 16 25

@julien51  
can leave as is? or change it to remove decimals when `0`?

unlock-app/src/components/interface/locks/List/elements/LockCard.tsx
```diff
@@ -160,15 +204,15 @@ export const LockCard = ({ lock, network, isLoading }: LockCardProps) => {
...
   const duration =
-    lock.expirationDuration === UNLIMITED_KEYS_DURATION ? (
+    lock?.expirationDuration === MAX_UINT ? (
       'Unlimited'
     ) : (
-      <Duration seconds={lock.expirationDuration} />
+      <Duration seconds={lock?.expirationDuration} />
     )
 
+  const keyPrice = ethers.utils.formatEther(lock?.price)
```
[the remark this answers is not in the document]
julien51: That's ok!

unlock-app/src/components/interface/locks/List/elements/LockCard.tsx
```diff
@@ -140,18 +175,27 @@ const LockIcon = ({ lock }: LockIconProps) => {
...
     successDuration: 2000,
   })
-
   const { explorer, baseCurrencySymbol } = networks?.[network] ?? {}
 
   const explorerUrl = explorer?.urls?.address(lockAddress) || '#'
 
-  const symbol = (lock as any)?.currencySymbol ?? baseCurrencySymbol
+  const getLockDetail = async () => {
+    return await web3service.getLock(lock.address, parseInt(network, 10))
```
julien51, line 190: No , if you do this it's going to be exactly like what we do now! Please just get the balance... not the whole lock object.

julien51 (changes requested):
In the current state, this is not changing the behavior! You need only to retrieve the balance using a new function on unlockjs.

julien51 (changes requested):
Kali, you cannot use `web3service.getLock` otherwise this PR does nothing :) 
The goal here is to _not_ query the chain to retrieve the lock's details but just use the subgraph.
The only exception is the balance which we can retrieve from the ERC20 contract.

clemsos:
Actually there are existing methods in the web3service to query balance M. I have seen it in unlock js integration test. See the withdraw test here 

´´´js

const getBalance = async (address, chainId, tokenAddress) =>
      tokenAddress === null
        ? await web3Service.getAddressBalance(address, chainId)
        : await web3Service.getTokenBalance(
            lock.currencyContractAddress,
            address,
            chainId
          )
´´´

unlock-app/src/components/interface/locks/List/elements/LockCard.tsx
```diff
@@ -140,18 +175,55 @@ const LockIcon = ({ lock }: LockIconProps) => {
...
+    tokenAddress: string
+  ) => {
+    return await web3service.getAddressBalance(
+      address,
+      chainId,
+      tokenAddress === DEFAULT_USER_ACCOUNT_ADDRESS ? undefined : tokenAddress
+    )
+  }
+
+  const getSymbol = async () => {
```
julien51, line 202: I wonder if things could be more elegant if combined in a single function that executes both promises?

julien51 (approved):
this looks good to me and is going to be MUCH faster!

PR #10053, opened 2022-10-26
feat(unlock-app): Add support for dynamic pricing

searchableguy:
# Description

CleanShot 2022-10-26 at 20 37 34@2x

Work in progress. 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/alpha/Checkout/Confirm.tsx
+++ b/unlock-app/src/components/interface/checkout/alpha/Checkout/Confirm.tsx
@@ -71,7 +71,7 @@ export function Confirm({
     name: lockName,
     keyPrice,
   } = lock!
-
+  let purchaseData = password || captcha || undefined
   const recurringPayment = paywallConfig?.locks[lockAddress]?.recurringPayments
   const totalApproval =
     typeof recurringPayment === 'string' &&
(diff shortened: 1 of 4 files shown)


julien51:
i am not a huge fan of the UI here... I think it migth be confusing to users.

julien51:
image

What about this ^^

searchableguy:
Yeah, I would put a modal and have a small button which shows the breakdown prices for each address. @julien51 

WDYT?

julien51:
The one caveat I would show is when a recipient has a discount (price is different from the "regular") price, we add a line below that says : `regular price j TOKEN`

julien51:
> Yeah, I would put a small button which shows the breakdown prices for each address in a modal. @julien51

Not sure we need to add a modal if we show line items + totals below them?

searchableguy:
For the individual breakdown for multiple addresses and price. 

We can show "Discount is applied to some or more purchases - see breakdown"

That way, we don't have any scrolling or too much information issue but also present way to see the detailed logs on how the total is calculated if discount is applied.

julien51:
> We can show "Discount is applied to some or more purchases - see breakdown"

In theory the discounts could be different for each addresses...
I say don't worry too much about things being too long. In the vast majority of cases the user is buying a single membership so let's optimize for this first :)

unlock-app/src/components/interface/checkout/alpha/Checkout/Confirm.tsx
```diff
@@ -496,71 +501,16 @@ export function Confirm({
     }
   }
 
+  console.log(prices)
```
julien51, line 499: 🧹

unlock-app/src/components/interface/checkout/alpha/Checkout/Confirm.tsx
```diff
@@ -149,6 +144,73 @@ export function Confirm({
...
 
+  const { isInitialLoading: isInitialDataLoading, data: purchaseData } =
+    useQuery(
+      ['purchaseData', lockAddress, lockNetwork, JSON.stringify(recipients)],
+      async () => {
+        let purchaseData = password || captcha || null
+        const dataBuilder =
+          paywallConfig.locks[lock!.address].dataBuilder ||
+          paywallConfig.dataBuilder
+        // if Data builder url is present, prioritize that above rest.
```
julien51, line 155: 👍

julien51 (approved):
I like it!

PR #10068, opened 2022-10-27
feat(subgraph) - add balance to subgraph

kalidiagne:
# Description
Add subgraph to `subgraph`

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/subgraph/schema.graphql
+++ b/subgraph/schema.graphql
@@ -6,6 +6,7 @@ type Lock @entity {
   expirationDuration: BigInt
   tokenAddress: Bytes!
   price: BigInt!
+  balance: BigInt!
   lockManagers: [Bytes!]!
   version: BigInt!
   totalKeys: BigInt!
--- a/subgraph/src/unlock.ts
+++ b/subgraph/src/unlock.ts
@@ -52,6 +52,11 @@ export function handleNewLock(event: NewLock): void {
     lock.maxNumberOfKeys = maxNumberOfKeys.value
   }
 
+  let balance = lockContract.try_balanceOf(lockAddress)
+  if (!balance.reverted) {
+    lock.balance = balance.value
+  }
+
   // store info from event
   lock.address = lockAddress
   lock.version = version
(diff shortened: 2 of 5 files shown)


clemsos:
I think there is a misunderstanding there. The `balanceOf` function does not return the ETH or ERC20 balance of the lock. It will return the number of valid keys for a specific user - see [here](https://docs.unlock-protocol.com/core-protocol/smart-contracts-api/PublicLock#balanceof).

What you want here is how much ETH or ERC20 is in the lock? I will say that is going to be kind of hard to track with the graph as we may lack some of the functions that affects the balance :/ 

I think it may be better to get it directly from the RPC / ERC20 contract directly, no? This is quite the critical info people are looking at and we dont want to have inaccurate data here. 

wdyt?

kalidiagne:
> I think there is a misunderstanding there. The `balanceOf` function does not return the ETH or ERC20 balance of the lock. It will return the number of valid keys for a specific user - see [here](https://docs.unlock-protocol.com/core-protocol/smart-contracts-api/PublicLock#balanceof).
> 
> What you want here is how much ETH or ERC20 is in the lock? I will say that is going to be kind of hard to track with the graph as we may lack some of the functions that affects the balance :/
> 
> I think it may be better to get it directly from the RPC / ERC20 contract directly, no? This is quite the critical info people are looking at and we dont want to have inaccurate data here.
> 
> wdyt?

Yes, what I mean here with `balance` is the total balance that can be withdrawn for the lock.
I get your point here, for now, I think we can get it directly from the contract for when needed.

@julien51 wdyt?

julien51:
> What you want here is how much ETH or ERC20 is in the lock? I will say that is going to be kind of hard to track with the graph as we may lack some of the functions that affects the balance :/

That's a good point. We _could_ still use the PublicLock events and then retrieve the balance from the ERC20 contract but you are correct that this would miss some data (for example if an ERC20 transfer to the lock happens _without_ a `purchase`).

So indeed I think the best way to get the balance of every lock is to use `getBalance` on web3Service for each lock? This should still be MUCH faster since it's a single call (maybe 2 to get decimals...) but at least were not querying _all_ of the PublicLock stuff.

julien51:
@kalidiagne Let me know if you need more clarification about this...

kalidiagne:
not needed, balanced added to `web3service`

Issue #10045, opened 2022-10-26
Broken checkout on mobile wallets


julien51:
One of the difficult things about several mobile wallets like MetaMask and others is that they don't inject the provider in iframes (only on the main window/frame).
That is the reason why the paywall app has a "proxy" provider. that basically passes RPC calls from the main frame to the iframe.
It appears that this is not working right now.

You can try from our home page or a demo page with mobile MetaMask and/or Coinbase's wallet application.

searchableguy:
It is working for me. I use wallet connect and it connects the account properly.  Can you record a clip?

PR #10078, opened 2022-10-27
feat(unlock-app): Use injected provider for detection on checkout

searchableguy:
# Description

fixes https://github.com/unlock-protocol/unlock/issues/10045

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/alpha/Connected.tsx
+++ b/unlock-app/src/components/interface/checkout/alpha/Connected.tsx
@@ -10,7 +10,6 @@ import { ConnectService } from './Connect/connectMachine'
 import { RiWalletFill as WalletIcon } from 'react-icons/ri'
 import { SiBrave as BraveWalletIcon } from 'react-icons/si'
 import { DownloadWallet } from '../../DownloadWallet'
-import { WalletLinkProvider } from 'walletlink'
 interface SignedInProps {
   onDisconnect?: () => void
   isUnlockAccount: boolean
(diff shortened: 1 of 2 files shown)


julien51 (approved):

unlock-app/src/components/interface/checkout/alpha/Connected.tsx
```diff
@@ -63,11 +64,13 @@ interface SignedOutProps {
     provider: 'METAMASK' | 'UNLOCK' | 'WALLET_CONNECT' | 'COINBASE'
   ): Promise<void>
   onUnlockAccount(): void
+  injectedProvider: WalletLinkProvider
```
julien51, line 67: Not sure this is the correct type?

unlock-app/src/components/interface/checkout/alpha/index.tsx
```diff
@@ -25,7 +25,7 @@ export function CheckoutPage() {
   const paywallConfig = communication.paywallConfig || paywallConfigFromQuery
 
   const injectedProvider =
-    communication.providerAdapter || selectProvider(config)
+    selectProvider(config) || communication.providerAdapter
```
julien51, line 28: Actually the order here is important! Let's reverse that.
This is because we should let the main frame "force" a wallet :)

julien51 (changes requested):
Please swap providers

unlock-app/src/components/interface/checkout/alpha/Connected.tsx
searchableguy: changed to any because I don't see any type which we use for this. We `any` in the window object.

julien51 (approved):

PR #10085, opened 2022-10-27
feature(smart-contracts): adding a check to see if the key exists before yielding tokenUri

julien51:
# Description

This seems to be causing issues for ThirdWeb... and is probably the right thing to do!

# Checklist:


clemsos:
It seems we had this check before but it was removed in v 9 or 10. Maybe to same space ?

julien51:
You are correct. Not worth doing.

PR #10089, opened 2022-10-28
feat(locksmith): Pass down default lock metadata to generated key metadata properly

searchableguy:
# Description

Currently, attributes aren't passed down from the lock default metadata since we override fields. This PR fixes that and make some optimisation to reduce latency from tens of seconds to a few seconds using subgraph. 

Bumped two version since another PR of mine also bumps the minor. 

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/operations/metadataOperations.ts
```diff
@@ -56,17 +54,33 @@ export const generateKeyMetadata = async (
...
+    attributes.push(...onChainKeyMetadata.attributes)
+  }
+
+  if (Array.isArray(baseTokenData?.attributes)) {
+    attributes.push(...baseTokenData.attributes)
+  }
+
+  if (Array.isArray(keyCentricData?.attributes)) {
+    attributes.push(...keyCentricData.attributes)
+  }
```
searchableguy, line 70: Let me know if do not want to merge the attributes.

locksmith/src/operations/metadataOperations.ts
```diff
@@ -125,34 +127,23 @@ const fetchChainData = async (
...
-  // Custom mappings
-  // TODO: move that to a datastore at some point...
-  Metadata.forEach((lockMetadata) => {
-    if (address.toLowerCase() == lockMetadata.address.toLowerCase()) {
-      defaultResponse.name = lockMetadata.name
-      defaultResponse.description = lockMetadata.description
-      defaultResponse.image = lockMetadata.image || defaultResponse.image
-    }
-  })
-
```
searchableguy, line 153: Removed this since it was old and we store metadata for lock now so not sure local files are useful.

julien51: Actually we may want to keep it because some old locks are still using this for their metadata and I don't think we can break them :/

julien51 (commented):
Please restore the stuff from the config file... after that, LG!

locksmith/src/operations/metadataOperations.ts
searchableguy: Restoring.

julien51: thx!

julien51 (approved):

Issue #10220, opened 2022-11-06
Fix subgraph RemoveLockManager


0xYee:
**Describe the bug**
When a Lock Manager is removed, the subgraph removes more managers than supposed to.

**To Reproduce**
1. Add 2 (or more) lock manager to a lock
2. Remove the first one that was added
3. The ones added after the first one will also be removed from the subgraph

**Expected behavior**
Only remove the lock manager mentioned in the `LockManagerRemoved` event.

**Screenshots**
If you go to the subgraph playground, it shows no lockManagers:
![image](https://user-images.githubusercontent.com/57074511/200188692-cca9246a-0df4-4ef3-aead-824f8042f157.png)

If you go directly to the contract, it shows that address `0x4281aE3793A513D2e1Bf600948B0b1b837Ab37ad` is indeed a Lock Manager:
![image](https://user-images.githubusercontent.com/57074511/200188643-e436d5eb-8313-4ac0-93c7-a179bff25837.png)

**To Replicate:**
https://mumbai.polygonscan.com/address/0xc3c8bfb8248f1cbee5bd58bb9254e02bcfa918e2#readProxyContract
with address `0x4281aE3793A513D2e1Bf600948B0b1b837Ab37ad`

On subgraph playground:
```{
  locks(where:{address:"0xc3c8BFb8248F1CbeE5bd58bb9254e02BcFa918e2"}) {
    lockManagers
  }
}
```

**Reason**:
On `public-lock.ts` in the function `handleLockManagerRemoved`. The splice logic is missing the second argument to indicate how many items to splice. Just adding a `1` should fix this. I'll add a PR for this:
https://github.com/unlock-protocol/unlock/blob/679bd7321b1bfcdc51094d54da9da146befcaa58/subgraph/src/public-lock.ts#L213

PR #10221, opened 2022-11-06
fix(subgraph): don't remove entire list on handleLockManagerRemoved

0xYee:
Updated the `splice` function to only remove i. Before it was removing all lock managers after i. 

(If `splice` doesn't have second argument (number to delete) it assumes to delete until it reaches the `array.length`)

# Description

PR to fix issue #10220 . Which removes more LockManagers on the subgraph than it should.

# Issues
#10220

### DIFF
--- a/subgraph/src/public-lock.ts
+++ b/subgraph/src/public-lock.ts
@@ -210,7 +210,7 @@ export function handleLockManagerRemoved(event: LockManagerRemovedEvent): void {
   if (lock && lock.lockManagers) {
     const lockManagers = lock.lockManagers
     const i = lockManagers.indexOf(event.params.account)
-    lockManagers.splice(i)
+    lockManagers.splice(i, 1)
     lock.lockManagers = lockManagers
     lock.save()
   }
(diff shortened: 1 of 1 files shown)


subgraph/src/public-lock.ts
```diff
@@ -210,7 +210,7 @@ export function handleLockManagerRemoved(event: LockManagerRemovedEvent): void {
   if (lock && lock.lockManagers) {
     const lockManagers = lock.lockManagers
     const i = lockManagers.indexOf(event.params.account)
-    lockManagers.splice(i)
+    lockManagers.splice(i, 1)
```
searchableguy, line 213: Wonder if we could use filter here to avoid mutation and make this more readable?

julien51:
Thanks @0xYee ! I will let @clemsos on our end review this because he has the most knowledge here!

clemsos:
thanks @0xYee ! Agree with @searchableguy there, maybe using filter will be more appropriate? wdyt?

PR #10224, opened 2022-11-07
feat(unlock-app) - allow 0 as maxNumber of keys

kalidiagne:
# Description

 Update maxNumber to ZERO keys

Screenshot 2022-11-07 at 16 07 23

Screenshot 2022-11-07 at 16 07 33

Screenshot 2022-11-07 at 16 08 23

# Issues

Fixes #
Refs #

# Checklist:


searchableguy:
I wonder we minimum by default should be whatever are existing number of keys distributed. :)

kalidiagne:
> I wonder if minimum by default should be whatever are existing number of keys distributed.

I think this can be tricky for lockManager standpoint, users can expire their keys at any time, for me the best is to keep it without any forced default value

julien51:
> I wonder if minimum by default should be whatever is the valid number of keys.

No, the valid number of keys changes when they expire anyway.

julien51 (approved):

PR #10227, opened 2022-11-07
feat(unlock-js): Add uniswap consult function in web3Service

searchableguy:
# Description

Adds function to query uniswap contract through unlock contract for price. 

```ts
const price = await web3Service.consultUniswap({
 network: 1,
 tokenInAddress: '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2',
 tokenOutAddress: '0x6B175474E89094C44Da98b954EedeAC495271d0F',
 value: '1',
})
```
# Issues

Fixes #
Refs #

# Checklist:


packages/unlock-js/src/web3Service.ts
```diff
@@ -609,4 +609,83 @@ export default class Web3Service extends UnlockService {
...
+    tokenInAddress: string
+    tokenOutAddress?: string
+    value: string
+    network?: number
+  }) {
+    const { network, tokenInAddress, value } = options
+    const networkId = network || 1
+    const networkConfig = this.networks[networkId] // By default, use mainnet
+    const tokenOutAddress =
+      options.tokenOutAddress ||
```
julien51, line 637: I like that, but lets' add also a `tokenInAddress` to default to the local chain's "wrapped" token, like Eth on mainnet, MATIC if the the network is polygon, BNB if the network is BSC... etc.
That means you need to add the "mainnet" ERC20 version of each chain's token but I think that's doable?

wdyt?

packages/unlock-js/src/web3Service.ts
```diff
@@ -609,4 +609,83 @@ export default class Web3Service extends UnlockService {
...
+      throw new Error('No unlock contract address found in the network config')
+    }
+
+    const unlockContract = await this.getUnlockContract(
+      networkConfig.unlockAddress!,
+      provider
+    )
+
+    // Get the uniswapContractAddress from unlock contract
+    const uniswapContractAddress = await unlockContract.uniswapOracles(
```
julien51, line 659: Ha no! We can't use the Unlock contract because it may not have all the tokens and we want to support any token that's support by Uniswap, not Unlock. We need to query the Uniswap contract(s) directly.

julien51: These are public and we can add to the config.

julien51: (we also save a call!)

julien51 (changes requested):
Let's find the routes directly without queryoing Unlock's contract. Maybe @clemsos can help?

packages/unlock-js/src/web3Service.ts
searchableguy: Okay will directly use the uniswap contracts :D

packages/networks/src/networks/arbitrum.ts
```diff
@@ -42,6 +42,9 @@ export const arbitrum: NetworkConfig = {
   isTestNetwork: false,
   description:
     'Arbitrum One is a Layer 2 (L2) chain running on top of Ethereum Mainnet that enables high-throughput, low cost smart contracts operations.',
+  uniswapV3: {
```
julien51, line 45: I am not sure we need these for anything but mainnet?

searchableguy: I added it because there was one. If someone else wants to use the uniswap service for other network. They can do it.

packages/unlock-js/src/web3Service.ts
julien51: Please don't forget!

julien51 (commented):
I think it is good, but I would trust when I ass a quick test suite testing all combos of a 4 or 5 tokens:
ETH (WETH), USDC, WBTC, MATIC, UDT.

wdyt?

packages/networks/src/networks/arbitrum.ts
julien51: fair point

packages/unlock-js/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changes
 
+# 0.31.2
+
+- Add uniswap service
+- Add consultUniswap method to web3 service for getting price of a token using uniswap pools
```
julien51, line 6:
```suggestion
- Add consultUniswap method to web3 service for getting price of a token using uniswap v3 pools
```

packages/unlock-js/src/__tests__/web3Service.test.js
```diff
@@ -1,32 +1,13 @@
...
-
 import PublicLockVersions from '../PublicLock'
+import networks from '@unlock-protocol/networks'
+import { getCoinbasePricing } from './helpers/coinbase'
 
-const host = process.env.CI ? 'eth-node' : '127.0.0.1'
-const port = 8545
-const provider = `http://${host}:${port}`
-
-const networks = {
```
julien51, line 9: thanks for cleanup!

packages/unlock-js/src/uniswapService.ts
```diff
@@ -0,0 +1,169 @@
+import { NetworkConfigs } from '@unlock-protocol/types'
+import { TickMath, FullMath } from '@uniswap/v3-sdk'
+import { abi as UniswapV3PoolABI } from '@uniswap/v3-core/artifacts/contracts/UniswapV3Pool.sol/UniswapV3Pool.json'
+import { abi as UniswapV3FactoryABI } from '@uniswap/v3-core/artifacts/contracts/UniswapV3Factory.sol/UniswapV3Factory.json'
+import { ethers } from 'ethers'
+import JSBI from 'jsbi'
```
julien51, line 6: 😭 Another Big number!

packages/unlock-js/src/uniswapService.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  // Calculates time-weighted means of tick for a uniswap pool. This does not provide for liqiduity.
+  async consult({
+    network,
+    tokenIn,
+    tokenOut,
+    // 5 min by default
+    range = [60 * 5, 0],
+  }: ConsultOptions) {
+    const UniswapV3Pool = await this.getPoolContract({
```
julien51, line 61: Let's pass this as an argument so we can cache it more efficiently?

julien51 (commented):
LG, I think we can optimize a few things to increase speed and do efficient caching.

searchableguy:
Moved to using subgraph because on-chain was significantly slow.

### On-chain queries
image

### Subgraph + on-chain query 
image

packages/networks/src/networks/celo.ts
```diff
@@ -40,6 +40,11 @@ export const celo: NetworkConfig = {
   previousDeploys: [],
   isTestNetwork: false,
   teamMultisig: '0xc293E2da9E558bD8B1DFfC4a7b174729fAb2e4E8',
+  uniswapV3: {
```
julien51, line 43: let's only use official subgraphs... built by the Uniswap team. This is obviously _very_ sensitive and we don't want to rely on data we can't reliably trust.

packages/unlock-js/src/uniswapService.ts
```diff
@@ -0,0 +1,150 @@
...
+    return {
+      networkConfig,
+      provider,
+    }
+  }
+
+  /**
+   * Get pool address for erc20 token in and token out pairs on uniswap. By default, we find the 5% fee pools.
+   */
+  async getPoolAddress({
```
julien51, line 40: Looking at the subgraph I think we can probably get the pool from there?

julien51 (commented):
Looking more closely at this I think it can be simplified further. There is no need to auery the contract to get the pool at all.
We can just use the subgraph for this.

That said it also does not provide a way to "route", because I am not sure pools exist for _all_ pairs to USDC. So that means we need to do "routing". 
I think the way to do this is to get pools with USDC and pool with the tokens and convert between the 2. I would assume WETH might be the intermediate pool between any 2 token?

packages/unlock-js/src/uniswapService.ts
searchableguy: As mentioned on the discord. It does not work for some tokens it should hence the replacement with on-chain calls. 

The where filter is unreliable.

searchableguy:
> Looking more closely at this I think it can be simplified further. There is no need to auery the contract to get the pool at all.
We can just use the subgraph for this.

As I mentioned. Queries involving token address does not always yield a result when there is one on the chain. 

For example, this does not yield any result on the subgraph when there are few. 

```graphql
{
  pools(
    where: {
    token0: "0xdac17f958d2ee523a2206206994597c13d831ec7"
    token1: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"
  }) {
    token1Price
  }
}
```
This happens only with some tokens, not all so hard to predict and I've not dig further as to why. 

> That said it also does not provide a way to "route", because I am not sure pools exist for all pairs to USDC. So that means we need to do "routing".

Uniswap already provides a package for this called smart router. I tried it 
CleanShot 2022-11-10 at 16 51 10@2x

It takes 22 seconds to get a price quote. 

> I think the way to do this is to get pools with USDC and pool with the tokens and convert between the 2. I would assume WETH might be the intermediate pool between any 2 token?

I'm not sure that alone is a good way because you also have to consider liquidity and volume. ETH pools are expensive so they do not have volume for many low erc20 tokens as opposed to stables. 

> I would assume WETH might be the intermediate pool between any 2 token?

No, afaik.

packages/unlock-js/src/uniswapService.ts
```diff
@@ -26,125 +26,85 @@ export class UniswapService {
       throw new Error(`No network found for the id: ${networkId}`)
     }
     const provider = new ethers.providers.JsonRpcProvider(
-      networkConfig.provider
+      'https://mainnet.infura.io/v3/957ccfce160e46c3bbc6906d2e72f2de'
```
julien51, line 29: Please make this is a config option

searchableguy: ah sorry, it was stashed between changes. I was checking if our RPC provider was slow.

julien51 (approved):
ok LGTM

PR #10030, opened 2022-10-25
feat(smart-contracts): tests with uniswap v3 oracle 

clemsos:
# Description

This allows to use Uniswap V3 as oracle to calculate the GNP for Unlock.

Note that that part of the tests will now run on a mainnet fork only, as we dont want to have to redeploy Uniswap v3

```
export ALCHEMY_API_KEY=<xxx>
RUN_MAINNET_FORK=1 yarn hardhat test test/Unlock/uniswapValue.mainnet.js


Running a mainnet fork...

  Contract: Unlock / uniswapValue
    ✓ weth is set correctly already
    ✓ sets oracle address correctly
    A supported token (USDC)
      ✓ pricing is set correctly
      Purchase keys
        ✓ GDP went up by the expected ETH value
        ✓ a GDP tracking event has been emitted
    A unsupported token
      Purchase key
        ✓ GDP did not change
        ✓ a GDP tracking event has been emitted
    ETH
      Purchase key
        ✓ GDP went up by the keyPrice
        ✓ a GDP tracking event has been emitted

```

# Issues

Fixes #
Refs #

# Checklist:

- [ ] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #10241, opened 2022-11-10
fix(ci): simplify dockerfile

clemsos:
# Description

Our `Dockerfile` is cumbersome, contains too many steps - which prevents proper caching. This PR simplifies the process 

TODO: ~~rebase once #10260 is merged~~ 

# Issues

Fixes #
Refs #

# Checklist:


Dockerfile
```diff
@@ -46,107 +15,47 @@ RUN mkdir -p /home/unlock
...
+    default-jdk \
+    openjdk-11-jre
 
 # install deps
 USER node
 
-# attempt to create dir only for non-packages
-RUN if echo ${BUILD_DIR} | grep -q "packages" ; then echo "skipping"; else mkdir /home/unlock/${BUILD_DIR}; fi
+# copy all code
+COPY --chown=node . /home/unlock/
```
julien51, line 38: doing this here means every PR will likely trigger anything below to be executed again because every PR contains file changes.
We need to copy only the package.json & yarn.lock first, then build the deps and finally copy the files so that dependency building can be cached.

Dockerfile
```diff
@@ -3,69 +3,99 @@
...
+WORKDIR ${DEST_FOLDER}
+RUN yarn
+
+#
+# 2. build packages and prepare image for testing/dev
+#
+FROM node:16-bullseye-slim as dev
+
+# install all deps required to build packages
+RUN apt-get update \
```
julien51, line 63: I think these change very little, so maybe move upward?

clemsos: we just copy the file from the stage above, so we have to install things to compile deps

julien51 (approved):
Ok, this _looks_ good to me but is unfortunately slower than what we currently have?

docker/docker-compose.ci.yml
```diff
@@ -21,7 +21,7 @@ services:
     platform: linux/amd64
     build:
       context: ../
-      target: subgraph
+      target: dev
       args:
         BUILD_DIR: subgraph
```
searchableguy, line 27: We actually have entries with custom command which assume they are inside the package folder such as `locksmith` and `locksmith-websub`. 

Ditto for some parts of CI scripts.

clemsos:
> Ok, this _looks_ good to me but is unfortunately slower than what we currently have?

yes the goal here is just to simplify the dockerfile so we can easily cache the image. So there are no gain on first build but should be much faster once we have docker caching

docker/docker-compose.ci.yml
clemsos: yes the "production" step needs to be adapted as we dont use that yarn prod-install plugin anymore. The best approach is to move the workdir into the right folder and run the yarn command without workspace prefix then.

clemsos: 144d0ae should fix it. wdyt?

julien51:
> my thinking was to add caching to another PR

Works for me!

clemsos:
I have started a draft for the docker cache here https://github.com/unlock-protocol/unlock/pull/10255

clemsos:
looking at it, I think the current design is a bit silly. It took 281.3s to copy files from stage deps to stage dev, which could be avoided altogether by just reusing the whole stage deps directly (which will also allow to run apt upgrade a single time). Just need to figure out the proper perms to sue `/home/unlock/` as the workdir so we keep things clean

clemsos:
so integration tests fails but logs appears to be unaccessible from CI 😓

Dockerfile
clemsos: actually better to build everything on first step (see https://github.com/unlock-protocol/unlock/pull/10241#issuecomment-1311623124)

clemsos:
before merging this, need to revert [cf2dd0b](https://github.com/unlock-protocol/unlock/pull/10241/commits/cf2dd0bdba0db61c32c8a347f8c82627bf1e3205) and

julien51:
ha! That commit was squashed. Can you just put that change in this one?

clemsos:
hmmm I messed up with branches here so better to merge #10030 and rebase to clean things up...

clemsos:
so the yarn install command takes 566.5s with the new docker image, while it took only 330.0s in the previous version. (We are installing all deps in the docker image VS installing just the ones required for the specific folder / workspace before). 

Hope the caching adds significant time improvement !

PR #10244, opened 2022-11-10
updated a few docstrings

julien51:
# Description

I wonder if the best way to do that wouldn't be to find a listing tool that checks all docstrings?
cc @clemsos 

# Checklist:

### DIFF
--- a/smart-contracts/contracts/interfaces/IPublicLock.sol
+++ b/smart-contracts/contracts/interfaces/IPublicLock.sol
@@ -282,27 +282,27 @@ interface IPublicLock
    * Determines how much of a fee would need to be paid in order to
    * transfer to another account.  This is pro-rated so the fee goes 
    * down overtime.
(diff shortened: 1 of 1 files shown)


clemsos:
> I wonder if the best way to do that wouldn't be to find a listing tool that checks all docstrings?

I am not aware of such tool for solidity. One thing I found out reading the uniswap contracts is that they do all documentation in the interface, then just inherit the docs within the contract - like shown [here](https://github.com/Uniswap/v3-core/blob/05c10bf6d547d6121622ac51c457f93775e1df09/contracts/UniswapV3Factory.sol#L34). Actually their doc strings are quite clean (even though very minimal). They use docusaurus too ([here](https://github.com/Uniswap/docs)) but have a different lib to parse dosctrings - `[solidity-docgen](https://github.com/OpenZeppelin/solidity-docgen)` developed by oz. Worth exploring imo

Issue #10005, opened 2022-10-24
Logic change to show the "pay with crypto" option


julien51:
Right now, te logic for "pay with crypto" checks whether the user has enough of the currency (and base currency) to pay for the tx (+gas) and only shows the "pay with crypto" option if they do.
We should change this for the following:
* if the user is an Unlock account, then do as we do above.
* if the user is *not* and Unlock account, then, show a _disabled_ "pay with crypto" option with the an explanation of why they can't use that option (either not enough of the coins used for the price, or not enough to pay for gas).

PR #10248, opened 2022-11-10
feat(unlock-app): Always show crypto button with amount needed to cover transaction

searchableguy:
# Description

fixes https://github.com/unlock-protocol/unlock/issues/10005

CleanShot 2022-11-11 at 20 06 48@2x

# Issues

Fixes #
Refs #

# Checklist:


searchableguy:
Is an explicit message required? I think it's self explanatory.

julien51:
> Is an explicit message required? I think it's self explanatory.

Yes I think we need an explicit message, especially because there may be a misunderstanding between the need for Eth (or similar) to pay for gas and the need for ERC20.

unlock-app/src/components/interface/checkout/main/Payment.tsx
```diff
@@ -29,24 +30,31 @@ interface Props {
   checkoutService: CheckoutService
 }
 
+interface AmountBadeProps {
```
julien51, line 33:
```suggestion
interface AmountBadgeProps {
```

unlock-app/src/components/interface/checkout/main/Payment.tsx
```diff
@@ -29,24 +30,31 @@ interface Props {
...
+  symbol: string
+  amount: string
+}
+
+const AmountBadge = ({ symbol, amount }: AmountBadeProps) => {
```
julien51, line 38:
```suggestion
const AmountBadge = ({ symbol, amount }: AmountBadgeProps) => {
```

searchableguy:
Updated to show message when user does not have gas fee to cover the transaction. 

For payment, we show the amount to be paid and balance. 

@julien51

julien51 (approved):
Very good!

PR #10280, opened 2022-11-16
fix(unlock-app): fixing isGasPayable

julien51:
# Description

The current calculation actually does not make a lot of sense.
We looked at gas price, but not the actual gas being spent...
For simplicity, we just look at whether the user as some token to pay for gas. Their wallet will actually block sending the transaction if they don't.

# Checklist:


searchableguy (approved):

searchableguy:
Calculating the gas fee on calling the actual method with arguments was too slow so I used the network base gas fee as an alternate minimum.

PR #10286, opened 2022-11-16
feat(unlock-js): Adding runEstimate for purchase

julien51:
# Description

This will actually be useful for the swap and purchase flow!

We will probably add this to more functions as well over time.

# Checklist:


packages/hardhat-plugin/src/networks.json
```diff
@@ -4,6 +4,7 @@
     "unlockAddress": "0x3d5409CcE1d45233dE1D4eBDEe74b8E004abDD13",
```
julien51, line 4: I'll keep these changes (I rebuilt locally!)

PR #10297, opened 2022-11-18
feat(unlock-app): optional recipient

julien51:
# Description

In an effort to simplify the checkout flow "by default", we want to skip the step where the user chooses the recipient, but we want to keep it as an option in the paywall config.

[This is a draft because I need to understand the state machine a bit more!]

# Checklist:


unlock-app/src/components/interface/checkout/main/Select.tsx
```diff
@@ -368,7 +376,8 @@ export function Select({ checkoutService, injectedProvider }: Props) {
                     lock,
                     existingMember: !!membership?.member,
                     skipQuantity,
-                    // unlock account are unable to renew
+                    skipRecipient,
+                    // unlock account are unable to renew : wut?
```
searchableguy, line 380: We don't have an endpoint for this because we would need to trigger the transaction on their behalf. 

We decided to discourage unlock account users to pay using crypto or fund their wallet created by us previously.

julien51: > We don't have an endpoint for this because we would need to trigger the transaction on their behalf.

I am confused. On the API call that calls `purchase`, we could easily check if the recipient already has an expired membership NFT and use `extendMembershipFor` if they do?

searchableguy: We could. We didn't add it at the time and disabled unlock accounts from using crypto version.

julien51: Ok I'll try to work on this in a different PR!

PR #10302, opened 2022-11-18
fix(ci): fix locksmith deployment on new version of the Dockerfile 

clemsos:
# Description

Following #10241 that had to be reverted, this should add a fix to deploy locksmith

# Issues

Fixes #
Refs #

# Checklist:


Dockerfile
```diff
@@ -1,164 +1,90 @@
...
-
-WORKDIR /app
-
-# copy package info
-COPY --from=build --chown=node /home/node/app .
+WORKDIR /home/unlock/${BUILD_DIR}
 
 # start command
-EXPOSE $PORT
+EXPOSE 3000
```
julien51, line 88: I know you had questions about this but are we sure we want to hardcode the port in the docker file?
Would be better in the dockercompose config maybe?

julien51 (commented):
If it runs, good for me!

julien51 (approved):

docker/docker-compose.ci.yml
```diff
@@ -137,8 +115,6 @@ services:
     build:
```
searchableguy, line 115: Does not specify port?

locksmith runs on 8080 by default, no?

PR #10347, opened 2022-11-23
Revert "fix(ci): fix locksmith deployment on new version of the Dockerfile "

searchableguy:
Reverts unlock-protocol/unlock#10302


clemsos:
let just merge #10348 instead of reverting

PR #10355, opened 2022-11-23
feat(unlock-app): Display Link in properties in the metadata form

searchableguy:
# Description

CleanShot 2022-11-24 at 00 43 45@2x

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/jest.config.js
+++ b/unlock-app/jest.config.js
@@ -43,7 +43,7 @@ module.exports = {
   coveragePathIgnorePatterns: ['src/stories/.*/*.stories.js'],
   coverageThreshold: {
     global: {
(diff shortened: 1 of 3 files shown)


julien51 (approved):

julien51:
Instead of link can you add the domain name for the link?

searchableguy:
@julien51 I tried that initially but it's hard to understand what the intent is because the property itself is called "twitter" so you have "twitter" and "twitter.com" when the actual link is `twitter.com/julien51`

julien51:
Well I think it's actually not bad to have something redundant, but let's get out with this and we'll improve later. Progress over perfection!

PR #10374, opened 2022-11-25
feat(unlock-app) - settings referrer fees

kalidiagne:
# Description
Screenshot 2022-11-25 at 17 20 19

Screenshot 2022-11-25 at 17 32 54

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/packages/unlock-js/src/walletService.ts
+++ b/packages/unlock-js/src/walletService.ts
@@ -891,4 +891,30 @@ export default class WalletService extends UnlockService {
       callback
     )
   }
+
+  /**
+   * Update referrer fee
+   * @param {*} params
+   * @param {*} callback
+   */
+  async setReferrerFee(
+    params: {
+      lockAddress: string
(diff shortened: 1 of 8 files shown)


searchableguy (approved):

packages/unlock-js/src/walletService.ts
```diff
@@ -891,4 +891,30 @@ export default class WalletService extends UnlockService {
...
+  /**
+   * Update referrer fee
+   * @param {*} params
+   * @param {*} callback
+   */
+  async setReferrerFee(
+    params: {
+      lockAddress: string
+      address: string
+      feeBasisPoint: number
```
searchableguy, line 904: same here. Instead of basis point, does it make sense to use percentage?

kalidiagne: maybe it causes confusion in this way, the value stores the basis point

PR #10477, opened 2022-12-01
feat(unlock-app) - restart when user changes network on the checkout

kalidiagne:
# Description

https://user-images.githubusercontent.com/20865711/205132341-4f8de683-8fbb-4a39-8265-ff91fa04336d.mov

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/main/index.tsx
+++ b/unlock-app/src/components/interface/checkout/main/index.tsx
@@ -39,7 +39,7 @@ export function Checkout({
     },
   })
   const [state] = useActor(checkoutService)
-  const { account } = useAuth()
+  const { account, network: connectedNetwork } = useAuth()
   const { mint, messageToSign } = state.context
   const matched = state.value.toString()
   const paywallConfigChanged = !isEqual(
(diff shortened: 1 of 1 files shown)


searchableguy (approved):

unlock-app/src/components/interface/checkout/main/index.tsx
```diff
@@ -39,7 +39,7 @@ export function Checkout({
     },
   })
   const [state] = useActor(checkoutService)
-  const { account } = useAuth()
+  const { account, network: connectedNetwork } = useAuth()
```
searchableguy, line 42: A more robust way would be to listen to provider.on change network events so any unnecessary state doesn't cause re-render and back to select. WDYT?

julien51 (approved):

unlock-app/src/components/interface/checkout/main/index.tsx
kalidiagne: yeah! that's a great idea! changed

julien51 (approved):

PR #10513, opened 2022-12-06
Added freeClaim property on the networks and updated the docs template

wonderwomancode:
Making it clear that we have free claims and which networks they are on.


julien51 (commented):
Could you please also update the locksmith file so we use the values set here ?

julien51 (commented):
This is a good start though!

packages/networks/src/networks/avalanche.ts
```diff
@@ -41,6 +41,7 @@ export const avalanche: NetworkConfig = {
   startBlock: 17188332,
   previousDeploys: [],
   isTestNetwork: false,
+  freeClaim: 1,
```
julien51, line 44:
```suggestion
  maxFreeClaimCost: 1,
```

wonderwomancode:
> Do you need help with the TS issue @wonderwomancode ?

Yes, now I'm not seeing why is giving me an error now that maxFreeClaimCost does not exist after I fixed the renamed ones I missed and added it to the NetworkConfig type definition.

PR #10514, opened 2022-12-06
feat(unlock-app) - warning lock name

kalidiagne:
# Description
Show a warning to inform that the lock name will not be shown if the metadata name is set.

Screenshot 2022-12-06 at 09 38 20

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/locks/Settings/forms/UpdateNameForm.tsx
+++ b/unlock-app/src/components/interface/locks/Settings/forms/UpdateNameForm.tsx
@@ -88,7 +88,8 @@ export const UpdateNameForm = ({
               <span>
                 <span className="flex gap-1">
                   <span>
-                    This value will be set on the contract but the NFT metadata will remain unchanged if you have set a value there. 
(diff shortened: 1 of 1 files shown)


unlock-app/src/components/interface/locks/Settings/forms/UpdateNameForm.tsx
```diff
@@ -70,6 +70,7 @@ export const UpdateNameForm = ({
           error={errors?.name && 'Lock name should have at least 3 characters.'}
           autoComplete="off"
           disabled={disabledInput}
+          description={`This value will not appear when NFT metadata name is changed.`}
```
julien51, line 73: Please add a link to the sectiion where people can update the NFT metadata.

julien51 (changes requested):
This is not good enoughg. We need to add a link, which means we need description to support JSX as well...

kalidiagne:
> This is not good enoughg. We need to add a link, which means we need description to support JSX as well...

Screenshot 2022-12-06 at 16 10 19

PR #10521, opened 2022-12-06
refactor(unlock-js): Use local node for generating subgraph service

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/.github/workflows/_tests.yml
+++ b/.github/workflows/_tests.yml
@@ -43,7 +43,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Setup integration environment
         run: scripts/setup-integration.sh
-      - name: Log failing tests
+      - name: Log errors inside containers
         if: ${{ failure() }}
         run: ./scripts/docker-compose-log.sh
       - name: Running ${{ matrix.target }} tests
--- a/.github/workflows/_tests.yml
+++ b/.github/workflows/_tests.yml
@@ -42,13 +42,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Setup integration environment
-        run: scripts/setup_integration.sh
+        run: scripts/setup-integration.sh
       - name: Log failing tests
         if: ${{ failure() }}
         run: ./scripts/docker-compose-log.sh
       - name: Running ${{ matrix.target }} tests
         run: ./scripts/tests.sh ${{ matrix.target }}
-        shell: bash
 
   integration-tests:
     name: Integration Tests
--- a/.github/workflows/_tests.yml
+++ b/.github/workflows/_tests.yml
@@ -41,13 +41,8 @@ jobs:
         target: ${{fromJson( inputs.changed )}}
     steps:
       - uses: actions/checkout@v3
-      - name: Run Integration Tests
-        run: scripts/integration-tests.sh
-      - name: Archive production artifacts
-        uses: actions/upload-artifact@v3
-        with:
-          name: screenshots
-          path: /tmp/screenshots
+      - name: Setup integration environment
+        run: scripts/setup_integration.sh
       - name: Log failing tests
         if: ${{ failure() }}
         run: ./scripts/docker-compose-log.sh
--- a/.github/workflows/_tests.yml
+++ b/.github/workflows/_tests.yml
@@ -41,8 +41,8 @@ jobs:
         target: ${{fromJson( inputs.changed )}}
     steps:
       - uses: actions/checkout@v3
-      - name: Setup integration environment
-        run: scripts/setup-integration.sh
+      - name: Setup local environment stack
+        run: scripts/run-stack-dockerized.sh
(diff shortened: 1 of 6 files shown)


julien51:
cc @clemsos

clemsos:
so we need to run a graph node here too. We have one running for integration scripts so the logic should be similar

scripts/setup-integration.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+
+# First this script will deploy
+REPO_ROOT=$(dirname "$0")/..
+INTEGRATION_TESTS_FOLDER=$REPO_ROOT/tests
+EXTRA_ARGS=$*
+
+sh $INTEGRATION_TESTS_FOLDER/bin/tests.sh
```
clemsos, line 9: that will work as long as we don't add locksmith etc to the `bin/tests.sh` script - which will change as soon as we have e2e tests.

I was mentioning the other day having a `scripts/run-stack.sh` that could be used for subgraph by both integration and unlock-js - as well as e2e later on.

We can leave it here for now and make these changes in another PR maybe?

clemsos (approved):
so far so good!

scripts/setup-integration.sh
searchableguy: Let me rename it to run-stack.sh - will leave rest as is. We can modify it to run stack only and move away from that in the tests.sh

searchableguy:
While this runs node in localhost. The endpoint from networks package does not return introspection data so graphql generator isn't working. Any idea why that is? I presumed it would be the same as on the hosted platform.

.github/workflows/_tests.yml
```diff
@@ -41,9 +41,13 @@ jobs:
         target: ${{fromJson( inputs.changed )}}
     steps:
       - uses: actions/checkout@v3
+      - name: Setup integration environment
+        run: scripts/setup-integration.sh
```
julien51, line 45: Let's rename this because it is **not** integration tests

julien51: `scripts/setup-stack.sh`

julien51: We should also rename `./bin/tests-local.sh`

julien51: `/Users/julien/repos/unlock/tests/bin/tests-local.sh` and `/Users/julien/repos/unlock/tests/bin/tests.sh` should be split between the "setup" part (that moves to `/unlock/scripts`) and the "runner" part that stays here.

clemsos:
What is happening here is that generating from localhost fails as it it part of the root level `yarn build` and that task is required to initialise the subgraph node. So we are calling it *before* we run the sbgraph itself, leading to the following error.

```
Generate to ./src/@generated/subgraph/index.ts
#27 165.2 ➤ YN0000: [STARTED] Load GraphQL schemas
#27 165.3 ➤ YN0000: [FAILED]
#27 165.3 ➤ YN0000: [FAILED]         Failed to load schema from http://localhost:8000/subgraphs/name/testgraph:
#27 165.3 ➤ YN0000: [FAILED]
#27 165.3 ➤ YN0000: [FAILED]         connect ECONNREFUSED 127.0.0.1:8000
#27 165.3 ➤ YN0000: [FAILED]         Error: connect ECONNREFUSED 127.0.0.1:8000
#27 165.3 ➤ YN0000: [FAILED]     at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16)
#27 165.3 ➤ YN0000: [FAILED]
#27 165.3 ➤ YN0000: [FAILED]         GraphQL Code Generator supports:
#27 165.3 ➤ YN0000: [FAILED]           - ES Modules and CommonJS exports (export as default or named export "schema")
#27 165.3 ➤ YN0000: [FAILED]           - Introspection JSON File
#27 165.3 ➤ YN0000: [FAILED]           - URL of GraphQL endpoint
#27 165.3 ➤ YN0000: [FAILED]           - Multiple files with type definitions (glob expression)
#27 165.3 ➤ YN0000: [FAILED]           - String in config file
#27 165.3 ➤ YN0000: [FAILED]
#27 165.3 ➤ YN0000: [FAILED]         Try to use one of above options and run codegen again.
#27 165.3 ➤ YN0000: [FAILED]
#27 165.3 ➤ YN0000: [FAILED]
```

I am investigating how to generate a JSON introspection file from subgraph folder - which may actually be easier.

julien51:
> I am investigating how to generate a JSON introspection file from subgraph folder - which may actually be easier.

Good idea!

clemsos:
> > I am investigating how to generate a JSON introspection file from subgraph folder - which may actually be easier.
> 
> Good idea!

drafted this approach in https://github.com/unlock-protocol/unlock/pull/11858 but I dont think it will work as the graph node does a bunch of private cooking of graphql under the hood which may be tedious to recreate

Issue #10511, opened 2022-12-05
Delay "switch network"


julien51:
In the checkout process, we ask users to switch networks pretty early in the flow when we probably don't really need users to switch until they actually send a transaction.

PR #10540, opened 2022-12-07
feat(unlock-app): Switch network at the end on the checkout only if required

searchableguy:
# Description

fixes https://github.com/unlock-protocol/unlock/issues/10511

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/main/Confirm.tsx
+++ b/unlock-app/src/components/interface/checkout/main/Confirm.tsx
@@ -239,7 +239,7 @@ export function Confirm({
 
   const SwitchNetwork = () => (
     <Button
-      disabled={isSwitchingNetwork}
+      disabled={isSwitchingNetwork || isLoading}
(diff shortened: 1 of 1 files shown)


unlock-app/src/components/interface/checkout/main/Renew.tsx
```diff
@@ -171,6 +176,21 @@ export function Renew({
     }
   }
 
+  const SwitchNetwork = () => (
```
julien51, line 179: I wonder if we should isolate and create a dedicated component?

julien51 (approved):
This looks good to me!
Let's do thorough tests on staging!

unlock-app/src/components/interface/checkout/main/Renew.tsx
searchableguy: If it's re-used more, we should.

PR #10542, opened 2022-12-08
fix(ci): add docker cache manifest to image

### LINKED ISSUE #10489: Fix docker caching
Our new Docker file does not work well for caching because we copy the files "too early" and every single pull request will modify at least one file, requiring the whole image to be rebuilt after that.

Let's move back to a world where we copy `package.json` and `yarn.lock` files earlier, install the dependencies and only copy the rest of the files _later_ in the flow.

Additionally, let's also make sure the `yarn` and `yarn build` steps are happening _inside_ of the `deps` stage so that the result of this can be cached though docker. (We can keep `RUN yarn apps:build` inside of `dev`).

Additionally we should explore ways to reduce the size of the image that gets cached (clocking at 6GB right now) by removing files once we have built all the required things. This should make caching much more useful.

clemsos:
# Description

Currently the docker image we are pushing doesn't contain the cache manifest, so docker can not use it as `cache_from`. This adds the correct [buildkit arg](https://docs.docker.com/engine/reference/builder/#buildkit-built-in-build-args) to actually allow the image to be used as cache.

# Issues

Fixes #10489
Refs #

# Checklist:

### DIFF
--- a/.github/workflows/_docker-cache.yml
+++ b/.github/workflows/_docker-cache.yml
@@ -40,6 +40,7 @@ jobs:
         with:
           context: .
           builder: ${{ steps.buildx.outputs.name }}
+          build-args: BUILDKIT_INLINE_CACHE=1
           # Build only desired target
           target: deps
           push: true 
--- a/.github/workflows/_docker-cache.yml
+++ b/.github/workflows/_docker-cache.yml
@@ -19,9 +19,8 @@ jobs:
         run: |
           TAG=$(echo $GITHUB_SHA | head -c7)
           IMAGE="unlockprotocol/unlock-dev"
-          echo ::set-output name=tagged_image::${IMAGE}:${TAG}
-          echo ::set-output name=tag::${TAG}
-          echo ::set-output name=latest::${IMAGE}:latest
+          echo "tagged_image=${IMAGE}:${TAG}" >> $GITHUB_OUTPUT
+          echo "latest=${IMAGE}:latest" >> $GITHUB_OUTPUT
 
       - name: Login to Docker Hub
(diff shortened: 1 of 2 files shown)


.github/workflows/_docker-cache.yml
```diff
@@ -19,9 +19,8 @@ jobs:
         run: |
           TAG=$(echo $GITHUB_SHA | head -c7)
           IMAGE="unlockprotocol/unlock-dev"
-          echo ::set-output name=tagged_image::${IMAGE}:${TAG}
-          echo ::set-output name=tag::${TAG}
-          echo ::set-output name=latest::${IMAGE}:latest
+          echo "tagged_image=${IMAGE}:${TAG}" >> $GITHUB_OUTPUT
+          echo "latest=${IMAGE}:latest" >> $GITHUB_OUTPUT
```
clemsos, line 23: the `set-ouput` is getting deprec on gh actions so upgrading that too here

julien51:
Let's test locally with the following flow:

From a _clean_ docker:

1. use `/script/tests.sh unlock-app`. This should take some time as it will build everything
2. once succeeded push the image to docker hub.
3. cleanup docker again locally
4. re-run `/script/tests.sh unlock-app`. This one should only download the layers from Docker Hub and not actually build anything locally, which.... depending on network speed, could be faster.

What do you think? If we have this, then the way to make things faster will be to make the docker image lighter!

julien51 (approved):

julien51:
I will merge for now just to see if this does indeed make things faster.

julien51:
Let's do it!

Issue #8179, opened 2022-02-01
rebuild all package deps on yarn build


clemsos:
> it looks like when someone builds a library it does not rebuild all of its dependencies.
> For example I just (re)built unlock-js locally but it did not build the contracts package first. I had to go rebuild it manually myself for unlock-js to be using the right values.
> Is there a way to tell yarn to rebuild all of the deps?

We could leverage https://yarn.build for that

RishiKumarRay:
can you give some context to solve this ?

julien51:
@RishiKumarRay ! Thanks for jumping in! We use a monorepo for our code. Right now one of the challenges is that when we build a piece of the code, and it has other pieces as dependencies, we do not necessarily trigger a build in the dependency which might lead to outdated dependencies.

clemsos:
The main idea is that without rebuilding the packages automatically, some part of the repo can become out of sync or go missing, therefore creating unexpected errors. 

The best way to reproduce is to just clone the repo and run a build in one of the folders (e.g. unlock-app). That will fail as the networks or contracts packages  are used but they donc get build automatically. 

Not sure what is the best way to solve this. On one hand, having an automatic rebuild will be useful. On the other hand, adding a complete rebuild of the packages will slow down the repo workflow when developing locally


One approach that @searchableguy pointed at was to use https://yarn.build . I try to be very cautious when adding new tools to the workflow, to avoid lock-ins and possible debt in the future but that is an option worth exploring maybe

julien51:
We could maybe have a Github action for this? 
It checks master out, rebuilds all the packages and opens a PR with changes.
This could run as a cron once a day for example

clemsos:
ok I just tried yarn.build and it does not seem to work.

here is the std of a dry run after issuing ```yarn plugin import https://yarn.build/latest``` (and removing the existing `build` command in the root package.json)


```
An error occured in yarn.build. TypeError: Cannot read property 'stderr' of undefined
    at Dc.checkIfRunIsRequired (/Users/clement/Dev/unlock/unlock-alt/.yarn/plugins/@ojkelly/plugin-all.cjs:118:5482)
    at async Dc.plan (/Users/clement/Dev/unlock/unlock-alt/.yarn/plugins/@ojkelly/plugin-all.cjs:79:968)
    at async Dc.addRunTarget (/Users/clement/Dev/unlock/unlock-alt/.yarn/plugins/@ojkelly/plugin-all.cjs:118:3529)
[...]

PR #10545, opened 2022-12-08
feat: introduce turbo as monorepo task manager

clemsos:
# Description

[Turbo](https://turbo.build/) seems like the new kids on the block when dealing with building complex apps in monorepo, with full yarn support etc. It will take care of parsing dependencies within workspaces when rebuilding packages, running tests, linting, etc

Relates to #8179 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "eslint-plugin-prettier": "latest",
     "husky": "8.0.2",
     "lint-staged": "13.0.4",
-    "prettier": "2.8.0"
+    "prettier": "2.8.0",
+    "turbo": "^1.6.3"
   },
   "workspaces": [
     "packages/**",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -22,7 +22,8 @@
     "eslint-plugin-prettier": "latest",
     "husky": "8.0.2",
     "lint-staged": "13.0.4",
-    "prettier": "2.8.0"
+    "prettier": "2.8.0",
+    "turbo": "^1.6.3"
```
searchableguy, line 26: pin?

searchableguy:
Nice. I don't have any experience with turbo but I've used https://nx.dev before which used to generate package.json individually for each package in the build so you could cache your dependencies easily and use their computational cache for the workflow instead of depending on git script.

I know turbo is pretty similar and supported by vercel.

julien51:
Closing as stale. We can revisit when the time comes!

PR #10556, opened 2022-12-09
refactor(locksmith): Replace jest with vitest

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:


locksmith/__tests__/controllers/metadataController/tokenData.test.ts
```diff
@@ -120,11 +126,11 @@ describe('Requesting Token Data', () => {
       },
     })
 
-    mockWeb3Service.isLockManager = jest.fn(() => Promise.resolve(false))
+    mockWeb3Service.isLockManager = vi.fn(() => Promise.resolve(false))
   })
 
   describe("when persisted data doesn't exist", () => {
-    it('returns wellformed data for Week in Ethereum News', async () => {
+    it.skip('returns wellformed data for Week in Ethereum News', async () => {
```
clemsos, line 133: is that normal for these to be skipped?

locksmith/server.js
```diff
@@ -1,66 +0,0 @@
-/* eslint no-console: 0 */
-console.log('Starting Locksmith...')
-const args = require('yargs').argv
```
clemsos, line 3: thats unused?

searchableguy: Yes. see `server.ts`

locksmith/__tests__/controllers/metadataController/tokenData.test.ts
searchableguy: Yes. 3 are skipped because they need to change. I will do it after this PR.

julien51: Could do these changes _before_ this PR so we don't forget?

locksmith/__tests__/controllers/priceController.test.ts
```diff
@@ -1,55 +0,0 @@
-import request from 'supertest'
```
julien51, line 1: Are we not using this controller anymore?

locksmith/config/config.js
```diff
@@ -48,9 +48,4 @@ if (process.env.DATABASE_URL) {
   }
 }
 
-if (process.env.UNLOCK_ENV === 'prod' || process.env.UNLOCK_ENV === 'staging') {
```
julien51, line 51: Hum, are you sure this is not needed?

locksmith/src/config/config.ts
```diff
@@ -0,0 +1,54 @@
...
+  config.database.port = Number(databaseConfigUrl.port)
+  config.database.database = databaseConfigUrl.pathname.substring(1)
+} else {
+  config.database.username = process.env.DB_USERNAME
+  config.database.password = process.env.DB_PASSWORD
+  config.database.database = process.env.DB_NAME
+  config.database.host = process.env.DB_HOSTNAME
+}
+
+if (process.env.UNLOCK_ENV === 'prod' || process.env.UNLOCK_ENV === 'staging') {
```
julien51, line 49: Oh I see it here now

julien51 (approved):

locksmith/__tests__/controllers/priceController.test.ts
searchableguy: It was skipped so far. We've not been running it at all for a while. 

The test needs a rewrite so I'm removing it. We already test the new price controller.

locksmith/src/config/config.ts
searchableguy: To explain, I've moved everything to start locksmith to typescript and ESM. 

but we still need the config file as plain cjs for sequelize-cli and using it in migration files which aren't typescript. 

So I duplicated parts I need for tests and migration in config folder. I will think of a way to remove the duplication but it's the most straight forward solution.

locksmith/config/config.js
searchableguy: Not required as this is not used by locksmith code anymore. Only sequelize-cli and sequelize migration files. I will add it back anyway so we keep it same.

locksmith/__tests__/controllers/priceController.test.ts
julien51: works for me!

julien51: Let's see if we can remove the old priceController completely?

PR #10591, opened 2022-12-15
feat(unlock-app) - add data-testid to form elements

kalidiagne:
# Description

Add `data-testid` to form elements in order to get items quickly during tests and avoid manually adding everytime 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/packages/ui/lib/components/Form/Input.tsx
+++ b/packages/ui/lib/components/Form/Input.tsx
@@ -58,6 +58,7 @@ export const Input = forwardRef(
       description,
       label,
       icon,
+      name,
       ...inputProps
     } = props
     const [isCopied, setCopy] = useClipboard(props.value as string)
@@ -103,6 +104,7 @@ export const Input = forwardRef(
             value={value}
             ref={ref}
             className={inputClass}
+            data-testid={name || label}
           />
           <div className="absolute inset-y-0 right-0 flex items-center pr-1 ml-4">
             {copy && !hidden && (
--- a/packages/ui/lib/components/Form/TextBox.tsx
+++ b/packages/ui/lib/components/Form/TextBox.tsx
@@ -39,6 +39,7 @@ export const TextBox = forwardRef(
       success,
       description,
       label,
+      name,
       ...restProps
     } = props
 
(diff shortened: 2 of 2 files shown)


searchableguy:
I'm not sure this is a good thing to add in UI library because you could have two different forms on the same page with same key "address" for example. 

Any test-id should be explicitly added on the page or component where you are testing things.

kalidiagne:
> I'm not sure this is a good thing to add in UI library because you could have two different forms on the same page with same key "address" for example. How do you differentiate in that case?
> 
> ```tsx
> // in one part of the page
> <Input label="Address" />
> // in the second part of the page.
> <Input label="Address" />
> ```
> 
> Any test-id should be explicitly added on the page or component in the app itself.

i see, the thing here is to have a specific identifier based on the `name`

in the case you are describing we can do something like 

`document.querySelector('#form1 [data-testid="field"]')`

`document.querySelector('#form2 [data-testid="field"]')`

kalidiagne:
> > I'm not sure this is a good thing to add in UI library because you could have two different forms on the same page with same key "address" for example. How do you differentiate in that case?
> > ```tsx
> > // in one part of the page
> > <Input label="Address" />
> > // in the second part of the page.
> > <Input label="Address" />
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Any test-id should be explicitly added on the page or component in the app itself.
> 
> i see, the thing here is to have a specific identifier based on the `name`
> 
> in the case you are describing we can do something like
> 
> `document.querySelector('#form1 [data-testid="field"]')`
> 
> `document.querySelector('#form2 [data-testid="field"]')`


wdyt?

searchableguy:
I think that adds complexity and would introduce unnecessary complexity in figuring out the tree to target the elements when an ID would be the easiest to go especially if you refactor anything later on - you may also be required to change the test. 

So we can't depend on the form test id anyway.

kalidiagne:
> I think would introduce unnecessary complexity in figuring out the tree to target the elements when an ID would be the easiest to go especially if you refactor anything later on - you may also be required to change the test.
> 
> So we can't depend on the form test id anyway.

Super, lets close it for now and add it later if needed

PR #10565, opened 2022-12-12
fix(ci): make `run all tests` tag run properly

clemsos:
# Description

Initiated in #10534 but PRs were hanging when not using the tag.

# Issues

Fixes #
Refs #

# Checklist:

- [ ] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #10604, opened 2022-12-18
Docker : attempting to reduce image size

julien51:
# Description

Our current CI setup means that our image is pulled from Docker Hub quite often (where it is cached).
In order to speed up the process, we're trying to reduce the size of that image.

# Checklist:


julien51:
Hum, I need to remove the commits from the state branch here

julien51:
@clemsos How do we run all tests again please?

clemsos:
So the run all tests command works, it just does not show feedback on the PR. You have to dig into the Actions tab to find the ci run. The opened PR will remove that command and make use only of the tag.

clemsos:
Let's merge that one then can use the tag directly https://github.com/unlock-protocol/unlock/pull/10565

julien51:
Approved then!

julien51:
https://github.com/unlock-protocol/unlock/actions/runs/3732855189 ? I don't think it ran tests?

Dockerfile
```diff
@@ -15,7 +15,7 @@ USER root
 # install all deps required to build modules
 RUN apt-get update \
     && DEBIAN_FRONTEND=noninteractive \
-    apt-get install --assume-yes \
+    apt-get install --no-install-recommends --assume-yes \
```
julien51, line 18: https://askubuntu.com/questions/65081/what-are-the-implications-of-no-install-recommends-apt-get-install

julien51:
What do you think @clemsos @searchableguy ?

Dockerfile
```diff
@@ -80,6 +83,13 @@ RUN yarn build
 # Build locksmith and other apps separately to reduce startup time on heroku
 RUN yarn apps:build
 
+# Cleanup up image to make it lighter
+USER root
+RUN apt-get autoremove
```
julien51, line 88: Removing packages we don't need anymore.

Dockerfile
```diff
@@ -80,6 +83,13 @@ RUN yarn build
...
+USER node
+
+# Remove yarn cache (re-installing dependencies will take more time, but image will be 500MB lighter)
+RUN rm -rf .yarn/cache/
```
julien51, line 92: Removing yarn cache (about 40% of image size before!)

clemsos (approved):
LG! Should be much smaller with all these 👍

julien51:
Merci! Let's see!

Issue #10398, opened 2022-11-29
AddressInput UI component


julien51:
There are multiple places in our web applications where we use input fields for people to enter addresses. We should create a single component for this that we can re-use everywhere. This component should be added to our `UI` library (along with appropriate storybook stories)
This component should do the following:
* Format validation
* ENS resolution (optional, defaults to true), but keep ENS "visible" (however the value used in forms should be the `0x...` address

It should work natively with `useForm` that we use in many places.

Examples of places where it needs to be used:
* `Referrer` setting on the Checkout Builder,
* `Recipient` or `Key Manager` on the Airdrop Keys form,
* Verifier or Lock Managers in the `Roles` settings.

SouravInsights:
hey, @julien51 i'm interested in working on this issue!

julien51:
Alright! I am assigning it to you and we'll add a 300 USDC bounty paid on Polygon!
Please make sure you copy @searchableguy and @kalidiagne  on the PR for reviews!

SouravInsights:
Hey, @julien51 👋 

Sorry for taking such a long time but here's a draft **[PR](https://github.com/unlock-protocol/unlock/pull/10644)**. But I still need to make some improvements on this so thought of opening up a draft PR first so everybody can see the progress & provide any feedback.

Tagging @searchableguy and @kalidiagne so that you can have a look! 

I'll be out on a vacation for some days but will get back to this as soon as possible.

julien51:
Hey @SouravInsights any chance we can close the loop on this one please? We'd love to use it in https://github.com/unlock-protocol/unlock/pull/10937 for example!

SouravInsights:
> Hey @SouravInsights any chance we can close the loop on this one please? We'd love to use it in #10937 for example!

Hey, Julian. I'm waiting for final approval from @kalidiagne & @searchableguy

julien51:
Thanks for your work on this @SouravInsights !

Issue #10213, opened 2022-11-05
Handle ENS in the unlockjs


searchableguy:
In unlockjs, we have many methods which interact with user addresses. We should support also ens for these and automatically resolve the address for the consumer.

This can elevate issues on our frontend where we handle ENS explicitly.

julien51:
Also, we should be smarter when we do resolutions: we should only assume the resolved account address is correct on mainnet and if there is a match but we're looking at a different network, we should only accept the match if it is an EOA.

julien51:
As I think more about this, I am actually not sure this is the job of unlock-js. I worry that it adds a bit of complexity and coupling for things which should not be coupled. I think resolution should be dealt with "at a higher level"  and UnlockJS only handles proper addresses. I also don't like the idea of picking "winners". I can imagine people want to do resolution using Unstoppable Domains or others down the line...
I know ethers.js does ENS resolution but I am not sure this is a great idea either (and is explained by the shared story between ENS and ethers!)

julien51:
I will close in favor of https://github.com/unlock-protocol/unlock/issues/10398

Issue #10212, opened 2022-11-05
Handling referrer as ENS in Checkout URL builder


julien51:
We need to handle the cases where the referrer is an ENS. When the user enters an ENS, we should resolve it and use the address in the paywall config.

julien51:
Related https://github.com/unlock-protocol/unlock/issues/10213

julien51:
Closing in favor of the more generic https://github.com/unlock-protocol/unlock/issues/10398

PR #10644, opened 2022-12-23
feat(ui): Addressinput UI component

SouravInsights:
# Description

This PR adds a new component called `AddressInput`. The component is supposed to be used in any Input filled that deals with eth addresses like:
- `Referrer` setting on the Checkout Builder,
- `Recipient` or `Key Manager` on the Airdrop Keys form,
- `Verifier` or `Lock Managers` in the Roles settings 

# Issues
This PR is associated with these issues: [#10213](https://github.com/unlock-protocol/unlock/issues/10213), [#10212](https://github.com/unlock-protocol/unlock/issues/10212), [#10398](https://github.com/unlock-protocol/unlock/issues/10398), etc. 

## Refs:
The component now looks like this in the LocksManager component
CleanShot 2022-12-23 at 16 50 14@2x

# Checklist:


searchableguy:
Open a pull request with your username in cla. :D I will approve.

packages/unlock-js/src/web3Service.ts
```diff
@@ -939,4 +939,57 @@ export default class Web3Service extends UnlockService {
...
+   */
+  async resolveEns(ensName: string) {
+    const provider = this.providerForNetwork(1)
+    try {
+      const ensAddress = await provider.resolveName(ensName)
+      if (ensAddress !== null) {
+        const isValid = await this.isValidEOA(ensAddress)
+        if (isValid === true) {
+          return ensAddress
+        } else throw 'The ens address is not a valid EOA'
```
searchableguy, line 969: Throw an instance of Error instead of string. That way, someone can check if instance of error and also get other properties attached to an error. 

Same below.

packages/unlock-js/src/web3Service.ts
```diff
@@ -939,4 +939,57 @@ export default class Web3Service extends UnlockService {
...
+
+  /**
+   * Returns the resolved ens name that's a valid EOA
+   */
+  async lookupAddress(address: string) {
+    const provider = this.providerForNetwork(1)
+    try {
+      const ensName = await provider.lookupAddress(address)
+      const isValid = await this.isValidEOA(address)
+      if (ensName !== null) {
```
searchableguy, line 985: use !ensName unless it's a number or also check for undefined here since that could be falsely.

packages/unlock-js/src/web3Service.ts
```diff
@@ -939,4 +939,57 @@ export default class Web3Service extends UnlockService {
...
+        if (isValid === true) {
+          return ensName
+        } else throw 'The ens address is not a valid EOA'
+      } else
+        throw 'The name does not exist, or the forward lookup does not match.'
+    } catch (error) {
+      console.error(error)
```
searchableguy, line 992: I don't think we should log it. Throw it as an error so user can catch it

SouravInsights:
> Thank you for your pull request and welcome to Unlock! We require contributors to sign our [Contributor License Agreement](https://github.com/unlock-protocol/unlock/blob/master/CLA.txt), and we don't seem to have the users @SouravInsights on file. In order for us to review and merge your code, please open _another_ pull request with a single modification: your github username added to the the file `.clabot`. Thank you!

Sure, here's my pr: https://github.com/unlock-protocol/unlock/pull/10647

packages/unlock-js/src/web3Service.ts
SouravInsights: Cool! Do you recommend throwing it like this: `throw new Error('The ens name is not a valid EOA')`? @searchableguy

packages/unlock-js/src/web3Service.ts
SouravInsights: Are you sure we should rethrow the error from a catch block? @searchableguy

packages/unlock-js/src/web3Service.ts
searchableguy: Yes.

packages/unlock-js/src/web3Service.ts
searchableguy: I mean you can remove the catch block and simply throw error instead of catching it here.

SouravInsights: ok, so will totally remove the `try`...`catch` block and instead have something like this:

```
  async lookupAddress(address: string) {
    const provider = this.providerForNetwork(1)
    const ensName = await provider.lookupAddress(address)
    console.log('ensName:', ensName)
    console.log('!ensName:', !ensName)
    const isValid = await this.isValidEOA(address)
    console.log('isValid:', isValid)
    if (!ensName) {
      if (isValid) {
        return ensName
      } else throw new Error('The ens name is not a valid EOA')
    } else
      throw new Error(
        'The name does not exist, or the forward lookup does not match.'
      )
  }
```

packages/unlock-js/src/web3Service.ts
SouravInsights: Cool, thanks!

packages/unlock-js/src/web3Service.ts
SouravInsights: Does this look good? @searchableguy

julien51:
@SouravInsights any progress please?

SouravInsights:
Hey, Julien. 👋 Yes, I have addressed almost all the comments by @searchableguy

SouravInsights:
Will push these changes in a while.

SouravInsights:
I think it's almost ready.. Can you please test this out & provide some feedback? @searchableguy @julien51

searchableguy:
I may be missing something here but the component itself does not do any resolution automatically. I thought that was the core part of this new component. You have to explicitly call the resolution function outside the component and manage the state as well as handling of it in form since it does not pass resolved address to form register. 

@julien51 can you confirm the intent?

Another thing is simply an optimization that we had in getAddressForName. It checks if someone passes a valid ethereum address and does not do any resolution  (we don't need to since it is not ENS).

unlock-app/src/hooks/useEns.ts
```diff
@@ -31,13 +31,16 @@ export const getNameOrAddressForAddress = async (
 export const getAddressForName = async (_name: string): Promise<string> => {
   try {
     const name = _name.trim()
+    console.log('name:', name)
```
searchableguy, line 34: Remove logs?

julien51:
@SouravInsights as @searchableguy noted, you need to handle resolution *inside* the component...

SouravInsights:
Yeah I did thought about it earlier, thanks for surfacing it @searchableguy. Will do the necessary changes..

SouravInsights:
Hey, @searchableguy & @julien51 I've tried a different approach that includes a new form component, `AddressInputForm`. Let me know your thoughts.

## Using this component:

```
<AddressInputForm
  inputName="manager" // The name of the Input being registered
  label="Add manager, please enter the wallet address of theirs."
  description="Enter a wallet address or an ens name"
  onSubmit={() => onAddLockManager()} // Pass the onSubmit handler function
  web3Service={web3Service}
  form={form} // Pass all the useForm methods
  withIcon
/>
```

## Visual changes:

The original spec suggested that when a user enters an ENS, we should keep it visible but the value used in forms should be the 0x... address. But with this new implementation, the input component will always show the 0x address & the ENS will be shown below it.

The reason to do this way is, in order to access the `resolvedAddress` in the `LockManagerForm` component, I had to dynamically set the value of the registered field to reflect the **resolved address** using the `setValue` method. 

CleanShot 2023-01-14 at 04 17 10@2x

##

searchableguy:
You can use controlled component to pass back resolved address on change.

https://react-hook-form.com/api/usecontroller/controller/

You can achieve what is written in spec that way. 

Failed resolution should also result in a move to an error state on input level which this does not handle. Look at the Input component which does handle different states in thr Form folder.

julien51:
@SouravInsights any progress on this? We'd love to close the loop and pay the bounty here!

SouravInsights:
> @SouravInsights any progress on this? We'd love to close the loop and pay the bounty here!

I have just pushed a [commit](https://github.com/unlock-protocol/unlock/pull/10644/commits/996400ad3466d86a70a8ad12f3caee060fb54127) that works as per the spec but I'm stuck with handling the error state..

Also, there's a minor bug with implementation which I just noticed, will try to fix it tomorrow. But feel free to review this & let me know your feedback.

@julien51 @searchableguy

packages/ui/lib/components/Form/AddressInput.tsx
```diff
@@ -0,0 +1,179 @@
...
+                <span className="absolute inset-y-0 left-0 flex items-center pl-3 pointer-events-none">
+                  <Icon size={size} icon={CustomizedIcon} />
+                </span>
+              )}
+              <input
+                {...inputProps}
+                id={label}
+                className={inputClass}
+                {...register(name, {
+                  setValueAs: (value: string) => handleResolver(value),
```
searchableguy, line 147: Wouldn't a debounce be useful given user will likely be typing an ens and each hit will fire a request to resolve? Does this not cause an issue currently? 

Also setValueAs is run after all validations passes instead of before so one cannot add any kind of validation on the resolved value. Do you think you can use onChange which runs before any validation happens?

packages/ui/lib/utils.ts
```diff
@@ -6,3 +6,20 @@ export const minifyAddress = (address: string) => {
     ? `${address.slice(0, 5)}...${address.slice(address.length - 5)}`
     : address
 }
+
+export const validateName = (address: string) => {
+  const checked = ethers.utils.isAddress(address)
+  return checked
+    ? `${address.slice(0, 5)}...${address.slice(address.length - 5)}`
+    : address
+}
```
searchableguy, line 15: We already have a function for this. minifyAddress. Use that.

packages/ui/lib/components/Form/AddressInput.tsx
```diff
@@ -0,0 +1,179 @@
...
+      withIcon = true,
+      isTruncated,
+      web3Service,
+      localForm,
+      name,
+      ...inputProps
+    } = props
+
+    if (!localForm) return null;
+    const { register } = localForm
```
searchableguy, line 74: You can use useFormContext() and wrap the external component inside provider instead of passing the methods around.

SouravInsights: I don't think we can do it since `AddressInput` lives inside the UI which is a separate package, that's not aware of form context. I did it try it before and `useFormContext()` always returns `null`. Also, I don't see why passing the methods around is bad. People at RaidGuild also follow this approach where they pass the whole useForm hook from the parent form component. You can check [this](https://github.com/raid-guild/design-system/blob/main/src/components/forms/Input/Input.tsx#L42) out. @searchableguy

packages/ui/lib/components/Form/AddressInput.tsx
SouravInsights: I will try adding the debounce function but regarding the latter, the goal is to pass the `resolvedAddress` back to parent, and as far as I have tried, `setValueAs` is the only way I was able to do it. And to be honest I don't see any scenarios where anyone would want to run validation on the `resolvedAddress`. It's a very edge-case scenario which I think we can add later.

searchableguy: Hah. I actually do this with current places where ENS input is used which is why I explicitly mentioned it if this component were to replace them.

packages/ui/lib/components/Form/AddressInput.tsx
searchableguy: It was a nit :D 

And you need to wrap the parent component in form context provider to use the useFormHook. 

Either way is fine.

searchableguy: Sorry, I wasn't calling you anything. It is a nit as in It is not consequential and either way is fine.

SouravInsights: Bummer, I totally misunderstood. I'm sorry, I thought you referred lit to the person at RaidGuild who followed that particular approach.

julien51:
@SouravInsights Any progress? We'd love to close that as we will be using this component in a new feature! 
https://github.com/unlock-protocol/unlock/issues/10852

SouravInsights:
hey, @kalidiagne thinks it's working fine now, please try testing it out & let me know if there's any issues!

kalidiagne:
> hey, @kalidiagne thinks it's working fine now, please try testing it out & let me know if there's any issues!

@SouravInsights thanks! i give you feedback after i tested it

packages/ui/package.json
```diff
@@ -37,8 +37,9 @@
     "@tw-classed/react": "1.3.2",
     "@unlock-protocol/networks": "workspace:./packages/networks",
     "@unlock-protocol/unlock-assets": "workspace:./packages/unlock-assets",
-    "ethers": "5.7.2",
+    "ethers": "^5.7.2",
     "prism-react-renderer": "1.3.5",
+    "react-hook-form": "^7.41.5",
```
kalidiagne, line 42:
```suggestion
    "react-hook-form": "7.41.5",
```

packages/ui/package.json
```diff
@@ -37,8 +37,9 @@
     "@tw-classed/react": "1.3.2",
     "@unlock-protocol/networks": "workspace:./packages/networks",
     "@unlock-protocol/unlock-assets": "workspace:./packages/unlock-assets",
-    "ethers": "5.7.2",
+    "ethers": "^5.7.2",
```
kalidiagne, line 40:
```suggestion
    "ethers": "5.7.2",
```

unlock-app/src/components/interface/locks/Settings/forms/LockManagerForm.tsx
```diff
@@ -304,10 +318,14 @@ export const LockManagerForm = ({
...
-              Add manager, please enter the wallet address of theirs.
-            </span>
-            <Input disabled={disableInput} {...register('manager')} />
+            <AddressInput
+              withIcon
+              name="manager"
+              label="Add manager, please enter the wallet address of theirs."
+              description="Enter a wallet address or an ens name"
+              web3Service={web3Service}
+              localForm={localForm}
```
kalidiagne, line 327: - We can manage `web3Service ` inside the component to make sure we use the correct web3service and is specific  `await web3Service.resolveName(address)`, because now the type is any, potentially fail because is not a required prop.
- same for `localForm` maybe
@searchableguy @julien51  wdyt?

packages/ui/lib/components/Form/AddressInput.stories.tsx
```diff
@@ -0,0 +1,60 @@
+import { AddressInput } from './AddressInput'
```
kalidiagne, line 1: can test this component by running `yarn storybook`, some props are missing

kalidiagne:
@SouravInsights added some comment

unlock-app/src/components/interface/locks/Settings/forms/LockManagerForm.tsx
SouravInsights: Oh nice catch, I can make the prop required. I initially thought of defining `web3Service` inside the UI package but I don't think it's that necessary. Cause other than the `AddressInput` component, no other component will be using it. So I think it will be much easier if we pass it down as a prop. @kalidiagne

SouravInsights: Regarding the `localForm`, @searchableguy did suggest using `useFormContext` but for some reason, I couldn’t get it to work. Maybe because our UI package lives in a separate repo, that's why react-hook-form is not aware of the `context`, not sure... 

But as I said earlier, I have seen people follow the same approach, passing down the form hook as a prop. For example, you can check [this](https://github.com/raid-guild/design-system/blob/main/src/components/forms/Input/Input.tsx#L42) out. @kalidiagne

SouravInsights: Hey, @kalidiagne @searchableguy @julien51 Are we good to move forward with this?

julien51:
Also, please rebase if you get a chance!

SouravInsights:
> Also, please rebase if you get a chance!

Sure, sure! Will do..

SouravInsights:
There are too many conflicts on my side after fetching the upstream so I'm thinking I'll close this PR & open a new one. @julien51

PR #10652, opened 2022-12-28
feat(wedlocks) - replace jest with vitest

kalidiagne:
# Description
Screenshot 2022-12-28 at 14 06 32

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/wedlocks/package.json
+++ b/wedlocks/package.json
@@ -43,8 +43,7 @@
   "scripts": {
     "build": "netlify build",
     "dev": "UNLOCK_ENV=dev netlify functions:serve",
-    "test": "UNLOCK_ENV=test jest",
-    "testvi": "UNLOCK_ENV=test vitest run --no-threads --coverage --environment=node",
+    "test": "UNLOCK_ENV=test vitest run --no-threads --coverage --environment=node",
     "lint": "eslint --resolve-plugins-relative-to ../packages/eslint-config --ext .ts,.js src/",
     "ci": "yarn test && yarn lint",
     "deploy-netlify": "./scripts/deploy-netlify.sh"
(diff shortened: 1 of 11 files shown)


wedlocks/src/__tests__/encrypter.test.js
```diff
@@ -1,26 +1,41 @@
...
-let publicKey
+beforeAll(() => {
+  vi.useFakeTimers()
+})
 
-// These tests are slow because we generate private keys
-jest.setTimeout(15000)
+afterAll(() => {
+  vi.clearAllTimers()
+})
```
searchableguy, line 11: Not sure we need this as well as below declaration.  I don't see any use of fake timer in the test. Could you please explain?

wedlocks/src/__tests__/handler.test.js
```diff
@@ -16,12 +18,14 @@ describe('handler', () => {
           throw error
         }
         expect(response.statusCode).toBe(204)
-        done()
+        new Promise((done) => {
+          done()
+        })
```
searchableguy, line 23: Instead of this. Use promise resolve to resolve the response value and capture it using variable.

This way, you can remove the expect and assertions outside this callback.

wedlocks/package.json
```diff
@@ -43,7 +32,7 @@
   "scripts": {
     "build": "netlify build",
     "dev": "UNLOCK_ENV=dev netlify functions:serve",
-    "test": "UNLOCK_ENV=test jest",
+    "test": "UNLOCK_ENV=test vitest run --no-threads --coverage --environment=node",
```
searchableguy, line 35: You can remove --no-threads flag. I added it in locksmith because our tests weren't written correctly for parallelism. I think it would work here. If not, add it back.

searchableguy (approved):
A few nits then LGTM

searchableguy:
> A few changes then LGTM

wedlocks/src/__tests__/encrypter.test.js
kalidiagne: you are right, not needed!

searchableguy (approved):

PR #10708, opened 2023-01-03
fix(contracts): update UnlockV11 interface

### LINKED ISSUE #10654: `@unlock-protocol/contracts` `IUnlock` interfaces do not have proper return types
https://github.com/unlock-protocol/unlock/blob/ef24d9cb9447df896fa1838d891c749f23d1405c/packages/contracts/src/contracts/Unlock/IUnlockV11.sol#L183-L187

`publicLockVersions`,  `publicLockImpls` in `IUnlockV11` doesn't have return types causing issues when using those methods

_Originally posted by @sebastiantf in https://github.com/unlock-protocol/unlock/issues/9877#issuecomment-1366758727_

clemsos:
# Description

Fix signature issues and improve formatting in the `UnlockV11` distributed by `@unlock-protocol/contracts`

# Issues

Fixes #10654
Refs #

# Checklist:


packages/contracts/src/contracts/Unlock/IUnlockV11.sol
```diff
@@ -1,90 +1,94 @@
...
 
-
 /**
  * @title The Unlock Interface
-**/
+ **/
 
-interface IUnlockV11
-{
+interface IUnlock {
```
julien51, line 8: Should we alias this? so people can use either?

julien51: I guess what's the way to include this from the package?

julien51 (approved):

packages/contracts/src/contracts/Unlock/IUnlockV11.sol
clemsos: yes better to keep the version here. We have an open issue for adding aliases to the package itself

PR #10717, opened 2023-01-05
feat(smart-contracts): use unlock provider to run local forks

clemsos:
# Description

Remove needs for ALchemy API key when running local fork

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/README.md
+++ b/smart-contracts/README.md
@@ -31,15 +31,12 @@ npx hardhat test --logs
 
 ### Run a mainnet fork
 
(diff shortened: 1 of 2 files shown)


julien51 (approved):
Yay!

julien51 (approved):
Technically we can fork _any_ network though, right? I think it's ok for this PR but something like using an Environment variable like `FORK_NETWORK=<networkId>` could be useful!

clemsos:
> Technically we can fork _any_ network though, right? I think it's ok for this PR but something like using an Environment variable like `FORK_NETWORK=<networkId>` could be useful!

yes I was thinking about the same. Can do in another PR

PR #10726, opened 2023-01-06
feat(smart-contracts): remove UDT mint on mainnet purchases

### LINKED ISSUE #10706: Stop minting from mainnet
The protocol behaves differently on mainnet and on other networks. On mainnet new UDT are minted while on other networks they are distributed.
Let's change this so that tokens are also distributed from the premine on mainnet.

As part of this change, we will need to redeploy the mainnet contract. This means that we need to include a mechanism in the `PublicLock` contract (and in its new version) so that when they are upgraded, they "point" to the new Unlock contract. Additionnaly the new Unlock needs to "accept" locks deployed from the older contract as "valid" locks.

clemsos:
# Description

Use the premint on mainnet to distribute tokens instead of minting new tokens

# Issues

Fixes #10706
Refs #

# Checklist:


smart-contracts/contracts/Unlock.sol
```diff
@@ -452,24 +452,13 @@ contract Unlock is UnlockInitializable, UnlockOwnable {
...
-              2 /
-              grossNetworkProduct;
-          }
+          // we distribute tokens using asymptotic curve between 0 and 0.5
+          // maxTokens = IMintableERC20(udt).balanceOf(address(this)).mul((valueInETH / grossNetworkProduct) / (2 + 2 * valueInETH / grossNetworkProduct));
+          uint maxTokens = (IMintableERC20(udt).balanceOf(
+              address(this)
+            ) * valueInETH) /
+            (2 + (2 * valueInETH) / grossNetworkProduct) /
+            grossNetworkProduct;
```
julien51, line 461: I wonder if there is a way to simplify this equation?

clemsos: that may mean change how UDT distribution works . Should write a proposal about tokenomics first as it is a change in the protocol itself?

julien51: No no, I want to keep the same formula but have the code be more legible!

clemsos: Ok. A simple thing is to reuse the `balance` variable below. We are hitting "stack too deep" so can not really add more named vars :/

julien51 (approved):

PR #10743, opened 2023-01-10
feature(smart-contracts): adding KeyManager contract

julien51:
# Description

To support wallet-less airdrops, we are introducing a new contract. It is not strictly part of the core protocol.
This contract will be deployed on all networks and be the `keyManager` for all airdropped keys. 
This contract then supports a single function `transfer` that can be called by anyone to transfer a key to themselves as long as locksmith (the signer) approves this transfer.

To make things more robust, the transfer is not strictly a transfer but a `lend`, which means that the `keyManager` contract will remain `keyManager` for the keys. We could eventually add a way to perform "full" transfers if the need arises.

# Checklist:


smart-contracts/test/KeyManager/transfer.js
```diff
@@ -0,0 +1,77 @@
...
+      owner: grantee,
+      deadline: OneHourFromNow,
+    };
+    const attackerSigner = await ethers.getSigner(attacker)
+    const signature = await attackerSigner._signTypedData(domain, types, transfer);
+    await reverts(keyManager.transfer(transfer.lock, transfer.token, transfer.owner, transfer.deadline, signature), `VM Exception while processing transaction: reverted with custom error 'NOT_AUTHORIZED()'`)
+  })
+
+
+  it('should lend the key to the sender if the signature matches', async () => {
```
julien51, line 62: @clemsos This test fails :/ For some reason the signature does not recover the same account in JS and in Solidity :/ any idea on how to fix it?

smart-contracts/test/KeyManager/upgrades.js
```diff
@@ -0,0 +1,9 @@
+contract('KeyManager', () => {
+
+  beforeEach(async () => {
```
julien51, line 3: Any suggestion on testing this @clemsos ?

searchableguy:
Could we have a getter to detect if transfer code has a key?

smart-contracts/test/KeyManager/transfer.js
julien51: It does not fail anymore!

julien51:
> Could we have a getter to detect if transfer code has a key? Maybe isTransferCodeValid or similar which detects a key exist for the following code to avoid people sending transactions and easier to query status on the UI.

Tell me more! I think the best way is to run estimates on te function itself?

smart-contracts/contracts/KeyManager.sol
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    return IPublicLockForKeyManager(lock).lendKey(owner, msg.sender, token);
+  }
+
+  /**
+   * Function to change the signer. This can only be called by the owner.
+   */
+  function setLocksmith(address _locksmith) public onlyOwner {
+    locksmith = _locksmith;
```
clemsos, line 76: Could also set locksmith in initializer to avoid a 2nd tx

smart-contracts/test/KeyManager/upgrades.js
clemsos: Why make it upgradeable in the first place? Should be easy enough to just redeploy a new one as there is no storage (except locksmith address).

smart-contracts/contracts/KeyManager.sol
julien51: Ha! I like. I'll do that.

julien51:
Deployed on Goerli at https://goerli.etherscan.io/address/0x862a839fcdba499bbfd6325ba1845787b4b44e61

smart-contracts/scripts/deployments/keyManager.js
```diff
@@ -0,0 +1,49 @@
...
+    console.log(`> Added ${locksmith} as signer on KeyManager`)
+  }
+
+  // Transfer ownership to multisig (based on networks package)
+  if (networks[chainId.toString()].multisig) {
+    await keyManager.transferOwnership(networks[chainId.toString()].multisig)
+    console.log(`> Transfered ownership of KeyManager to ${networks[chainId.toString()].multisig}`)
+  }
+
+  // TODO: transfer proxy admin to multisig (based on networks package?)
```
julien51, line 32: @clemsos Any idea how to do that??

clemsos: yes we have a script for that already https://github.com/unlock-protocol/unlock/blob/master/smart-contracts/scripts/multisig/transferOwnership.js

Also you can use the CLI directly `yarn hardhat gnosis:transfer` if that helps.

clemsos: proxyAdmin is here, you need ot transfer ownership  https://goerli.etherscan.io/address/0x604fa8712e61fe65db326f97905efcb087c631cc#writeContract

smart-contracts/test/KeyManager/upgrades.js
clemsos: There is a possible example with the `TestPublicLockUpgraded` contract. Basically you have to import yout own contract into a new one like:

```solidity
import './KeyManager.sol';

contract KeyManagerV2 is KeyManager {
  function someNewFeature() {... }
}
```

then you can test with sth like that :

```js
const { ethers, upgrades } = require('hardhat')

const KeyManager = await ethers.getContractFactory('KeyManager')
const proxy = await upgrades.deployProxy(KeyManager, args)

const KeyManagerV2 = await ethers.getContractFactory('KeyManagerV2')

await upgrades.upgradeProxy(proxy.address, KeyManagerV2)
```

Here is an example with a similar pattern https://github.com/unlock-protocol/unlock/blob/0d8e0a5d93513c5d6a52a7f5f5340a0aef603e33/smart-contracts/test/Lock/upgrades/v11.js#L91

julien51: Perfect! Thanks a lot!

julien51: Done!

smart-contracts/scripts/deployments/keyManager.js
julien51: I added a few things to the script.

searchableguy (approved):

PR #10744, opened 2023-01-11
fix(subgraph): another round of subgraph fixes

julien51:
# Description

I realized the way we initialize daily data is wrong because it resets to 0 if there is no lock deployed on a given day (we initialized with the values from the day before).

# Checklist:


julien51:
I dont understand what is going on with the integration tests here :(

PR #10747, opened 2023-01-11
feat(unlock-js): Add key manager to unlock js

searchableguy:
# Description

Will replace with proper key manager address on networks once deployed and remove it for those we aren't deploying.

# Issues

Fixes #
Refs #

# Checklist:


packages/networks/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@unlock-protocol/networks",
-    "version": "0.0.8",
+    "version": "0.0.9",
```
julien51, line 3: Any reason for the bump?

julien51: Oh dang. Of course. It's because there's a new property!

packages/unlock-js/src/KeyManager.ts
```diff
@@ -0,0 +1,167 @@
...
+      deadline,
+      transferSignature
+    )
+    return tx
+  }
+
+  /**
+   * This function sets the locksmith address.
+   */
+  async setLocksmith({
```
julien51, line 123: I'm not sure at need this in the Unlockjs library? 
We will probably not do that though code...

packages/unlock-js/src/KeyManager.ts
```diff
@@ -0,0 +1,167 @@
...
+    return tx
+  }
+
+  /**
+   * This function creates a wallet address from a lock address and an email address.
+   */
+  createTransferAddressWallet({
+    params: { lockAddress, email },
+  }: CreateTransferAddressKey) {
+    const privateKey = ethers.utils.keccak256(
```
julien51, line 158: No no let's not do that because anyone could recreate the private key and impersonate the key owner.
Lets just do a hash and format as address

julien51 (changes requested):
Generally, in not 100% sure we need this video to be in unlockJs and i worry about increasing the size/surface area but i can see how it is useful to share some code between locksmith and unlock app so let's keep it here (but still remove the admin functions like set locksmith)

packages/unlock-js/src/KeyManager.ts
searchableguy: ok. Created hash using id and truncated to 42 characters to work as an address.

julien51 (approved):

julien51 (approved):

packages/unlock-js/src/KeyManager.ts
```diff
@@ -0,0 +1,158 @@
...
+  }: TransferOptions) {
+    const KeyManagerContract = this.getContract({ network, signer })
+
+    const tx = await KeyManagerContract.transfer(
+      lock,
+      token,
+      owner,
+      deadline,
+      transferSignature
+    )
```
searchableguy, line 129: I seem to get gas errors here on goerli. Anything stands out? @julien51

PR #10748, opened 2023-01-11
feature(provider): logging, tests and refactor

julien51:
# Description

Addinglogging to sentry + basic tests and refactored a bit.

# Checklist:


provider/src/index.ts
```diff
@@ -27,120 +9,22 @@ export default {
...
-    // Reject requests that are not POST
-    if (request.method !== 'POST') {
-      return new Response(JSON.stringify({ message: 'Method not supported' }), {
-        status: 400,
-        headers: {
-          'content-type': 'application/json',
-        },
-      })
-    }
+    context.passThroughOnException()
```
julien51, line 14: This should pass the errors down to the front-end apps.

provider/src/supportedNetworks.ts
```diff
@@ -0,0 +1,31 @@
+import { Env } from './types'
+
+// This is the list of networks currently supported
+const supportedNetworks = (env: Env, networkId: string) => {
```
julien51, line 4: Moving to env vars (already configured in Cloudflare) to avoid exposing them publicly.

searchableguy: Wouldn't it be easier to get rid of the switch statement and simply use object then return object[id]?

searchableguy (approved):
LGTM

provider/src/supportedNetworks.ts
julien51: Ha, we need to pass `env` though :/

searchableguy: That would be fine, right?

```ts
const object = { 
  "1": env.MAINNET, 
  "5": env.GOERLI 
}

return object[id]
```

inside the function 

It's just a little easier to read than a switch case maze.

julien51: Ha! Sorry i had misunderstood. Yes. That works too. I'll fix it.

PR #10762, opened 2023-01-16
feat(locksmith): Add create transfer code endpoint

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/fulfillment/dispatcher.ts
```diff
@@ -14,7 +18,7 @@ interface KeyToGrant {
 export default class Dispatcher {
   async getPurchaser(network: number) {
     const provider = new ethers.providers.JsonRpcProvider(
-      networks[network].publicProvider
+      networks[network].provider
```
julien51, line 21: So, when sending transactions, it's better to use the `publicProvider` recommended by the networks themselves.
In the past we have seen issues where some of our transactions were queued because they were sent from Alchemy.

searchableguy: oh, ok. I thought public ones get rate limited or have lower throughout so I changed to using paid ones. Will revert.

locksmith/src/routes/v2/airdrop.ts
```diff
@@ -0,0 +1,23 @@
+import express from 'express'
+import { createTransferCode } from '../../controllers/AirdropController'
+import { createRateLimitMiddleware } from '../../utils/middlewares/rateLimit'
+import { captchaMiddleware } from '../../utils/middlewares/recaptchaMiddleware'
+import { authenticatedMiddleware } from '../../utils/middlewares/auth'
+const router = express.Router({ mergeParams: true })
+
+const rateLimiter = createRateLimitMiddleware({
```
julien51, line 8: Is this global?

locksmith/src/controllers/AirdropController.ts
```diff
@@ -0,0 +1,78 @@
...
+    )
+  })
+
+  if (!email) {
+    return response.status(404).send({
+      message: 'No email address found for this user',
+    })
+  }
+
+  const validPeriod = 86400 // 1 day in seconds
```
julien51, line 36: Let's reduce this to 15 minutes please!

julien51, line 36:
```suggestion
  const validPeriod = 60 * 60 * 15 // 15 minutes
```

locksmith/src/controllers/AirdropController.ts
```diff
@@ -0,0 +1,78 @@
...
+  transferUrl.searchParams.set('transferCode', transferCode)
+  transferUrl.searchParams.set('network', network.toString())
+
+  await sendEmail('transferCode', 'transferCode', email, {
+    lockName: key.lock.name || 'Unlock Lock',
+    network: network.toString(),
+    transferCode,
+    keyId,
+    transferUrl: transferUrl.toString(),
+    validPeriod: '24 hours',
```
julien51, line 72:
```suggestion
    validPeriod: '15 minutes',
```

locksmith/src/fulfillment/dispatcher.ts
julien51: yeah when "reading" we want to use ours when when sending tx we want to use the public ones.

locksmith/src/routes/index.ts
```diff
@@ -67,6 +67,7 @@ router.use('/v2/applications', applicationRouter)
 router.use('/v2/purchase', purchaseRouterV2)
 router.use('/v2/subscriptions', subscriptionRouter)
 router.use('/v2/claim', claimV2Router)
+router.use('/v2/airdrop', airdropRouter)
```
julien51, line 70: Let's add a captcha on that please :)

julien51 (commented):
That's a start!

locksmith/src/routes/index.ts
searchableguy: Already added at router level, no?

locksmith/src/routes/v2/airdrop.ts
searchableguy: No. It's specific to this endpoint.

julien51: ho I meant, is this for every user?

locksmith/src/routes/index.ts
julien51: good!

locksmith/src/controllers/AirdropController.ts
```diff
@@ -0,0 +1,79 @@
...
+  transferUrl.searchParams.set('transferCode', transferCode)
+  transferUrl.searchParams.set('network', network.toString())
+  transferUrl.searchParams.set('transfer', JSON.stringify(transfer))
+
+  await sendEmail('transferCode', 'transferCode', email, {
+    lockName: key.lock.name || 'Unlock Lock',
+    network: network.toString(),
+    transferCode,
+    keyId,
+    transferUrl: transferUrl.toString(),
```
julien51, line 74: Not sure we need to pass this... (or build it) per our earlier convo.

locksmith/src/controllers/AirdropController.ts
```diff
@@ -0,0 +1,79 @@
...
+  const owner = normalizer.ethereumAddress(key.owner)
+
+  const transfer = {
+    owner,
+    lock: lockAddress,
+    token: keyId,
+    deadline: deadline,
+  }
+
+  const transferCode = await dispatcher.createTransferCode(network, transfer)
```
julien51, line 60: We may want to "reduce" the size of this string by using a Base64 number instead of the hex one?

julien51 (approved):
A few changes because we don't need to pass the URL and ideally we "reduce" the size of the transferCode displayed as string by using the base64 representation of the number (not of the string!)

locksmith/src/routes/v2/airdrop.ts
searchableguy: No. It is per IP.

julien51: Perfect!

locksmith/src/controllers/AirdropController.ts
searchableguy: As communicated. The difference isn't huge and not worth the added hoop. I look for something which could be much smaller. 

For approximate reference:

Current: 140
Base36: 110
Base64: 90

Base64 also contain characters like `=/` . It's not something you can type. You will need to copy paste.

locksmith/src/controllers/AirdropController.ts
searchableguy: I think I explained the reasoning for having it as fallback in case the user closes the tab accidentally or want to open it from a different device if they have email logged in there. 

I can still remove it but I think it's useful to have a link you can quickly open.

wedlocks/src/templates/transferCode.js
```diff
@@ -1,11 +1,19 @@
...
-  html: `<h1>A transfer code for a NFT membership was generated for you to claim!</h1>
+  subject: 'Your Transfer Code is Here',
+  html: `
+<h1> Your Transfer Code is Here </h1>
 
-<p>A transfer code for NFT #{{keyId}} to the lock <strong>{{lockName}}</strong> was just generated for you!</p>. It is valid for {{validPeriod}}.
+<p> Your transfer code for {{lockName}} NFT has been successfully generated and is ready for you to use. This code will enable you to transfer your NFT for a limited period of {{validPeriod}}. </p>
 
-Transfer code: {{transferCode}}
+<p> Your transfer code is <code>{{transferCode}}</code> and it will expire in 15 minutes from now. </p>
```
julien51, line 8:
```suggestion
<p> Your transfer code is <code>{{transferCode}}</code>. </p>
```

PR #10764, opened 2023-01-17
feat(locksmith) - receipts controllers

kalidiagne:
# Description

- [x] add `receiptsController`/`receiptsBaseController`
- [x] add `save`/`get` functions to retrieve and get receipts for specific controllers 
- [x] add `lockAddress`, `network` and `hash` attributes to models  

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/models/receiptsBase.ts
```diff
@@ -21,6 +22,11 @@ export class ReceiptsBase extends Model<
   declare state: string
   declare country: string
 
+  // lock & transaction details
+  declare network: number
+  declare lockAddress: string
```
julien51, line 27: We may not need the lockAdress because we can look up by transaction but I think it;s ok!

locksmith/src/controllers/v2/receiptsBaseController.ts
```diff
@@ -0,0 +1,145 @@
...
+        servicePerformed = '',
+        addressLine1 = '',
+        addressLine2 = '',
+        state = '',
+        city = '',
+        zip = '',
+        country = '',
+      } = (request.body as ReceiptsBase) ?? {}
+
+      if (!supplier.length) {
```
julien51, line 49: For now let's make all the fields as optional please.

julien51 (changes requested):
Can we please use zod to stay consistent with the rest of the application?

locksmith/src/models/receiptsBase.ts
```diff
@@ -10,7 +10,8 @@ export class ReceiptsBase extends Model<
   InferAttributes<ReceiptsBase>,
   InferCreationAttributes<ReceiptsBase>
 > {
-  declare id: number
+  // receipts details
+  declare hash: string
```
kalidiagne, line 14: @julien51  can we use only `lockAddress` here to get supplier details?

locksmith/src/controllers/v2/receiptBaseController.ts
```diff
@@ -0,0 +1,122 @@
...
+          const newSupplier = await this.createSupplier(props, {
+            lockAddress,
+            network: `${network}`,
+          })
+          return response.status(201).json({
+            ...newSupplier?.dataValues,
+          })
+        } else {
+          // update the existing details
+          const [{ dataValues }] = await ReceiptBase.upsert(
```
julien51, line 74: If you use `upsert` there is no need to distinguish the 2 cases because `upsert` means update OR insert

julien51 (changes requested):
Please refactor. Since you are using `upsert` there is no need to check if the objects exist first.
So either you keep `upsert` and then remove the `create` logic, or change `upsert` in `update`

julien51 (approved):
much cleaner!

PR #10799, opened 2023-01-23
fix(unlock-app): adding a better error message when password is not correct

julien51:
# Description

Instead of showing `Nan` when the transaction price could not be determined, we show a better error message.
When password is set, we assume it is because password is not correct.

Screenshot 2023-01-23 at 1 59 33 PM

# Checklist:


searchableguy (approved):

julien51:
Thanks! Hotfixed it!

PR #10808, opened 2023-01-24
refactor(locksmith): Optimize the performance on ticket endpoints

searchableguy:
# Description

Reduces time in some cases by 3 seconds. Down to 1.5-2 second per generation.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/operations/metadataOperations.ts
+++ b/locksmith/src/operations/metadataOperations.ts
@@ -7,10 +7,8 @@ import networks from '@unlock-protocol/networks'
 import { Verifier } from '../models/verifier'
 import Normalizer from '../utils/normalizer'
 import * as lockOperations from './lockOperations'
-import * as Asset from '../utils/assets'
 import { Attribute } from '../types'
 import metadata from '../config/metadata'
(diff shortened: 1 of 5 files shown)


locksmith/src/operations/metadataOperations.ts
```diff
@@ -7,10 +7,8 @@ import networks from '@unlock-protocol/networks'
 import { Verifier } from '../models/verifier'
 import Normalizer from '../utils/normalizer'
 import * as lockOperations from './lockOperations'
-import * as Asset from '../utils/assets'
 import { Attribute } from '../types'
 import metadata from '../config/metadata'
-const baseURIFragement = 'https://assets.unlock-protocol.com'
```
searchableguy, line 13: So the endpoint does not work and all the methods which use these to generate an some kind of base metadata endpoint do not work either so I removed it. Was this intended to be aws s3?

julien51: No this is legacy stuff... ok to remove!

julien51 (approved):

PR #10809, opened 2023-01-24
feat(locksmith): Add endpoint to transfer all data associated with email recipient

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/fulfillment/dispatcher.ts
+++ b/locksmith/src/fulfillment/dispatcher.ts
@@ -16,10 +16,13 @@ interface KeyToGrant {
   expiration?: number
 }
 export default class Dispatcher {
-  async getPurchaser(network: number) {
-    const provider = new ethers.providers.JsonRpcProvider(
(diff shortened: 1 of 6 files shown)


locksmith/src/fulfillment/dispatcher.ts
```diff
@@ -286,4 +273,25 @@ export default class Dispatcher {
     })
     return transferCode
   }
+
+  isTransferSignedByLocksmith(
```
julien51, line 277: I think we could make this more generic? `params` could be anything (not just a transfer code)

julien51 (approved):
LGTM

locksmith/src/fulfillment/dispatcher.ts
searchableguy: For this, I don't think it can be any more generic. You need also need domain and types for the typed signature derivation which adds a couple more lines.

Issue #10790, opened 2023-01-20
Create event landing page


julien51:
If events details are entered in the metadata field, let's automagically create an "event" landing page on which there is a checkout button to buy the ticket.
We should add a little event description as well.

PR #10823, opened 2023-01-26
feature(unlock-app): Adding an event landing page 

julien51:
# Description

When the event metadata is configured, we add a landing page for the event.
This is an MVP!
Still missing:
* [x] adding actual checkout
* [x] add to calendar button
* [x] add to twitter button

More:
* [ ] landing page for `/event` that shows some "marketing" details about how to do that!

# Issues

Refs #10790

# Checklist:


julien51:
Screenshot 2023-01-26 at 5 22 49 PM

unlock-app/src/index.css
```diff
@@ -5,3 +5,22 @@
...
+
+.markdown ul {
+  margin-top: 1em;
+  margin-bottom: 1em;
+  list-style: disc outside none;
+}
+
+.markdown ul li {
+  margin-left: 2em;
+  display: list-item;
```
searchableguy, line 25: You can actually use tailwind classes with `@apply`

unlock-app/src/components/content/EventContent.tsx
```diff
@@ -0,0 +1,201 @@
...
+              </button>
+            </Tooltip>
+          </li>
+        </ul>
+      </section>
+      <section className="mb-8">
+        {account && (
+          <Button
+            className="mt-8 w-full"
+            style={{ backgroundColor: eventData.accentColor }}
```
searchableguy, line 191: Also need to set appropriate text color otherwise this would result in an inaccessible button text (eg, accent color is white and text is also white)

unlock-app/package.json
```diff
@@ -52,6 +52,7 @@
     "react-icons": "4.7.1",
     "react-jazzicon": "1.0.4",
     "react-jwt": "1.1.7",
+    "react-markdown": "^8.0.5",
```
searchableguy, line 55:
```suggestion
    "react-markdown": "8.0.5",
```

unlock-app/src/components/content/EventContent.tsx
```diff
@@ -0,0 +1,201 @@
...
+import LoadingIcon from '../interface/Loading'
+import { AddressLink } from '../interface/AddressLink'
+import { ReactMarkdown } from 'react-markdown/lib/react-markdown'
+import { FaCalendar, FaRegCalendarPlus, FaClock } from 'react-icons/fa'
+import { GoLocation } from 'react-icons/go'
+import { FiTwitter } from 'react-icons/fi'
+import Link from 'next/link'
+import { useAuth } from '~/contexts/AuthenticationContext'
+
+const formatEventData = (metadata: any) => {
```
searchableguy, line 17: Don't we have a function to format data into a standard form we can use already?

I think we can reuse it or make it more reusable if it isn't.

unlock-app/src/index.css
julien51: Please, tell me more!

unlock-app/src/components/content/EventContent.tsx
```diff
@@ -0,0 +1,201 @@
...
+    return <LoadingIcon></LoadingIcon>
+  }
+
+  return (
+    <AppLayout showLinks={false} authRequired={false} title="">
+      <Head>
+        <title>{pageTitle('Event')}</title>
+      </Head>
+      <div className="md:w-3/4 m-auto">
+        {(!router.query.lockAddress || !router.query.network) && (
```
searchableguy, line 64: More readable to extract network and lock address at the top.

unlock-app/src/components/content/EventContent.tsx
```diff
@@ -0,0 +1,201 @@
...
+            </Link>
+          </li>
+        </ul>
+        <div className="markdown">
+          {/* eslint-disable-next-line react/no-children-prop */}
+          <ReactMarkdown children={eventData.description}></ReactMarkdown>
+        </div>
+      </section>
+      <section className="flex flex flex-col items-center">
+        <img alt={eventData.title} className="mb-5" src={eventData.image} />
```
searchableguy, line 155: Add aspect ratio, overflow, cover, and max size classes to avoid any overflow issues.

unlock-app/src/components/content/EventContent.tsx
julien51: Do you know of a lib that we would use and would generate the highest contrast for any color (between black and white) since users can select any accent color?

unlock-app/package.json
julien51: oupsie! Thx

unlock-app/src/components/content/EventContent.tsx
julien51: Ho Maybe we do. Good catch! I;ll find it!

unlock-app/src/components/content/EventContent.tsx
julien51: good point

unlock-app/src/components/content/EventContent.tsx
julien51: The problem is that this is the NFT image and we don't actually have these details :/

searchableguy: We don't require them. I'm only asking to add some defensive styling to avoid any responsiveness issues, not actually setting any size/width/height explicitly.

unlock-app/src/index.css
searchableguy: .markdown li {
@apply mb-2
}

So can create composeable classes like this. Mostly to remain consistent and reduce the lines of code.

searchableguy: Also have you tried using "prose prose-md". This basically handles base typography for markdown. We use it on our static website for blog articles which are in markdown.

julien51: Let me look!

unlock-app/src/components/content/EventContent.tsx
julien51: HA! Ok let me add that!

unlock-app/src/components/content/EventContent.tsx
searchableguy: https://www.npmjs.com/package/font-color-contrast

unlock-app/src/index.css
julien51: I could not make `@apply` work :/ Also looked at prose and prose-md but was confused :(

unlock-app/src/components/content/EventContent.tsx
julien51: Ho nice!

unlock-app/src/components/content/EventContent.tsx
julien51: I could not find this anywhere else!

unlock-app/src/components/content/EventContent.tsx
julien51: Done!

julien51:
ok @searchableguy @kalidiagne I think this is ready for another review. PTAL!

unlock-app/src/components/content/EventContent.tsx
searchableguy: https://github.com/unlock-protocol/unlock/blob/master/unlock-app/src/components/interface/locks/metadata/utils.ts

unlock-app/src/components/content/event/TweetItButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+      undefined,
+      {
+        weekday: 'long',
+        year: 'numeric',
+        month: 'long',
+        day: 'numeric',
+      }
+    )}. \n\nGet your ticket with @unlockProtocol!`
+  )
+  tweetIntent.searchParams.set('url', window.location.toString())
```
searchableguy, line 28: nit: refactor into a function and move outside the component.

unlock-app/src/components/content/event/AddToCalendarButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+              <SiYahoo className="w-8 h-8" />
+            </Link>
+            <Link
+              target="_blank"
+              className="p-4 hover:bg-blue-200 bg-blue-100 rounded border"
+              href={ics(calendarEvent)}
+            >
+              <BsCalendarDate className="w-8 h-8" />
+            </Link>
+          </div>
```
searchableguy, line 81: maybe not for this PR but I will add name of the respective calendar service in the button and make it one per line instead of small icons. This would be much more mobile friendly. 

I think we can safely remove yahoo one since it has very small user base and it adds to noise.

unlock-app/src/components/content/EventContent.tsx
julien51: Thx! Fixed!

PR #10825, opened 2023-01-27
feat(subgraph) - Add more details for receipts

kalidiagne:
# Description
We need `keyId` and `owner` for receipts to be able to get all receipts for a specific `keyId` of an owner 

# Issues

Fixes #
Refs #

# Checklist:


subgraph/src/public-lock.ts
```diff
@@ -239,7 +238,11 @@ export function handleKeyExtended(event: KeyExtendedEvent): void {
     key.save()
   }
   // create receipt
-  createReceipt(event)
+  if (key && key.owner) {
+    createReceipt(event, key.owner, event.params.tokenId)
+  } else {
+    createReceipt(event, Address.fromString(''), event.params.tokenId)
```
clemsos, line 244: not sure when you end up with an empty key owner on renewal? Also this `Address.fromString('')` gives a null address?

subgraph/src/public-lock.ts
```diff
@@ -269,7 +272,11 @@ export function handleRenewKeyPurchase(event: RenewKeyPurchaseEvent): void {
   }
 
   // create receipt
-  createReceipt(event)
+  if (key && key.owner) {
+    createReceipt(event, key.owner, tokenId.value)
+  } else {
+    createReceipt(event, Bytes.empty(), tokenId.value)
```
clemsos, line 278: Same as above, when does a key without an owner happen in renewal? We should not pass empty Bytes but at least a null address

clemsos (commented):
Seems weird to store an empty key owner in a receipt. People buy or renew a key for someone?

subgraph/schema.graphql
```diff
@@ -110,4 +110,8 @@ type Receipt @entity {
   amountTransferred: BigInt
   "Total gas paid"
   gasTotal: BigInt!
+  "Refer to token id"
+  tokenId: BigInt
+  "Key owner of the receipt"
+  owner: String
```
julien51, line 116: That's a bad idea as a single receipt can apply to multiple keys for multiple owners.

julien51:
Kali I removed the `owner` on purpose because there can be a single receipt for multiple okeys and owners. I don't understand why you need to add these values? Please clarify.

kalidiagne:
> Kali I removed the `owner` on purpose because there can be a single receipt for multiple okeys and owners. I don't understand why you need to add these values? Please clarify.

to retrieve all the receipts by `owner +keyId` in receipt page details.
I can just use the hash of the transactions eventually.

julien51 (changes requested):
Can we closer this please as this is not the right approach?

kalidiagne:
yes, let's close it!

PR #10837, opened 2023-01-29
feature(ui): adding a prop to make a modal that's empty

julien51:
# Description

Using this to create a way to "embed" the Checkout modal on the event landing page.

# Checklist:


searchableguy:
I would add another component named sheet and then reuse it here instead of using a prop to toggle. That's a more common pattern. 

So you can use sheet irrespective of modal for situation like checkout.

searchableguy (approved):
LGTM though.

julien51:
Great idea! In order to keep things tractable, I will work on this after that one is merged (see https://github.com/unlock-protocol/unlock/issues/10843)

PR #10841, opened 2023-01-29
chores: remove node-fetch

julien51:
#10838  Description

Now that we use node 18, it has native support for the `fetch` API and we don't need node-fetch anymore!

# Issues

Fixes #
Refs #

# Checklist:


docker/development/eth-node/package.json
```diff
@@ -11,7 +11,6 @@
     "dependencies": {
         "@nomiclabs/hardhat-ethers": "2.2.1",
         "@unlock-protocol/hardhat-plugin": "0.0.18",
-        "cross-fetch": "3.1.5",
```
julien51, line 14: I don't think this was used!

packages/networks/src/networks/palm.ts
```diff
@@ -0,0 +1,74 @@
...
+    {
+      name: 'Dai',
+      symbol: 'DAI',
+      decimals: 18,
+      address: '0x4C1f6fCBd233241bF2f4D02811E3bF8429BC27B8',
+    },
+  ],
+}
+
+export default palm
```
searchableguy, line 74: This does not seem intended to be in this PR.

julien51: Oupsie

searchableguy (approved):
LGTM. Also let's make sure we use node 18 everywhere to run the code including netlify (we don't use docker image there).

julien51:
Since we now have the node engine in the package file, we should be good correct?

PR #10933, opened 2023-02-01
feat(subgraph): Set cancelled to false by default

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/subgraph/src/public-lock.ts
+++ b/subgraph/src/public-lock.ts
@@ -33,6 +33,7 @@ function newKey(event: TransferEvent): void {
   key.tokenId = event.params.tokenId
   key.owner = event.params.to
   key.createdAtBlock = event.block.number
+  key.cancelled = false
 
   const lockContract = PublicLock.bind(event.address)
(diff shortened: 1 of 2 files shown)


julien51 (approved):

subgraph/src/public-lock.ts
```diff
@@ -127,12 +128,16 @@ export function handleTransfer(event: TransferEvent): void {
...
-      key.expiration = getKeyExpirationTimestampFor(
+      const expiration = getKeyExpirationTimestampFor(
         event.address,
         event.params.tokenId,
         event.params.to
       )
 
+      key.expiration = expiration
+      // Mark as cancelled if the key has expired
+      key.cancelled = expiration.lt(event.block.timestamp)
```
julien51, line 139: I think we should removed the `cancelled` property as it creates a lot of confusion.

searchableguy: They are different. Cancelled is only true if the key was cancelled or transferred while expiration always follows the key expiration.

PR #10937, opened 2023-02-01
feat(unlock-app) - customize recipients for withdrawal

kalidiagne:
# Description

**Modal** 

Screenshot 2023-02-09 at 17 13 41

Screenshot 2023-02-09 at 17 17 11

**Form validations** 

Screenshot 2023-02-02 at 12 55 10

# Issues

Fixes #
Refs #

# Checklist:


julien51:
For the address can we put the address of the current user by default? and call that `beneficiary`?
For the amount let's use a `<input type="range">` between 0 and the current balance and add a warning indicating that if the user withdraws too much users would not be able to cancel their memberships.

unlock-app/src/components/interface/locks/Manage/elements/WithdrawFundModal.tsx
```diff
@@ -101,23 +111,26 @@ export const WithdrawFundModal = ({
...
+            {...register('beneficiary', {
               required: {
                 value: true,
                 message: 'This field is required.',
               },
               minLength: {
                 value: 3,
-                message: 'Address should be 3 characters long at least.',
+                message:
+                  'Beneficiary address should be 3 characters long at least.',
```
julien51, line 122: That's not correct. it should be an ethereum address or ENS. Please use the component that we have that supports that?

kalidiagne: We don't actually have the component merged, I felt a comment to replace it when it is ready!

kalidiagne:
> For the address can we put the address of the current user by default? and call that `beneficiary`? For the amount let's use a `<input type="range">` between 0 and the current balance and add a warning indicating that if the user withdraws too much users would not be able to cancel their memberships.

Updated 

Screenshot 2023-02-02 at 18 54 37

searchableguy (commented):
Can remove the top notification far since we have a loading button with text? WDYT? less clutter.

kalidiagne:
> Can remove the top notification far since we have a loading button with text? WDYT? less clutter.

Updated!

kalidiagne:
@julien51 added `AddressInput` component

julien51 (approved):

julien51:
LG!

Issue #10185, opened 2022-11-02
Slow "renewal" loop


julien51:
Our current script to find keys to renew runs every minute but only looks at keys that have expired recently (30 minutes).
We need to also make sure we are able to renew keys that have expired more than 30 minutes and that have become renewable in the meantime (for example, the user added some ERC20 to their balance... etc).

For this, we don't need to run as frequently. For example I suggest a script that runs *once a day* and looks at _all_ keys that have expired in the month and a script that runs once a week that looks at all keys that have expired _ever_.

PR #10966, opened 2023-02-06
feat(locksmith): Run renewal jobs weekly and monthly

searchableguy:
# Description

fixes https://github.com/unlock-protocol/unlock/issues/10185

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/graphql/datasource/keysToRenew.ts
+++ b/locksmith/src/graphql/datasource/keysToRenew.ts
@@ -1,58 +1,50 @@
-import gql from 'graphql-tag'
-import { GraphQLDataSource } from 'apollo-datasource-graphql'
-import networks from '@unlock-protocol/networks'
+import { SubgraphService } from '@unlock-protocol/unlock-js'
+import logger from '../../logger'
 
-export class KeysToRenew extends GraphQLDataSource {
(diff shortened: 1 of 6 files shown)


julien51:
Please add a daily cycle too.

julien51:
For weekly, we run on _all_ the keys that have ever expired.
For daily we run on all of the keys that have expired in the month.
We add one that runs every 3 hours and that runs on all the keys that have expired within the day.
The current one runs every minute and renews anything expired within 30 minutes

Let's start with that.

Later we will add subgraphs that let us identify *when* someone who had an expired key has re-approved by stopping up their balance of ERC20 or bumped the approval.

searchableguy:
Added 

weekly - keys within the last year. This is a good cut off instead of forever which can be unexpected from a user's perspective too if they authorized balance but didn't have any until a year later and suddenly get cut. 

daily - within the last week
every 5 minute - within the last hour

julien51 (approved):
Lgtm! 

Quick note: do you think we should have a way to "report" when the script ran with something like:
" Ran reveals for keys expired between x and y and renewed z successfully"

We could then pay to sentry for example?

PR #10980, opened 2023-02-07
feat(wedlocks) - refactor and use email-template package

kalidiagne:
# Description

- [x] Move all email templates to `email-templates` package
- [x] Move email template tests to `email-templates` package

# Issues

Fixes #
Refs #

# Checklist:


packages/email-templates/package.json
```diff
@@ -8,7 +8,11 @@
   ],
   "main": "dist",
   "devDependencies": {
-    "typescript": "4.9.4"
+    "@babel/plugin-transform-runtime": "7.19.6",
+    "@vitest/coverage-c8": "0.28.3",
+    "regenerator-runtime": "0.13.11",
```
julien51, line 13: Are these new dependencies?

packages/email-templates/src/index.ts
```diff
@@ -30,7 +30,22 @@ export const EmailTemplates: Record<Partial<Template>, EmailTemplateProps> = {
   debug,
   transferCode,
   keyAirdropped,
-  ...LockTemplates,
 }
 
-export const getEmailTemplate = (template: Template) => EmailTemplates[template]
+const templates = {}
+Object.keys(LockTemplates).forEach((template: any) => {
+  // @ts-ignore
```
searchableguy, line 37: No need for ts-ignore if you type the const template declaring with a record.

packages/email-templates/src/index.ts
```diff
@@ -30,7 +30,22 @@ export const EmailTemplates: Record<Partial<Template>, EmailTemplateProps> = {
...
+})
+
+Object.keys(EmailTemplates).forEach((template: string) => {
+  // @ts-ignore
+  templates[template.toLowerCase()] = EmailTemplates[template]
+})
+
+const getEmailTemplate = (template: Template) => EmailTemplates[template]
+// @ts-ignore
+const getLockTemplate = (template: any) => LockTemplates[template]
```
julien51, line 48: Why are we introducing a new method here? I thin this will be error prone. Let's just have `getEmailTemplate` return *all* templates (generic AND lock specific).

julien51 (commented):
This is a big change and we don't have a way to test in staging. Please test extensively.

searchableguy:
If you can remove the vitest change from this PR and the associated dependencies, that would be really helpful.

packages/email-templates/package.json
kalidiagne: we dont need them actually, just `vitest` here

kalidiagne:
> If you can remove the vitest change from this PR and the associated dependencies, that would be really helpful.

Sorry, what do you mean? by moving all into `email-templates` i also need to move some tests and deps update

searchableguy:
I mean if you can open a separate PR for those changes, that would be much easier to review. If it is not possible or too much work, that's OK.

kalidiagne:
> I mean if you can open a separate PR for those changes, that would be much easier to review. If it is not possible or too much work, that's OK.

Oh, nice, ill work on it

kalidiagne:
> I mean if you can open a separate PR for those changes, that would be much easier to review. If it is not possible or too much work, that's OK.

@searchableguy feel free to check now :)

julien51 (approved):
LGTM if you are confident this won't break emails!

PR #10990, opened 2023-02-08
feat(unlock-app): Add getWalletService and remove explicit network switch handling from code

searchableguy:
# Description

This can be used to automatically change network then prompt user to sign the transaction. 

```ts

const walletService = await getWalletService(network)
walletService.xyz()
```

# Issues

Fixes #
Refs #

# Checklist:


unlock-app/src/components/interface/keychain/CancelAndRefundModal.tsx
```diff
@@ -40,7 +40,7 @@ export const CancelAndRefundModal = ({
   network,
   onExpireAndRefund,
 }: CancelAndRefundProps) => {
-  const walletService = useWalletService()
```
julien51, line 43: Should we remove that hook completely to avoid confusion?

unlock-app/src/components/interface/keychain/CancelAndRefundModal.tsx
```diff
@@ -73,6 +73,7 @@ export const CancelAndRefundModal = ({
       lockAddress,
       tokenId,
     }
+    const walletService = await changeNetwork(network)
```
julien51, line 76: What do you think of actually naming this `getWalletServiceFor(network)`  and just using it everywhere in the app when we need a walletService?

julien51 (commented):
@searchableguy I still find this pretty confusion TBH :/

unlock-app/src/components/content/event/NewEvent.tsx
```diff
@@ -58,8 +58,7 @@ export const NewEvent = () => {
           }),
         }
       )
-      console.log(lockResponse)
-
+      console.debug(lockResponse)
```
julien51, line 61: We can probably clean that up...

unlock-app/src/components/interface/checkout/main/Confirm.tsx
```diff
@@ -85,8 +84,7 @@ export function Confirm({
   communication,
 }: Props) {
   const [state, send] = useActor(checkoutService)
-  const { account, network, isUnlockAccount, changeNetwork } = useAuth()
-  const walletService = useWalletService()
+  const { account, network, getWalletService } = useAuth()
```
julien51, line 87: Are we using `network` ?

I think we should _remove_ the `network` from the `useAuth` hook completely because it is a code smell IMO. It means a component needs to know on what network the user is, which is probably not right.

unlock-app/src/components/interface/checkout/main/Renew.tsx
```diff
@@ -161,7 +156,11 @@ export function Renew({
   const onSign = async () => {
     setIsSigningMessage(true)
     try {
-      const signature = await signMessage(messageToSign!)
+      const walletService = await getWalletService()
+      const signature = await walletService.signMessage(
```
julien51, line 160: Ha! We probably need to change this... (later PR!)

julien51: But can you add a `TODO` please?

unlock-app/src/components/interface/keychain/Key.tsx
```diff
@@ -66,8 +65,7 @@ export interface Props {
 function Key({ ownedKey, account, network }: Props) {
   const { lock, expiration, tokenId, isExpired, isExtendable, isRenewable } =
     ownedKey
-  const { network: accountNetwork } = useAuth()
-  const walletService = useWalletService()
+  const { network: accountNetwork, getWalletService } = useAuth()
```
julien51, line 68: Same here.

unlock-app/src/components/interface/locks/Create/CreateLock.tsx
```diff
@@ -80,40 +71,46 @@ export const CreateLockSteps = () => {
...
-          ? UNLIMITED_KEYS_DURATION
-          : expirationInSeconds,
-        maxNumberOfKeys: unlimitedQuantity
-          ? UNLIMITED_KEYS_COUNT
-          : maxNumberOfKeys,
-        currencyContractAddress,
-        keyPrice: `${keyPrice}`, // must be a string
-      }
-      return createLockPromise(payload)
+      const lockAddress = await walletService.createLock(
```
julien51, line 88: Thank you for moving this here from the hook!

unlock-app/src/hooks/usePrice.ts
```diff
@@ -19,8 +19,8 @@ export const useGetPrice = ({
   currencyContractAddress,
   hash,
 }: GetPriceProps) => {
-  const walletService = useWalletService()
-  const provider = walletService.providerForNetwork(network)
+  const web3Service = useWeb3Service()
+  const provider = web3Service.providerForNetwork(network)
```
julien51, line 23: Good catch!

julien51 (approved):
Ok LG!

I realized thru it that we need to keep exposing the `network` as part of the `useAuth` for the very small number of places where we need this (lock creation...)

unlock-app/src/components/interface/checkout/main/Renew.tsx
searchableguy: It's already in TODO.

unlock-app/src/components/interface/checkout/main/Confirm.tsx
searchableguy: We still use it in some places. I think we can remove it eventually but to minimize changes, I'm keeping it as it is.

PR #11000, opened 2023-02-08
feature(smart-contracts): adding new task to deploy from package

julien51:
# Description

I realized we should **not** deploy from the "current" version of the code in the contracts, but only based on packages.
This PR adds that task! 
It should make adding new networks trivial!

# Checklist:

### DIFF
--- a/smart-contracts/scripts/deployments/deploy-from-package.js
+++ b/smart-contracts/scripts/deployments/deploy-from-package.js
@@ -1,0 +1,58 @@
+const { ethers, upgrades, unlock } = require('hardhat')
+const { networks } = require('@unlock-protocol/networks')
+const { getNetworkName } = require('../../helpers/network')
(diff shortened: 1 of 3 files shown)


julien51:
The problem though is that this does not seem to add the files that we need to keep track of? Any idea @clemsos ?

smart-contracts/scripts/deployments/deploy-from-package.js
```diff
@@ -0,0 +1,58 @@
...
+
+async function main(owner) {
+  const [deployer] = await ethers.getSigners()
+  console.log(`Deploying from ${deployer.address}.`)
+
+  // fetch chain info
+  const chainId = await deployer.getChainId()
+  const networkName = getNetworkName(chainId)
+
+  await unlock.deployProtocol();
```
clemsos, line 14: that task was intended mostly for local dev, not sure we should use it for prod purposes

PR #11022, opened 2023-02-13
feat(smart-contracts): implement protocol fee

clemsos:
# Description

This PR implements a fee for the protocol to be paid by locks 

```
setProtocolFee
      ✓ default to zero
      ✓ can be changed
      ✓ can be changed only by owner
    Pay protocol fee using ETH
      ✓ fee is set correctly in Unlock
      pays fees to Unlock correctly when
        ✓ purchasing a single key
        ✓ purchasing multiple keys
        ✓ extending a key
    Pay protocol fee using ERC20
      ✓ fee is set correctly in Unlock
      pays fees to Unlock correctly when
        ✓ purchasing a single key
        ✓ purchasing multiple keys
        ✓ extending a key
        ✓ renewing a key
```

# Issues

Fixes #
Refs #

# Checklist:


smart-contracts/contracts/Unlock.sol
```diff
@@ -114,6 +114,9 @@ contract Unlock is UnlockInitializable, UnlockOwnable {
   mapping(uint16 => address) private _publicLockImpls;
   uint16 public publicLockLatestVersion;
 
+  // protocol fee
+  uint public fee;
```
julien51, line 118:
```suggestion
  uint public protocolFee;
```

Let's more precise here since we already have referrerFees in the locks.

smart-contracts/contracts/Unlock.sol
```diff
@@ -532,6 +533,14 @@ contract Unlock is UnlockInitializable, UnlockOwnable {
     return 11;
   }
 
+  /**
+   * Set the fee used by the protocol
+   * @param _protocolFee fee in basic point
+   */
+  function setProtocolFee(uint _protocolFee) external onlyOwner {
+    fee = _protocolFee;
```
julien51, line 541:
```suggestion
    protocolFee = _protocolFee;
```

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -44,6 +44,8 @@ contract MixinPurchase is
 
   mapping(address => uint) public referrerFees;
 
+  error TransferFailed();
```
julien51, line 47: ?

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -134,6 +138,12 @@ contract MixinPurchase is
         address(unlockProtocol)
       );
     }
+
+    // get fee from Unlock 
+    uint protocolFee = (_keyPrice * unlockProtocol.fee()) / BASIS_POINTS_DEN;
+
+    // pay fee to Unlock
+    _transfer(tokenAddress, payable(address(unlockProtocol)), protocolFee);
```
julien51, line 146: Wow, that's easy :D

julien51, line 146: Let's only call this if `unlockProtocol.fee()` is  `> 0` to save gas...

clemsos: we have one issue though. Currently that fails with ERC20 because we receive all the ERC20 tokens in a single call at the end of the purchase loop BUT we trigger the `_recordKeyPurchase`  for each key within the loop. 

That means we attempt to transfer tokens that have not been received yet - which fails. 

It works fine with native tokens hat arrives directly to the contract via `payable` call.

The solution will be to take the protocol fee payment logic out of `_recordKeyPurchase` and call it once at the end of the purchase call. We already have a `_payReferrer` that could be reused for that maybe

clemsos: I was worried about contract size but seems to work fine after adding a `_payProtocol` internal func

smart-contracts/contracts/interfaces/IUnlock.sol
```diff
@@ -311,4 +311,8 @@ interface IUnlock {
    * Can only be called by the current owner.
    */
   function transferOwnership(address newOwner) external;
+
+  // TODO: document these
```
julien51, line 315: Let's do it as part of the PR please :D

julien51:
Ok this looks good but we're missing the swap and burn mechanism :) Please add that too!
I think it is just a function that can be called by anyone (important) and which takes a token address (or `0x0` for the native token) and amount and that will then use Uniswap to swap the token for UDT at the current price and then burn the UDT received.

Importantly, the token **cannot** be UDT (ie you can swap UDT for UDT).

julien51:
In fact, let's move that burn mechanism to another PR!

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -104,15 +106,42 @@ contract MixinPurchase is
...
+   * @param _baseAmount the total amount to calculate the fee
+   * @dev internal function to execute the payments to referrers if any is set
+   */
+  function _payProtocol(uint _baseAmount) internal {
+
+    // get fee from Unlock 
+    uint protocolFee;
+    try unlockProtocol.protocolFee() {
+      // calculate fee to be paid
+      protocolFee = (_baseAmount * unlockProtocol.protocolFee()) / BASIS_POINTS_DEN;
```
julien51, line 119: Is this making 2 calls to `unlock` contract? Maybe better to use a variable so we avoid spending gas?

julien51 (approved):

clemsos:
`trying to deploy a contract whose code is too large` 😢

clemsos:
can rebase once  #11155 is merged

PR #11051, opened 2023-02-15
feature(locksmith): Adding verification url

julien51:
# Description

In order to add passes, we actually need the URL (not the QR code)

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/ticketsController.tsx
+++ b/locksmith/src/controllers/v2/ticketsController.tsx
@@ -193,7 +193,7 @@ export class TicketsController {
       })
     } catch (err) {
       logger.error(err)
-      return response.sendStatus(500).send({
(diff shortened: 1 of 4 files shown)


searchableguy (approved):

locksmith/src/controllers/v2/ticketsController.tsx
```diff
@@ -169,6 +169,35 @@ export class TicketsController {
...
+        lockAddress,
+        tokenId,
+      })
+
+      return response.status(200).send({
+        verificationUrl,
+      })
+    } catch (err) {
+      logger.error(err)
+      return response.sendStatus(500).send({
```
searchableguy, line 196:
```suggestion
      return response.status(500).send({
```

PR #11076, opened 2023-02-20
feat(unlock-app) - improve AddressInput validation

kalidiagne:
# Description

- [x] Improve `AddressInput` validation 

# Issues

Fixes #
Refs #

# Checklist:


packages/ui/lib/components/Form/AddressInput.stories.tsx
```diff
@@ -40,6 +42,13 @@ const Template: ComponentStory<typeof AddressInput> = () => {
     console.log('error value', error)
   }
 
+  const isAddressOrEns = (address = '') => {
+    return (
+      address?.toLowerCase()?.includes('.eth') ||
```
julien51, line 47: ENS are actually not limited to `.eth`.

unlock-app/src/utils/strings.tsx
```diff
@@ -37,3 +39,11 @@ export const getValidNumber = (value: string | number): number | undefined => {
     ? parseInt(`${value}`)
     : undefined
 }
+
+/** Check if string is an address or ens */
+
+export const isAddressOrEns = (address = '') => {
```
julien51, line 45: Let's just export the other one so we don't duplicate logic?

kalidiagne: sure, you mean from the component or the other in codebase?

julien51: From the component. And if we use in either places just make sure we only have a single way of doing this.

packages/ui/lib/components/Form/AddressInput.stories.tsx
```diff
@@ -31,6 +32,7 @@ const Template: ComponentStory<typeof AddressInput> = () => {
   })
 
   const onSubmit = (form: any, e: any) => {
+    console.log('form')
```
julien51, line 35: Remove please

packages/ui/lib/utils.ts
```diff
@@ -6,3 +6,9 @@ export const minifyAddress = (address: string) => {
     ? `${address.slice(0, 5)}...${address.slice(address.length - 5)}`
     : address
 }
+
+export const isAddressOrEns = (address = '') => {
+  return (
+    address?.toLowerCase()?.includes('.eth') || ethers.utils.isAddress(address)
```
julien51, line 12: Kali, ENS names don't necessarily end up with `.eth`.... So I don't think that's right. If you want to check if something is a valid ENS you need to check that it resolves correctly.

kalidiagne: making this an async function will cause a side-effect in the component, the idea here is to have a result without calling `resolveName` and waiting for the result.

julien51:
https://github.com/unlock-protocol/unlock/issues/11092

julien51:
For now let's just use `.eth` so we can ship and we can add more resolution later

julien51 (approved):

PR #11098, opened 2023-02-22
feat(ui) - Detail component

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:


packages/ui/lib/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,78 @@
...
+  append?: React.ReactNode
+  loading?: boolean
+  inline?: boolean
+  labelSize?: Size
+  valueSize?: Size
+  truncate?: boolean
+}
+
+const DetailValuePlaceholder = () => {
+  return <div className="w-10 h-5 animate-pulse bg-slate-200"></div>
```
searchableguy, line 22: Use place holder component

packages/ui/lib/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,78 @@
...
+    medium: 'text-lg',
+    large: 'text-2xl',
+  }
+
+  const VALUE_STYLES: SizeStyleProp = {
+    tiny: 'text-base',
+    small: 'text-base',
+    medium: 'text-lg',
+    large: 'text-2xl md:text-4xl',
+  }
```
searchableguy, line 50: You can use tw class for these since we added that as a package.

packages/ui/lib/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <span className={labelClass}>{label}</span>
+      </div>
+      {loading ? (
+        <DetailValuePlaceholder />
+      ) : (
+        <Tooltip tip={value} label={label} side="bottom">
+          <div className="flex items-center gap-2 text-right">
+            {prepend && <>{prepend}</>}
+            <span className={valueClass}>{value ?? '-'}</span>
+            {append && <>{append}</>}
```
searchableguy, line 72: I would remove append and prepend prop and allow children instead of a value so you can pass a custom component instead with whatever you want to prepend or append. 

It may be a little more code for implementator but provide more flexibility. You can provide a default children wrapper if string passed.

kalidiagne: Updated

packages/ui/lib/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,101 @@
...
+          large: 'text-2xl md:text-4xl',
+        },
+      },
+      defaultVariants: {
+        size: 'small',
+      },
+    }
+  )
+
+  const Label = classed.span('relative text-gray-700', {
```
searchableguy, line 57: Move it outside the component and put truncate as a boolean variant.

packages/ui/lib/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { Icon } from '../Icon/Icon'
+import { Size, SizeStyleProp } from '~/types'
+import { Placeholder } from '../Placeholder'
+import { classed } from '@tw-classed/react'
+
+export interface DetailProps {
+  label: React.ReactNode
+  icon?: IconType
+  iconSize?: number
+  value?: React.ReactNode
```
searchableguy, line 12: Change to children as name because that way, you can wrap the component.

packages/ui/lib/components/Detail/Detail.tsx
```diff
@@ -0,0 +1,101 @@
...
+  loading?: boolean
+  inline?: boolean
+  labelSize?: Size
```
searchableguy, line 11: Size it according to the entire component when possible. I think label can be like that.

searchableguy (approved):

packages/ui/lib/components/Detail/Detail.tsx
kalidiagne: i added `labelSize` and `valueSize` to make it more flexible as possible, for example in lock card we have this: 

```
- label: tiny
- value: large 
```

**Example**

Screenshot 2023-02-22 at 15 30 47

PR #11147, opened 2023-02-27
feature(paywall): switch tooling for paywall

julien51:
# Description

Moving away from webpack, per @searchableguy 's reco!

# Checklist:


julien51:
I am doing this so we're able to provide the Paywall package in more than just umd so i can use it in a vite application.
It also raises the question of how we handle the CSS.
Maybe @searchableguy has ideas?

searchableguy:
@julien51 as mentioned. Tsup does not support bundling css modules as injectable js module directly. So need to use a plugin. I couldn't find any which does from a cursory exploration.

Alternative option which works is keep webpack for umd and generate esm/cjs files using tsup separately and export them from package.json 

This will require minimal changes for making this work.

We can remove webpack completely later but this will unblock you.

I'm making changes to bundling to support WP so will look into this along side if it's not an immediate blocker today.

julien51:
Closing because @searchableguy will work on a better version of this!

PR #11225, opened 2023-03-01
feat(unlock-js) - Add support password hook contract

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:


julien51:
@kalidiagne I am not sure this is the right approach and this duplicates a lot of code. 
I think for now it is better to add to `walletService` and `web3Service` the functions we need and once we have an API for the locks add the API there,

searchableguy:
Maybe create a CustomHookService which can handle all of these hooks. 

I think adding it to walletService changes the scope since so far it's only used for unlock contracts.

searchableguy (approved):

packages/unlock-js/src/customHookService.ts
```diff
@@ -0,0 +1,64 @@
...
+      lockAddress: string
+      signerAddress: string
+      network: number
+    },
+    signer: Signer
+  ) {
+    const { lockAddress, signerAddress, network } = params ?? {}
+    const contract = await this.getContract({
+      network,
+      address: lockAddress,
```
searchableguy, line 58: This should be the password hook address

PR #11226, opened 2023-03-01
fix(unlock-app): using a valid referrer

julien51:
# Description

Making sure we use the correct referrer when applicable!

# Checklist:


searchableguy (approved):
LGTM. Also resolve an issue with contract bug, I see.

searchableguy:
I also wonder if unlock js should set referrer to recipient when it is not defined because this contract bug will also affect anyone using unlock-js.

julien51:
> I also wonder if unlock js should set referrer to recipient when it is not defined because this contract bug will also affect anyone using unlock-js.

I am not sure this is really a bug as I want people to set referrers :D

PR #11260, opened 2023-03-06
feat(unlock-js) - refactor customHookService

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:


julien51 (approved):

julien51:
I still think this is the wrong approach and we should add that logic to web3Service...

julien51 (changes requested):
Please add this to web3Service.

searchableguy:
@julien51 shouldn't it be walletService though given you already have signer by default there and it is the one used for transaction.

kalidiagne:
> @julien51 shouldn't it be walletService though given you already have signer by default there and it is the one used for transaction.

already changed, but since we are going to use add more custom hooks, maybe the things is using the new custom service as before

packages/unlock-js/src/web3Service.ts
```diff
@@ -963,4 +964,49 @@ export default class Web3Service extends UnlockService {
       return ''
     }
   }
+
+  /**
+   * Set signer for `Password hook contract`
+   */
+  async setPasswordHookSigner(
```
julien51, line 971: This should be in WalletService as it modifies the state.

packages/unlock-js/src/web3Service.ts
```diff
@@ -963,4 +964,49 @@ export default class Web3Service extends UnlockService {
...
+      signer,
+    })
+    const tx = await contract.setSigner(lockAddress, signerAddress)
+    return tx
+  }
+
+  /**
+   * Get signer for `Password hook contract`
+   */
+  async getPasswordHookSigners(
```
julien51, line 995: This should remain here as it only reads the state

packages/unlock-js/src/web3Service.ts
kalidiagne: moved to `WalletService `

julien51 (approved):

PR #11267, opened 2023-03-07
feature(locksmith): adding config for logtail

julien51:
# Description

We may as well use their "native" transport so that we can do better searches in their UI.

# Checklist:

### DIFF
--- a/locksmith/src/config/config.ts
+++ b/locksmith/src/config/config.ts
@@ -51,6 +51,7 @@ const config = {
       process.env.STORAGE_PUBLIC_HOST || defaultConfig.storage.publicHost,
   },
   recaptchaSecret: process.env.RECAPTCHA_SECRET,
+  logtailSourceToken: process.env.RECAPTCHA_SECRET,
 }
 
 if (process.env.ON_HEROKU) {
--- a/locksmith/src/controllers/purchaseController.ts
+++ b/locksmith/src/controllers/purchaseController.ts
@@ -249,7 +249,7 @@ export class PurchaseController {
     const fulfillmentDispatcher = new Dispatcher()
 
     if (pricing.keyPrice !== undefined && pricing.keyPrice > 0) {
-      return res.status(500).send('Lock is not free')
+      return res.status(400).send('Lock is not free')
     }
 
     const hasEnoughToPayForGas =
--- a/locksmith/src/controllers/v2/claimController.ts
+++ b/locksmith/src/controllers/v2/claimController.ts
@@ -29,7 +29,7 @@ export const claim: RequestHandler = async (request, response) => {
   }
 
   if (pricing.keyPrice !== undefined && pricing.keyPrice > 0) {
-    return response.status(500).send({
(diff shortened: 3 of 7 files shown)


julien51:
@clemsos I think I will need your help to set up the env vars :/

locksmith/src/controllers/purchaseController.ts
```diff
@@ -249,7 +249,7 @@ export class PurchaseController {
     const fulfillmentDispatcher = new Dispatcher()
 
     if (pricing.keyPrice !== undefined && pricing.keyPrice > 0) {
-      return res.status(500).send('Lock is not free')
+      return res.status(400).send('Lock is not free')
```
julien51, line 252: These feels like they are 400 more than 500

locksmith/src/controllers/v2/claimController.ts
```diff
@@ -29,7 +29,7 @@ export const claim: RequestHandler = async (request, response) => {
   }
 
   if (pricing.keyPrice !== undefined && pricing.keyPrice > 0) {
-    return response.status(500).send({
+    return response.status(400).send({
```
julien51, line 32: Same

locksmith/src/config/config.ts
```diff
@@ -51,6 +51,7 @@ const config = {
       process.env.STORAGE_PUBLIC_HOST || defaultConfig.storage.publicHost,
   },
   recaptchaSecret: process.env.RECAPTCHA_SECRET,
+  logtailSourceToken: process.env.LOGTAIL_SOURCE_TOKEN,
```
julien51, line 54: I am confused about both the config.js and config.ts.
Do we really need both @searchableguy ?

searchableguy: yeah because one is used by sequelize-cli and it doesn't support running typescript config file. You don't need to add these to that one.

locksmith/src/logger.ts
```diff
@@ -13,7 +16,10 @@ export const logger = winston.createLogger({
 // No output in tests
 logger.add(
   new winston.transports.Console({
-    silent: process.env?.NODE_ENV === 'test',
+    silent: !!(
+      process.env?.NODE_ENV &&
+      ['test', 'production'].indexOf(process.env?.NODE_ENV) > -1
```
searchableguy, line 21: not on staging?

locksmith/src/config/config.ts
julien51: ha! Good point. Ok I think we indeed need to "normalize"... but that's for later!

locksmith/src/logger.ts
julien51: I am not sure I want to murky the production logs... but maybe we can add later?

julien51:
I found how the env var is added! Automatically by the plugin!
Cool!
Screenshot 2023-03-08 at 12 03 07 PM


Ready for a final review @searchableguy PTAL!

searchableguy (approved):

PR #11275, opened 2023-03-07
fix(unlock-app): removing network where it does not make sense

julien51:
# Description

Also fixing the checkout where we use the user's network instead of the locks' network.

# Checklist:

### DIFF
--- a/unlock-app/src/components/content/RecoverContent.tsx
+++ b/unlock-app/src/components/content/RecoverContent.tsx
@@ -25,7 +25,6 @@ export const RestoreAccount = ({
   config,
   email,
   recoveryKey,
-  network,
 }: RestoreAccountProps) => {
   const { setProvider, provider } = useContext(ProviderContext)
   const [loading, setLoading] = useState(false)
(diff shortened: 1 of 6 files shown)


unlock-app/src/components/content/RecoverContent.tsx
```diff
@@ -48,7 +47,7 @@ export const RestoreAccount = ({
 
           // We need to log the user in from the recoveryPhrase + recoveryKey!
           // TODO: what network do we pick???
-          const unlockProvider = new UnlockProvider(config.networks[network])
+          const unlockProvider = new UnlockProvider(config.networks[1])
```
searchableguy, line 50: maybe use config.preferredNetwork since we encourage that everywhere?

searchableguy (approved):

unlock-app/src/components/content/RecoverContent.tsx
julien51: Hum, what is `preferredNetwork` ?

searchableguy: One we show as default in selection and other places (It's polygon currently)

PR #11277, opened 2023-03-07
fix(subgraph) commenting tokenUri update because it breaks on large collections

julien51:
# Description

When a collection is (very?) large, we can't loop over the keys and update them all without breaking indexing on our subgragh.

For now we'll skip updating the tokenURI but we probably need something better.

# Checklist:


searchableguy:
Can you also update any place where subgraph service is used with tokenURI in keys?

julien51:
> Can you also update any place where subgraph service is used with tokenURI in keys?

Well we still have the `tokenUri` property... but it will just be outdated. I am trying to find what the best option is from The Graph team

clemsos:
A workaround would be to assess what "very" large means and then bypass the operation if the `totalKeys` is let's more than 1000. Not ideal but can save us some pain until graph team sorts out the issue

julien51:
I think this is a good idea. The problem is we have to "guess" what the timeout is :/

clemsos:
> I think this is a good idea. The problem is we have to "guess" what the timeout is :/

do we know which lock is failing?

julien51:
> do we know which lock is failing?

Yes: https://polygonscan.com/token/0xafcfff71f3e717fcdb0b6a1bf20026304fd41bee (40k keys!)

clemsos:
Isn't 100 a bit low? Also shouldn't we check to have this condition only on Polygon (if that only fails there)?

Issue #11246, opened 2023-03-03
Add staging environment for wedlocks


julien51:
All in the title.

julien51:
Done!

PR #11280, opened 2023-03-08
chore(wedlocks): adding staging for wedlocks

julien51:
# Description

We recently had failure on wedlocks because of dependencies which broke email sending. 
We do not use a staging server for wedlocks, which means we really have no way to check if these emails would be sent.

I suggest adding a staging environment for wedlocks so that we can test how emails are being sent.

# Checklist:

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -1,4 +1,4 @@
-name: "Production Branch"
+name: 'Production Branch'
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


julien51:
I added the env vars... or at least I think I did, but I am not sure how to check that!

julien51:
https://github.com/unlock-protocol/unlock/issues/11246

PR #11288, opened 2023-03-08
feat(unlock-app) - check for password on checkout

kalidiagne:
# Description

- [x] Get `promo`, `password`, `discount code` status by checking if hook is present 
- [x] Remove from checkout build because not are handled automatically by the checkout 

# Issues

Fixes #
Refs #

# Checklist:


julien51 (commented):
Let's check with @searchableguy who has the most checkout knowledge please

kalidiagne:
> Let's check with @searchableguy who has the most checkout knowledge please

Perfect! waiting for @searchableguy!

searchableguy:
One thing I'm confused about is why are we doing this check on checkout. 

I thought we would do this on checkout builder and automatically set the config there. 

Doing this check on checkout means it will slow things down and add complexity when we want to change anything.

searchableguy:
Okay. Got it. 🤔 

It does result in people not able to use a modified version of password hook but I guess that's a very edge case as most won't deploy their own.

unlock-app/src/components/interface/checkout/main/checkoutMachine.ts
```diff
@@ -47,6 +47,11 @@ export interface SelectLockEvent {
   skipRecipient?: boolean
   recipients?: string[]
   keyManagers?: string[]
+  hook?: {
+    isCaptcha?: boolean
+    isPassword?: boolean
+    isPromo?: boolean
+  }
```
searchableguy, line 54: Do we really to make a boolean state?

Our system only supports one hook at a time so add it as a string.

hook = password | promocode | captcha

kalidiagne:
PR updated and tested the 3 cased on checkout flow 

Screenshot 2023-03-09 at 16 40 47

Screenshot 2023-03-09 at 16 41 07

julien51:
> It does result in people not able to use a modified version of password hook but I guess that's a very edge case as most won't deploy their own.

I actually think we should keep the options in the JSON config, but not in the builder UI anymore. If you are an "advanced" user who creates your own hook then you can arguably use your own paywall config.

kalidiagne:
> > It does result in people not able to use a modified version of password hook but I guess that's a very edge case as most won't deploy their own.
> 
> I actually think we should keep the options in the JSON config, but not in the builder UI anymore. If you are an "advanced" user who creates your own hook then you can arguably use your own paywall config.

Added `paywallConfig` as fallback and hide field as suggested

unlock-app/src/components/interface/checkout/main/useCheckoutHook.ts
```diff
@@ -0,0 +1,95 @@
...
+  const [state] = useActor(service)
+  const web3Service = useWeb3Service()
+
+  let lockHookMapping: LockHookProps = {}
+  const { paywallConfig } = state.context
+
+  const queries = useQueries({
+    queries: [
+      ...Object.entries(paywallConfig.locks).map(
+        ([lockAddress, { network = 1 }]) => {
```
searchableguy, line 23: hm this way, we can't batch the request. Do you think it would make sense to use Promise.all under a single query?

searchableguy (approved):

unlock-app/src/components/interface/checkout/main/useCheckoutHook.ts
kalidiagne: Actually the `Promise.all` it what i had before this comment https://github.com/unlock-protocol/unlock/pull/11288#discussion_r1130865210 ? 

wdyt?

julien51 (approved):

unlock-app/src/components/interface/checkout/main/useCheckoutHook.ts
kalidiagne: Merging for now! ill change it later eventually

searchableguy: I think you got confused. What I meant was separate the request code into a separate promise.all call.

PR #11290, opened 2023-03-08
feature(unlock-app): Improved UI to support a large number of chains!

julien51:
# Description

If we want to add support for more and more chains, we need to provide a better UI where the user does not have to pick in a _very_ long list of chains.

Here I am adding a way to pick in a subset of _featured_ networks while still supporting the ability for users to connect to a specific chain and pick it if we support it.

As part of this PR I modified the `tooltip` component to add a `dark` variant, as well as the `Select` to add a tooltip. (we should probably add this to all input fields...)

# Checklist:


julien51:
Screenshot 2023-03-08 at 3 04 02 PM
Screenshot 2023-03-08 at 3 04 13 PM

searchableguy (approved):

PR #11328, opened 2023-03-16
fix(keychain): Remove use of window.open and use static image url

searchableguy:
# Description

Browser blocks pop up because we do some fetching and generation before opening the window. 
Replaced it with windows.location.assign to avoid this. 

Uses static locksmith url for image instead of passing it as a prop.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/main/Minting.tsx
+++ b/unlock-app/src/components/interface/checkout/main/Minting.tsx
@@ -130,10 +130,6 @@ export function Minting({
 
   const stepItems = useCheckoutSteps(checkoutService)
 
-  const tokenImage = `${config.services.storage.host}/lock/${
-    lock!.address
-  }/icon?id=${tokenId}`
-
   const hasTokenId = !!tokenId
 
   return (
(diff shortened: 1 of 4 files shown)


kalidiagne (approved):

unlock-app/src/components/interface/checkout/main/Minting.tsx
```diff
@@ -195,10 +191,9 @@ export function Minting({
                     network={lock!.network}
                     lockAddress={lock!.address}
                     tokenId={tokenId}
-                    image={tokenImage}
                     name={lock!.name}
                     handlePassUrl={(url: string) => {
-                      window.open(url, '_')
+                      window.location.assign(url)
```
julien51, line 196: Does that open in a new tab though? Maybe ok...

searchableguy: Yeah. Alternative workaround would be to pop up a window when user clicks immediately and then update it later but that is also confusing UX.

PR #11329, opened 2023-03-17
feat(unlock-app) - fix zoom on input focus mobile

kalidiagne:
# Description
Prevent zooming page when an input is focused

**Issue**

Screenshot 2023-03-17 at 11 03 37

**Fix**

Screenshot 2023-03-17 at 11 06 42

# Issues

Fixes #
Refs #

# Checklist:


searchableguy:
This prevents any kind of zooming though. I don't think this is good for accessibility.

kalidiagne:
> This prevents any kind of zooming though. I don't think this is good for accessibility.

This will prevent the zoom when input is focused, user will be able to zoom with "pinch to zoom" action

searchableguy:
This will prevent zooming on android phones and other browsers. 

I think a real solution is to set font size to 16px at least on input which would avoid this zoom.

julien51:
@kalidiagne What's the stastus here? Can you do what @searchableguy suggested and see if that works better?

kalidiagne:
> @kalidiagne What's the stastus here? Can you do what @searchableguy suggested and see if that works better?

Added font size to 16px on mobile to prevent zooming 

Screenshot 2023-03-24 at 14 33 47

julien51 (approved):

PR #11330, opened 2023-03-17
feat(unlock-app,locksmith) - customize app + emails more for events 

kalidiagne:
# Description

- [x] Add event details to email so it can be added to the calendar directly
- [x] Make it more obvious how lock managers can set verifiers
    - [x] Add verifiers details in event page when lock manager is logged
- [x] Add tests for `email-templates`

**Add verifier section in event page for lock manager** 

Screenshot 2023-03-17 at 12 35 56

**KeyMined Email**

Screenshot 2023-03-17 at 16 25 11

**Add to calendar**

Screenshot 2023-03-17 at 16 26 31

# Issues

Fixes #
Refs #

# Checklist:


searchableguy (commented):
So far looks good.

packages/email-templates/src/templates/helpers/eventDetails.ts
```diff
@@ -0,0 +1,81 @@
...
+        <div>
+          <strong>Date:</strong> ${eventDate}
+        </div>
+      `
+    }
+
+    if (!!eventTime) {
+      eventDetails = `${eventDetails}
+        <div>
+          <strong>Time:</strong> ${eventTime}
```
searchableguy, line 48: Mention timezone?

julien51:
@kalidiagne Thanks! This is a great start!
I think there is a way to add stuff in the email directly for it to be added as a calendar event without having the user click on anything from the email body itself by adding add `.ics` attachment. We need to do that because your link only works with Google.

julien51:
Check this https://stackoverflow.com/questions/49354486/how-to-get-ics-calendar-invitation-to-automatically-add-to-calendar

packages/email-templates/package.json
```diff
@@ -8,6 +8,7 @@
   ],
   "main": "dist",
   "dependencies": {
+    "calendar-link": "2.4.0",
```
julien51, line 11: Let's not use a link approach but adding the ics file as attachement

packages/email-templates/src/templates/keyMined.ts
```diff
@@ -1,8 +1,10 @@
 import handlebars from 'handlebars'
 import { customContentStyle } from './helpers/customContentStyle'
 import { links } from './helpers/links'
+import { eventDetails } from './helpers/eventDetails'
```
julien51, line 4: ok so rather than re-using the `keyMined` template here we should have a whole new template `eventKeyMined` that locksmith calls ... (with failover as `keyMined`)

kalidiagne:
ICS attachments added + support for end date

julien51 (approved):

PR #11338, opened 2023-03-20
feat(unlock-app): Add ability to save checkout config

searchableguy:
# Description

https://user-images.githubusercontent.com/73341821/226451404-18df12e6-2cfa-4a69-816a-df0a59cbd930.mp4

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/checkoutController.ts
+++ b/locksmith/src/controllers/v2/checkoutController.ts
@@ -100,8 +100,11 @@ export const deleteCheckoutConfig: RequestHandler = async (
   })
   if (checkoutConfig) {
     await checkoutConfig.destroy()
-  }
-  return response.status(200).send({
+     return response.status(200).send({
     message: 'Config deleted',
+  })
+  }
+  return response.status(404).send({
+    message: 'Not found',
   })
 }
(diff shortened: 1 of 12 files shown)


searchableguy:
I think maybe better to just disallow people from creating same name config. 

That's a good UX point. I'm not sure the last updated information is relevant to users. We sort by last updated in the background.

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -52,14 +70,38 @@ export const getCheckoutConfigsByUser: RequestHandler = async (
...
+      id,
+      createdBy: userAddress,
+    },
+  })
+  if (checkoutConfig) {
+    await checkoutConfig.destroy()
+  }
+  return response.status(200).send({
+    message: 'Config deleted',
+  })
```
kalidiagne, line 106:
```suggestion
  if (checkoutConfig) {
    await checkoutConfig.destroy()
     return response.status(200).send({
    message: 'Config deleted',
  })
  }
  return response.status(404).send({
    message: 'Not found',
  })
```

julien51 (approved):
I wonder if we could auto-save? I worry that users will "forget" to save...

locksmith/src/controllers/v2/checkoutController.ts
kalidiagne: here `'Config deleted'` is returned even if the config is not found

PR #11346, opened 2023-03-21
fix(smart-contracts): use `_transferFrom` to burn a key

### LINKED ISSUE #11331: Burn function should update balance
Currently the burn function seem to not update balances correctly.

> Question about this error when trying to grant an address a token - 
https://dashboard.tenderly.co/tx/polygon/0x0367bf2379ae76c89b155498392923a051095106df6fcccb0eeb2e369f5315de
> 
> The contract on polygonscan:
https://polygonscan.com/token/0xdb37fe42e529f5f686865c23b61fdddc1fa6573e
> 
> The address that I am trying to airdrop a token to was airdropped one previously, that was then burned (via the burn method on the contract). On polygonscan, the token seems to have been successfully transferred to the null address / the burn transaction was 
> 
> `Looking at the contract, it doesn't seem like the _balances map is updated when the burn / _cancelKey methods are called, as ownerOf is just updated directly, instead of calling "transfer" (which does update the balance). Wondering if this is a bug or if this is intentional? 
> 
> I think in the short term we can just increase the maxKeysPerAddress on the contract, and manage that limit off-chain, but would be good to know if there is something else to do.

clemsos:
# Description

The burn function in PublicLock was not updating correctly the balances. With this PR, burn now relies on calling `_transferFrom` to the zero address solving the issue.

# Issues

Fixes #11331
Refs #

# Checklist:


clemsos:
@julien51 one remaining question is do we need to keep the safeguard in `balanceOf` or `transferFrom`? See this commit [764922a](https://github.com/unlock-protocol/unlock/pull/11346/commits/764922a61872a6346cd2b28013e8774a13ed6f62)

smart-contracts/test/Lock/erc721/balanceOf.js
```diff
@@ -10,7 +10,7 @@ contract('Lock / erc721 / balanceOf', (accounts) => {
     lock = await deployLock()
   })
 
-  it('should fail if the user address is 0', async () => {
+  it.skip('should fail if the user address is 0', async () => {
```
julien51, line 13: Let's remove these tests if they are not applicable anymore!

julien51 (approved):
Let's cleanup the tests and then LG!

PR #11353, opened 2023-03-22
fix(locksmith): returning partial results when we paginate

julien51:
# Description

I realized loading members for locks with more than 5000 keys.
When debugging I found that The Graph does not allow the "Skip" param to be larger than 5000.
I am adding that partial fix to return "partial" results, but I really want to understand _why_ we are loading _all_ the results when in practice the front-end would only show 20 anyway.

Any idea @kalidiagne @searchableguy ?

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/keyController.ts
+++ b/locksmith/src/controllers/v2/keyController.ts
@@ -46,7 +46,7 @@ export default class KeyController {
     } catch (error) {
       logger.error(error.message)
       return response.status(500).send({
-        message: 'Keys list could not be retrived.',
+        message: 'Keys list could not be retrieved.',
       })
     }
   }
--- a/locksmith/src/graphql/datasource/keysByQuery.ts
+++ b/locksmith/src/graphql/datasource/keysByQuery.ts
@@ -3,6 +3,7 @@ import networks from '@unlock-protocol/networks'
 import gql from 'graphql-tag'
 import { DocumentNode } from 'graphql'
 import { getValidNumber } from '../../utils/normalizer'
+import logger from '../../logger'
 
 export type KeyFilter = 'all' | 'active' | 'expired' | 'tokenId'
 const keyholdersByTokedIdQuery = gql`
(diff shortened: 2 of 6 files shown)


locksmith/src/controllers/v2/keyController.ts
```diff
@@ -46,7 +46,7 @@ export default class KeyController {
     } catch (error) {
       logger.error(error.message)
       return response.status(500).send({
-        message: 'Keys list could not be retrived.',
+        message: 'Keys list could not be retrieved.',
```
julien51, line 49: typo fixes

locksmith/src/graphql/datasource/keysByQuery.ts
```diff
@@ -159,6 +159,8 @@ export class keysByQuery extends GraphQLDataSource {
 
       const getData = async (getFromPage = page) => {
         const skip = parseInt(`${getFromPage}`, 10) * first
+        // The Graph does not support skipping more than 5000
+        // https://thegraph.com/docs/en/querying/graphql-api/#pagination
```
julien51, line 163: So this will fail, but that's ok because now we handle the failure in a way that allows us to return partial results.

julien51:
I also realized the formatting of our logs on Logtail is still messy. I am not sure why but I will try to debug that.

searchableguy (approved):

locksmith/src/graphql/datasource/keysByQuery.ts
```diff
@@ -180,20 +182,25 @@ export class keysByQuery extends GraphQLDataSource {
...
+          getForNextPage = nextPageKeys?.length === first
+        } catch (error) {
+          console.error(error)
+          getForNextPage = false // When we have an error, we stop paginating, results will be partial
+        }
       }
 
       return locks
     } catch (error) {
+      console.error(error)
```
searchableguy, line 203: use logger otherwise won't be logged

PR #11357, opened 2023-03-22
feat(locksmith) - email notification when are about to expire

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/websub/jobs/keys.ts
+++ b/locksmith/src/websub/jobs/keys.ts
@@ -1,13 +1,22 @@
 import { Op } from 'sequelize'
 import { networks } from '@unlock-protocol/networks'
 import { Key } from '../../graphql/datasource'
-import { Hook, ProcessedHookItem } from '../../models'
+import { Hook, ProcessedHookItem, UserTokenMetadata } from '../../models'
 import { TOPIC_KEYS_ON_LOCK, TOPIC_KEYS_ON_NETWORK } from '../topics'
 import { notifyHook, filterHooksByTopic } from '../helpers'
-import { notifyNewKeysToWedlocks } from '../../operations/wedlocksOperations'
+import {
+  notifyNewKeysToWedlocks,
+  sendEmail,
+} from '../../operations/wedlocksOperations'
 import { logger } from '../../logger'
+import { SubgraphService, Web3Service } from '@unlock-protocol/unlock-js'
+import * as Normalizer from '../../utils/normalizer'
+import { ethers } from 'ethers'
+import dayjs from 'dayjs'
 
 const FETCH_LIMIT = 25
+const MAX_UINT =
+  '115792089237316195423570985008687907853269984665640564039457584007913129639935'
 
 async function fetchUnprocessedKeys(network: number, page = 0) {
   const keySource = new Key(network)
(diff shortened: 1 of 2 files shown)


locksmith/src/websub/jobs/keys.ts
```diff
@@ -102,3 +111,153 @@ export async function notifyOfKeys(hooks: Hook[]) {
...
+          lockName,
+          keyId,
+          network: networkId,
+          currency,
+          expirationDate,
+          keychainUrl: 'https://app.unlock-protocol.com/keychain',
+          isRenewable: isRenewable && !isAutoRenewable ? 'true' : '',
+          isAutoRenewable: isAutoRenewable ? 'true' : '',
+          isRenewableIfRePurchased: '',
+          isRenewableIfReApproved: '',
```
kalidiagne, line 201: @julien51, how can I get these values?

locksmith/src/websub/jobs/keys.ts
```diff
@@ -102,3 +111,157 @@ export async function notifyOfKeys(hooks: Hook[]) {
...
+  end.setDate(now.getDate() + 1)
+
+  await Promise.allSettled(
+    // get expiring keys for every network
+    Object.keys(networks).map(async (networkId: string) => {
+      const subgraph = new SubgraphService()
+      // get keys from now to 24h in the future
+      const keys = await subgraph.keys(
+        {
+          first: 1000, // more than 1000 limit? need to handle it
```
julien51, line 130:
```suggestion
          first: 1000, // TODO: handle more than 1000 keys
```

locksmith/src/websub/jobs/keys.ts
```diff
@@ -102,3 +111,157 @@ export async function notifyOfKeys(hooks: Hook[]) {
...
+            expiration_lt: end.getTime().toString(),
+          },
+        },
+        {
+          networks: [Number(networkId)],
+        }
+      )
+
+      await Promise.allSettled([
+        keys?.map(async (key: any) => {
```
julien51, line 142: That's bad if keys has 1000 objects it will break by sending all networks requests at the same time. Make it sequential.

locksmith/src/websub/jobs/keys.ts
```diff
@@ -102,3 +111,157 @@ export async function notifyOfKeys(hooks: Hook[]) {
...
+          const keyId = key?.tokenId ?? ''
+
+          const userTokenMetadataRecord = await UserTokenMetadata.findOne({
+            where: {
+              tokenAddress: lockAddress,
+              userAddress: ownerAddress,
+            },
+          })
+
+          const protectedData = Normalizer.toLowerCaseKeys({
```
julien51, line 156: Maybe better to write this in function that does this since we seem to use that in multiple places. Use `UserTokenOperations`

locksmith/src/websub/jobs/keys.ts
```diff
@@ -102,3 +111,157 @@ export async function notifyOfKeys(hooks: Hook[]) {
...
+
+          const currency = isERC20
+            ? (await web3Service.getTokenSymbol(
+                tokenAddress,
+                Number(networkId)
+              )) ?? ''
+            : networks?.[networkId]?.baseCurrencySymbol
+
+          const addressBalance =
+            web3Service.getAddressBalance(ownerAddress, Number(networkId)) ?? 0
```
julien51, line 180: Why are you getting the balance of native currency???

locksmith/src/websub/jobs/keys.ts
```diff
@@ -102,3 +111,157 @@ export async function notifyOfKeys(hooks: Hook[]) {
...
+
+          // send expiring email
+          await sendEmail(`keyExpiring`, `keyExpiring`, recipient, {
+            lockName,
+            keyId,
+            network: networkId,
+            currency,
+            expirationDate,
+            keychainUrl: 'https://app.unlock-protocol.com/keychain',
+            isRenewable: isRenewable && !isAutoRenewable ? 'true' : '',
```
julien51, line 199: This is wrong. 
```suggestion
            isRenewable
```

locksmith/src/websub/jobs/keys.ts
```diff
@@ -102,3 +111,157 @@ export async function notifyOfKeys(hooks: Hook[]) {
...
+            isAutoRenewable: isRenewable && isAutoRenewable ? 'true' : '',
```
julien51, line 200:
```suggestion
            isAutoRenewable,
```

julien51 (changes requested):
I think you still don't understand how renewals work :(

locksmith/src/websub/jobs/keys.ts
```diff
@@ -102,3 +122,233 @@ export async function notifyOfKeys(hooks: Hook[]) {
...
+                network: networkId,
+                keychainUrl: `${config.unlockApp}/keychain`,
+                currency,
+                expirationDate,
+                isRenewable: isRenewable ? 'true' : '',
+                isAutoRenewable: isAutoRenewable ? 'true' : '',
+                isRenewableIfRePurchased: isRenewableIfRePurchased
+                  ? 'true'
+                  : '',
+                isRenewableIfReApproved: isRenewableIfReApproved ? 'true' : '',
```
kalidiagne, line 274:
```suggestion
                isRenewable: isRenewable ? 'true' : '',
                isAutoRenewable: isAutoRenewable ? 'true' : '',
                isRenewableIfRePurchased: isRenewableIfRePurchased
                  ? 'true'
                  : '',
                isRenewableIfReApproved: isRenewableIfReApproved ? 'true' : '',
```

I dont think we need to use boolean here, not accepted as valid value, if we look at `params` the typedefinition is `[key: string]: string | number | undefined`: 

also when values are set as `false` it will show in the template 

Screenshot 2023-03-23 at 12 43 11

we need to have an empty string or `undefined` to hide it

Screenshot 2023-03-23 at 12 44 42

julien51: > I dont think we need to use boolean here, not accepted as valid value, if we look at params the typedefinition is [key: string]: string | number | undefined:

Then we need to change the type :) It does not make sense to use `"false"` or `"true"` :)

locksmith/src/operations/subscriptionOperations.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+interface GetSubscriptionsProps {
+  tokenId: string
+  lockAddress: string
+  owner: string
+  network: number
+}
+
+export const getSubscriptions = async ({
```
julien51, line 43:
```suggestion
export const getSubscriptionsForLockByOwner = async ({
```

locksmith/src/operations/metadataOperations.ts
```diff
@@ -248,6 +250,32 @@ export const getKeysMetadata = async ({
...
+  userAddress: string
+}) => {
+  const userTokenMetadataRecord = await UserTokenMetadata.findOne({
+    where: {
+      tokenAddress: lockAddress,
+      userAddress,
+    },
+  })
+
+  logger.info(
```
julien51, line 267: You may want to handle the case when there is none...

locksmith/src/operations/subscriptionOperations.ts
julien51: For clarity

locksmith/src/operations/subscriptionOperations.ts
```diff
@@ -0,0 +1,150 @@
...
+  const [userBalance, decimals, userAllowance, symbol] = await Promise.all([
+    getErc20BalanceForAddress(key.lock.tokenAddress, owner, provider),
+    getErc20Decimals(key.lock.tokenAddress, provider),
+    getAllowance(key.lock.tokenAddress, key.lock.address, provider, owner),
+    getErc20TokenSymbol(key.lock.tokenAddress, provider),
+  ])
+
+  const balance = ethers.utils.formatUnits(userBalance, decimals)
+
+  const price = ethers.BigNumber.from(key.lock.price)
```
julien51, line 83: That's not correct we should use the decimals for the lock's keyPrice here.

locksmith/src/operations/subscriptionOperations.ts
```diff
@@ -0,0 +1,150 @@
...
+  const next =
+    key.expiration === ethers.constants.MaxUint256.toString()
+      ? null
+      : dayjs.unix(key.expiration).isBefore(dayjs())
+      ? null
+      : parseInt(key.expiration)
+
+  // Approved renewals
+  const approvedRenewalsAmount =
```
julien51, line 92: const numberOfRenewalsApproved =

locksmith/src/operations/subscriptionOperations.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+export const getSubscriptions = async ({
+  tokenId,
+  lockAddress,
+  owner,
+  network,
+}: GetSubscriptionsProps): Promise<Subscription[]> => {
+  const subgraphService = new SubgraphService(networks)
+  const key = await subgraphService.key(
```
julien51, line 50: I don't think we should use the subgraph here. It is too britle IMO.
Let's retrieve the fields we need from it? `keyPrice` and `tokenAddress` and `version` ?

julien51 (changes requested):
I think there are things that don't work well here.
Also I would love to get @searchableguy 's eyes on this because it is a big change

kalidiagne:
any update here?

julien51:
@kalidiagne I left a few comments :) please address them!

@searchableguy please take a look!

julien51:
> This would mean if we deploy and restart the worker for any reason. It might end up re-sending duplicates.

@kalidiagne This is an issue. It could be fairly easy to store info about any reminder that we sent so we know we don't have to send again for the _current_ expiration (remember that a key will expire multiple times... so I think you need to store the lock address, the key id, but also the expiration and the "type" of reminder that was sent. So we don't send the "same" reminder twice.

locksmith/src/operations/subscriptionOperations.ts
kalidiagne: for now use of subgraph because we don't have the we to get specific params from `getLock`

locksmith/src/controllers/v2/subscriptionController.ts
```diff
@@ -1,37 +1,8 @@
...
-export interface Subscription {
-  next: number | null
-  balance: Amount
-  price: Amount
-  possibleRenewals: string
-  approvedRenewals: string
-  type: 'Crypto' | 'Stripe'
-}
+import * as subscriptionOperations from '../../operations/subscriptionOperations'
+import { Subscription } from '../../operations/subscriptionOperations'
```
julien51, line 5: Why the 2 imports here?

locksmith/migrations/20230425123140-key-expiration-reminder.js
```diff
@@ -0,0 +1,52 @@
...
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+    })
+    await queryInterface.addIndex('KeyExpirationReminders', {
+      fields: ['id', 'lockAddress', 'network', 'tokenId', 'type', 'expiration'],
```
julien51, line 43: I don't think it makes sense to add the `id` in there. 
When you create an index is it because you will use the indexed fields to make a query.
Here you will likely retrieve by network, lockAddress, tokenId, type and expiration, but NOT by `id`.

julien51 (commented):
There is a lot of things going on. Let's please split this in several smaller PR with their corresponding ytests.

For example a first PR could add the new object.
Then another PR would the SubscriptionsOperation file...

etc

kalidiagne:
> There is a lot of things going on. Let's please split this in several smaller PR with their corresponding ytests.
> 
> For example a first PR could add the new object. Then another PR would the SubscriptionsOperation file...
> 
> etc

ok, splitting in small PR

kalidiagne:
this PR was split in small ones, and this will solve https://github.com/unlock-protocol/unlock/pull/11840

PR #11376, opened 2023-03-24
fix(unlock-app): remove .enable on providers call by default and using it only for WalletConnect an…

### LINKED ISSUE #11374: `.enable` has been deprecated
We call `.enable` on the provider, but it looks like it has been deprecated. We should probably not do it anymore... (not sure what is the recommended API these days!).

julien51:
# Description

`.enable` has been deprecated even though it is still used by WalletConnect and WalletLink.  I removed from the "default" path and just added it to both of these.

# Issues

Fixes #11374

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useAuthenticate.ts
+++ b/unlock-app/src/hooks/useAuthenticate.ts
@@ -72,7 +72,15 @@ export function useAuthenticate(options: AuthenticateProps = {}) {
     const walletConnectProvider = new WalletConnectProvider({
       rpc: rpcForWalletConnect(config),
     })
-    return authenticate(walletConnectProvider)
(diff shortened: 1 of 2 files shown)


searchableguy (approved):

searchableguy:
enabled was an alias for `ethereum.request({ method: 'eth_requestAccounts' })`

julien51:
> enabled was an alias for ethereum.request({ method: 'eth_requestAccounts' }) - would this work for all of the above? If so, wouldn't it be better to use this?

Well we actually do that in the walletService... so I don't think it's worth doing it twice. The weird thing is that WalletConnect and WalletLink still are waiting for this `enable` call to actually show their UI...

PR #11383, opened 2023-03-27
Ccarfi grant templates

ccarfi:
added grant templates for grant application and grant completion


.github/ISSUE_TEMPLATE/grant_application.md
```diff
@@ -0,0 +1,79 @@
...
+
+If you have a detailed pitch deck, please provide a link or attach as a comment in the discussion of this proposal
+
+## What dev languages or frameworks will you use?
+
+Development languages that will be used
+
+## Is this project open source?
+
+Yes | No
```
julien51, line 55:
```suggestion
* [ ] Yes
```

.github/ISSUE_TEMPLATE/grant_completion.md
```diff
@@ -0,0 +1,70 @@
+---
```
julien51, line 1: I am not sure we will need that anymore because IMO it's better if all of the grant lifecycle happens on the same Github issue. If we have a special template for completion, this would likely open another issue which is not what we'd want. 

That said I think it would be good to have a way to ask the user to complete that as part of the initial grant application thread/issue?

PR #11392, opened 2023-03-28
feat(locksmith): Add support for dynamic pricing

### LINKED ISSUE #10472: Use purchasePriceFor when computing credit card prices
Our contract will return the price that a specific user has to pay for memberships on a given lock.
We should use that rather than `keyPrice`.

searchableguy:
# Description

# Issues

Fixes #10472

# Checklist:

### DIFF
--- a/locksmith/src/utils/pricing.ts
+++ b/locksmith/src/utils/pricing.ts
@@ -277,7 +277,7 @@ export const createTotalCharges = async ({
       gasCost,
       unlockServiceFee: 0,
       creditCardProcessingFee: 0,
-      isCreditPurchasable: false,
+      isCreditCardPurchasable: false,
     }
   }
   const subtotal = Math.round(pricing.priceInAmount * 100)
@@ -288,7 +288,7 @@ export const createTotalCharges = async ({
   const result = {
     ...fees,
     subtotal,
-    isCreditPurchasable: fees.total > 50,
+    isCreditCardPurchasable: fees.total > 50,
   }
   return result
 }
@@ -309,6 +309,6 @@ export const createPricingForPurchase = async (options: KeyPricingOptions) => {
     ...fees,
     recipients,
     gasCost,
-    isCreditPurchasable: fees.total > 50,
+    isCreditCardPurchasable: fees.total > 50,
   }
 }
--- a/locksmith/src/utils/pricing.ts
+++ b/locksmith/src/utils/pricing.ts
@@ -2,6 +2,7 @@ import { networks } from '@unlock-protocol/networks'
 import { Web3Service, getErc20Decimals } from '@unlock-protocol/unlock-js'
 import { ethers } from 'ethers'
 import logger from '../logger'
+import GasPrice from './gasPrice'
 
 // Stripe's fee is 30 cents plus 2.9% of the transaction.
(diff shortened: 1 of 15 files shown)


julien51 (approved):

locksmith/src/utils/pricing.ts
```diff
@@ -57,8 +75,176 @@ export async function defiLammaPrice({
...
+  const serviceFee = getUnlockServiceFee(subtotal)
+  const total = subtotal + serviceFee
+  // This is rounded up to an integer number of cents.
+  const percentageFee = Math.ceil(total * stripePercentage)
+  return baseStripeFee + percentageFee
+}
+
+// Fee denominated in cents
+export const getUnlockServiceFee = (cost: number) => {
+  return Math.ceil(cost * 0.1) // Unlock charges 10% of transaction.
```
julien51, line 231: Maybe add the cost of gas here... ideally as a simulation of the tx to be sent?

julien51 (commented):
This is looking good but missing one significant part : gas.

In an ideal world we would run a simulation to get the actual amount in the chain's native currency and then we convert this to $.

locksmith/src/utils/pricing.ts
```diff
@@ -57,8 +75,176 @@ export async function defiLammaPrice({
...
+
+const fromDecimal = (num: string, decimals: number) => {
+  return parseFloat(
+    ethers.utils
+      .formatUnits(ethers.BigNumber.from(num), decimals)
+      .replace(/\.0$/, '')
+  )
+}
+
+export const getLockKeyPricing = async ({
```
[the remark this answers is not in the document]
searchableguy: So I tried simulation and put the code inside another branch. It becomes to slow to be usable. 30+ seconds on the full API call. 

I think we will simulate it on frontend for checking errors so we can get the gas estimate from there and pass. I changed the API to use frontend values for calculating the total price and for the actual verification - we do calculate on backend on confirmation.

locksmith/src/utils/pricing.ts
searchableguy: Added the same gas cost above we have on API.

locksmith/src/utils/pricing.ts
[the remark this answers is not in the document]
julien51: Hum, I don't understand why it would take 30+ seconds because it should just be 1 RPC call and it returns the amount of gas consumed?

locksmith/src/utils/pricing.ts
```diff
@@ -1,4 +1,14 @@
...
+import { Web3Service, getErc20Decimals } from '@unlock-protocol/unlock-js'
+import { ethers } from 'ethers'
+import logger from '../logger'
+import GasPrice from './gasPrice'
+
+// Stripe's fee is 30 cents plus 2.9% of the transaction.
+const baseStripeFee = 30
+const stripePercentage = 0.029
+export const GAS_COST = 200000 // hardcoded : TODO get better estimate, based on actual execution
+export const GAS_COST_TO_GRANT = 250000
```
julien51, line 11: We already define this somewhere else, please re-use otherwise it is going to be a pain to maintain. (mauyve mobve to constants if that's easier, same for vars above)

PR #11405, opened 2023-03-28
feat(unlock-app) - hotfix sendemail

kalidiagne:
# Description
Correctly pass `newtorkId` for send email

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/operations/wedlocksOperations.ts
+++ b/locksmith/src/operations/wedlocksOperations.ts
@@ -31,6 +31,7 @@ type Params = {
   keychainUrl?: string
   lockName: string
   network: string
+  networkId: number
   lockAddress: string
   txUrl?: string
   openSeaUrl?: string
@@ -78,7 +79,7 @@ export const sendEmail = async (
   // prevent send email when is not enabled
   const { sendEmail: canSendEmail } = await getLockSettings(
     params.lockAddress,
-    Number(params.network)
+    Number(params.networkId)
   )
 
   if (!canSendEmail) {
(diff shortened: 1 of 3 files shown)


locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -31,6 +31,7 @@ type Params = {
   keychainUrl?: string
   lockName: string
   network: string
+  networkId: number
```
julien51, line 34: We already have network. Why are you passing networkId too?

julien51 (changes requested):
@kalidiagne Please explain hat you are trying to do and why you are not using `network` ?

kalidiagne:
> @kalidiagne Please explain hat you are trying to do and why you are not using `network` ?

we are passing network name and not the id, we need the id to correctly use it in the queries 

ref: 
```
 networkId: network!,
 network: networks[network!]?.name ?? '',
```

locksmith/src/operations/wedlocksOperations.ts
kalidiagne: we are passing network name and not the id, we need the id to correctly use it in the queries

ref:

```
 networkId: network!,
 network: networks[network!]?.name ?? '',
 ```

julien51: But the templates don't need it! I think we should not pollute the `params` object with stuff that's not needed by the templates...

julien51 (approved):

PR #11411, opened 2023-03-29
fix(unlock-app): EthPass button only shown if user is logged in

julien51:
# Description

EthPass requires a user signer from the `owner` of the key. So we are not showing the EthPass buttons when the connected user is not the owner of the token.

# Checklist:

### DIFF
--- a/packages/networks/src/networks/goerli.ts
+++ b/packages/networks/src/networks/goerli.ts
@@ -43,7 +43,7 @@ export const goerli: NetworkConfig = {
   startBlock: 7179039,
   previousDeploys: [],
   isTestNetwork: true,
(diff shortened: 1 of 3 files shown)


packages/networks/src/networks/goerli.ts
```diff
@@ -43,7 +43,7 @@ export const goerli: NetworkConfig = {
   startBlock: 7179039,
   previousDeploys: [],
   isTestNetwork: true,
-  maxFreeClaimCost: 10000,
+  maxFreeClaimCost: 100000,
```
julien51, line 46: Bumped becuase gas is high rigth now on Goerli

PR #11426, opened 2023-03-30
feat(unlock-app, networks)Add link to OpenSea profile

kalidiagne:
# Description

Screenshot 2023-03-30 at 16 59 06

Screenshot 2023-03-30 at 16 59 12

# Issues

Fixes #
Refs #

# Checklist:


searchableguy (approved):

julien51:
Hum, I don't think it makes sense to put the link there... Because that menu is for things that are specific to the key itself.
I think a much better place for this would be at the top of the page, below the "keychain" menu.
Screenshot 2023-03-30 at 12 18 57 PM

julien51:
Maybe adding a menu like this one on the right with the OpenSea logo to link to OpenSea and maybe even to Etherscan for that address?
Screenshot 2023-03-30 at 12 19 38 PM

kalidiagne:
> Hum, I don't think it makes sense to put the link there... Because that menu is for things that are specific to the key itself. I think a much better place for this would be at the top of the page, below the "keychain" menu. Screenshot 2023-03-30 at 12 18 57 PM

Not sure, the thing is for which network? Also for the `etherscan` link, if we put it outside the card, we need to have something generic, and the links are different for every network (eg. `https://goerli.etherscan.io/` vs `https://mumbai.polygonscan.com/`)

kalidiagne:
> > Not sure, the thing is for which network?
> 
> Good point. since the link is the same for _all_ networks but testnets let's just use that?
> 
> As for Etherscan, let's use `https://blockscan.com/address/...` like we already do on the members page

**OpenSea**

Screenshot 2023-04-03 at 13 09 23

**Blockscout**

Screenshot 2023-04-03 at 13 09 29

julien51:
Very cool! Thanks!

julien51 (approved):

unlock-app/src/components/content/KeychainContent.tsx
```diff
@@ -2,13 +2,44 @@ import React from 'react'
...
+                rel="noreferrer"
+                className="hover:text-brand-ui-primary"
+              >
+                <WorldIcon size={25} />
+              </a>
+            </Tooltip>
+
+            <Tooltip tip="View Opensea Profile" label="View Opensea Profile">
+              <a
+                href={`https://opensea.io/${account}`}
```
julien51, line 32: Actually can use please add this as a te,plate in the networks file?

unlock-app/src/components/content/KeychainContent.tsx
```diff
@@ -2,13 +2,44 @@ import React from 'react'
...
   return (
     <AppLayout
-      title="Member Keychain"
+      title={
+        <div className="flex justify-between">
+          <h1 className="text-4xl font-bold">Member Keychain</h1>
+          <div className="flex gap-3">
+            <Tooltip tip="" label="">
+              <a
+                href={`https://blockscan.com/address/${account}`}
```
julien51, line 21: Please add this to the networks configs?

julien51 (approved):

PR #11510, opened 2023-04-01
fix(smart-contracts): `receive()` function safety upgrade

abdulhakim-altunkaya:
# Adding a require() function inside receive() function to prevent zero or negative transactions. A safety precaution against malicious or unexpected behaviour.

# receive() function can be exloited or it can cause unexpected behaviour.

Fixes #
Refs #

# Checklist:


abdulhakim-altunkaya:
CLA signed

abdulhakim-altunkaya:
Hello Clement,

Sure, I agree with you about customs errors. They are better alternatives
compared to require. I used "require" just to draw attention to that small
improvement area.

And about the exploit, receive() function can be triggered even with zero
value transactions. So, when a zero-value transaction is sent to a
contract, it still requires a certain amount of gas to be processed by the
network. This means that if an attacker sends a large number of zero-value
transactions to a contract, it can quickly consume all available gas
for the block, preventing other transactions from being executed. This
attack is called as Denial-of-Service (DoS) attack.

In other words, if an attacker sends a large number of zero value
transactions to a contract, the gas consumed by these transactions will add
up and eventually exceed the maximum gas limit for the current block. This
will cause the remaining transactions in the block to fail and any changes
made by those transactions will be undone. In this way, the attacker can
cause a Denial-of-Service (DoS) attack by preventing other transactions
from being executed.

Also, I could not see a function (at least in Unlock.sol) which will later
help you to withdraw the funds coming with "receive()"

clemsos:
Yes, the network DDOS makes sense. Actually the `receive` function is not in use anymore (as you noticed) - as the logic to unwrap wETH has been moved to another standalone contract. Thanks for the scrutiny, that's very useful !

Closing this in favour of #11511 11511

clemsos (approved):
LG!

PR #11540, opened 2023-04-05
show recurring only when applicable

kalidiagne:
# Description

**Lock with recurring applicable**

Screenshot 2023-04-05 at 12 18 56

**Lock without recurring applicable**

Screenshot 2023-04-05 at 12 10 41

# Issues

Fixes #
Refs #

# Checklist:


julien51 (commented):
Did you check that the checkout still works when recurring is set to 0? 
I suspect this may have broken things.

unlock-app/src/components/interface/locks/CheckoutUrl/elements/LocksForm.tsx
```diff
@@ -90,11 +98,39 @@ export const LocksForm = ({
...
   const [recurringUnlimited, setRecurringUnlimited] = useState(false)
+  const [lockRecurring, setLockRecurring] = useState<RecurringByLock>({})
 
   const { getIsRecurringPossible } = useLockSettings()
 
   const [locks, setLocks] = useState<LocksProps>(locksDefault)
 
+  // preload default and set recurring (es. saved config)
+  useEffect(() => {
+    const getRecurringCb = async () => {
```
julien51, line 109: Please use a better function name

julien51: Also I think I have already seen you write code to check if a lock is recurring somewhere? Are you sure your code is not duplicated?

kalidiagne: We need both: 
- when we add a new lock to the list, check if recurring is possible 
- for an existing saved config with locks, we fetch with the code above

kalidiagne:
> Did you check that the checkout still works when recurring is set to 0? I suspect this may have broken things.

Yes, there is no issue on this side.
The only is on Metamask side with the `default value` on approve: without setting the default value but a custom value it fails :/ (also when renewals > 0)

julien51 (approved):

julien51 (approved):

PR #11548, opened 2023-04-05
fix(locksmith) : adding migration to move from integer to biginteger

julien51:
# Description

The Palm chain id is bigger than Postgres' INTEGER :/

# Checklist:

- [] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/locksmith/__tests__/websub/renewKey.test.ts
+++ b/locksmith/__tests__/websub/renewKey.test.ts
@@ -5,7 +5,7 @@ import { vi } from 'vitest'
 import { renewKey, isWorthRenewing } from '../../src/websub/helpers/renewKey'
 
 const renewalInfo = {
-  network: 31137,
+  network: 11297108109,
   keyId: '1',
   lockAddress: '0xaaa',
 }
@@ -64,15 +64,6 @@ const mockWalletService = {
     hash: 'txhash',
   }),
 }
-
-vi.mock('@unlock-protocol/networks', () => {
-  return {
-    default: {
-      1: {},
-      31137: {},
-    },
-  }
-})
 
 vi.mock('ethers', async () => {
   const original = await vi.importActual<any>('ethers')
--- a/locksmith/migrations/20230405185518-change-integer-to-big-integer.js
+++ b/locksmith/migrations/20230405185518-change-integer-to-big-integer.js
@@ -68,6 +68,9 @@ module.exports = {
       }),
       queryInterface.changeColumn('LockSettings', 'network', {
         type: 'pg_chain_id',
+      }),
+      queryInterface.changeColumn('HookEvents', 'network', {
+        type: 'pg_chain_id',
       })
     ])
   },
@@ -136,6 +139,9 @@ module.exports = {
         type: Sequelize.INTEGER,
       }),
       queryInterface.changeColumn('LockSettings', 'network', {
+        type: Sequelize.INTEGER,
+      }),
+      queryInterface.changeColumn('HookEvents', 'network', {
         type: Sequelize.INTEGER,
       })
     ])
--- a/locksmith/migrations/20230405185518-change-integer-to-big-integer.js
+++ b/locksmith/migrations/20230405185518-change-integer-to-big-integer.js
@@ -3,78 +3,81 @@
 /** @type {import('sequelize-cli').Migration} */
 module.exports = {
   async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.query('CREATE DOMAIN pg_chain_id AS BIGINT;')
     return Promise.all([
-      queryInterface.changeColumn('Blocks', 'chain', {
-        type: Sequelize.BIGINT,
-      }),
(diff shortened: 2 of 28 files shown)


julien51:
Tested locally and it runs...  I want to see it run on staging!

locksmith/migrations/20230405185518-change-integer-to-big-integer.js
```diff
@@ -0,0 +1,134 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    return Promise.all([
+      queryInterface.changeColumn('Transactions', 'chain', {
```
searchableguy, line 7: Are we doing this for tables which are not used anymore? If so, for consistency - there is also `Blocks`.

julien51: Good catch! Let me add `Blocks`! (and yes for consistency!).

julien51:
Hum. I do not like these tests failing. Investigating.

locksmith/migrations/20230405185518-change-integer-to-big-integer.js
julien51: I can't because it is a primary key... also not a huge deal since we don't use this anymore?

locksmith/migrations/20230405185518-change-integer-to-big-integer.js
```diff
@@ -0,0 +1,144 @@
...
+    await Promise.all([
+      // Can't change as it is a primary key
+      // queryInterface.changeColumn('Blocks', 'chain', {
+      //   type: Sequelize.INTEGER,
+      // }),
+      queryInterface.changeColumn('Transactions', 'chain', {
+        type: Sequelize.INTEGER,
+      }),
+      queryInterface.changeColumn('Locks', 'chain', {
+        type: Sequelize.INTEGER,
```
searchableguy, line 85: Why not use our new type for consistency so there is no mismatch in migration either?

locksmith/src/models/sequelize.ts
```diff
@@ -1,9 +1,43 @@
...
+
+// @ts-expect-error Property 'postgres' does not exist on type 'typeof import("/Users/julien/repos/unlock/node_modules/sequelize/types/data-types")'.
+const PgTypes = DataTypes.postgres
+
+class PgNetworkId extends PgTypes.BIGINT {
+  static parse(x: string): number {
+    return Number.parseInt(x)
+  }
+}
+
```
searchableguy, line 35: Confused about this pg part. Is this needed? Shouldn't sequelize class handle parsing?

locksmith/src/models/sequelize.ts
```diff
@@ -1,9 +1,43 @@
...
+ *
+ */
+// @ts-expect-error Base constructors must all have the same return type.
+class NETWORK_ID extends DataTypes.BIGINT {
+  // Optional: parser for values received from the database
+  static parse(value: string): number {
+    return Number.parseInt(value)
+  }
+}
+
```
searchableguy, line 24: Do we also need to change validate, _sanitize, and stringify to only accept numbers and then convert to string before passing?

locksmith/migrations/20230405185518-change-integer-to-big-integer.js
julien51: That is the `down` (revert) part of the migration so it goes back to its original state :)

locksmith/src/models/sequelize.ts
julien51: Technically things will break when the `value` is for a number that's larger than JS's Number limit. For the rest I am ok to rely on `BIGINT`...

julien51: Actually let me throw here when the number is too large.

locksmith/src/models/sequelize.ts
julien51: Yes. Sequelize needs both a "generic" type and a parsing _per_ database engine.

locksmith/__tests__/websub/renewKey.test.ts
```diff
@@ -5,7 +5,7 @@ import { vi } from 'vitest'
 import { renewKey, isWorthRenewing } from '../../src/websub/helpers/renewKey'
 
 const renewalInfo = {
-  network: 31137,
```
julien51, line 8: Added this to make sure the fix works (and it does!)

locksmith/__tests__/websub/renewKey.test.ts
```diff
@@ -65,15 +65,6 @@ const mockWalletService = {
   }),
 }
 
-vi.mock('@unlock-protocol/networks', () => {
```
julien51, line 68: Removed... we can use the package!

searchableguy (approved):

PR #11567, opened 2023-04-06
feat(locksmith) - Generate certification image

kalidiagne:
# Description

Add endpoint to generate certification 

- [x] Add new endpoint to generate certificates
- [x] Add tests 
- [x] Refactor `Certificate` component with inline style 
- [x] Add email attachments 
 
Screenshot 2023-04-11 at 20 30 34

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/utils/certification.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { addressMinify } from './strings'
+import satori from 'satori'
+import dayjs from 'dayjs'
+
+import { readFileSync } from 'fs'
+const inter400 = readFileSync('src/fonts/inter-400.woff')
+const inter700 = readFileSync('src/fonts/inter-700.woff')
+
+export const MAX_UINT =
+  '115792089237316195423570985008687907853269984665640564039457584007913129639935'
```
searchableguy, line 16: move it to a separate file? I thought this was already a constant.

locksmith/src/utils/certification.tsx
```diff
@@ -0,0 +1,104 @@
...
+        width: '100vw',
+      }}
+    >
+      <Certificate
+        name={metadata?.name || ''}
+        description={metadata.description}
+        tokenId={metadata?.tokenId}
+        owner={metadata?.owner}
+        issuer={object?.certification_issuer}
+        image={metadata?.image}
```
searchableguy, line 76: satori earlier had some issues with fetching mime data. Is this working properly now after update? Can you try with a few different image/metadata?

locksmith/src/utils/certification.tsx
kalidiagne: not in locksmith, only in `unlock-app`, maybe we can add a shared lib later

locksmith/src/utils/strings.ts
```diff
@@ -0,0 +1,5 @@
+export function addressMinify(address: string): string {
+  return `${address?.substring(0, 6)}...${address?.substring(
+    address.length - 4
+  )}`
+}
```
searchableguy, line 5: remove. Import from ui library 'minifyAddress'

kalidiagne: oh awesome, importing from there :)

locksmith/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const MAX_UINT =
```
julien51, line 1: Do we not have this already? If not please use `ethers` which already defines it.

locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -214,6 +215,20 @@ const getAttachments = async ({
...
+  if (isCertification && network && tokenId) {
+    const certificate = await createCertificate({
+      network,
+      lockAddress,
+      tokenId,
+    })
+    const svg = new resvg.Resvg(certificate)
+    const pngData = svg.render()
+    const pngBuffer = pngData.asPng()
+    const dataURI = `data:image/png;base64,${pngBuffer.toString('base64')}`
```
julien51, line 228: Why do you need to render as png and not use svg since you are doing a data uri anyway?

locksmith/src/utils/certification.tsx
```diff
@@ -0,0 +1,117 @@
...
+        description={metadata.description}
+        tokenId={metadata?.tokenId}
+        owner={metadata?.owner}
+        issuer={object?.certification_issuer}
+        image={imageBase64}
+        lockAddress={minifyAddress(lockAddress)}
+        network={network}
+        networkName={networks[network]?.name}
+        // can't use minifyAddress because is checking if the address is valid
+        transactionsHash={`${transactionHash.slice(
```
julien51, line 89: Please create a util function and add to the core package?

locksmith/src/utils/image.ts
```diff
@@ -0,0 +1,14 @@
+export const imageToBase64 = async (url: string) => {
+  const response = await fetch(url, {
```
julien51, line 2: Why are we fetching the image? It wastes resources because this image is already stored here, Let's just retrieve it.

packages/ui/lib/components/Certificate/Certificate.tsx
```diff
@@ -18,11 +17,12 @@ interface CertificateProps {
   badge?: string
   transactionsHash?: ReactNode
   externalUrl?: string
+  isMobile?: boolean
 }
 
 interface Props {
   children?: ReactNode
-  size?: Size
+  size?: string
```
julien51, line 25: Why the change?

locksmith/src/utils/certification.tsx
kalidiagne: we have it already in `ui`:

```
export const minifyAddress = (address: string) => {
  const checked = ethers.utils.isAddress(address)
  return checked
    : address
}
```
should not work for transaction hash because we verify that we pass an address for minify

locksmith/constants.ts
kalidiagne: not present in `locksmith`, replacing with `ethers`

locksmith/src/operations/wedlocksOperations.ts
kalidiagne: refactored, but still i think we need to keep it as is, the SVG need to be decoded otherwise

i think @searchableguy  can give more details about the implementation here

packages/ui/lib/components/Certificate/Certificate.tsx
kalidiagne: ah! restored

locksmith/src/operations/wedlocksOperations.ts
searchableguy: I'm confused. If you are talking about the reason why I convert svg to png for tickets, that's for displaying the ticket in emails and make it easy to print.

locksmith/src/utils/certification.tsx
julien51: ok but there is no value in displaying a truncated version if there is no link, because nobody can do anything useful with a truncated hash... so either you show the full hash or you point to the block explorer.

kalidiagne: In this case for the generated image the link is not present, but is present on the certification detail https://github.com/unlock-protocol/unlock/blob/master/unlock-app/src/components/content/certification/CertificationDetails.tsx#L215

Issue #11555, opened 2023-04-05
Single Stripe account


julien51:
When connecting to Stripe, if the user has previously connected any of their lock, let's let them select that Stripe Account, rather than ask them to go thru the Stripe onboarding again.

PR #11575, opened 2023-04-07
feature(unlock-app, locksmith, unlock-js): Single Stripe account

julien51:
# Description

Letting lock managers re-use a single stripe account for all their locks.

Refs #11555

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/stripeController.ts
+++ b/locksmith/src/controllers/v2/stripeController.ts
@@ -48,13 +48,15 @@ export const getConnectionsForManager = async (req: Request, res: Response) => {
     userAddress
   )
   if (!connections) {
-    return []
+    return {
+      result: [],
+    }
   }
-  return res.json(
-    connections.map(({ lock, chain, stripeAccount }) => ({
+  return res.json({
+    result: connections.map(({ lock, chain, stripeAccount }) => ({
       lock,
       chain,
       stripeAccount,
-    }))
-  )
+    })),
+  })
 }
--- a/locksmith/src/controllers/v2/stripeController.ts
+++ b/locksmith/src/controllers/v2/stripeController.ts
@@ -8,6 +8,7 @@ export const connectStripe = async (req: Request, res: Response) => {
   const network = Number(req.params.network)
   const userAddress = req.user!.walletAddress
   const baseUrl = Normalizer.getURL(req.body.baseUrl)?.toString()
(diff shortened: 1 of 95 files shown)


locksmith/src/operations/stripeOperations.ts
```diff
@@ -135,54 +136,74 @@ export const disconnectStripe = async ({
 }
 
 /**
- * @deprecated
```
julien51, line 138: I don't think this is deprecated @searchableguy ? It is the method used by our main flow to connect Stripe accounts...

searchableguy: We have a v2 version. This is deprecated. Check v2 folder.

searchableguy: smh, we have too many connectStripe

https://github.com/unlock-protocol/unlock/blob/173cb49f9bc07bbc15b854639061ba87a94a6d64/locksmith/src/controllers/lockController.ts#L10

locksmith/src/controllers/v2/stripeController.ts
```diff
@@ -36,3 +41,20 @@ export const connectStripe = async (req: Request, res: Response) => {
...
+
+export const getConnectionsForManager = async (req: Request, res: Response) => {
+  const userAddress = req.user!.walletAddress
+  const connections = await stripeOperations.getConnectionsForManager(
+    userAddress
+  )
+  if (!connections) {
+    return []
+  }
+  return res.json(
```
searchableguy, line 53: want to send something like `result`. We use this everywhere to be consistent so if we have any error but not complete failure, we could attach it as error or additional information later without changing the format.

julien51: yes I can! Good point. Thanks!

unlock-app/src/components/interface/locks/Settings/forms/CreditCardForm.tsx
```diff
@@ -124,11 +129,20 @@ export const CreditCardForm = ({
     }
   )
 
+  const {
+    isFetching: isFetchingPreviouslyConnectedLocks,
+    data: previouslyConnectedLocks,
+  } = useQuery(['connectedLocks'], async () => {
```
searchableguy, line 135: Pass the account since it is dependent on account for cache.

searchableguy (approved):

PR #11586, opened 2023-04-10
feat(unlock-app, unlock-js, locksmih) - strict check lock type from attributes

kalidiagne:
# Description
We need a better way to check if a lock is an event/certification ecc...
Added an endpoint that returns the lock types based on the attributes properties.

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -192,8 +195,10 @@ const getAttachments = async ({
     attachments.push({ path: dataURI })
   }
 
+  const { isEvent } = types ?? {}
+
   // Add ICS attachment when event is present
-  if (event) {
+  if (isEvent && event) {
```
kalidiagne, line 201: add `event` attachment only when lock is eventType

unlock-app/src/components/content/event/EventDetails.tsx
```diff
@@ -74,7 +79,7 @@ export const EventDetails = ({ lockAddress, network }: EventDetailsProps) => {
     )
   }
 
-  if (!metadata?.attributes) {
+  if (!metadata && !types?.isEvent) {
```
kalidiagne, line 82: show details only when lock `isEvent` and have metadata

locksmith/src/controllers/v2/metadataController.ts
```diff
@@ -246,3 +246,17 @@ export const updateUsersMetadata: RequestHandler = async (
     error,
   })
 }
+
+export const getLockTypeByMetadata: RequestHandler = async (
```
julien51, line 250: Let's do this check on the front-end

julien51 (changes requested):
It's a good idea but there is no reason to do it on locksmith. Can we do it on the front-end please? This will save resources and reduce our reliance on the backend.

kalidiagne:
> It's a good idea but there is no reason to do it on locksmith. Can we do it on the front-end please? This will save resources and reduce our reliance on the backend.

Actually, we need to do the check also from locksmith, to check for the event type before adding the attachments.

unlock-app/src/components/content/event/utils.tsx
```diff
@@ -24,3 +25,26 @@ export const getEventEndDate = (ticket: any): Date | null => {
...
   return null
 }
+
+export interface LockTypes {
+  isEvent: boolean
+  isCertification: boolean
+  isStamp: boolean
+}
+
+export const getLockTypeByMetadata = (
```
kalidiagne, line 35: check on frontend with metadata without endpoint call

locksmith/src/operations/metadataOperations.ts
```diff
@@ -272,3 +272,32 @@ export const getLockMetadata = async ({
...
   return lockData?.data
 }
+
+export interface LockTypes {
+  isEvent: boolean
+  isCertification: boolean
+  isStamp: boolean
+}
+
+export const getLockTypeByMetadata = async ({
```
kalidiagne, line 282: helper to check metadata attributes on locksmith

kalidiagne:
> It's a good idea but there is no reason to do it on locksmith. Can we do it on the front-end please? This will save resources and reduce our reliance on the backend.

Update and remove the endpoint

locksmith/src/operations/metadataOperations.ts
```diff
@@ -272,3 +272,32 @@ export const getLockMetadata = async ({
...
+    return attributes.some((attribute) =>
+      attribute?.trait_type?.startsWith(name)
+    )
+  }
+
+  return {
+    isEvent: hasAttribute('event'),
+    isCertification: hasAttribute('certification'),
+    isStamp: hasAttribute('stamp'),
+  }
```
searchableguy, line 302: Wanna move this to `core` library?

julien51:
> Actually, we need to do the check also from locksmith, to check for the event type before adding the attachments.

Ok but then do not make an API for this. Just a re-usable library.

packages/core/src/utils/index.ts
```diff
@@ -0,0 +1,24 @@
+import { z } from 'zod'
+
+const LockTypeSchema = z.object({
+  isEvent: z.boolean().default(false),
+  isCertification: z.boolean().default(false),
+  isStamp: z.boolean().default(false),
+})
+
+export const getLockTypeByMetadata = (metadata: any): LockType => {
```
julien51, line 9: Please add tests for this.

locksmith/src/operations/wedlocksOperations.ts
julien51: Great!

julien51 (approved):
Please add tests to the `core` library function you're adding.
This is pretty central.

kalidiagne:
Test in place!

packages/core/package.json
```diff
@@ -4,13 +4,18 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "build": "tsc -p ."
+    "build": "tsc -p .",
+    "test": "UNLOCK_ENV=test vitest run --coverage --environment=node"
```
julien51, line 8: Please also add `ci` like we do for other packages (and make sure it runs the linter as well!)

kalidiagne:
includes also https://github.com/unlock-protocol/unlock/issues/11594

julien51:
I don't think tests are running for the core package

PR #11587, opened 2023-04-10
feat(unlock-app) - max keys per address fix

kalidiagne:
# Description
By default, `MAX_KEY_PER_ADDRESS` is set to 1, we don't need to show the toggle, should be enable by default 

**Default value**

Screenshot 2023-04-10 at 16 59 47

**Value updated**

Screenshot 2023-04-10 at 17 02 22

# Issues

Fixes #
Refs #

# Checklist:


kalidiagne:
> Did you check that changing the value works too?

i cant reproduce the issue with value changing, can give me more details ?

julien51:
I think there is an issue when the user is on the wrong network. We should just prompt the user switch network and then send the transaction,

Please check that other settings are not broken in the same way as well. ok?

Issue #11551, opened 2023-04-05
Key Manager is wrong!


julien51:
For [this lock](https://goerli.etherscan.io/address/0x38181253e065b18df0006644f5fa9c2fda46cc75) (but it is not specific to this lock!) on key `2`, the Key Manager is this: 

<img width="1351" alt="Screenshot 2023-04-05 at 5 59 38 PM" src="https://user-images.githubusercontent.com/17735/230221301-ffc0ae17-1915-41f7-8760-78bcfd809f1c.png">

But our dashboard shows:

<img width="1147" alt="Screenshot 2023-04-05 at 5 59 53 PM" src="https://user-images.githubusercontent.com/17735/230221375-511c1b8c-03ab-4263-b39f-0f066d3be95a.png">

kalidiagne:
I think issue is related to the subgraph, for the same key the lock manager is `null`

<img width="1310" alt="Screenshot 2023-04-06 at 01 42 48" src="https://user-images.githubusercontent.com/20865711/230237232-7ba2e158-c271-4d02-b7f8-76848bc5a9bc.png">

clemsos:
I think this was introduced by https://github.com/unlock-protocol/unlock/pull/11356/files

clemsos:
Looking into it, the subgraph may ecounter a sort of "race condition" when parsing events in the same tx. 

The keyManager for token 2 was set at creation (see [this tx ](https://goerli.etherscan.io/tx/0x4e4f57a44b316db662338b8214814c8c177e806e59943aa9c402166b4053a870#eventlog ). Both `Transfer` and `KeyManagerChanged` are fired together. We check that the key exists when storing our key manager here

https://github.com/unlock-protocol/unlock/blob/d9f5b835e0bda800da2e6d568daa5720d29c5cc1/subgraph/src/public-lock.ts#L212-L215

What could have happen is that the key still not created when the `KeyManagerChanged` was parsed, leading the field to stay `null`.

ot 100% sure, but that looks like it

julien51:
I think the easiest solution here is to read from the chain rather than use the event.

julien51:
Reading further I don't think it is a race condition because we don't even set the manager on `Transfer`...

PR #11636, opened 2023-04-13
fix(locksmith): fixing isAirdroppedRecipient and improved error messages

julien51:
# Description

`isAirdroppedRecipient` is either because the email address matches or because the manager is the key manager contract.

This PR also improves the log messages so they are more readable..

# Checklist:

### DIFF
--- a/locksmith/src/operations/wedlocksOperations.ts
+++ b/locksmith/src/operations/wedlocksOperations.ts
@@ -374,7 +374,7 @@ export const notifyNewKeyToWedlocks = async (
 
   const isAirdroppedRecipient =
     airdroppedRecipient?.toLowerCase() === ownerAddress?.toLowerCase() ||
-     manager?.toLowerCase()?.trim() ===
+    manager?.toLowerCase()?.trim() ===
       networks[network!]?.keyManagerAddress?.toLowerCase()?.trim()
 
   logger.info(`Sending ${recipient} key: ${lockAddress}-${tokenId}`)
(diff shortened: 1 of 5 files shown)


locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -373,7 +373,7 @@ export const notifyNewKeyToWedlocks = async (
   })
 
   const isAirdroppedRecipient =
-    airdroppedRecipient?.toLowerCase() === ownerAddress?.toLowerCase() &&
+    airdroppedRecipient?.toLowerCase() === ownerAddress?.toLowerCase() ||
     manager?.toLowerCase().trim() ===
       networks[network!]?.keyManagerAddress?.toLowerCase()?.trim()
```
searchableguy, line 378: Is this intended? We changed this to make sure manager also matches.

locksmith/src/operations/wedlocksOperations.ts
```diff
@@ -373,7 +373,7 @@ export const notifyNewKeyToWedlocks = async (
   })
 
   const isAirdroppedRecipient =
-    airdroppedRecipient?.toLowerCase() === ownerAddress?.toLowerCase() &&
+    airdroppedRecipient?.toLowerCase() === ownerAddress?.toLowerCase() ||
     manager?.toLowerCase().trim() ===
```
searchableguy, line 377:
```suggestion
     manager?.toLowerCase()?.trim() ===
```

locksmith/src/operations/wedlocksOperations.ts
julien51: Yes it is. Since the `manager` is not set correctly right now in the subgraph see https://github.com/unlock-protocol/unlock/issues/11551 that @kalidiagne is working on we need to do an OR.

Once the subgraph has been fixed, we can remove the check on the matching recipient address.

searchableguy (approved):

PR #11637, opened 2023-04-13
fix(ui): adding QueryClient inside of AddressComponent

julien51:
# Description

This should remove the need to downgrade.
I am surprised the library can't use the provider from `_app`, but I think it is cleaner this way anyway because it won't expect other apps to have a provider set

# Checklist:


julien51:
> Wouldn't it be better to remove the use of react-query from ui library? It should be refactored out. We already pass onResolveName as function.

I guess it _could_ work but we would still need to component to know when resolution is "happening" so it changes its UI... so we would need to implement a logic similar to one of the mutation function...

julien51:
I'll merge to unblock... we can still change the approach later.

PR #11638, opened 2023-04-13
feat(subgraph) - set key manager when key is mined

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/subgraph/src/helpers.ts
+++ b/subgraph/src/helpers.ts
@@ -75,3 +75,16 @@ export function loadOrCreateUnlockDailyData(
 
   return unlockDailyData
 }
+
+export function getKeyManagerOf(
+  lockAddress: Address,
+  tokenId: BigInt,
+  owner: Address
+): Address {
+  const lockContract = PublicLockV7.bind(lockAddress)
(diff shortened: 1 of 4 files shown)


subgraph/src/helpers.ts
```diff
@@ -75,3 +75,16 @@ export function loadOrCreateUnlockDailyData(
 
   return unlockDailyData
 }
+
+export function getKeyManagerOf(
+  lockAddress: Address,
+  tokenId: BigInt,
+  owner: Address
+): Address {
+  const lockContract = PublicLockV7.bind(lockAddress)
```
julien51, line 84: I am a bit confused about why you use `PublicLockV7` here? But i think that could work? @clemsos wdyt?

julien51 (approved):
I _think_ that is good. Would love to get a quick check by @clemsos

subgraph/src/helpers.ts
kalidiagne: there is no difference in the implementation of `keyManagerOf` from `v7` and `v11`, will works fine

clemsos (approved):
LGTM

PR #11643, opened 2023-04-13
feature(networks, unlock-app) : publicLockVersionToDeploy

julien51:
# Description

Adding the version for each network so we can make deploy progressively...

# Issues

Fixes #
Refs #

# Checklist:


kalidiagne (approved):

julien51:
> is it possible that we could have one version on 13 and another on 12? Or normally should be always the same version for all network

Yes it is possible when we're doing upgrades as we can't guarantee they will all happen at the same time.

unlock-app/src/components/interface/locks/Create/CreateLock.tsx
```diff
@@ -95,7 +96,7 @@ export const CreateLockSteps = () => {
             : maxNumberOfKeys,
           currencyContractAddress,
           keyPrice: keyPrice?.toString(),
-          publicLockVersion: 12, // default to latest version supported by Unlock JS
+          publicLockVersion: networks[network].publicLockVersionToDeploy,
```
searchableguy, line 99: Add a fallback too?

searchableguy (approved):

clemsos (approved):
LG 👍

PR #11655, opened 2023-04-14
feat(locksmith) - correctly check for event

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/operations/eventOperations.ts
```diff
@@ -52,8 +55,7 @@ export const getEventDetail = async (
     )
   }
 
-  // This is an event, collect event information
-  if (attributes) {
+  if (types.isEvent) {
```
kalidiagne, line 58: We already have this to prevent but is still needed here in the case we move this function 
```
// get event details only when lock is event
  if (isEvent) {
    eventDetail = await getEventDetail(lockAddress, network)
  }
```

julien51 (approved):

PR #11664, opened 2023-04-15
fix(smart-contracts): proper Lock upgrade tests

julien51:
# Description

Adding more tests for upgrades.
These are failing when they should not.

# Checklist:


smart-contracts/test/Unlock/upgradeLock.js
```diff
@@ -130,3 +130,114 @@ describe('upgradeLock (deploy template with Proxy)', () => {
     assert.equal(await lock.sayHello(), 'hello world')
   })
 })
+
+
+describe('upgrades', async () => {
```
julien51, line 135: I added this which deploys a lock v9 and upgrades consecutively... It does not test everything (we can add later more things) but it currently fails, which helps debug.

clemsos:
Need to rebase once https://github.com/unlock-protocol/unlock/pull/11665 is merged

smart-contracts/test/Lock/upgrades/v13.js
```diff
@@ -101,7 +106,8 @@ describe('PublicLock upgrade v12 > v13', () => {
 
       // deploy new implementation
       lock = await upgrades.upgradeProxy(lock.address, PublicLockLatest, {
-        unsafeSkipStorageCheck: true, // UNSECURE - but we need the flag as we are resizing the `__gap`
```
clemsos, line 104: here is one main reason we didn't catch that error... 😓

julien51: Actually I think we need to change `scripts/lock/testUpgrade.js` to make sure it runs with any/all versions

clemsos: In the test/Lock/upgrades we have all versions increment so that should cover it, but yes we should merge all that into a single one. The only thing is that we may have specific migrate logic so that's why they are separated

PR #11680, opened 2023-04-19
feat(unlock-js): Add more fields to lock queries

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/packages/unlock-js/src/subgraph/schema.graphql
+++ b/packages/unlock-js/src/subgraph/schema.graphql
@@ -17,9 +17,12 @@ query allLocks(
     name
     expirationDuration
     tokenAddress
+    symbol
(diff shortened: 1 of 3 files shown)


packages/unlock-js/src/subgraph/schema.graphql
```diff
@@ -17,9 +17,12 @@ query allLocks(
     name
     expirationDuration
     tokenAddress
+    symbol
```
julien51, line 20: @searchableguy Note that the symbol here is the ERC721 symbol (`KEY` by default), not the currency symbol... I do think it would be useful to add the `currencyName` and `currencySymbol` too though.

PR #11683, opened 2023-04-19
feat(subgraph) - fix broken receipts

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:


subgraph/src/public-lock.ts
```diff
@@ -440,11 +441,17 @@ export function createReceipt(event: ethereum.Event): void {
       // including one for the ERC20 transfer
       for (let i = 0; i < logs.length; i++) {
         const txLog = logs[i]
-        if (
-          txLog.address == tokenAddress &&
```
kalidiagne, line 444: `txLog.address == tokenAddress` this check should not be right looking at this logs we have 

**address**: 0x38181253e065b18df0006644f5fa9c2fda46cc75
**hashed signature**: 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef

https://goerli.etherscan.io/tx/0xaa9a7112a5a2738e6d45b3966f2df358d25c535fa53441639cf2ab5ecedad4ed#eventlog

subgraph/src/public-lock.ts
```diff
@@ -440,11 +441,17 @@ export function createReceipt(event: ethereum.Event): void {
...
-          // Do we always have txLog.topics[0] ?
-          txLog.topics[0].toHexString() == ERC20_TRANSFER_TOPIC0
-        ) {
+
+        const hashedSignature = txLog.topics[0].toHexString()
+        const decodedSignature = ethereum.decode(
+          '(address,address,uint256)',
+          txLog.data
+        )
```
kalidiagne, line 449:
```suggestion
        const decodedSignature = ethereum.decode(
          '(address,address,uint256)',
          txLog.data
        )
```

@julien51 @clemsos can i review here?

subgraph/src/public-lock.ts
```diff
@@ -440,11 +441,17 @@ export function createReceipt(event: ethereum.Event): void {
...
+
+        const isTransfer =
+          hashedSignature === decodedSignature!.toAddress().toHexString()
+
```
kalidiagne, line 453:
```suggestion
        const hashedSignature = txLog.topics[0].toHexString()
        const decodedSignature = ethereum.decode(
          '(address,address,uint256)',
          txLog.data
        )

        const isTransfer =
          hashedSignature === decodedSignature!.toAddress().toHexString()

```

@julien51 @clemsos can i review here?

subgraph/src/public-lock.ts
julien51: @kalidiagne indeed, this transaction is NOT on an ERC20 lock... so the `payer` should be the sender of the transaction... so the problem is not about the `Transfer` event!

subgraph/src/public-lock.ts
julien51: That is not good. The signature of the event includes the event name `Transfer(address,address,uint256)` as we discussed earlier today.

kalidiagne:
PR updated with fix

julien51 (changes requested):
@kalidiagne Can you please add tests for this?
I want to avoid regressions here. I believe we already have some tests for that functions.

PR #11705, opened 2023-04-26
fix(locksmith): adding timeout middleware

julien51:
# Description

Let's timeout _inside_ the app so we have a lot more control... and avoid having HEroku kill our processes because they timeout.

# Checklist:

### DIFF
--- a/locksmith/package.json
+++ b/locksmith/package.json
@@ -52,6 +52,7 @@
     "@unlock-protocol/ui": "workspace:^",
     "@unlock-protocol/unlock-js": "workspace:./packages/unlock-js",
     "color-scheme": "1.0.1",
+    "connect-timeout": "1.9.0",
     "cookie-parser": "1.4.6",
     "cors": "2.8.5",
     "dayjs": "1.11.7",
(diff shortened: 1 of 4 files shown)


locksmith/src/app.ts
```diff
@@ -7,9 +7,12 @@ import * as Tracing from '@sentry/tracing'
 import cookieParser from 'cookie-parser'
 import router from './routes'
 import { errorHandler } from './utils/middlewares/error'
+import timeout from 'connect-timeout'
 
 const app = express()
 
+app.use(timeout('3s'))
```
searchableguy, line 14: Add this to config file and increase it to 25s? We have some slow endpoints. Until we can optimize them, shouldn't break them.

searchableguy (approved):

locksmith/package.json
```diff
@@ -44,6 +44,7 @@
     "@resvg/resvg-js": "2.4.1",
     "@sentry/node": "7.48.0",
     "@sentry/tracing": "7.48.0",
+    "@types/connect-timeout": "^0.0.36",
```
searchableguy, line 47:
```suggestion
    "@types/connect-timeout": "0.0.36",
```

locksmith/src/app.ts
julien51: ok I think we should try to identify these slow endpoints and see how to speed them up :)

PR #11717, opened 2023-04-28
feature(smart-contracts): Card purchaser

julien51:
# Description

Introducing the card purchaser contract.
This contract has a single (stateless) purpose: spend a user's USDC to buy from a lock:
It take 2 signatures:
- one by the user to spend USDC
- one by the user to purchase from a lock
as well as a `callData` from a lock.


julien51:
Tests coming once I get a first validation about the approach!

smart-contracts/contracts/CardPurchaser.sol
```diff
@@ -0,0 +1,209 @@
...
+  /**
+   * Constructor
+   * We double the `name` and `version` fields from EIP712 so we can query the contract to get them and OZ's Ownable does not expose them
+   */
+  constructor(
+    address _owner,
+    address _unlockAddress,
+    address _usdc
+  ) EIP712("Card Purchaser", "1") Ownable() {
+    name = "Card Purchaser";
```
clemsos, line 57: You should be able to access is using `super.name` no?

clemsos (approved):
LGTM

smart-contracts/contracts/CardPurchaser.sol
julien51: Let me try but `name` is a `private` var in `EIP712` :(

PR #11726, opened 2023-05-01
fix(unlock-app): Fixing UX for credit cards connections

julien51:
# Description

We now show the stripe account name for clarity.

# Checklist:

### DIFF
--- a/packages/unlock-js/openapi.yml
+++ b/packages/unlock-js/openapi.yml
@@ -2450,14 +2450,9 @@ paths:
                 type: object
                 properties:
                   result:
-                    type: object
-                    properties:
-                      lock:
-                        type: string
-                      chain:
-                        type: number
-                      stripeAccount:
-                        type: string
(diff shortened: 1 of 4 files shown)


packages/unlock-js/openapi.yml
```diff
@@ -2450,14 +2450,9 @@ paths:
...
-                    properties:
-                      lock:
-                        type: string
-                      chain:
-                        type: number
-                      stripeAccount:
-                        type: string
+                    type: array
+                    items:
+                      type: object
```
searchableguy, line 2455: maybe declare the type for object here?

searchableguy (approved):

packages/unlock-js/openapi.yml
julien51: It's actually a pretty deep object (stripe API) and was not sure of the right approach :(

Issue #10583, opened 2022-12-14
Handling non-convertible currencies for Credit card payments


julien51:
When locks with "exotic" currencies are used we cannot get a $ price which means the credit card payments won't work :(
We should do the following:
* All lock managers to set a $ price to be charged (regardless of the conversion price)
* Use the hardcoded conversion price "first" and if none is set, use conversion
* If we don't have a conversion price, indicate it on the payment settings.

julien51:
As a follow up please look at this as well https://github.com/unlock-protocol/unlock/issues/12167

PR #11831, opened 2023-05-08
feat(locksmith, unlock-app) - handle credit card custom price BE

kalidiagne:
# Description

Handle credit card custom price as part of https://github.com/unlock-protocol/unlock/issues/10583 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/utils/keyPricer.ts
+++ b/locksmith/src/utils/keyPricer.ts
@@ -12,6 +12,7 @@ import {
   baseStripeFee,
   GAS_COST_TO_GRANT,
 } from './constants'
+import * as lockSettingOperations from '../operations/lockSettingOperations'
 
 const ZERO = ethers.constants.AddressZero
 
@@ -56,6 +57,16 @@ export default class KeyPricer {
     if (lock.keyPrice === '0') {
       return 0
     }
+
+    const { creditCardPrice } = await lockSettingOperations.getSettings({
+      lockAddress,
+      network,
+    })
+
+    if (creditCardPrice) {
+      return parseInt((creditCardPrice * 100).toFixed(0)) // returns the USD cents price
+    }
+
     const priceConversion = new PriceConversion()
     const usdPrice = await priceConversion.convertToUSD(
       symbol.toUpperCase(),
--- a/locksmith/src/utils/pricing.ts
+++ b/locksmith/src/utils/pricing.ts
@@ -4,6 +4,8 @@ import { ethers } from 'ethers'
 import logger from '../logger'
 import GasPrice from './gasPrice'
 import { GAS_COST, stripePercentage, baseStripeFee } from './constants'
+import * as lockSettingOperations from '../operations/lockSettingOperations'
+import * as Normalizer from '../utils/normalizer'
 
 export interface Options {
   amount?: number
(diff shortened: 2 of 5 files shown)


locksmith/src/utils/keyPricer.ts
```diff
@@ -56,6 +57,16 @@ export default class KeyPricer {
...
       return 0
     }
+
+    const { creditCardPrice } = await lockSettingOperations.getSettings({
+      lockAddress,
+      network,
+    })
+
+    if (creditCardPrice) {
+      return parseInt((creditCardPrice * 100).toFixed(0)) // returns the USD cents price
```
kalidiagne, line 67: return `creditCardPrice`only when is set from settings

locksmith/src/utils/pricing.ts
```diff
@@ -275,7 +296,11 @@ export const createTotalCharges = async ({
       isCreditCardPurchasable: false,
     }
   }
-  const subtotal = Math.round(pricing.priceInAmount * 100)
+
+  // prioritize to credit card total if present
+  const price = creditCardTotal || pricing.priceInAmount
```
kalidiagne, line 301: `price` from for `createTotalCharges` and used for stripe charge, we use creditCardTotal as default or fallback to the converted one

packages/unlock-js/openapi.yml
```diff
@@ -2676,6 +2676,20 @@ paths:
...
 
+        - in: query
+          name: lockAddress
+          required: false
+          description: The address of the lock
+          schema:
+            type: string
+
+        - in: query
+          name: keysToPurchase
```
kalidiagne, line 2687: need this to know how many keys we are purchasing in order to have the total price to charge in `USD`

unlock-app/src/hooks/useFiatChargePrice.ts
```diff
@@ -6,6 +6,8 @@ interface Options {
   tokenAddress?: string
   amount: number
   enabled?: boolean
+  lockAddress?: string
```
julien51, line 9: Since these are optional let's just add that change to +another+ PR to make sure we're not breaking anything?

julien51 (changes requested):
I am confused about why you had to add `lockSettingOperations.getSettings` in 2 different places?

We should have a *single* function on the backend that yields prices and that function should be the one in which we do `lockSettingOperations.getSettings`.

Can you refactor to simplify the code please?

kalidiagne:
> I am confused about why you had to add `lockSettingOperations.getSettings` in 2 different places?
> 
> We should have a _single_ function on the backend that yields prices and that function should be the one in which we do `lockSettingOperations.getSettings`.
> 
> Can you refactor to simplify the code please?

I can simplify by adding `priceOperations` here but we still need to change in 2 different places: 

in the UI we use this 2 endpoints actually:

1)`/v2/api/purchase/total` (Get the total price for a purchase (including fees))
 - Used to show credit card fees breakdown 
 - Includes values as total, subtotal, gasCost, creditCardProcessingFee 
 

2)`'/fiat/:lockAddress'` (keyPricer.ts) -> use `keyPricer` this class has this note
```
"@deprecated - Remove once no longer used anywhere. Use functions in pricing.ts instead"
```
but is not still replaced everywhere yet, needs a bit refactor for it the best thing is to add it in a different PR later.

kalidiagne:
will be fixed here https://github.com/unlock-protocol/unlock/pull/11905

PR #11845, opened 2023-05-10
refactor(unlock-app): simplified the stepper

julien51:
# Description

As I was working on the credit stuff I simplified our stepper.

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/Stepper.tsx
+++ b/unlock-app/src/components/interface/checkout/Stepper.tsx
@@ -80,21 +80,16 @@ export interface StepItem {
 
 interface StepperProps {
   items: StepItem[]
-  position: number
   service: CheckoutService | UnlockAccountService
   disabled?: boolean
 }
 
-export const Stepper = ({
-  items,
-  position,
-  service,
-  disabled,
-}: StepperProps) => {
-  const index = items.findIndex((item) => item.id === position)
+export const Stepper = ({ items, service, disabled }: StepperProps) => {
+  const index = items.findIndex((item) => item.to === service.state?.value)
   const step = items[index]
   const base = items.slice(0, index).filter((item) => !item?.skip)
   const rest = items.slice(index + 1).filter((item) => !item?.skip)
+
   return (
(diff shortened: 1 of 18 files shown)


unlock-app/src/components/interface/checkout/Stepper.tsx
```diff
@@ -79,22 +81,26 @@ export interface StepItem {
...
-  position,
-  service,
-  disabled,
-}: StepperProps) => {
-  const index = items.findIndex((item) => item.id === position)
+export const Stepper = ({ service, disabled }: StepperProps) => {
+  const [state] = useActor(service)
+
+  const isUnlockAccount = service.id === 'unlockAccount'
+  // @ts-expect-error Property 'renew' does not exist on type 'UnlockAccountMachineContext'.
```
julien51, line 92: I am not sure how to avoid these @searchableguy ?

searchableguy: I think you need instance of or type of check to narrow down here.

unlock-app/src/components/interface/checkout/Stepper.tsx
```diff
@@ -79,22 +81,26 @@ export interface StepItem {
...
+  const isRenew = service.id === 'checkout' && !!state.context?.renew
+
+  const items = useStepperItems(service, { isUnlockAccount, isRenew })
+
+  const index = items.findIndex(
```
searchableguy, line 97: Why not use find instead of find index?

unlock-app/src/components/interface/checkout/Stepper.tsx
julien51: > 'CheckoutService' only refers to a type, but is being used as a value here.ts(2693)

unlock-app/src/components/interface/checkout/Stepper.tsx
julien51: becuase we actually need the index to construct the `base` and `rest` arrays...

PR #11880, opened 2023-05-12
feature(locksmith): Universal card locksmith pricing

julien51:
# Description

Adding the pricing API for universal credit cards.

# Checklist:


locksmith/src/controllers/v2/priceController.ts
```diff
@@ -36,3 +42,56 @@ export const total: RequestHandler = async (request, response) => {
...
+ * @param request
+ * @param response
+ * @returns
+ */
+export const universalCard: RequestHandler = async (request, response) => {
+  const network = Number(request.params.network)
+  const lockAddress = request.params.lock
+  const { recipients: recipientQ = [], purchaseData = [] } = request.query
+
+  // Setup values
```
julien51, line 58: It is so annoying that we have to do this because `recipientQ` can be a string pr an array...

Is there a way to "force" it to be an array @searchableguy ?

searchableguy: You would need type the request type which is also annoying so nope.

searchableguy (approved):
Approving but open API definition?

julien51:
HA! Good point. Let me pull that and add it here

(I wish we could automate!)

searchableguy (approved):

PR #11893, opened 2023-05-15
fix(unlock-js): use introspection JSON to generate service

clemsos:
# Description

Generate graphql introspection JSON directly from Goerli subgraph and check that JSON into subgraph repo. Use that local JSON file to parse Unlock JS subgraph graphql client, so we dont depend on Goerli during the tests/build process

replace #11858

closes #10521

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/subgraph/bin/introspect.js
+++ b/subgraph/bin/introspect.js
@@ -1,0 +1,31 @@
+const path = require('path')
+const fs = require('fs-extra')
+const { getIntrospectionQuery } = require('graphql')
+const { networks } = require('@unlock-protocol/networks')
+
+const introspectionJSONFilePath = path.join(
+  __dirname,
+  '..',
+  'introspection.json'
+)
+
+const endpoint = networks['5'].subgraph.endpointV2
+console.log(`Fetching graphql introspection from ${endpoint}`)
+async function main() {
+  const query = getIntrospectionQuery()
+  const req = await fetch(endpoint, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+    },
+    body: JSON.stringify({ query }),
+  })
+
+  const json = JSON.parse(await req.text())
+  await fs.writeJSON(introspectionJSONFilePath, json, { spaces: 2 })
+}
+
+main()
+  .then(() => console.log(`Saved at ${introspectionJSONFilePath}`))
+  .catch((err) => console.error(err))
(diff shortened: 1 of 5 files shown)


julien51 (approved):

julien51:
I will merge!

subgraph/package.json
```diff
@@ -17,6 +17,7 @@
     "create-local": "graph create --node http://localhost:8020/ ",
     "remove-local": "graph remove --node http://localhost:8020/ ",
     "deploy-local": "graph deploy --node http://localhost:8020/ --ipfs http://localhost:5001",
+    "introspect": "node bin/introspect.js",
```
julien51, line 20: Actually can we make this run before _any_ `yarn build`?

subgraph/bin/introspect.js
```diff
@@ -0,0 +1,31 @@
...
+
+const introspectionJSONFilePath = path.join(
+  __dirname,
+  '..',
+  'introspection.json'
+)
+
+const endpoint = networks['5'].subgraph.endpointV2
+console.log(`Fetching graphql introspection from ${endpoint}`)
+async function main() {
```
julien51, line 14: Can we just make this run every time but skip if CI=true?

julien51: This way we can make sure never have an outdated version?
And also ideally instead of running agasing Goerli, we run against the "local" subgraph?

clemsos: hmmm i think we can add a flag to run against localhost but we should not default to localhost, as this will require running the whole stack locally to have a deployed subgraph and then query it for introspection.

What we can also add is a recurring ci task (such as the one for yarn lock or npm publish) that will run on prod branch and open a PR if the graph introspection file is outdated.

I think both of these are doable but can be addressed in another PR

subgraph/package.json
clemsos: will add rn

subgraph/bin/introspect.js
clemsos: added a `yarn introspect:local` task that can be used locally to fetch against localhost

clemsos:
I added a `INTROSPECT_NETWORK` env to pick a subgraph to query from its chain id (default to Goerli). 

Made me think that we could use this to check consistency btw all subgraphs by doing on prod branch : 

1) deploying localhost version (which should be latest) to get "authority" introspection json file
2) looping through all networks and getting introspection json file for each
3) comparing them and checking if there are any differences

Also that will address https://github.com/unlock-protocol/unlock/pull/11893#discussion_r1194228577 as we can update the json in the repo if its outdated

PR #11894, opened 2023-05-15
feat(unlock-js) - add `transferFrom` to walletService

kalidiagne:
# Description
- add `transferFrom` to walletService, needed for https://github.com/unlock-protocol/unlock/issues/11836 

# Issues

Fixes #
Refs #

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [x] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #11299, opened 2023-03-10
fix(unlock-app) : unlimited memberships can still be soulbound

julien51:
# Description

Fixing a bug where unlimited memberships could not be made soulbound.
They just can't have fees on transfers...

# Issues

Fixes #
Refs #

# Checklist:

- [x] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [x] I have performed a self-review of my own code
- [x] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #11895, opened 2023-05-15
feat(unlock-app) - let lock manager transfer a key from members list

kalidiagne:
# Description
- [x] ability to transfer key by lock manager from the members list

needs https://github.com/unlock-protocol/unlock/pull/11894 

**Max keys reached**

Screenshot 2023-06-09 at 11 31 24

**Address is owner** 

Screenshot 2023-06-09 at 11 31 39

**Transfer ownership from the members list**

Screenshot 2023-06-09 at 11 13 20

Screenshot 2023-06-09 at 11 13 15

**Transfer ownership disabled (soul-bound)**

Screenshot 2023-06-09 at 11 18 11

**Ownership transferring** 

Screenshot 2023-06-09 at 12 03 19

# Issues

Fixes #
Refs #

# Checklist:


julien51:
>  ability for key owner to transfer key from keychain (@julien51 keep or remove this one? is added as extra)

It is a good idea but there are more things to check here... for example if the key is actually transferable and if the owner of the key is indeed the manager for example.

unlock-app/src/components/interface/keychain/Key.tsx
```diff
@@ -306,6 +316,23 @@ function Key({ ownedKey, account, network }: Props) {
                       </MenuButton>
                     )}
                   </Menu.Item>
+                  {!isKeyExpired && (
```
julien51, line 319: We need to check more things here, as the key may _not_ be transferable. for example please check that the key owner is also the key manager (or no key manager is set), as well as that there is no 100% fee on transfers!

unlock-app/src/components/interface/keychain/TransferKeyDrawer.tsx
```diff
@@ -0,0 +1,155 @@
...
+      to: newOwner!,
+      lockAddress,
+      tokenId,
+    })
+  }
+
+  const transferFromMutation = useMutation(onTransferFrom)
+
+  const onTransferKeyOwnership = async () => {
+    const promise = transferFromMutation.mutateAsync()
```
julien51, line 82:
```suggestion
    const transferPromise = transferFromMutation.mutateAsync()
```

unlock-app/src/components/interface/keychain/TransferKeyDrawer.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    await ToastHelper.promise(promise, {
+      loading: `Transferring key ownership to ${addressMinify(newOwner)}.`,
+      success: 'Key successfully transferred.',
+      error: 'There is an issue with transfer, please try again.',
```
julien51, line 87: We may need to show more details about why it failed.

kalidiagne: Agree, for this I think we can open a different PR and handle it for all the`ToastHelper.promise`

kalidiagne:
i mark it as `Blocked` until we fix the strange behavior on the tests as discussed yesterday in the Eng. Meeting with @clemsos

unlock-app/src/components/interface/locks/Manage/elements/MetadataCard.tsx
```diff
@@ -350,9 +364,20 @@ export const MetadataCard = ({
 
   const ownerIsManager = owner?.toLowerCase() === manager?.toLowerCase()
   const showManager = !ownerIsManager && manager !== ADDRESS_ZERO
+  const canTransferKeyOwnership =
+    isLockManager && isTransferAllowed && !isLoadingTransferFee
```
kalidiagne, line 368: make sure that transfer is visible only if the key is transferable & the user is the lock manager

kalidiagne:
PR update and make more clear the buttons with actions, in some case we only have just `Change` but we can have different buttons in one line

**Update** 
Screenshot 2023-06-09 at 11 13 20

unlock-app/src/hooks/useTransferFee.ts
```diff
@@ -29,3 +31,32 @@ export const useTransferFee = ({ lockAddress, network }: TransferFeeProps) => {
...
+      const [transferFeeBasisPoints, lock] = await Promise.all([
+        web3Service.transferFeeBasisPoints(lockAddress, network),
+        web3Service.getLock(lockAddress, network),
+      ])
+
+      const transferFeePercentage = (transferFeeBasisPoints ?? 0) / 100
+      const unlimitedDuration =
+        lock?.expirationDuration === UNLIMITED_KEYS_DURATION
+
+      // unlimited memberships could not be made soul-bound
```
julien51, line 52: I dont understand this? 
Unlimited means they don't expire. Why would they be non transferable?

unlock-app/src/components/interface/keychain/TransferKeyDrawer.tsx
```diff
@@ -0,0 +1,170 @@
...
+      const total = await web3Service.totalKeys(lockAddress, newOwner, network)
+
+      return maxKeysPerAddress <= total
+        ? 'This address already owns max number of keys.'
+        : true
+    }
+    return true
+  }
+
+  const isTransferable = async () => {
```
julien51, line 119: Rather than handle this here, you should **not** show the button on the keychain when transfers are not possible.

julien51 (changes requested):
This needs more work

unlock-app/src/components/interface/keychain/TransferKeyDrawer.tsx
kalidiagne: this is already handled in the frontend, this is an extra check

unlock-app/src/hooks/useTransferFee.ts
kalidiagne: ref from this PR https://github.com/unlock-protocol/unlock/pull/11299 

I was added `unlimitedDuration` with `allowTransfer`

julien51: That still does not make sense. The fact that they do not expire does not mean they can';t be transferred. It just means the transfer cannot have fees. 

Lock manager should be able to transfer a key, whether there are fees or not.

kalidiagne:
> This needs more work

PR updated and removed `useGetTransferFeeBasisPoints`, i will open a different one for this and use it for where is needed

PR #11905, opened 2023-05-16
feat(locksmith) - credit card pricing BE

kalidiagne:
# Description

To support for custom credit card, `pricingOperations` with `getUsdPricingForLock` was added; this will prioritize the credit card price if set. Code also changed changed for support  custom credit card for this endpoints : 

- `/:network/price` endpoint 
- `/purchase/total`
- `/price/:network/:lock/card`

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/priceController.ts
+++ b/locksmith/src/controllers/v2/priceController.ts
@@ -1,17 +1,13 @@
 import { RequestHandler } from 'express'
-
-import {
-  createPricingForPurchase,
-  createTotalCharges,
-  defiLammaPrice,
-} from '../../utils/pricing'
+import * as priceOperations from '../../operations/pricingOperations'
+import { createPricingForPurchase } from '../../utils/pricing'
 import { ethers } from 'ethers'
 import { getCreditCardEnabledStatus } from '../../operations/creditCardOperations'
 import * as Normalizer from '../../utils/normalizer'
 import { Web3Service } from '@unlock-protocol/unlock-js'
 import networks from '@unlock-protocol/networks'
+import { MIN_PAYMENT_STRIPE } from '../../utils/constants'
 
-const MIN_PAYMENT_STRIPE = 100
 export const amount: RequestHandler = async (request, response) => {
   const network = Number(request.params.network || 1)
   const amount = parseFloat(request.query.amount?.toString() || '1')
@@ -20,7 +16,7 @@ export const amount: RequestHandler = async (request, response) => {
     ? erc20Address
     : undefined
 
-  const result = await defiLammaPrice({
+  const result = await priceOperations.getUsdPricingForLock({
     network,
     amount,
     address,
@@ -38,9 +34,9 @@ export const total: RequestHandler = async (request, response) => {
     ? erc20Address
     : undefined
 
-  const charge = await createTotalCharges({
-    network,
+  const charge = await priceOperations.getTotalCharges({
     amount,
+    network,
     address,
   })
 
@@ -110,7 +106,7 @@ export const isCardPaymentEnabledForLock: RequestHandler = async (
   const web3Service = new Web3Service(networks)
   const lock = await web3Service.getLock(lockAddress, network)
 
-  const result = await defiLammaPrice({
+  const result = await priceOperations.getDefiLammaPrice({
     network,
     address: lock?.currencyContractAddress,
     amount: Number(`${lock.keyPrice}`),
--- a/locksmith/src/operations/pricingOperations.ts
+++ b/locksmith/src/operations/pricingOperations.ts
@@ -1,0 +1,173 @@
+import networks from '@unlock-protocol/networks'
+import { getFees, getGasCost } from '../utils/pricing'
+import * as lockSettingOperations from './lockSettingOperations'
+import { MIN_PAYMENT_STRIPE } from '../utils/constants'
+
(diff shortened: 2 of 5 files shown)


locksmith/src/utils/constants.ts
```diff
@@ -3,3 +3,4 @@ export const baseStripeFee = 30
 export const stripePercentage = 0.029
 export const GAS_COST = 200000 // hardcoded : TODO get better estimate, based on actual execution
 export const GAS_COST_TO_GRANT = 250000
+export const MIN_PAYMENT_STRIPE = 0.5 // 0.50 $
```
kalidiagne, line 6: @julien51 i see that you set this value as `100`, around the app we use `50 cents` as value? keep 1$ or 0.5$ here?

locksmith/src/operations/pricingOperations.ts
```diff
@@ -0,0 +1,173 @@
...
+  network: number
+}
+
+export type PriceResults = Partial<
+  Price & {
+    priceInAmount: number
+  }
+>
+
+export async function getDefiLammaPrice({
```
kalidiagne, line 38: `defiLammaPrice` renamed and moved here

locksmith/src/operations/pricingOperations.ts
```diff
@@ -0,0 +1,173 @@
...
+    ...item,
+    priceInAmount,
+  }
+}
+
+/**
+ * Get lock pricing with priority to credit card total if present
+ * @returns
+ */
+export async function getUsdPricingForLock({
```
kalidiagne, line 95: this is used where we get the pricing with priority to `credit card`

locksmith/src/operations/pricingOperations.ts
```diff
@@ -0,0 +1,173 @@
...
+ */
+export const getTotalCharges = async ({
+  lockAddress,
+  network,
+  address,
+  amount = 1,
+  keysToPurchase = 1,
+}: LockPricingProps): Promise<LockPricingValues> => {
+  const [pricing, gasCost] = await Promise.all([
+    getUsdPricingForLock({
```
kalidiagne, line 139:
```suggestion
    getUsdPricingForLock({
```

replaced `defiLammaPrice` in order to get custom credit card if present here

locksmith/src/utils/constants.ts
kalidiagne: Screenshot 2023-05-12 at 21 55 22

locksmith/src/operations/pricingOperations.ts
```diff
@@ -0,0 +1,173 @@
...
+  if (lockAddress) {
+    const { creditCardPrice } = await lockSettingOperations.getSettings({
+      lockAddress,
+      network,
+    })
+
+    if (creditCardPrice) {
+      return {
+        confidence: 1,
+        priceInAmount: creditCardPrice * keysToPurchase,
```
searchableguy, line 112: You need to pass a few more things to be consistent with defilamma pricing output here.

locksmith/src/controllers/v2/priceController.ts
```diff
@@ -20,10 +16,20 @@ export const amount: RequestHandler = async (request, response) => {
     ? erc20Address
     : undefined
 
-  const result = await defiLammaPrice({
+  const _lockAddress = request.query.lockAddress?.toString()
+  const keysToPurchase = Number(request.query.keysToPurchase?.toString() || 1)
```
searchableguy, line 20: I'm confused. Isn't this what amount is for?

searchableguy (changes requested):
Need more clarification on two things.

locksmith/src/controllers/v2/priceController.ts
kalidiagne: `amount` is the amount related to the lock price eg . `50 USDC` has `50` as amount, and this is used to get the correct price conversion 

`keysToPurchase` in the numbers of keys we are purchasing, is needed to multiply the custom credit card price to have the correct final price

locksmith/src/operations/pricingOperations.ts
kalidiagne: actually non needed the return type is 

```
export type PriceResults = Partial<
  Price & {
    priceInAmount: number
  }
>
```

and is the same for `getUsdPricingForLock` and `getDefiLammaPrice`

searchableguy: That is only if it cannot get the price which is why it is partial otherwise if it can find the price, it will return all so you do need to provide.

kalidiagne: updated!

locksmith/src/controllers/v2/priceController.ts
julien51: @kalidiagne Please let's use a better name for that function. I agree with @searchableguy that this is unclear.

locksmith/src/utils/constants.ts
julien51: I am confused? Where did I set 100?

julien51 (commented):
Please split this PR. I don't understand what it does right now.

kalidiagne:
changes in this PR https://github.com/unlock-protocol/unlock/pull/11973

PR #11914, opened 2023-05-17
feat(locksmith) - rename address to erc20address

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/packages/unlock-js/openapi.yml
+++ b/packages/unlock-js/openapi.yml
@@ -1324,7 +1324,7 @@ paths:
             type: number
 
         - in: query
-          name: erc20Address
+          name: address
           required: false
           description: The erc20 address you want to get the price for. If nothing is passed, the API will return the native token price.
           schema:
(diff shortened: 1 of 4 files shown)


packages/unlock-js/openapi.yml
```diff
@@ -1324,7 +1324,7 @@ paths:
             type: number
 
         - in: query
-          name: address
+          name: erc20Address
```
searchableguy, line 1327: I would not change these. Just the internal naming, that's all.

kalidiagne: It will not be a breaking change here we use it with the `operationId` in `unlock-app`

searchableguy: It is still a breaking change externally outside of our storage service and I don't see good reason to do that so I will keep it as it is but internally refactor.

kalidiagne: restored

Issue #11889, opened 2023-05-12
Show payment options based on user address


julien51:
In some cases, a lock may use a hook that "discounts" the price of a membership for a specific address.
Espceially, some locks will make the memberships "free" when they use a certain lock (for example thru an allowList or because their wallet matches certain criteria).
When that's the case we should show the claim option to users.

PR #11915, opened 2023-05-17
feat: Allow claim when discounted price is 0

searchableguy:
# Description

fixes https://github.com/unlock-protocol/unlock/issues/11889

# Issues

Fixes #
Refs #

# Checklist:


locksmith/src/controllers/purchaseController.ts
```diff
@@ -313,35 +327,49 @@ export class PurchaseController {
...
+            userAddress: recipient,
+            network,
+            data: parsed?.data?.[index] || '0x',
+            referrer: parsed.referrers?.[index] || recipient,
+          })
+        })
+      )
+
+      const totalAmount = prices.reduce((a, b) => a + b, 0)
+      console.log(totalAmount, prices)
```
julien51, line 348:
```suggestion
```

unlock-app/src/components/interface/checkout/main/Payment.tsx
```diff
@@ -90,21 +109,19 @@ export function Payment({ injectedProvider, checkoutService }: Props) {
...
-  const isWaiting = isLoading || isClaimableLoading || isBalanceLoading
+  const isWaiting = isLoading || isBalanceLoading
 
   const isReceiverAccountOnly =
     recipients.length <= 1 &&
     recipients[0]?.toLowerCase() === account?.toLowerCase()
 
   const enableCrypto = !isUnlockAccount || !!balance?.isPayable
 
-  const forceClaim = lock.network === 42161
```
searchableguy, line 101: Removed this since we are no longer subsidising on that network.

locksmith/src/utils/claim.ts
```diff
@@ -0,0 +1,36 @@
...
+}: Options) => {
+  const web3Service = new Web3Service(networks)
+  const prices = await Promise.all(
+    recipients.map((recipient, index) => {
+      return web3Service.purchasePriceFor({
+        lockAddress,
+        userAddress: recipient,
+        network,
+        data: data?.[index] || '0x',
+        referrer: recipient,
```
julien51, line 26: In the future we may wat to use the *real* referrer, because in theory the lock could change its price based on it... but let's not worry too much about it right now

unlock-app/src/components/interface/checkout/main/Payment.tsx
```diff
@@ -63,17 +64,35 @@ export function Payment({ injectedProvider, checkoutService }: Props) {
...
-  })
-
   const { isLoading: isBalanceLoading, data: balance } = useBalance({
     account: account!,
     network: lock.network,
     currencyContractAddress: lock.currencyContractAddress,
   })
 
+  const { data: purchaseData, isLoading: isPurchaseDataLoading } =
+    usePurchaseData({
```
julien51, line 74: I assume this is cached right? we seem to do that in many many places :)

julien51 (approved):
Can we also hide the card payment when claim is possible to avoid users being confused?

unlock-app/src/components/interface/checkout/main/Payment.tsx
searchableguy: Yes

locksmith/src/utils/claim.ts
searchableguy: I actually have the code for this. I added it previously but then removed it because when we do any TX ourselves, we use our purchaser address as referrer.

PR #11916, opened 2023-05-17
refactor(unlock-app): dry-ing up SIWE messages

julien51:
# Description

All in the title!

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useSIWE.tsx
+++ b/unlock-app/src/hooks/useSIWE.tsx
@@ -1,7 +1,7 @@
 import { ReactNode, createContext, useContext, useState } from 'react'
 import { useSession } from './useSession'
 import { useAuth } from '~/contexts/AuthenticationContext'
-import { SiweMessage } from 'siwe'
+import { SiweMessage, generateNonce } from 'siwe'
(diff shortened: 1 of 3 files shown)


unlock-app/src/hooks/useSIWE.tsx
```diff
@@ -118,10 +123,13 @@ export const SIWEProvider = ({ children }: Props) => {
       await queryClient.refetchQueries()
       await refetchSession()
```
searchableguy, line 124: You need to refactor this signIn into two functions - signInMessage and signIn. The signIn function correctly also re-authentication with our locksmith using the new message which we shouldn't because we wouldn't allow anyone to change that message.

searchableguy (changes requested):
Need to refactor a little.

unlock-app/src/hooks/useSIWE.tsx
julien51: good point! We actually need to talk about this more tomorrow during our 1:1 because I realize if locksmith is not running users can't event login which I think we should avoid,

searchableguy (approved):

PR #11930, opened 2023-05-18
feat(locksmith) - change expiration type

kalidiagne:
# Description

By testing `email notification` I realize that the current type of `expiration` is incorrect.

When the value of expiration is too big, it gives an error, as documentation suggest i change it to string 

> If your SQL type supports integer values outside this range, we recommend using [bigint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) or [string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) to represent your integers.

@searchableguy need to create another migration?

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/migrations/20230425123140-key-expiration-reminder.js
+++ b/locksmith/migrations/20230425123140-key-expiration-reminder.js
@@ -39,6 +39,9 @@ module.exports = {
         type: Sequelize.DATE,
       },
     })
+    await queryInterface.changeColumn('KeyExpirationReminders', 'expiration', {
+      type: Sequelize.STRING,
+    })
     await queryInterface.addIndex('KeyExpirationReminders', {
       fields: ['lockAddress', 'network', 'tokenId', 'type', 'expiration'],
       unique: true,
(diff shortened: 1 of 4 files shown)


locksmith/migrations/20230425123140-key-expiration-reminder.js
```diff
@@ -39,6 +39,9 @@ module.exports = {
         type: Sequelize.DATE,
       },
     })
+    await queryInterface.changeColumn('KeyExpirationReminders', 'expiration', {
```
julien51, line 42: You need to create a new migraion file. This one has already been executed and will not be executed again...

julien51:
I am not sure this change is required? I assume your are storing timestamps? What is the largest number you can store and what date is that?

kalidiagne:
> I am not sure this change is required? I assume your are storing timestamps? What is the largest number you can store and what date is that?

You are right, I was testing with all keys that include `MAX_UINT`, i close

PR #11951, opened 2023-05-19
feat(unlock-app) - fix autocomplete for emails

kalidiagne:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/main/Metadata.tsx
+++ b/unlock-app/src/components/interface/checkout/main/Metadata.tsx
@@ -187,7 +187,7 @@ export const MetadataInputs = ({
                   }}
                   ref={ref}
                   onBlur={onBlur}
-                  autoComplete={label}
+                  autoComplete="email"
                 />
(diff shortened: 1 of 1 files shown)


unlock-app/src/components/interface/checkout/main/Metadata.tsx
```diff
@@ -187,7 +187,7 @@ export const MetadataInputs = ({
                   }}
                   ref={ref}
                   onBlur={onBlur}
-                  autoComplete={label}
+                  autoComplete="email"
```
searchableguy, line 190: should take into account that this box can be an text input too. This doesn't currently. See above.

searchableguy (approved):
Need one change otherwise LGTM

unlock-app/src/components/interface/checkout/main/Metadata.tsx
kalidiagne: good point!

PR #11954, opened 2023-05-22
feat(locksmith) - add get user email recipient

kalidiagne:
# Description
Add `getUserEmailRecipient` to metadata operations 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/operations/userMetadataOperations.ts
+++ b/locksmith/src/operations/userMetadataOperations.ts
@@ -52,6 +52,26 @@ export async function getMetadata(
   return data ? data.data : data
 }
 
+export async function getUserEmailRecipient({
+  lockAddress,
+  ownerAddress,
+}: {
+  lockAddress: string
+  ownerAddress: string
+}): Promise<string | undefined> {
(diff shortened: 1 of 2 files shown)


locksmith/src/operations/userMetadataOperations.ts
```diff
@@ -52,6 +52,26 @@ export async function getMetadata(
...
 }
 
+export async function getUserEmailRecipient({
+  lockAddress,
+  ownerAddress,
+}: {
+  lockAddress: string
+  ownerAddress: string
+}): Promise<string | undefined> {
+  const ownerMetadata = await getMetadata(
```
searchableguy, line 62: Does this only get metadata? 🤔 You should only use function which makes a database query.

searchableguy (approved):

locksmith/src/operations/userMetadataOperations.ts
kalidiagne: yes, this gets only user metadata with a query

PR #11980, opened 2023-05-24
fix(unlock-app): Document referer being undefined

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useSIWE.tsx
+++ b/unlock-app/src/hooks/useSIWE.tsx
@@ -80,26 +80,30 @@ export const SIWEProvider = ({ children }: Props) => {
   const siweSign = async (
     nonce: string,
     statement: string,
-    opts: any = {}
+    opts: {
+      resources?: string[]
+    } = {}
   ): Promise<{ message: string; signature: string } | null> => {
     try {
-      setStatus('loading')
       if (!connected) {
         throw new Error('No wallet connected.')
       }
       const walletService = await getWalletService()
 
       const address = await walletService.signer.getAddress()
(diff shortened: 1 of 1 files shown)


unlock-app/src/hooks/useSIWE.tsx
```diff
@@ -80,26 +80,30 @@ export const SIWEProvider = ({ children }: Props) => {
...
         throw new Error('No wallet connected.')
       }
       const walletService = await getWalletService()
 
       const address = await walletService.signer.getAddress()
+      const insideIframe = window.location.href !== window.parent.location.href
 
       const parent = new URL(
-        window.location != window.parent.location
-          ? document.referrer
```
searchableguy, line 96: @julien51 Can you tell me the reason for using document.referrer here? 

I think this was part of a previous refactor to fix the metamask warnings but this is undefined in some context which leads to error being thrown.

julien51: That's when it's embedded as an iframe. Siwe normally requires us to show it as the URI

julien51 (approved):

unlock-app/src/hooks/useSIWE.tsx
searchableguy: yeah, I got that. I'm wondering why referrer instead of parent.location?

julien51: I think it was because referrer is more "accessible" than `parent.location` if the origin is different.
I think I took this from there https://stackoverflow.com/questions/16413903/difference-between-document-referrer-and-window-parent-location-href

Issue #11993, opened 2023-05-25
Wrong base URI on avalanche


julien51:
See messgae on discord, we're missing the final `/`: https://discord.com/channels/462280183425138719/835883531438653460/1111312412826664990

julien51:
This was fixed!

PR #11994, opened 2023-05-25
fix(locksmith): fix for broken base URLs

julien51:
# Description

This is a temporary fix to handle #11993 while the base on the contract is being fixed.

# Checklist:

### DIFF
--- a/locksmith/src/routes/index.ts
+++ b/locksmith/src/routes/index.ts
@@ -52,6 +52,17 @@ router.use('/', lockRouter)
 router.use('/users', userRouter)
 router.use('/purchase', purchaseRouter)
 router.use('/claim', claimRouter)
+
+// Redirects for broken base URLs
+router.use(
+  '/api/key/:chain([0-9]{1,12}):lock(0x[0-9A-Fa-f]{40}):rest(/[0-9]{0,12})?',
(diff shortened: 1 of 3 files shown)


searchableguy (approved):
CleanShot 2023-05-25 at 23 16 33@2x

LGTM

locksmith/src/routes/index.ts
```diff
@@ -52,6 +52,17 @@ router.use('/', lockRouter)
 router.use('/users', userRouter)
 router.use('/purchase', purchaseRouter)
 router.use('/claim', claimRouter)
+
+// Redirects for broken base URLs
+router.use(
+  '/api/key/:chain([0-9]{1,12}):lock(0x[0-9A-Fa-f]{40}):rest(/[0-9]{0,12})?',
```
searchableguy, line 58:
```suggestion
  '/api/key/:chain([0-9]{1,12}):lock(0x[0-9A-Fa-f]{40}):rest(/[0-9]{0,12})?',
```
I wonder, if you do not need to escape `/` in the rest here. Does it work without it?

julien51: Yes, I tested without `/`, with just `/` and with `/213`... all work!

PR #12102, opened 2023-06-05
fix(ui): Remove default import from svgr

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:


searchableguy:
Any idea how to fix this big lint issues? The lint staged fails because of generated files. We should ignore them I think.

julien51 (approved):

julien51:
> Any idea how to fix this big lint issues? The lint staged fails because of generated files. We should ignore them I think.

Yeah, I have been wondering about that. Maybe we can even remove these files from git? wdyt @clemsos ?

clemsos:
we are using a js parser to output that file so the best will be to make that parser output the correct lint format. It shouldn't be too complicated

PR #12125, opened 2023-06-07
fix(unlock-app): stripe onramp not visible in iframe

julien51:
# Description

Prompted by the Stripe team, we're not loading inside an iframe.

# Checklist:


unlock-app/src/components/interface/checkout/main/Payment.tsx
```diff
@@ -126,6 +126,7 @@ export function Payment({ injectedProvider, checkoutService }: Props) {
   // Universal card is enabled if credit card is not enabled by the lock manager and the lock is USDC
   const USDC = networkConfig?.tokens?.find((t: any) => t.symbol === 'USDC')
   const universalCardEnabled =
+    window.top === window &&
```
searchableguy, line 129: I think we repeat this in many places. Could you move this to a variable in one of the const files so we can import the check anywhere we want.

PR #12141, opened 2023-06-08
fix(unlock-app): using the parent as the origin

julien51:
# Description

After much investigation I found that indeed we are sometimes (inconsistently but that's another issue) using the parent's injected provider and we need to use its domain for the SIWE message.

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useCheckoutCommunication.ts
+++ b/unlock-app/src/hooks/useCheckoutCommunication.ts
@@ -44,7 +44,7 @@ export interface MethodCallResult {
 
 // Taken from https://github.com/ethers-io/ethers.js/blob/master/src.ts/providers/web3-provider.ts
 export type AsyncSendable = {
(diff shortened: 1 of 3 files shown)


searchableguy (approved):

unlock-app/src/hooks/useCheckoutCommunication.ts
```diff
@@ -197,6 +196,10 @@ export const useCheckoutCommunication = () => {
 
   if (useDelegatedProvider && !providerAdapter) {
     setProviderAdapter({
+      parentOrigin: () => {
+        // @ts-expect-error Property 'parentOrigin' does not exist on type 'ChildAPI'.ts(2339)
```
searchableguy, line 200: 🤔

julien51: yeah postmates does not have good TS support!

Issue #12136, opened 2023-06-08
Allow for hidden metadata field!


julien51:
We should be able to add the ability to pass metadata fields for which there is a value that is added to the user's metadata, but not "changeable" (or visible) to the user when they checkout. This would allow for user mapping when using checkout.

julien51:
Similarly we should add the ability to pass a `data` field from the paywall config.

PR #12155, opened 2023-06-12
feature(unlock-app): adding the ability to pass hidden metadata

julien51:
# Description

This PR adds the ability to add a new type of metadata input called `hidden`. If this one is used, the implementer can pass an arbitrary value that gets saved as part of the owner's metadata and that the implemeter can then used to "map" owners with users in their own application.

Refs #12136

# Checklist:

### DIFF
--- a/packages/ui/lib/components/Form/FieldLayout.tsx
+++ b/packages/ui/lib/components/Form/FieldLayout.tsx
@@ -12,6 +12,7 @@ export interface Props {
   description?: ReactNode
   children: ReactNode
   append?: ReactNode
+  hidden?: boolean
 }
 
 const SIZE_STYLES: SizeStyleProp = {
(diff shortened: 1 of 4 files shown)


packages/ui/lib/components/Form/FieldLayout.tsx
```diff
@@ -37,13 +38,18 @@ export function FieldLayout(props: Props) {
...
   } = props
   const labelSizeStyle = SIZE_STYLES[size!]
   const labelClass = twMerge('px-1', labelSizeStyle)
   const descriptionClass = twMerge('text-gray-600', TEXT_SIZE[size])
   const errorClass = twMerge('text-red-500', TEXT_SIZE[size])
   const successClass = twMerge('text-green-500', TEXT_SIZE[size])
 
+  if (hidden) {
+    return null
+  }
```
searchableguy, line 51: I'm confused why both this and the hidden type. Hidden type should make input invisible already, no?

julien51: yes but the label and such are still rendered... hence this

PR #12157, opened 2023-06-12
fix(locksmith): Remove use of owner in filtering key in subscription endpoint

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:


julien51 (approved):

locksmith/src/routes/v2/subscriptions.ts
```diff
@@ -13,6 +14,7 @@ router.get(
   '/:network/locks/:lockAddress/keys/:keyId',
```
julien51, line 14: Is this endpoint used by verifiers when the scan tickets?

searchableguy: That's a good question. I don't think we use it in the verifiers since that's only for ticket which doesn't require subscription.

searchableguy: We don't.

Issue #12002, opened 2023-05-26
Send locksmith transactions from Defender Relay


julien51:
We now send transactions ourselves manually which is suboptimal and creates issues around gas price calculations, nonce management, retries... etc that we should not have to deal with.

[Defender Relay](https://docs.openzeppelin.com/defender/relay) could be a solution worth exploring to manage these things. They integrate with ethers which should limit the risk of writing "custom" code for them.
 
<img width="810" alt="Screenshot 2023-05-26 at 10 41 51 AM" src="https://github.com/unlock-protocol/unlock/assets/17735/f71bb892-8534-4a7f-a9aa-dfb91db9d1ff">

julien51:
this has shipped!

PR #12158, opened 2023-06-13
feature(locksmith): Using OpenZeppelin's Relay to send transactions and sign messages

julien51:
# Description

This was easier than I thought but probably needs a lot more testing.
Also, right now this is on a _per network_ basis, which is ok I guess, but we need to find a way to avoid having 1 API key/secret pair for each network or this will get hard to manage pretty quickly IMO.

# Checklist:

### DIFF
--- a/locksmith/src/config/config.ts
+++ b/locksmith/src/config/config.ts
@@ -87,7 +87,6 @@ const config = {
   requestTimeout: '30s',
   defenderRelayCredentials,
   databaseUrl: process.env.DATABASE_URL || '',
-  databaseUrl: '',
   sentry: {
     dsn: 'https://30c5b6884872435f8cbda4978c349af9@o555569.ingest.sentry.io/5685514',
   },


packages/unlock-js/src/KeyManager.ts
```diff
@@ -13,7 +13,7 @@ export interface TransferObject {
   deadline: number
 }
 
-type Signer = ethers.Wallet | ethers.providers.JsonRpcSigner
+type Signer = ethers.Signer | ethers.providers.JsonRpcSigner
```
julien51, line 16: Changed to signer to be more accurate!

julien51:
First transaction sent by the Relayer! https://goerli.etherscan.io/tx/0x0e8a00b37373074c50a1ced32c8ec19669d8470bc3b382bb783df300b724dc49

julien51:
Bad news: we need to use differenyt API keys for every network. That's going to be quite a bore :(

julien51:
Using 1password for secrets should work! https://github.com/unlock-protocol/unlock/pull/12162

julien51:
Since we will be adding so many secrets, we need a better secret approach: https://github.com/unlock-protocol/unlock/issues/12002

locksmith/src/config/config.ts
```diff
@@ -58,7 +58,8 @@ const config = {
   recaptchaSecret: process.env.RECAPTCHA_SECRET,
   logtailSourceToken: process.env.LOGTAIL,
   sessionDuration: Number(process.env.SESSION_DURATION || 86400 * 60), // 60 days
-  requestTimeout: '30s',
+  requestTimeout: '25s',
```
julien51, line 61:
```suggestion
  requestTimeout: '30s',
```

searchableguy (approved):

Issue #12134, opened 2023-06-08
Yield token id from Paywall?


julien51:
A dev wants to be able to get access to the token Id (on top of the transaction hash) when they embed the paywall and the config is set to be `pessimistic`.

PR #12161, opened 2023-06-13
feat(unlock-app): Transmit token Id and metadata in confirmation event to paywall

searchableguy:
# Description

fixes https://github.com/unlock-protocol/unlock/issues/12134

Also adds metadata to confirmation event to simplify 

# Issues

Fixes #
Refs #

# Checklist:


packages/unlock-js/src/web3Service.ts
```diff
@@ -1055,4 +1056,39 @@ export default class Web3Service extends UnlockService {
...
+      network: number
+      lockAddress: string
+    }
+  }) {
+    const provider = this.providerForNetwork(network)
+    const lockContract = await this.getLockContract(lockAddress, provider)
+    const txReceipt = await provider.getTransactionReceipt(hash)
+    const parser = lockContract.interface
+    const events = txReceipt.logs.map((log) => {
+      if (log.address.toLowerCase() !== lockAddress.toLowerCase()) return // Some events are triggered by the ERC20 contract
```
julien51, line 1074:
```suggestion
      if (log.address.toLowerCase() !== lockAddress.toLowerCase()) return // Filter events not emitted by the lock contract
```

packages/unlock-js/src/web3Service.ts
```diff
@@ -1055,4 +1056,39 @@ export default class Web3Service extends UnlockService {
       estimatedGasUsedUSD,
     }
   }
+
+  async getTokenIdFromTx({
```
julien51, line 1060: I wonder if this will work for any version of the lock (it should). Would you be able to add a test in the integration test suite for this? It will easily run for all versions.

julien51 (approved):
LGTM. Please add tests though!

packages/unlock-js/src/web3Service.ts
searchableguy: Added in integration test.

PR #12164, opened 2023-06-14
feat(unlock-app) - fix guild error

kalidiagne:
# Description
Fix issue found on `Guild` step component during some test

**Error**

Screenshot 2023-06-14 at 11 56 35

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/main/Guild.tsx
+++ b/unlock-app/src/components/interface/checkout/main/Guild.tsx
@@ -1,6 +1,6 @@
 import { CheckoutService } from './checkoutMachine'
 import { Connected } from '../Connected'
-import { Button, minifyAddress } from '@unlock-protocol/ui'
+import { Button, Placeholder, minifyAddress } from '@unlock-protocol/ui'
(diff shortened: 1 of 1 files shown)


julien51 (commented):
I think we need to do better. Here, if `guild` is falsy then we show a link to nowhere... Can you at least show an error or loading message?

unlock-app/src/components/interface/checkout/main/Guild.tsx
```diff
@@ -48,25 +48,30 @@ export function Guild({ injectedProvider, checkoutService }: Props) {
 
   const disabled = data && data.filter((d) => !d).length > 0
 
+  const guildUrl =
+    guild && guild.urlName ? `https://guild.xyz/${guild?.urlName}` : '#'
+
+  const guildName = guild && guild?.name ? guild.name : undefined
```
searchableguy, line 54: there is no need to do this right. guild?.name itself is enough. It will be undefined if name property or guild does not exist.

PR #12184, opened 2023-06-15
fix(unlock-app, unlock-protocol-com, paywall): fixing syntax for netlify deploys

julien51:
# Description

Netlify cli changed... this adds support for the latest changes.

# Checklist:

- [X] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [X] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #12188, opened 2023-06-16
Revert "fix(unlock-app, unlock-protocol-com, paywall): fixing syntax for netlify deploys"

julien51:
Reverts unlock-protocol/unlock#12184


julien51:
Reverting this

PR #12257, opened 2023-06-28
feat(smart-contracts): swap and burn UDTs with Unlock collected fees

### LINKED ISSUE #11677: Swap Fee and Burn
Now that fee switch is in place, let's work on swap fee and burn!

clemsos:
# Description

This PR adds a `UnlockSwapBurner` which is in charge of swapping tokens collected by the Unlock contract to UDT. The contract has no transfer function and therefore acts as a burn address, making the received supply of UDT unusable

# Issues

Fixes #11677
Refs #

# Checklist:


smart-contracts/contracts/utils/UnlockSwapBurner.sol
```diff
@@ -0,0 +1,138 @@
...
+    if (tokenAddress != address(0)) {
+      // Approve the router to spend src ERC20
+      TransferHelper.safeApprove(tokenAddress, swapRouter, amount);
+
+      // approve PERMIT2 to manipulate the token
+      IERC20(tokenAddress).approve(permit2, amount);
+    }
+
+    // issue PERMIT2 Allowance
+    IPermit2(permit2).approve(
```
julien51, line 108: Maybe better to do in a separate function for unlimited amounts? This way we an save gas?

julien51 (commented):
We also need to add a function to `Unlock` for it to send the funds to that contract?

I am trying to wrap my head around malicious uses. I guess someone could create a "bad pool" where the price of UDT would be MUCH higher and then they could burn a very small amount of UDT for a high amount of fees?

clemsos:
> We also need to add a function to `Unlock` for it to send the funds to that contract?

Yes, for Unlock it will be a simple transfer wrapper.


> I am trying to wrap my head around malicious uses. I guess someone could create a "bad pool" where the price of UDT would be MUCH higher and then they could burn a very small amount of UDT for a high amount of fees?

We can hardcode the recipient address as address(this).

clemsos:
The better approach for a UDT swap is to pass directly the path to the contract that can execute the swap (with hardcoded UDT pool address at the end). The issue here is that the UDT pool still a v2 pool which makes parsing the swap calldata much more complicated. 

With a v3 UDT pool, we could use a [multihop swap](https://docs.uniswap.org/contracts/v3/guides/swaps/multihop-swaps#exact-input-multi-hop-swaps) which is quite straightforward in approach and implemntation. 

Then with a v2 pool, the smart router will return a "mixed" route (that includes v2 and v3 pools) which requires a much more complicated approach to parse calldata (as we will have to executed each hop separately). A js implementation is [available here](https://github.com/Uniswap/router-sdk/blob/d8eed164e6c79519983844ca8b6a3fc24ebcb8f8/src/swapRouter.ts#L270-L322) but looks quite tricky to reproduce correctly in solidity.

A simpler approach may be to create UDT v3 pools (convert exiting ones?) so we can have v3 paths and execute directly using a multihop swap.

clemsos:
See #12660 for Uniswap V3 pool creation

smart-contracts/contracts/Unlock.sol
```diff
@@ -642,11 +673,23 @@ contract Unlock is UnlockInitializable, UnlockOwnable {
...
+    _transfer(token, to, amount);
+  }
+
+  /**
+   * Send tokens held by this contract to the UDT SwapBurner contract. The tokens sent to the
+   * contract can later be swapped for UDT and will remain in the burner contract in order to
+   * decrease UDT suplt in circulation.
+   * @notice This function can be called by anyone (not only the contract owner)
+   * @param token the address of the tokem (zero address for native)
+   * @param amount the amount of tokens to send
```
julien51, line 685:
```suggestion
   * Send tokens held by this contract to the UDT SwapBurner contract. The tokens sent to the
   * contract are then swapped for UDT and UDT itself will be sent to a burner address.
   * @notice This function can be called by anyone (not only the contract owner) as a way to ensure decentralization.
   * @param token the address of the token (zero address for native) to swap and burn
   * @param amount the amount of tokens to swap and burn
```

smart-contracts/contracts/interfaces/IUnlock.sol
```diff
@@ -314,4 +314,25 @@ interface IUnlock {
...
    */
   function removeLock(address lock) external;
+
+  /**
+   * Send tokens held by this contract to the UDT SwapBurner contract. The tokens sent to the
+   * contract can later be swapped for UDT and will remain in the burner contract in order to
+   * decrease UDT suplt in circulation.
+   * @notice This function can be called by anyone (not only the contract owner)
+   * @param token the address of the tokem (zero address for native)
+   * @param amount the amount of tokens to send (use zero to send entire balance)
```
julien51, line 324:
```suggestion
   * Send tokens held by this contract to the UDT SwapBurner contract. The tokens sent to the
   * contract are then swapped for UDT and UDT itself will be sent to a burner address.
   * @notice This function can be called by anyone (not only the contract owner) as a way to ensure decentralization.
   * @param token the address of the token (zero address for native) to swap and burn
   * @param amount the amount of tokens to swap and burn
```

smart-contracts/contracts/utils/UnlockSwapBurner.sol
```diff
@@ -0,0 +1,151 @@
...
+  // addresses on current chain
+  address public unlockAddress;
+
+  // required by Uniswap Universal Router
+  address public permit2;
+  address public uniswapRouter;
+
+  // dead address to burn
+  address public constant burnAddress =
+    0x000000000000000000000000000000000000dEaD;
```
julien51, line 36: Should we use the 0x0 address rather?

smart-contracts/contracts/utils/UnlockSwapBurner.sol
```diff
@@ -0,0 +1,151 @@
...
+
+  // events
+  event SwapBurn(address tokenAddress, uint amountSpent, uint amountBurnt);
+
+  // errors
+  error UDTSwapFailed(address uniswapRouter, address tokenIn, uint amount);
+  error InsufficientBalance();
```
julien51, line 43: Not used

```suggestion
```

smart-contracts/contracts/utils/UnlockSwapBurner.sol
```diff
@@ -0,0 +1,151 @@
...
+  error UnautorizedRouter(address routerAddress);
```
julien51, line 44: Unused!

```suggestion
```

smart-contracts/contracts/utils/UnlockSwapBurner.sol
```diff
@@ -0,0 +1,151 @@
...
+      udtAddress
+    );
+
+    // executes the swap token > WETH > UDT
+    ISwapRouter.ExactInputParams memory params = ISwapRouter.ExactInputParams({
+      path: tokenAddress == wrappedAddress
+        ? defaultPath
+        : abi.encodePacked(tokenAddress, poolFee, defaultPath),
+      recipient: address(this),
+      deadline: block.timestamp,
```
julien51, line 127: 1 minute for consistency?

julien51 (commented):
What happens if someone sends UDT to this contract? Is it stuck there then?

smart-contracts/test/Unlock/swapBurner.js
```diff
@@ -0,0 +1,140 @@
...
+      await token.mint(unlock.address, amount)
+      const balanceBefore = await getBalance(swapBurner.address, token.address)
+      await unlock.connect(randomSigner).sendToSwapBurner(token.address, amount)
+      const balanceAfter = await getBalance(swapBurner.address, token.address)
+      assert.equal(
+        balanceAfter.minus(balanceBefore).toString(),
+        amount.toString()
+      )
+    })
+    it('can send the entire balance by passing 0 as amount param', async () => {
```
julien51, line 133: Not supported anymore, rigth?

smart-contracts/contracts/utils/UnlockSwapBurner.sol
```diff
@@ -0,0 +1,148 @@
...
+    if (amountUDTOut == 0) {
+      revert UDTSwapFailed(uniswapRouter, tokenAddress, tokenAmount);
+    }
+
+    // burn the UDT
+    bool success = IERC20(udtAddress).transfer(burnAddress, amountUDTOut);
+    if (success == false) {
+      revert UDTSwapFailed(uniswapRouter, tokenAddress, tokenAmount);
+    } else {
+      emit SwapBurn(tokenAddress, tokenAmount, amountUDTOut);
```
julien51, line 140: Can we emit the same event as the eRC20 contract would maybe?

julien51 (approved):
One quick question about events...

smart-contracts/contracts/utils/UnlockSwapBurner.sol
clemsos: That will be a `Transfer` event. Just will make it harder to read maybe (as the erc20 contract will emit one anyway)?

julien51: I was wondering if there was a specific `Burn` event in ERC20 contracts, but maybe not... in an y case, nbd!

clemsos: No its usually a simple transfer to address zero (here to "dead" address)

PR #12260, opened 2023-06-28
feat(locksmith): Don't cache authenticated requests

searchableguy:
# Description

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/utils/middlewares/cacheMiddleware.ts
+++ b/locksmith/src/utils/middlewares/cacheMiddleware.ts
@@ -26,6 +26,12 @@ export const createCacheMiddleware = (option: Partial<Options> = {}) => {
     if (req.method !== 'GET') {
       return next()
     }
+
+    // We don't cache authenticated requests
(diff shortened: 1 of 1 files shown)


locksmith/src/utils/middlewares/cacheMiddleware.ts
```diff
@@ -26,6 +26,12 @@ export const createCacheMiddleware = (option: Partial<Options> = {}) => {
     if (req.method !== 'GET') {
       return next()
     }
+
+    // We don't cache authenticated requests
+    if (req.user?.walletAddress) {
```
julien51, line 31: I think that would not apply to `app` users? Can we just make it if `req.user` is defined?

julien51, line 31:
```suggestion
    if (req.user) {
```

searchableguy: It does apply to app users.

PR #12382, opened 2023-07-18
feat(unlock-app) - Update destination for CTA buttons on EVENTS

kalidiagne:
# Description
- Update destination for CTA buttons on EVENTS 

# Issues

Fixes #
Refs #

# Checklist:


ccarfi:
@kalidiagne i clicked on the CTA buttons in the netlify staging build, and they still seem to go to the old (open metamask) location instead of the new (land on hubspot lead capture form) location. can you please confirm this opens the hubspot form when either of the in-page CTA buttons are clicked? thanks.

searchableguy (approved):

julien51 (changes requested):
I don't think it would be that hard to add the form to the actual existing flow AND save it into Hubspot. I think it would significantly reduce friction/weirdness. Any reason why you would not want to do it?

julien51 (approved):

PR #12393, opened 2023-07-24
feat(unlock-app) - checkout builder auto-save

kalidiagne:
# Description

- [x] Add auto-save when the form of the checkout builder has changes
- [x] Add check for validation to save the form only when the form is valid

Needs https://github.com/unlock-protocol/unlock/pull/12392

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
+++ b/unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
@@ -175,28 +175,6 @@ export const CheckoutUrlPage = () => {
     }
   }, [checkoutConfigList?.length])
 
-  const onConfigSave = useCallback(async () => {
-    if (!checkoutConfig?.id) return // prevent save if not config is set
-    const updated = await updateConfig({
-      config: checkoutConfig.config,
-      name: checkoutConfig.name,
-      id: checkoutConfig.id,
-    })
-    setCheckoutConfig({
-      id: updated.id,
-      name: updated.name,
-      config: updated.config as PaywallConfig,
-    })
-
-    // Show toast message when configuration is changed after first render
-    if (!firstRender.current) {
-      firstRender.current = true
-    } else {
-      ToastHelper.success('Configuration updated.')
-      await refetchConfigList()
-    }
-  }, [checkoutConfig, updateConfig, refetchConfigList])
-
   const onConfigRemove = useCallback(async () => {
     if (!checkoutConfig.id) {
       setDeleteConfirmation(false)
@@ -218,6 +196,7 @@ export const CheckoutUrlPage = () => {
     DEFAULT_CONFIG,
     setDeleteConfirmation,
   ])
+
   useEffect(() => {
     const checkout = checkoutConfigList?.[0]
     if (!checkout) return
@@ -269,9 +248,6 @@ export const CheckoutUrlPage = () => {
       }
     })
   }
-
-  const addLockMutation = useMutation(onAddLocks)
-  const onBasicConfigChangeMutation = useMutation(onBasicConfigChange)
 
   const TopBar = () => {
     return (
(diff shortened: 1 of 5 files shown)


unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
```diff
@@ -184,8 +187,14 @@ export const CheckoutUrlPage = () => {
       name: updated.name,
       config: updated.config as PaywallConfig,
     })
-    ToastHelper.success('Configuration updated.')
-    await refetchConfigList()
+
+    // Show toast message when configuration is changed after first render
```
julien51, line 191: I am not sure I understand this part?

unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
```diff
@@ -411,9 +442,13 @@ export const CheckoutUrlPage = () => {
                   description:
                     'Customize the checkout modal interaction & additional behavior',
                   disabled: !hasSelectedConfig,
+                  loading,
                   children: (
                     <BasicConfigForm
-                      onChange={onBasicConfigChange}
+                      onChange={(fields: any, isValid = true) => {
+                        if (!isValid) return // trigger update form only when is valid to prevent update auto-update with invalid values
```
julien51, line 449: Please add a logging statement when the config is invalid as it should not happen.

unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
```diff
@@ -423,9 +458,13 @@ export const CheckoutUrlPage = () => {
                   description:
                     'Select the locks that you would like to featured in this configured checkout modal',
                   disabled: !hasSelectedConfig,
+                  loading,
                   children: (
                     <LocksForm
-                      onChange={onAddLocks}
+                      onChange={(fields: any, isValid = true) => {
+                        if (!isValid) return // trigger update form only when is valid to prevent update auto-update with invalid values
```
julien51, line 465: Please add a logging statement when the config is invalid as it should not happen.

unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
kalidiagne: the form already have validation with error messages in place, need to avoid to call the endpoint to save the config in that case.

Screenshot 2023-07-25 at 10 46 03

unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
kalidiagne: as before, prevent auto-save when form is invalid, eg. `url invalid`

Screenshot 2023-07-25 at 10 46 58

unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
kalidiagne: When the checkout builder page loads, we set the last one the user updated as current configuration.
This will trigger the `onConfigSave` to auto-save, on the first-page load we don't need to show 'Configuration updated.' toast

julien51: ok that makes sense

unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
julien51: I understand but please still add the logging statement so we have something to look at if for some the reason the UI does not show an error.

unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
```diff
@@ -323,6 +337,20 @@ export const CheckoutUrlPage = () => {
...
 
+  /**
+   * Save checkout config when fields have changed, This is done with delays invoking a function until after wait milliseconds have passed
+   * to avoid calling the endpoint multiple times.
+   */
+  const [_isReady] = useDebounce(
+    () => {
+      onConfigSave()
+    },
+    5000, // 5 seconds of delay after edit's to trigger auto-save
```
julien51, line 348: This seems large. If the user is done, they make click thru to the next step in their journey and this would not be saved.
Any reason why you want 5 seconds?

searchableguy (approved):
Maybe change delay to 3s instead of 5 but otherwise LGTM

unlock-app/src/components/interface/locks/CheckoutUrl/index.tsx
julien51, line 348:
```suggestion
    2000, // 2 seconds of delay after edit's to trigger auto-save
```

PR #12421, opened 2023-07-31
fix(paywal): updated script to detect localstorage

julien51:
# Description

In incognito, we cannot access `localstorage` so we need to load it _inside_ the `try/catch`.
Taken from https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability

# Checklist:


packages/paywall/src/utils/localStorage.ts
```diff
@@ -1,8 +1,9 @@
 // copied from the MDN docs as the best way to detect localStorage presence
 // https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Testing_for_availability
 export default function localStorageAvailable(): boolean {
-  const storage = localStorage
+  let storage
```
searchableguy, line 4: Shouldn't this have a type?

PR #12488, opened 2023-08-02
feature(locksmith): adding the ability to bypass captcha for some addresses

julien51:
# Description

Our friends at Protocol Labs want to be able to use our API without the front-end captcha. 
I am suggesting we add an allowlist of addresses that can bypass the captcha requirement if they are authenticated.

# Checklist:


locksmith/src/routes/claim.ts
```diff
@@ -1,6 +1,7 @@
 import express from 'express'
 import { PurchaseController } from '../controllers/purchaseController'
 import { createGeoRestriction } from '../utils/middlewares/geoRestriction'
```
julien51, line 3: Note: I am not sure this route is still in use :(

locksmith/src/utils/middlewares/recaptchaMiddleware.ts
```diff
@@ -2,11 +2,25 @@ import { RequestHandler } from 'express'
 import fetch from 'isomorphic-fetch'
 import config from '../../config/config'
 
+/**
+ * A list of authenticated users who are making calls for which we should bypass the captcha veification
+ */
+const allowList = ['0x61be315032235Ac365e39705c11c47fdaee698Ee'].map(
+  (address: string) => address.toLowerCase()
+)
+
```
searchableguy, line 11: Probably better to put this in the database directly so we can modify this on the fly.

searchableguy (approved):

locksmith/src/utils/middlewares/recaptchaMiddleware.ts
julien51: You are right. I'll do it.

PR #12493, opened 2023-08-03
refactor(locksmith): the public provider should only ever be used to send transactions

julien51:
# Description

Publick providers are heavily monitored and and tend to block/be slow if used to "read" state.
We had an issue where we used the public provider to read the gas settings...

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/priceController.ts
+++ b/locksmith/src/controllers/v2/priceController.ts
@@ -105,12 +105,15 @@ export const isCardPaymentEnabledForLock: RequestHandler = async (
   const network = Number(request.params.network)
 
   const web3Service = new Web3Service(networks)
-  const lock = await web3Service.getLock(lockAddress, network)
+
+  const lock = await web3Service.getLock(lockAddress, network, {
+    fields: ['keyPrice', 'currencyContractAddress'],
+  })
 
   const [defiLammaPricing, settingsPricing] = await Promise.all([
     pricingOperations.getDefiLammaPrice({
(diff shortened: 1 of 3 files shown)


locksmith/src/controllers/v2/priceController.ts
```diff
@@ -105,12 +105,15 @@ export const isCardPaymentEnabledForLock: RequestHandler = async (
   const network = Number(request.params.network)
 
   const web3Service = new Web3Service(networks)
-  const lock = await web3Service.getLock(lockAddress, network)
+
+  const lock = await web3Service.getLock(lockAddress, network, {
```
julien51, line 109: No need to retrieve the rest of the fields :)

locksmith/src/controllers/v2/priceController.ts
```diff
@@ -105,12 +105,15 @@ export const isCardPaymentEnabledForLock: RequestHandler = async (
...
+    fields: ['keyPrice', 'currencyContractAddress'],
+  })
 
   const [defiLammaPricing, settingsPricing] = await Promise.all([
     pricingOperations.getDefiLammaPrice({
       network,
-      erc20Address: lock?.currencyContractAddress,
+      erc20Address: lock.currencyContractAddress,
```
searchableguy, line 116: Why the removal of ? Maybe it could be empty no?

julien51: no, it would fail

PR #12496, opened 2023-08-04
refactor(locksmith): changing the deployment scripts for locksmith

julien51:
# Description

As I was looking for ways to make sure that Heroku does proper blue/green deployment, I found we need to use what Heroku calls the "release phase", so I started to work on adding this.
I then realized we could also refactor our "worker" approach by having a single application that runs both `web` and `worker` dynos (and `release`).

# Checklist:

### DIFF
--- a/locksmith/scripts/release.sh
+++ b/locksmith/scripts/release.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+
+echo "I was released!"
--- a/locksmith/package.json
+++ b/locksmith/package.json
@@ -4,6 +4,7 @@
   "description": "an unlock storage service",
   "main": "server.js",
   "scripts": {
+    "release": "./scripts/release.sh",
     "prod": "tsx -r ./config/envLoader.js ./src/server.ts",
     "start": "./scripts/start.sh server",
     "dev": "NODE_ENV=development yarn db:migrate && tsx watch -r ./config/envLoader.js src/server.ts",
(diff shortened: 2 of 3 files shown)


julien51:
Suspiciously it looks like the job `deploy-locksmith-staging` did not run.

julien51:
HA! Of course, it only runs when merged on `master`, not on PR!

locksmith/scripts/release.sh
```diff
@@ -0,0 +1,32 @@
...
+env_file="./.op.env.$environment"
+
+# Check if the environment file exists
+if [ ! -f "$env_file" ]; then
+    echo "Environment file $env_file not found."
+    exit 1
+fi
+
+# First: run the migrations!
+
```
searchableguy, line 20: missing something here?

locksmith/scripts/release.sh
```diff
@@ -0,0 +1,32 @@
...
+# Prepare the command based on the argument
+if [ "$1" == "worker" ]; then
+    command="op run --env-file=\"$env_file\" -- yarn run db:migrate"
+else
+    command="op run --env-file=\"$env_file\" -- yarn run db:migrate"
+fi
```
searchableguy, line 26: isn't it the same in both cases?

searchableguy (approved):

locksmith/scripts/release.sh
julien51: good point!

Issue #12331, opened 2023-07-06
Can not withdrawal money from contract


boskiv:
**Describe the bug**
Uncaught (in promise) Error: fractional component exceeds decimals [ See: https://links.ethers.org/v5-errors-NUMERIC_FAULT ] (fault="underflow", operation="parseFixed", code=NUMERIC_FAULT, version=bignumber/5.7.0)

**Additional context**
Lock address: 0xb6db2c1923dad7411dad2c0c16c9997e5a481fe8

boskiv:
Able to do it with polygonscan write proxy method. But you should fix it on your side :)

julien51:
Related: https://github.com/unlock-protocol/unlock/issues/11944

PR #12498, opened 2023-08-04
fix(unlock-app): changing withdraw amount input from range to number.

### LINKED ISSUE #11944: "fixed" amount withdraws
Our withdraw UI lets the user chose what amount they want to withdraw with a bar.
We also need to let the user enter a fixed amount. That amount needs to bet between `0` and the full balance, rather than only decimal amounts.

For example on the screen below I want to withdraw 20 USDC exactly but I cannot.

Screenshot 2023-05-19 at 2 29 41 PM

julien51:
# Description

This fixes #11944 and #12331
Screenshot 2023-08-04 at 10 52 16 AM

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/locks/Manage/elements/WithdrawFundModal.tsx
+++ b/unlock-app/src/components/interface/locks/Manage/elements/WithdrawFundModal.tsx
@@ -230,7 +230,6 @@ export const WithdrawFundModal = ({
                 {...register('amount', {
                   validate: (value) => {
                     const formattedValue = parseFloat(`${value}`)
-                    console.log(formattedValue)
 
                     if (formattedValue <= 0) {
                       return 'The amount to withdraw should be greater than 0.'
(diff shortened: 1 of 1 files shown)


unlock-app/src/components/interface/locks/Manage/elements/WithdrawFundModal.tsx
```diff
@@ -220,18 +220,18 @@ export const WithdrawFundModal = ({
               />
 
               <Input
-                label={`Balance to transfer: ${amountToTransfer} ${symbol}`}
+                label={`Balance to transfer: ${amountToTransfer} of ${balance} ${symbol}`}
                 size="small"
-                type="range"
+                type="number"
```
searchableguy, line 225: Change this to text and add a number pattern to input.

Set valueAsNumber to true in register options.

searchableguy (approved):

unlock-app/src/components/interface/locks/Manage/elements/WithdrawFundModal.tsx
julien51: Please explain so I learn why!

julien51: Will replace across the board in another PR!

PR #12522, opened 2023-08-09
adjusting for how prettier locates solidity plugin

adeelhasan:
recent changes in prettier have altered how default plugins are located; the solidity plugin is passed as a command line param accordingly, though possibly it could also be a config change elsewhere.


julien51 (approved):
Thank you!

PR #12527, opened 2023-08-10
Gas optimizations use oz strings

adeelhasan:
this replaces calls to the UnlockUtils.sol file with the OZ Strings library.

the UnlockUtils.sol file can be removed, but when we remove references from older versions, it brings up the question of how to maintain those versions. We can leave that another PR / further discussion, or if its safe to just update those references, then I could just change them.

For right now the code will be more efficient in terms of gas, and since UnlockUtils calls are no longer referenced, they will _probably_ not be a part of the byte output of the compiler ... and basically not bloat the final contract size. However, this last bit is not 100% for me, let me know your thoughts. In the meantime I am testing in Remix how unused imports may be consequential or not, you may already know the answer.


adeelhasan:
I tested in Remix and the unused references / imports have no bearing on the final contract size.

clemsos:
The `UnlockUtils` contract is also used in some test contracts so we need to replace it there as well before we can remove the file from the repo.

Apart from that this seems like a legit improvement to the contract.

clemsos (approved):
Let's merge this. I'll do the cleanup of UnlockUtils on another PR after

PR #12534, opened 2023-08-11
fix(unlock-app): fixing race condition

julien51:
# Description

If no `provider` exists when we receive the first message from the "parent" window, the message was lost.
This PR adds an `incomingBuffer` to checkout communication to handle this. This gets flushed once we receive have a provider connected.

# Checklist:

### DIFF
--- a/packages/paywall/src/Paywall.ts
+++ b/packages/paywall/src/Paywall.ts
@@ -111,7 +111,9 @@ export class Paywall {
       clientId: window.location.origin.toString(),
     }
   ) => {
-    this.provider = new PaywallProvider(this, unlockUrl, config)
+    if (!this.provider) {
+      this.provider = new PaywallProvider(this, unlockUrl, config)
+    }
     return this.provider
   }
 
@@ -226,7 +228,8 @@ export class Paywall {
       )
     } else {
       console.error(
-        'unknown method to call provider! Please make sure you use and EIP1193 provider!'
+        'unknown method to call provider! Please make sure you use and EIP1193 provider!',
+        { provider }
       )
     }
   }
--- a/unlock-app/src/hooks/useCheckoutCommunication.ts
+++ b/unlock-app/src/hooks/useCheckoutCommunication.ts
@@ -124,16 +124,13 @@ export const useCheckoutCommunication = () => {
   const { provider } = useProvider(AppConfig)
   const [user, setUser] = useState<string | undefined>(undefined)
 
-  const pushOrEmit = useCallback(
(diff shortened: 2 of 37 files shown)


packages/paywall/src/Paywall.ts
```diff
@@ -111,7 +111,9 @@ export class Paywall {
       clientId: window.location.origin.toString(),
     }
   ) => {
-    this.provider = new PaywallProvider(this, unlockUrl, config)
+    if (!this.provider) {
```
julien51, line 114: This is to avoid creating multiple provider objects.

unlock-app/src/hooks/useCheckoutCommunication.ts
```diff
@@ -123,31 +124,24 @@ export const useCheckoutCommunication = () => {
   const { provider } = useProvider(AppConfig)
   const [user, setUser] = useState<string | undefined>(undefined)
 
-  // @ts-expect-error - attach to window
-  window.px = provider
```
julien51, line 127: Cleaned up

julien51 (commented):
I want to also add support for account change and for network change events here too.

unlock-app/src/hooks/useCheckoutCommunication.ts
```diff
@@ -123,31 +124,21 @@ export const useCheckoutCommunication = () => {
...
-        redirectUri: config?.redirectUri ?? '',
-      })
-      setPaywallConfig(undefined)
-    },
-    resolveMethodCall,
-    resolveOnEvent,
-    resolveOnEnable,
-    handleMethodCallEvent: async ({ id, params, method }: MethodCall) => {
-      // @ts-expect-error but we know it's there
-      const px = window.px
```
searchableguy, line 148: This was attached to window because otherwise it wouldn't be be available when you try to access provider. Is that resolved with your changes?

searchableguy (approved):

unlock-app/src/hooks/useCheckoutCommunication.ts
julien51: Hum, it was a debugging thing, right? We don't actually need window.px

searchableguy: Iirc, the issue was that initially, provider is empty and when you initiate function in a new post instance - it would keep the provider as undefined due to closure since no new instance is ever created again due to the limitation of the library with multiple instances listening to same window. 

So provider was attached to window.px and updated later on when provider is not undefined without causing issues due to closure.

PR #12542, opened 2023-08-14
feat(smart-contracts): Add unchecked scope for some additions / subtractions

adeelhasan:
this puts some of the additions and subtractions in the file under unchecked scope; solidity > 0.8 has math safety checks built-in which incur gas overhead. This extra check can be turned off and gas saved in contexts where the risk of an overflow is not there. Particularly in the case of loops. there are other spots in the code base, but we can start reviewing with this file.


clemsos (approved):
LGTM

PR #12548, opened 2023-08-14
Fix checkout modal: available quantity and overflow

xdamman:
Fixing two issues:

- overflow of the ticket image
- misleading label: "quantity" implies it's the quantity that you want to order. But there is no way to change that. So it's misleading and the user may hesitate to press next until they find a way to change the quantity

Before:

Screenshot 2023-08-14 at 2 12 02 PM

After:
Screenshot 2023-08-14 at 2 11 43 PM

### DIFF
--- a/unlock-app/src/components/interface/checkout/main/Select.tsx
+++ b/unlock-app/src/components/interface/checkout/main/Select.tsx
@@ -128,7 +128,7 @@ const LockOption = ({ disabled, lock }: LockOptionProps) => {
                   )}
                   {formattedData?.formattedKeysAvailable !== 'Unlimited' && (
                     <LabeledItem
-                      label="Tickets available"
+                      label="Available"
                       icon={QuantityIcon}
                       value={
                         formattedData?.isSoldOut
--- a/unlock-app/src/components/interface/checkout/main/Select.tsx
+++ b/unlock-app/src/components/interface/checkout/main/Select.tsx
@@ -72,7 +72,7 @@ const LockOption = ({ disabled, lock }: LockOptionProps) => {
           <Fragment>
             <div className="flex w-full gap-x-4">
               <div>
-                <Avatar.Root className="inline-flex items-center justify-center w-14 h-14 rounded-xl">
+                <Avatar.Root className="inline-flex items-center justify-center w-14 h-14 rounded-xl overflow-hidden">
                   <Avatar.Image src={lockImageURL} alt={lock.name} />
                   <Avatar.Fallback className="bg-gray-50">
                     {lock.name.slice(0, 2).toUpperCase()}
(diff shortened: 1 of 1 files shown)


unlock-app/src/components/interface/checkout/main/Select.tsx
```diff
@@ -128,7 +128,7 @@ const LockOption = ({ disabled, lock }: LockOptionProps) => {
                   )}
                   {formattedData?.formattedKeysAvailable !== 'Unlimited' && (
                     <LabeledItem
-                      label="Quantity"
```
julien51, line 131: The checkout is used by many more things that are mot tickets so we can't set "Tickets" but can we just use "Available" ?

unlock-app/src/components/interface/checkout/main/Select.tsx
```diff
@@ -72,7 +72,7 @@ const LockOption = ({ disabled, lock }: LockOptionProps) => {
           <Fragment>
             <div className="flex w-full gap-x-4">
               <div>
-                <Avatar.Root className="inline-flex items-center justify-center w-14 h-14 rounded-xl">
```
julien51, line 75: This is already fixed in https://github.com/unlock-protocol/unlock/pull/12545

unlock-app/src/components/interface/checkout/main/Select.tsx
```diff
@@ -72,7 +72,7 @@ const LockOption = ({ disabled, lock }: LockOptionProps) => {
...
+                <Avatar.Root className="inline-flex items-center justify-center w-14 h-14 rounded-xl overflow-hidden">
```
julien51, line 75:
```suggestion
                <Avatar.Root className="inline-flex items-center justify-center w-14 h-14 rounded-xl">
```

unlock-app/src/components/interface/checkout/main/Select.tsx
```diff
@@ -128,7 +128,7 @@ const LockOption = ({ disabled, lock }: LockOptionProps) => {
                   )}
                   {formattedData?.formattedKeysAvailable !== 'Unlimited' && (
                     <LabeledItem
-                      label="Quantity"
+                      label="Tickets available"
```
julien51, line 131:
```suggestion
                      label="Available"
```

unlock-app/src/components/interface/checkout/main/Select.tsx
xdamman: This is a good example where having a separate repo would make sense to avoid such issue.
It's always annoying where you know what the right thing is for the end user, but you just cannot do it.
Anyway.

2nd best option is to show "10 left" like on the right of this screenshot. But then we cannot reuse the same <LabeledItem> component.
photo_2023-08-14 14 29 45

3rd best option is to just use "Available" (that will already be better than current situation)

julien51: We can't have 2 sperate repo and maintain 2 checkout modals :D We're barely able to do it with 1!

julien51: We can do `10 Left` :) That's easy!

PR #12568, opened 2023-08-17
feature(wedlocks, email-templates): Adding more template bases for emails

julien51:
# Description

We use a different template base for emails sent for events.

# Checklist:


packages/email-templates/src/index.ts
```diff
@@ -63,4 +63,4 @@ Object.keys(EmailTemplates).forEach((template: string) => {
 const getEmailTemplate = (template: Template | string) => templates[template]
 
 export default templates
-export { getEmailTemplate, base }
+export { getEmailTemplate, bases }
```
searchableguy, line 66: I'm a little worried about this change since base is not under bases which may break something.

searchableguy (approved):

wedlocks/src/route.js
```diff
@@ -6,15 +6,14 @@ import templates from '@unlock-protocol/email-templates'
...
-console.log({ config })
-
 /**
  * Builds the template and params
  * @param {*} args
  * @returns
  */
 const getTemplateAndParams = async (args, opts) => {
   let template = templates[args.template.toLowerCase()]
+  console.log(template)
```
julien51, line 16:
```suggestion
```

packages/email-templates/src/index.ts
julien51: Not sure what you mean?

searchableguy: the exports are changed. If you want to import base now, you have to import it from under bases.

```ts
import { bases } from "./templates.ts"
const { base } = bases
```

PR #12596, opened 2023-08-31
feat(subgraph) - referrer fees

easyrun42:


subgraph/src/public-lock.ts
```diff
@@ -399,3 +406,16 @@ export function handleLockMetadata(event: LockMetadataEvent): void {
...
+
+  if (lock) {
+    const hash = event.transaction.hash.toHexString()
+    const referrerFee = new ReferrerFee(hash)
+    referrerFee.referrer = event.params.referrer
+    referrerFee.fee = event.params.fee
+    referrerFee.lock = lock.id
+    referrerFee.save()
+  }
+}
```
easyrun42, line 421: Here is how we will handle the referrer fees on the subgraph

subgraph/subgraph.yaml
```diff
@@ -109,5 +110,8 @@ templates:
         - event: LockMetadata(string,string,string)
           handler: handleLockMetadata
           receipt: true
+        - event: ReferrerFee(indexed address,uint256)
+          handler: handleReferrerFees
+          receipt: true
```
easyrun42, line 115: This will be the event we will listen to 👍

subgraph/tests/referrer-utils.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  referrerEvent.parameters.push(
+    new ethereum.EventParam('referrer', ethereum.Value.fromAddress(referrer))
+  )
+
+  referrerEvent.parameters.push(
+    new ethereum.EventParam('fee', ethereum.Value.fromUnsignedBigInt(fee))
+  )
+
+  return referrerEvent
+}
```
easyrun42, line 28: Simple utils test helper function for referrer fees

subgraph/tests/referrer.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  lockAddress,
+  lockManagers,
+  nullAddress,
+  referrerFee,
+} from './constants'
+import { Address, BigInt, Bytes } from '@graphprotocol/graph-ts'
+import { handleReferrerFees } from '../src/public-lock'
+import { Lock } from '../generated/schema'
+
+describe('Referrer', () => {
```
easyrun42, line 22: Added some unit tests too

subgraph/schema.graphql
```diff
@@ -117,3 +119,14 @@ type Receipt @entity {
...
+type ReferrerFee @entity {
+  "Transaction Hash"
+  id: ID!
+  "Address of the referrer."
+  referrer: Bytes!
+  "Fee attributed to the referrer. E.g., 200 represents 20%."
+  fee: BigInt!
+  "In the Unlock ecosystem, a “Lock” is a smart contract that creates (or “mints”) NFTs"
+  lock: Lock!
+}
```
easyrun42, line 132: Here is the `ReferrerFee` entity. Similar to what we do with `LockStats` entity or `Receipt` entity:

id - `Transaction Hash`
referrer - `The address you gave the referral to`
fee - `Fee you gave to the referrer`
lock - `The lock the which referrer fee belongs to`

image

easyrun42:
@searchableguy Any idea why i can't add reviewers here ?

subgraph/schema.graphql
julien51: I am a bit confused about the use of the transaction hash? Is that the transaction that "sets" the fee? What happens if the lock manager changes the fee then? do you remove the `ReferrerFee` or change it?

I think a better model would be to just use the `referrer` address as the `id` since a referrer fee is unique per lock and referrer address.

subgraph/src/public-lock.ts
```diff
@@ -399,3 +406,16 @@ export function handleLockMetadata(event: LockMetadataEvent): void {
     lock.save()
   }
 }
+
+export function handleReferrerFees(event: ReferrerFeeEvent): void {
+  const lock = Lock.load(event.address.toHexString())
+
+  if (lock) {
+    const hash = event.transaction.hash.toHexString()
+    const referrerFee = new ReferrerFee(hash)
```
julien51, line 415: What happens here if the fee already exists and is just changed? We should NOT create a new object.

subgraph/tests/referrer.test.ts
julien51: Great. Please add a test where a fee is set and then changed and make sure only a single entity exists

julien51 (changes requested):
There is a logic issue around changing fees. Let's change this please!

subgraph/schema.graphql
easyrun42: Good idea. I wasn't aware of that lock manager can change the fee for a given address. Let me change that 👍

subgraph/tests/referrer.test.ts
easyrun42: New test added here with a single entity 🧼: https://github.com/unlock-protocol/unlock/pull/12596/commits/7f641bf0176fb07ecf071031760906f8b1906121

subgraph/src/public-lock.ts
easyrun42: Right this will make a new referrer which is not good. I fixed it here as well 👍: https://github.com/unlock-protocol/unlock/pull/12596/commits/7f641bf0176fb07ecf071031760906f8b1906121

subgraph/schema.graphql
easyrun42: Fixed 👍

subgraph/src/public-lock.ts
```diff
@@ -399,3 +406,22 @@ export function handleLockMetadata(event: LockMetadataEvent): void {
...
+  const lock = Lock.load(event.address.toHexString())
+
+  if (lock) {
+    const referrerAddress = event.params.referrer.toHexString()
+
+    let referrerFee = ReferrerFee.load(referrerAddress)
+
+    if (!referrerFee) {
+      referrerFee = new ReferrerFee(referrerAddress)
+    }
```
easyrun42, line 420: @julien51 Here we either load an existing `ReferrerFee` by `referrer address` address or create a new one if it doesn't exist.

subgraph/tests/referrer.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    const newReferrerEvent = createReferrerEvent(
+      Address.fromString(defaultMockAddress),
+      BigInt.fromU32(100)
+    )
+
+    // Assign the referrer transfer event with the same lock address ✅
+    newReferrerEvent.address = Address.fromString(lockAddress)
+
+    // Handle the referrer fee as if it was in the subgraph
+    handleReferrerFees(newReferrerEvent)
```
easyrun42, line 79: @julien51 Here is new test added for changing the referrer fee part 🙏

julien51: Very cool!

julien51 (approved):

julien51:
Thanks! Merged!

PR #12648, opened 2023-09-01
Revert "fix(deps): update dependency @unlock-protocol/hardhat-plugin to v0.1.1"

searchableguy:
Reverts unlock-protocol/unlock#12631

seems to break integration tests.


julien51:
Indeed! Thanks @searchableguy

PR #12658, opened 2023-09-06
cleanup(unlock-app): remove extra saving of metadata in the confirmation steps

julien51:
# Description

I checked and we are indeed saving the metadata in the `Metadata` screen, so this was duplicated.
Copying @searchableguy who may have a bit more knowledge about _why_ we were also doing it in the confirmation steps.

# Checklist:


julien51:
One concern worth looking it per @searchableguy is to check what happens when there is hidden metadata and the metadata step is skipped. I will test that.

julien51:
Dang. I merged by accident! I will reopen.

PR #12669, opened 2023-09-12
fix(locksmith): fixing OG images

julien51:
# Description

SVG are not supported so we have to render as PNG.

# Checklist:

### DIFF
--- a/locksmith/src/controllers/v2/og.tsx
+++ b/locksmith/src/controllers/v2/og.tsx
@@ -152,6 +152,7 @@ export const OGEvent = ({
   )
 }
 
+// TODO: replaced by routes on unlock-app since we use Vercel for hosting but only once axos supports the Edge runtime!
 export const eventOGHandler: RequestHandler = async (request, response) => {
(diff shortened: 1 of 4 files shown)


locksmith/src/controllers/v2/og.tsx
```diff
@@ -206,10 +210,15 @@ export const eventOGHandler: RequestHandler = async (request, response) => {
     }
   )
 
+  // OG cannot be SVG
+  const svg = new Resvg(eventBannerSVG)
+  const pngData = svg.render()
```
clemsos, line 215: Maybe there is a need to handle a potential error here is the markup is wrong somehow ?

clemsos (approved):
A small error handling maybe but LGTM !

locksmith/src/controllers/v2/og.tsx
julien51: That's a good question. The markup is generated by `satori`, so I am not sure it could be invalid, but there could always be issue. I'll add a try/catch

PR #12671, opened 2023-09-12
fix(unlock-app): simplifying swap and purchase to support discount codes

julien51:
# Description

Using the actual amount for swap and purchase, rather than just the lcok's key price.

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/usePricing.ts
+++ b/unlock-app/src/hooks/usePricing.ts
@@ -7,7 +7,7 @@ import { useWeb3Service } from '~/utils/withWeb3Service'
 
 interface Options {
   currencyContractAddress?: string | null
-  symbol: string
+  symbol: string | undefined | null
   lockAddress: string
   network: number
   recipients: string[]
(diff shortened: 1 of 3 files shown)


unlock-app/src/hooks/usePricing.ts
```diff
@@ -17,6 +17,45 @@ interface Options {
   payment?: any
 }
 
+export const purchasePriceFor = async (
```
julien51, line 20: extracting from `usePricing`

unlock-app/src/hooks/useUniswapRoutes.ts
```diff
@@ -18,21 +18,108 @@ export interface UniswapRoute {
...
+        network: lock.network,
+        recipients,
+        data: purchaseData || recipients.map(() => ''),
+        paywallConfig,
+        currencyContractAddress: lock.currencyContractAddress,
+        symbol: lock.currencySymbol,
+      })
+
+      // compute total
+      const price = prices.reduce((acc, item) => acc + item.amount, 0)
```
clemsos, line 52: Maybe a check is needed here if the returned prices are null or NaN?

clemsos (approved):
A small check to make sure will fail gracefully if prices or routes are not found but otherwise LG

PR #12678, opened 2023-09-14
fix(hardhat-plugin): lock version bump fix

clemsos:
# Description

This adds a fix for https://github.com/unlock-protocol/unlock/commit/2023e81a5d1ececd84e6e857ce8d9413d5a96322 that apparently broke integration tests

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/docker/development/eth-node/package.json
+++ b/docker/development/eth-node/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@nomiclabs/hardhat-ethers": "2.2.3",
-    "@unlock-protocol/hardhat-plugin": "0.0.18",
+    "@unlock-protocol/hardhat-plugin": "latest",
     "eslint": "8.48.0",
     "ethers": "5.7.2",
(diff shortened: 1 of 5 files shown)


docker/development/eth-node/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@nomiclabs/hardhat-ethers": "2.2.3",
-    "@unlock-protocol/hardhat-plugin": "0.0.18",
+    "@unlock-protocol/hardhat-plugin": "latest",
```
clemsos, line 13: it's good to use `latest` here so we make sure we are up to date in our test pipeline. Maybe that requires some tweak in renovate config to prevent it from opening PR ?

julien51: HA! The problem with "latest" is that it means we can have inconsistent builds because the `yarn.lock` file will be outdated?

julien51 (approved):

PR #12699, opened 2023-09-25
feat(smart-contracts): add `setKeyExpiration` 

### LINKED ISSUE #12553: Potential overflow when merging keys
I haven't recreated this in the UI, but based on the code for the function _timeMachine, if a key's is merged into a non-expiring key, then line 547 will overflow because it will be adding to uint max. It's worth testing this out in practice, perhaps such a possibility is blocked on the UI.

https://github.com/unlock-protocol/unlock/blob/6abe9661d4c6b3982f9b7af95a48cfa481a9dfdc/smart-contracts/contracts/mixins/MixinKeys.sol#L535-L554

### LINKED ISSUE #12681: Lock managers should be able to set the expiration date to _any_ time
A lock manager should be able to change an individual key's expiration date, rather than just expire. They should be able to set it in the past or the future as well.

clemsos:
# Description

Introduces `setKeyExpiration` to allow a lock manager to freely update the timestamp of any existing keys

# Issues

Fixes #12681 and fixes #12553
Refs #

# Checklist:


smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -619,6 +619,23 @@ contract MixinKeys is MixinErrors, MixinLockCore {
...
   }
 
+  /**
+   * Set the expiration of a key
+   * @param _tokenId the id of the key
+   * @param _newExpiration the new timestamp to use
+   */
+  function setKeyExpiration(uint _tokenId, uint _newExpiration) public {
+    _onlyLockManager();
+    _keys[_tokenId].expirationTimestamp = _newExpiration;
```
julien51, line 629: We probably do that in multiple places (expire, cancellation... etc)
I wonder if we should try to reuse some of the code?

clemsos: This seems similar to the issue identified in  #12553

clemsos:
The tests files may be a bit hard to review as I had taken the opportunity to remove old patterns (remove legacy web3 and bigumber.js) and replace all with ethers.

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -555,6 +555,10 @@ contract MixinKeys is MixinErrors, MixinLockCore {
     uint formerTimestamp = _keys[_tokenId].expirationTimestamp;
 
     if (_addTime) {
+      // cant add to a non-expiring key
+      if (formerTimestamp == type(uint).max) {
```
julien51, line 559: I am a bit worried about this change. Does that break anything?

julien51 (approved):

smart-contracts/contracts/mixins/MixinKeys.sol
```diff
@@ -619,6 +623,41 @@ contract MixinKeys is MixinErrors, MixinLockCore {
...
+   * @param _tokenId the id of the key
+   * @param _newExpiration the new timestamp to use
+   * @param _delta time difference in sec (zero is unknown)
+   * @param _timeAdded boolean `true` = time added, `false` = time removed
+   * @notice is _delta is null and _timeAdded is `false`, then the time diff is considered unknown
+   */
+  function _setKeyExpiration(
+    uint _tokenId,
+    uint _newExpiration,
+    uint _delta,
```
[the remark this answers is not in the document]
clemsos: The goal is was to unified everything and reuse the existing `ExpirationChanged` event - which requires both params. The event is used in our `_timeMachine` and follows that formatting.

There are two solutions 

- keep the params here (as now)
- simplify the `ExpirationChanged` event (which will require a subgraph update) - and potentially add a more specific event for time machine

The 2nd solution will make the contract simpler but will require to change the graph

smart-contracts/contracts/mixins/MixinKeys.sol
clemsos: not that I know of. It just prevents from adding to max uint time. We can also not add it and just let the tx revert.

smart-contracts/contracts/mixins/MixinKeys.sol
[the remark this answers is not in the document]
julien51: I like the 2nd option better : let's simplify `ExpirationChanged` to _not_ include `_delta` or `_timeAdded`, but ideally the expiration before the change as well as the new one.

clemsos:
need to update the subgraph once the event is changed

clemsos:
Ok this seems good to ship. Just need to add the event to the subgraph but better to do it in another PR

julien51:
Thanks!

julien51 (approved):

PR #12786, opened 2023-10-06
fix(unlock-app): fixing transfers when user is on a different chain

julien51:
# Description

I found that transfers were broken when the wallet is on the wrong chain.

# Checklist:


julien51:
There are still a few places where `useWalletService` is used... we should clean that up.

searchableguy (approved):

PR #12827, opened 2023-10-17
feature(locksmith): Multiple signers

julien51:
# Description

As we're working on using 3rd party APIs (like #12002 ) to send transactions, I realized that a good prerequisite work would be to add support to having multiple signers work at the same time in locksmith (right now our code assumes there is just one).

This means doing several changes:
* adding support for multiple signers to the hooks that look at signers
  * Captcha
  * Guild
  In both cases, we look up the lock's hook contract and check if any of the current signers is a signer on the hook and then use that one.
* Adding support for multiple signers to the step to purchase by card
  * for each of the current signer, we check if they are key granters. If they are, we use that one. Otherwise we show an error.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/locksmith/src/config/config.ts
+++ b/locksmith/src/config/config.ts
@@ -84,7 +84,7 @@ const config = {
   recaptchaSecret: process.env.RECAPTCHA_SECRET,
   logtailSourceToken: process.env.LOGTAIL,
   sessionDuration: Number(process.env.SESSION_DURATION || 86400 * 60), // 60 days
-  requestTimeout: '10s',
+  requestTimeout: '25s',
   defenderRelayCredentials,
   databaseUrl: process.env.DATABASE_URL || '',
   sentry: {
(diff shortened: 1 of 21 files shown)


locksmith/src/config/config.ts
```diff
@@ -84,7 +84,7 @@ const config = {
   recaptchaSecret: process.env.RECAPTCHA_SECRET,
   logtailSourceToken: process.env.LOGTAIL,
   sessionDuration: Number(process.env.SESSION_DURATION || 86400 * 60), // 60 days
-  requestTimeout: '10s',
```
julien51, line 87: This was different in our 2 configs... restoring 25s for consistency.

locksmith/src/controllers/purchaseController.ts
```diff
@@ -72,17 +72,18 @@ export class PurchaseController {
       purchaseType,
     } = await PaymentCaptureBody.parseAsync(request.body)
     const dispatcher = new Dispatcher()
-    const hasEnoughToPayForGas = await dispatcher.hasFundsForTransaction(
-      network
-    )
+
+    const [hasEnoughToPayForGas, soldOut] = await Promise.all([
```
julien51, line 76: Parallelizing to make the response snappier.

julien51:
Once this is merged I will do extensive testing on staging to make sure we have everything (still) working as expected..

clemsos (approved):
LG 👍

Issue #12835, opened 2023-10-19
Download _all_ as CSV


julien51:
Our download CSV feature is currently paginated 100 by 100 for scalability reasons.
We need to provide the ability to download "all" at once. This will need to be an async job.

PR #12840, opened 2023-10-19
fix(unlock-app): fixed pagination for CSV downloads until we have full export

julien51:
# Description

Fixing the CSV pagination until we fix it allowing full download. (https://github.com/unlock-protocol/unlock/issues/12835)
This just makes things consistent by letting users download page by page.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/locks/Manage/elements/Members.tsx
+++ b/unlock-app/src/components/interface/locks/Manage/elements/Members.tsx
@@ -11,8 +11,7 @@ import { subgraph } from '~/config/subgraph'
 import { storage } from '~/config/storage'
 import { Placeholder } from '@unlock-protocol/ui'
 import { useWeb3Service } from '~/utils/withWeb3Service'
-
-const itemsPerPage = 30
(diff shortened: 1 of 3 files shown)


unlock-app/src/components/interface/locks/Manage/elements/Members.tsx
```diff
@@ -11,8 +11,7 @@ import { subgraph } from '~/config/subgraph'
 import { storage } from '~/config/storage'
 import { Placeholder } from '@unlock-protocol/ui'
 import { useWeb3Service } from '~/utils/withWeb3Service'
-
-const itemsPerPage = 30
```
clemsos, line 15: Good to store the constant outside of the file but maybe could keep the current `itemsPerPage` pattern with sth like `const itemsPerPage = MEMBERS_PER_PAGE` ?

That will simplify if this is to become dynamic at some point with a selector of how many items per page should be shown

clemsos (approved):
a small comment otherwise lg

PR #12849, opened 2023-10-24
feature(locksmith): yielding a checkout config when returning an event

julien51:
# Description

As am I working on simplifying how events are stored, I am adding a checkout config object to the API that returns events. This makes things much easier on the front-end!

# Checklist:


julien51:
I ended up cleaning types because we had a Paywallconfig type both in `types` and `core`... and the `core` one is more complete so I moved to using that one!

clemsos (approved):
Lg

PR #12850, opened 2023-10-24
refactor(unlock-app, locksmith): Changed data model for events

julien51:
# Description

We used to assume 1 event per lock and all of our event application was built around lock where the metadata for the event is stored in the lock object.
We now have an `event` object that itself points to locks (through a checkout config) and we can build a better UX.

This is a draft..

# Checklist:


julien51:
This needs https://github.com/unlock-protocol/unlock/pull/12849 to be rebased.

unlock-app/src/hooks/useEventOrganizer.ts
```diff
@@ -0,0 +1,37 @@
...
+            lockAddress,
+            account!,
+            checkoutConfig.locks[lockAddress].network
+          )
+        )
+      )
+      return !!isManagerByLock.find((isManager) => !!isManager)
+    }
+  )
+}
```
searchableguy, line 37: Don't we already have a similar hook?

unlock-app/src/pages/event/[slug].tsx
```diff
@@ -10,47 +12,26 @@ interface Params {
...
-  const { lockAddress, network, metadata } = p.pageProps
-  return (
-    <EventContentWithProps
-      lockAddress={lockAddress}
-      network={network}
-      metadata={metadata}
-    ></EventContentWithProps>
-  )
+const EventPage = (props: EventPageProps) => {
+  console.log(props.pageProps)
```
Comment (searchableguy) [inline unlock-app/src/pages/event/[slug].tsx:33]: remove

unlock-app/src/services/ethpass.ts
```diff
@@ -26,46 +26,6 @@ export const isEthPassSupported = (network: number) => {
   ].includes(network)
 }
 
-export const applePass = (
```
julien51, line 29: This is not used, so I am cleaning it up!

unlock-app/src/hooks/useEventOrganizer.ts
julien51: Not exactly. We have something to check for a single lock. I was initially wondering if we could re-use, but I thought it was also more useful to have a different name/function for clarity.

unlock-app/src/pages/event/[slug].tsx
Comment (julien51) [reply unlock-app/src/pages/event/[slug].tsx]: Done :)

unlock-app/src/components/content/EventContent.tsx
```diff
@@ -9,6 +9,8 @@ import EventDetails from './event/EventDetails'
 import { EventLandingPage } from './event/EventLandingPage'
 import { useRouterQueryForLockAddressAndNetworks } from '~/hooks/useRouterQueryForLockAddressAndNetworks'
 import { useMetadata } from '~/hooks/metadata'
+import { toFormData } from '~/components/interface/locks/metadata/utils'
+import { Event, PaywallConfigType } from '@unlock-protocol/core'
 
 export const EventContent = () => {
   const {
```
julien51, line 16: This is the legacy path (before slugs existed on locks...) We have ti reconstruct the whole object from the lock address.

PR #12851, opened 2023-10-25
chore(smart-contracts): remove chai + bignumber pattern from tests

clemsos:
# Description

This PR starts removing the chai lib from tests - also because renovate https://github.com/unlock-protocol/unlock/pull/12604 is failing 

# Issues

Fixes #
Refs #

# Checklist:


smart-contracts/test/UnlockDiscountToken/governance.js
```diff
@@ -1,166 +1,146 @@
...
+  notExpectEvent,
+  compareBigNumbers,
+  compareBigNumberArrays,
+} = require('../helpers')
 
-    return receipts
-  } finally {
-    // enable auto-mining
-    await network.provider.send('evm_setAutomine', [true])
-  }
```
clemsos, line 49: all that logic was intended to test that the gov contract could only add a single endpoint per block. This was adapted from OZ tests and is actually tested already by OZ so we don't need to test it again here as we rely on their work.

packages/contracts/package.json
```diff
@@ -30,7 +30,7 @@
     "@unlock-protocol/networks": "workspace:^",
     "@unlock-protocol/tsconfig": "workspace:./packages/tsconfig",
     "@unlock-protocol/types": "workspace:^",
-    "chai": "4.3.7",
```
julien51, line 33: Ha, should we remove it rather than update it?

clemsos: Actually here are like 25 other tests in the folder that uses chai currently, so it is a bit more work to remove it... We can either rename that PR and merge with all tests passing, or continue to work here to remove chai entirely. 

I personally dont think removing chai is a high priority as we removed the thing that was more pronve tp break (the `to.be.bignumber...` pattern).

I will advise we keep chai for the other and merge this PR as it is. Chai in itself is actually very stable

julien51 (approved):

PR #12853, opened 2023-10-25
feature(locksmith): Updated the events API and adding middleware

julien51:
# Description

We're removing the "locks" object from the `EventData` object as we are using a checkout config which itself contains locks. 

# Checklist:

### DIFF
--- a/locksmith/migrations/20231025111934-add_checkout_config_to_events.js
+++ b/locksmith/migrations/20231025111934-add_checkout_config_to_events.js
@@ -1,0 +1,18 @@
+'use strict'
+
+/** @type {import('sequelize-cli').Migration} */
+
+const table = 'EventData'
+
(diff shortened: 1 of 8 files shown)


locksmith/migrations/20231025111934-add_checkout_config_to_events.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+/** @type {import('sequelize-cli').Migration} */
+
+const table = 'EventData'
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn(table, 'checkoutConfigId', {
```
julien51, line 9: Adding `checkoutConfigId` which will be used to point to locks but also store a specific config that can easily be updated!

PR #12943, opened 2023-11-03
refactor(locksmsith): using the public provider to send tx

julien51:
# Description

We only use the OpenZeppelin relay provider when actually sending transactions.
This will reduce the risk of reaching the rate limit.

# Checklist:

### DIFF
--- a/locksmith/src/operations/authorizedLockOperations.ts
+++ b/locksmith/src/operations/authorizedLockOperations.ts
@@ -11,7 +11,7 @@ export const hasAuthorization = async (
   const lockAddress = Normalizer.ethereumAddress(address)
   const web3Service = new Web3Service(networks)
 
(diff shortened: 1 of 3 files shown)


locksmith/src/operations/authorizedLockOperations.ts
```diff
@@ -11,7 +11,7 @@ export const hasAuthorization = async (
   const lockAddress = Normalizer.ethereumAddress(address)
   const web3Service = new Web3Service(networks)
 
-  const { wallet } = await getPurchaser({ network })
+  const wallet = await getPurchaser({ network })
```
clemsos, line 14: it seems the function still return `{ wallet, provider}` so why remove the brackets?

clemsos (approved):
lg

PR #12952, opened 2023-11-06
deps(smart-contracts): bump to OpenZeppelin contracts 5.0

clemsos:
# Description

This PR adds required changes to upgrade our contracts to latest OZ package. 

# Issues

Fixes  #12926 and fixes  #12927
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/contracts/hooks/CaptchaHook.sol
+++ b/smart-contracts/contracts/hooks/CaptchaHook.sol
@@ -8,7 +8,7 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 contract CaptchaHook is Ownable {
   mapping(address => bool) public signers;
 
-  constructor() {}
+  constructor() Ownable(msg.sender) {}
 
   function addSigner(address signer) public onlyOwner {
     signers[signer] = true;
--- a/smart-contracts/contracts/hooks/CaptchaHook.sol
+++ b/smart-contracts/contracts/hooks/CaptchaHook.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.0;
 
 import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
 import "@unlock-protocol/contracts/dist/PublicLock/IPublicLockV12.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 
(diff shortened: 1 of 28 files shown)


clemsos:
Following recent conversations, we may need to flatten several OZ dependencies as the `__gap` pattern we use in mixins is now deprecated in 5.0 release.

julien51:
Do we really want to support 5.0 then? I worry this is adding a lot of complexity... for no useful reason. Let's talk about it.

smart-contracts/contracts/hooks/CaptchaHook.sol
```diff
@@ -2,13 +2,14 @@
...
 import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
 import "@unlock-protocol/contracts/dist/PublicLock/IPublicLockV12.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 
 contract CaptchaHook is Ownable {
   mapping(address => bool) public signers;
 
-  constructor() {}
+  constructor() Ownable(msg.sender) {}
```
basememara, line 12: Maybe use `_msgSender()` from the underlying `Context` contract, just in case `msg.sender` is overridden somewhere downstream by OpenZeppelin? Also for `GuildHook.sol`.

basememara:
> Do we really want to support 5.0 then? I worry this is adding a lot of complexity... for no useful reason. Let's talk about it.

The main benefit I've seen from v5 is lots of gas optimizations if this is something painful right now. And of course lots of ergonomics and alignment between 721 and 1155. And there's always security but would have to dig deeper if anything directly affects the project that's worth it.

Also something to note that v5 requires Solidity 0.8.20 so that has to be minimum version. But because of the new `PUSH0` opcode bonanza, some chains like Polygon mainnet doesn't have that push code yet last I checked recently so will fail deployments there. Therefore, have to downgrade the `evm` version to `paris` so the `PUSH0` opcode isn't used with 0.8.20 compilation. Here's a long discussion at Hardhat that sheds a little bit of light on the nuance: https://github.com/NomicFoundation/hardhat/issues/4391

All this to say upgrading to v5 is great and welcomed, but requires much more testing and deployment testing across testnets and mainnets.

julien51:
thanks for your input @basememara !

> All this to say upgrading to v5 is great and welcomed, but requires much more testing and deployment testing across testnets and mainnets.

Indeed, at this point, I don't think we're ready for v5. I will close this PR.

PR #12997, opened 2023-11-23
feature(locksmith): adding temporary handling of filbangalore tickets

julien51:
# Description

We're adding a mechanism to render the FIL Bangalore tickets in the NFT image.

# Checklist:


locksmith/src/utils/ticket.tsx
```diff
@@ -106,3 +107,43 @@ export const createTicket = async ({
...
+  console.log({ network, lockAddress, tokenId })
+  // We need to generate the icon from SVG + QR code!
+  const qrCode = await generateQrCode({
+    network,
+    lockAddress,
+    tokenId,
+  })
+  // Let's now generate the SVG
+
+  console.log(qrCode)
```
julien51, line 129:
```suggestion
```

locksmith/src/utils/ticket.tsx
```diff
@@ -106,3 +107,42 @@ export const createTicket = async ({
...
+export const ticketForFilBangalore = async ({
+  network,
+  lockAddress,
+  tokenId,
+}: {
+  network: number
+  lockAddress: string
+  tokenId: string
+}) => {
+  console.log({ network, lockAddress, tokenId })
```
julien51, line 120:
```suggestion
```

locksmith/src/controllers/lockController.ts
```diff
@@ -76,18 +80,30 @@ export const stripeConnected = async (req: Request, res: Response) => {
...
+    // Check if there is a custom icon for this key
+    const keyIcon = await getKeyIcon({
+      network,
+      lockAddress,
+      keyId: req.query.id.toString(),
+    })
+    if (keyIcon) {
+      icon = keyIcon
+    }
+  }
```
basememara, line 100: I know this is temporary but maybe keep it encapsulated and private within `getKeyIcon` and do the logic there?

```ts
// lockOperation.ts
export const getLockIcon = async ({
  lockAddress,
  original,
  requestUrl,
  key,
}: {
  lockAddress: string
  original?: boolean
  requestUrl: string
  key?: { id: string; network: number }
}): Promise<IconType> => {
  if (key?.id && key?.network) {
    const icon = await getCustomLockIcon({
      network: key.network,
      lockAddress,
      keyId: key.id,
    })
    if (icon) {
      return icon
    }
  }

  if (original) {
    return getGeneratedLockIcon(lockAddress)
  }

  ...
}
```

Notice this enables the `getCustomKeyIcon` to be private so the temp code doesn't get leaked into `lockController` and can even keep the `icon` variable immutable there as well. But also the `getLockIcon` logic doesn't get called unnecessarily sometimes if it knows there's a custom icon available:

```ts
// lockController.ts
const icon = await getLockIcon({
  lockAddress,
  original: original === '1',
  requestUrl: Normalizer.getRequestURL(req).toString(),
  key: {
    id: String(req.query.id || ''), // Safer then toString in case undefined
    network: parseInt(req.params.network) || 0, // Used parseInt since never expecting floats
  },
})

...
```

locksmith/src/controllers/lockController.ts
```diff
@@ -76,18 +80,30 @@ export const stripeConnected = async (req: Request, res: Response) => {
...
 
-  if (lockIcon.isURL) {
-    return res.redirect(lockIcon.icon)
+  if (icon.isURL) {
+    return res.redirect(icon.icon)
   } else {
-    res.setHeader('Content-Type', lockIcon.type)
-    return res.send(lockIcon.icon)
+    res.setHeader('Content-Type', icon.type!)
```
basememara, line 105: Does the `icon.type` have to be forced because I notice sometimes in `getLockIcon` it can be `null`. Maybe some defensive coding?

```ts
if (icon.type) {
  res.setHeader('Content-Type', icon.type)
}

// or fallback to default
res.setHeader('Content-Type', icon.type ?? 'application/octet-stream')
```

locksmith/src/operations/lockOperations.ts
```diff
@@ -106,6 +113,34 @@ export const getLockIcon = async ({
...
+    const icon = await ticketForFilBangalore({
+      network,
+      lockAddress,
+      tokenId: keyId,
+    })
+
+    return { type: 'image/svg+xml', icon, isURL: false }
+  }
+  // By default, no icon
+  return
```
basememara, line 141: Nitpick but I think returning `null` makes more sense since it means the function always returns a object instead of sometimes it does and sometimes it doesn't at all.

locksmith/src/controllers/lockController.ts
julien51: So the `icon` type is actually not great, because it can either be an actual icon, with a type, or just a URL...

locksmith/src/operations/lockOperations.ts
julien51: fair point!

locksmith/src/controllers/lockController.ts
julien51: I actually think it is imoportant to distinguish a `keyIcon` (individual NFT) from the `lockIcon` (icon for the contract that will be used for all keys if no custom key one is set).

locksmith/src/controllers/lockController.ts
julien51: and if it is a URL we should not ave a content-type at all.

locksmith/src/controllers/lockController.ts
basememara: Ah I see, after straightening out keys and locks in my head, I understand the intent here better and makes perfect sense to keep them separate. Thx for explanation!

locksmith/src/controllers/lockController.ts
basememara: Oh right, that's what the `isURL` is for. Food for thought later is maybe `getLockIcon` should return `IconType | string` then can use a type guard. But I see now in the meantime `type` should safely have a value if `isURL` is false.

PR #12998, opened 2023-11-27
feat(smart-contracts): move Lock proxy to a standalone contract

clemsos:
# Description

This creates a `LockProxy` standalone (flatten) contract. That will make verification process easier across chains.

# Issues

Fixes #
Refs #

# Checklist:


smart-contracts/contracts/Unlock.sol
```diff
@@ -276,11 +276,7 @@ contract Unlock is UnlockInitializable, UnlockOwnable {
     }
 
     // deploy a proxy pointing to impl
-    TransparentUpgradeableProxy proxy = new TransparentUpgradeableProxy(
-      publicLockImpl,
-      proxyAdminAddress,
-      data
-    );
+    LockProxy proxy = new LockProxy(publicLockImpl, proxyAdminAddress, data);
```
julien51, line 279: Ho wow, we already had the getter/setter?

julien51 (commented):
I am confused a bit here. Will it actually change anything on our end?

smart-contracts/contracts/utils/LockProxy.sol
```diff
@@ -0,0 +1,954 @@
+// Sources flattened with hardhat v2.18.3 https://hardhat.org
```
julien51, line 1: Is this just the OZ one? or did we add anything custom?
If not, maybe it's better to use the OZ one?

smart-contracts/contracts/utils/LockProxy.sol
```diff
@@ -0,0 +1,954 @@
...
+        require(msg.value == 0);
+    }
+}
+
+
+// File contracts/utils/LockProxy.sol
+
+interface ILockProxy is ITransparentUpgradeableProxy {}
+
+contract LockProxy is TransparentUpgradeableProxy {
```
julien51, line 948: yeah it looks like we could just use `ITransparentUpgradeableProxy` directly?

clemsos:
reopening as the approach in #13088 won't work until we maintain a custom proxy

smart-contracts/contracts/utils/LockProxy.sol
clemsos: yes we can remove the concept of `LockProxy` and keep only the native (flatten) `TransparentUpgradeableProxy` from Open Zeppelin

julien51 (approved):

PR #13093, opened 2023-12-13
feat(unlock-app): Check if image of key can be played as video if image load fails

sudheerDev:
# Description

If an image of lock is a video then the Avatar component renders fallback text. This PR checks if it can be played as video and shows the video without playback controls if it can be rendered. 

* Use callback `onLoadingStatusChange` of Avatar component and check if a metadata image can be played as video.
* If an image can be played as video remove the fallback of text render

Dec-13-2023 14-59-55

# Issues

Fixes #
Refs #

# Checklist:


searchableguy:
If too many people use video for their NFTs, it can impact the performance and overwhelm the user. Some browsers also block auto play even if you use mute so using a gif would be more appropriate if someone wants to use a loop image that is visible everywhere.

sudheerDev:
> If too many people use video for their NFTs, it can impact the performance and overwhelm the user.
This is subjective about overwhelming user isnt it? 

Does adding video support in unlock make the creators do nfts in video? because most platforms already support it and their choice of this is not so dependent on unlock IMO. Like in our case, we will go ahead with video either way. 

Even with performance that is likely with quite a few high frame rate and really high quality resolution, as an example https://superrare.com/anyma each file here is about 49mb. 

> Some browsers also block auto play even if you use mute so using a gif would be more appropriate
yes there are reasons why they do it, but asking someone to use gifs because of the reason seems not appropriate, because it should be freedom of choice for the user as it is not like a thumbnail and more of an nft

At the end it is a platform choice, i will leave it up to you guys to take a call and close the PR if it does not fit well with your maintenance or your platform needs

searchableguy:
Sorry if I wasn't super clear, this is a great addition. I'm commenting more on the auto play part since I'm not sure if having videos auto play by default is a good option as opposed to only playing it when a user hovers over it.

sudheerDev:
@searchableguy Aah okay, thanks. ya i understand, removed autoPlay and loop too.  i am not sure of the test fail here, is that a flaky test or failing here on this pr particularly?

sudheerDev:
> Thanks @sudheerDev ! What's failing is not a test but Typescript :)
> 
> ```
> ./src/components/interface/keychain/Key.tsx:204:21
> Type error: Property 'play' does not exist on type 'never'.
> ```

aah ya, I have to check this again. Not sure of this works now

julien51:
Hey @sudheerDev , we're doing a bit of cleanup here. Please post an update here or we will close this PR iun the next week :(

sudheerDev:
@julien51 hey, will fix the test and conflicts this week. Should not take much time

julien51:
Thanks @sudheerDev ! Please keep us posted!

julien51 (approved):

PR #13095, opened 2023-12-13
fix(unlock-app): looking up ens names from addresses (#13090)

### LINKED ISSUE #13090: ENS reverse resolution
We show address in _many_ places inside of the dashboard.
Let's add support for ENS reverse resolution (from an address, show the ENS) and let's apply that in every place where we show addresses.

teawaterwire:
# Description

Fixing the reverse lookup of ENS names (from an Ethereum address). 

image

image

image

# Issues

Fixes #13090


julien51 (approved):
so ENS only exists on ethereum mainnet for now... so I think we should ignore the network that the user is connected to and just use `1` for lookups?

teawaterwire:
> Ho wow that was quick :) Can you please check all the places were we display addresses and see if we can show the ENS ( I am thinking about the lock settings with lock managers, the members UI... etc)

i replaced it where `useEns` was used (see screenshots)

> so ENS only exists on ethereum mainnet for now... so I think we should ignore the network that the user is connected to and just use 1 for lookups?

if we use `1` then it doesn't work on Goerli

julien51:
> if we use 1 then it doesn't work on Goerli

Well that's the bug :) We need to make sure can do the reverse resolution anywhere on the app, whatever network the user is connected to :)

teawaterwire:
ok now resolving always on mainnet: 
image

julien51 (approved):

julien51 (approved):

PR #13097, opened 2023-12-13
fix(locksmith): better handling of timeouts when retrieving data from subgraphs

julien51:
# Description

As I was investigating why locksmith was serving some 500 from time to time I realized we are not properly handling timeouts when retrieving data from the subgraphs. This PR adds the ability to timeout "earlier".

I also found that we are making 2 requests to subgraphs (one to get the expiration and one to get the lock name) which is very wasteful. We will have to clean that up in another PR.

# Checklist:


locksmith/src/utils/promises.ts
```diff
@@ -0,0 +1,18 @@
+export const waitOrTimeout = async (
```
julien51, line 1: A simple util to race with a timeout and yield back control to the caller when that happens.

locksmith/src/utils/promises.ts
```diff
@@ -0,0 +1,18 @@
...
+  promise: Promise<any>,
+  delay: number,
+  responseIfTimeout: () => any
```
clemsos, line 4: Do these args really need to be passed? `delay` could have directly `TIMEOUT_ON_SUBGRAPH` set as default value, then just a smaller error message string

locksmith/src/utils/metadata.ts
```diff
@@ -1,23 +1,45 @@
...
-      {
-        network,
+        {
+          network,
+        }
+      ),
+      TIMEOUT_ON_SUBGRAPH,
+      () => {
+        logger.error(
+          `Timed out after 3s while retrieving info from Subgraph for lock metadata ${lockAddress} on ${network}`
```
clemsos, line 39: the `3s` should be parsed from `TIMEOUT_ON_SUBGRAPH` for consistency

clemsos (approved):
few comments for simplification but lg otherwise

locksmith/src/utils/metadata.ts
julien51: Great point! I'll do that!

PR #13198, opened 2024-01-11
fix(contracts): allow template initialization directly after deployment

clemsos:
# Description

This allows to initialise PublicLock directly after deploying the contract or during the setting of the template in Unlock. This will also allow to reuse a template - before the tx will revert when trying to set a template that has already been set.

# Issues

Fixes #
Refs #

# Checklist:


clemsos:
The Unlock contract needs to be re-packaged before merging but better to do it after #13197 is merged

smart-contracts/contracts/Unlock.sol
```diff
@@ -195,10 +195,14 @@ contract Unlock is UnlockInitializable, UnlockOwnable {
    * @dev This will initialize the template and revokeOwnership.
    */
   function addLockTemplate(address impl, uint16 version) public onlyOwner {
-    // First claim the template so that no-one else could
-    // this will revert if the template was already initialized.
-    IPublicLock(impl).initialize(address(this), 0, address(0), 0, 0, "");
-    IPublicLock(impl).renounceLockManager();
+    // if the template has not been initialized,
+    // claim the template so that no-one else could
+    try IPublicLock(impl).initialize(address(this), 0, address(0), 0, 0, "") {
```
julien51, line 200: 👍 Great idea to use try/catch here!

smart-contracts/contracts/Unlock.sol
```diff
@@ -195,10 +195,14 @@ contract Unlock is UnlockInitializable, UnlockOwnable {
...
+      // renounce the lock manager role that was added during initialization
+      IPublicLock(impl).renounceLockManager();
+    } catch {
+      // failure means that the template is already initialized
```
julien51, line 204: Would there be an easy way to verify that the contract does not have any lock nanager? Probably not?
That would be a check we can do here.

julien51 (approved):

smart-contracts/contracts/Unlock.sol
clemsos: No, there is no way to list the lock managers atm

julien51: ok, thanks!

PR #13078, opened 2023-12-06
feature(networks): Adding deployment on zkSync

julien51:
# Description

This is work in progress to add support for zkSync!
Since zksync is not perfectly EVM compatible, we have to follow the steps shown there https://era.zksync.io/docs/tools/hardhat/migrating-to-zksync.html

replaces #13217
# Checklist:

- [ ] 1 PR, 1 purpose: my Pull Request applies to a single purpose
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the docs to reflect my changes if applicable
- [ ] I have added tests (and stories for frontend components) that prove my fix is effective or that my feature works
- [ ] I have performed a self-review of my own code
- [ ] If my code involves visual changes, I am adding applicable screenshots to this thread

PR #13233, opened 2024-01-24
feat(hardhat-helpers): standalone contracts deploy functions

clemsos:
# Description

This refactoring adds some helpers function for contract deployments so it is easier to deploy on zk chains (see #13078 )

# Issues

Fixes #
Refs #

# Checklist:


packages/hardhat-helpers/src/deploy.js
```diff
@@ -0,0 +1,20 @@
+export const deployContract = async (
```
julien51, line 1: Do we also need a `deployUpgradableContract` helper?
Maybe in a follow up PR?

julien51 (commented):
Did you replace in all the scripts to deploy contracts in the `governance` folder?

julien51:
Do we have tests for `governance`? Can we just add a file that runs in CI that deploys all contracts on a "local" node so we can make sure the scripts are working and we don't end up finding that they don't work when we need to deploy a new network/contract?

packages/hardhat-helpers/src/deploy.js
clemsos: yes, added too here

clemsos:
> Do we have tests for `governance`? Can we just add a file that runs in CI that deploys all contracts on a "local" node so we can make sure the scripts are working and we don't end up finding that they don't work when we need to deploy a new network/contract?

That will be good indeed. We have tests already in `governance` folder but they are js unit tests. Will be good to have a way to test the cli tasks

governance/scripts/deployments/uniswap-v2.js
clemsos, line 1: I removed all tasks and scripts related to Uniswap V2 deployment

julien51 (approved):
LGTM!

julien51:
> That will be good indeed. We have tests already in governance folder but they are js unit tests. Will be good to have a way to test the cli tasks

Want to add that in another PR maybe?

clemsos:
> > That will be good indeed. We have tests already in governance folder but they are js unit tests. Will be good to have a way to test the cli tasks
> 
> 
> 
> Want to add that in another PR maybe?

Yes I added a basic one here but can add more coverage in a separate PR

PR #13358, opened 2024-02-16
Update WalletlessRegistration.tsx

ccarfi:
remove additional text. recommendation from snax. walletless folks in particular don't need to know about gas fees.


julien51 (changes requested):
Then let's just just remove the gas fees part. I think it's important to mention *why* we ask for the wallet address

julien51:
I fixed it in another PR! Thanks!

PR #13419, opened 2024-03-05
feat(governance): Improve scripts to inspect and execute bridge txs

clemsos:
# Description

Adds README instructions and improved scripts to manipulate bridged txs and Zodiac delay module actions 

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/governance/README.md
+++ b/governance/README.md
@@ -218,3 +218,49 @@ Edit directly the amounts and prices in the script
 ```
 yarn run scripts/uniswap/addLiquidity.js
 ```
+
+## Cross-Chain DAO Proposals
+
+To maintain the integrity of the protocol accross various chains, we use a pattern of DAO proposals that allows execution on multiple chains. Messaging is sent accross the [Connext bridge](https://connext.network) to all supported chains.
+
+### Prepare a cross-chain proposal
+
+#### Write a cross-chain DAO proposal
+
+Read the explanations and follow the template in [`./proposals/006-cross-bridge-proposal.js`](./proposals/006-cross-bridge-proposal.js) to submit a cross-chain proposal to the DAO.
+
+#### Test a cross-chain DAO proposal
+
+To make sure all calls can be executed properly, you can use Tenderly forks to test execution of calls on each destination chains.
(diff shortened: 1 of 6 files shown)


julien51:
Does this include the script that "batches" all multisigs tx into a single one for the bridge?

clemsos:
> Does this include the script that "batches" all multisigs tx into a single one for the bridge?

No, but the logic is present in the multisig approach we are using (using the safe-global libs). I think we can merge this one which is more about tooling and add the batched bridge txs in a separate PR (maybe for next cross chain proposal).

governance/helpers/bridge/xCall.js
```diff
@@ -0,0 +1,153 @@
...
+        ...prev,
+        [curr.governanceBridge.domainId]: curr,
+      }),
+      {}
+    )
+}
+
+const connextSubgraphIds = {
+  1: `FfTxiY98LJG6zoiAjCXdT34pAmCKDEP8vZRVuC8D5Gf`,
+  137: `7mDXK2K6UfkVXiJMhXU8VEFuh7qi2TwdYxeyaRjkmexo`, //plygon
```
julien51, line 107:
```suggestion
  137: `7mDXK2K6UfkVXiJMhXU8VEFuh7qi2TwdYxeyaRjkmexo`, //polygon
```

governance/helpers/bridge/xCall.js
```diff
@@ -0,0 +1,153 @@
...
+    .reduce(
+      (prev, curr) => ({
+        ...prev,
+        [curr.governanceBridge.domainId]: curr,
+      }),
+      {}
+    )
+}
+
+const connextSubgraphIds = {
```
julien51, line 105: Maybe we should mobe these to the networks file?

julien51 (approved):

governance/helpers/bridge/xCall.js
clemsos: not sure we will need them someone else than here for now, so better to keep them here imo

PR #13425, opened 2024-03-07
feat(unlock-app): enable renewals for Crossmint

julien51:
# Description

Allowing renewals for crossmint.
We need to do extensive testing on this:
1. It is not sure this would work for any collection
2. we should test with the existing Forta locks

# Checklist:


packages/unlock-js/src/PublicLock/v10/getTokenIdForOwner.js
```diff
@@ -11,7 +11,7 @@ export default async function (lockAddress, owner, network) {
 
   const validTokens = []
   const allTokens = []
-  const balanceOfTokens = (await lockContract.balanceOf(owner)).toNumber()
```
julien51, line 14: That was a bug in `unlock-js`!

unlock-app/src/components/interface/checkout/main/Payment.tsx
```diff
@@ -101,7 +101,7 @@ export function Payment({ injectedProvider, checkoutService }: Props) {
       recipients,
     })
 
-  const enableCrossmint = !state.context.renew && crossmintEnabled
```
julien51, line 104: We allow crossmint for renewals as well!

PR #13426, opened 2024-03-08
feature(smart-contracts): Gitcoin Passport hook

### LINKED ISSUE #13208: Gitcoin Passport hook
Can we create a Hook for `onKeyPurchase` that would require the recipient of the NFT to have a Gitcoin passport?

0xTxbi:
# Description
This PR introduces a Gitcoin Passport verification hook to the onKeyPurchase function. The addition serves to restrict NFT purchases to users holding a valid Gitcoin Passport.

# Issues

Fixes #13208
Refs #13208

# Checklist:


julien51:
Thanks a lot @0xTxbi ! Please add yourself to the CLA bot config file!

.clabot
```diff
@@ -58,7 +58,8 @@
     "sudheerDev",
     "SVell",
     "JoaoCampos89",
-    "jchanolm"
+    "jchanolm",
+    "0xTxbi"
```
julien51, line 62: @0xTxbi Please move this change to another PR!

julien51: We can't do it for you because of course that would not mean you agree to it :)

0xTxbi: very true. done. moved it to another PR here: #13432

smart-contracts/contracts/hooks/GitcoinHook.sol
```diff
@@ -0,0 +1,94 @@
...
+    address recipient,
+    address /* referrer */,
+    bytes calldata signature /* data */
+  ) external view returns (uint256 minKeyPrice) {
+    string memory message = toString(recipient);
+    require(checkIsSigner(message, signature), "WRONG_SIGNATURE");
+    if (address(msg.sender).code.length > 0) {
+      return IPublicLockV13(msg.sender).keyPrice();
+    }
+    return 0;
```
julien51, line 35: We should probably never return `0` here because that means the user could pay `0` instead.
I think a safer approach is just to revert here?

julien51 (changes requested):
Let's make sure we do NOT return `0` in any case.
Also I would love to have @clemsos check this!

smart-contracts/contracts/hooks/GitcoinHook.sol
0xTxbi: you're right, that was an oversight. great observation! now fixed!

julien51 (approved):

PR #13469, opened 2024-03-14
feat(hardhat-plugin,contracts): bump contracts versions

clemsos:
# Description

This updates default versions to Unlock 13 and PublicLock 14 when deploying protocol from hardhat plugin (currently used in integration tests)

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/packages/hardhat-plugin/src/constants.ts
+++ b/packages/hardhat-plugin/src/constants.ts
@@ -1,5 +1,5 @@
-export const UNLOCK_LATEST_VERSION = 11
-export const PUBLIC_LOCK_LATEST_VERSION = 13
+export const UNLOCK_LATEST_VERSION = 13
+export const PUBLIC_LOCK_LATEST_VERSION = 14
 
 // task names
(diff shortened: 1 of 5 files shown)


packages/hardhat-plugin/src/constants.ts
```diff
@@ -1,5 +1,5 @@
-export const UNLOCK_LATEST_VERSION = 11
-export const PUBLIC_LOCK_LATEST_VERSION = 13
+export const UNLOCK_LATEST_VERSION = 13
```
julien51, line 1: I wonder if our "contracts" package should just have some `latest` fields?

julien51 (approved):

packages/hardhat-plugin/src/constants.ts
clemsos: yes it will be the easiest and safest way. We need an alias in our exports, there is an existing opened issue on these I referenced  here

clemsos:
closing at is has been replaced by #13484

Issue #13363, opened 2024-02-21
Renew keys shortly before expiry


haseebrabbani:
**Is your feature request related to a problem? Please describe.**
We are using Unlock keys to gate access to the Forta API which provides threat intelligence data, but when our users purchase an Unlock key using a credit card (i.e. Crossmint) they are currently not able to automatically renew the key. Those users have to wait until the key expires, and then immediately purchase again. This can cause some short but potential downtime for those users.

**Describe the solution you'd like**
The feature request is to be able to renew a key shortly _**before**_ expiry (maybe a few days) so that downtime can be avoided.

**Describe alternatives you've considered**
We've considered allowing more than one key per user, but this complicates things a lot. We would prefer single keys per user that can be renewed before expiry

julien51:
@clemsos can you confirm that the contract will support that? Iirc we made it possible to renew slightly better expiration?

julien51:
If so it should be just a pretty simple change to the worker to look at keys expiring "soon" vs just the one recently expired.

clemsos:
Hello @haseebrabbani , thanks for the feedback ! 

Currently our contract will flag a key as "renewable" if it has reached 90% of its total expiration time (see the lines below). One can check if a key can be renewed by calling `isRenewable` directly on the contract.

https://github.com/unlock-protocol/unlock/blob/8d422f7d80444525842aceeb391564892bada419/smart-contracts/contracts/mixins/MixinPurchase.sol#L197-L200

Does the issue only happen with credit card? Or also with crypto subscriptions?

julien51:
@clemsos Can you look and see if we can update our scripts so that renewals are kicked at 90% instead of 100%? This will avoid any "downtime", correct?

PR #13471, opened 2024-03-14
feat(locksmith) trying to renew keys before they actually expire

julien51:
# Description

The membership contract supports renewals before the actual expiration (starting at 90% of the duration to be precise). In order to avoid a "gap" in valid memberships, we now run the cronjob to renew so that it checks for keys that "will soon expire", rather than just the expired ones.

# Issues

Fixes #https://github.com/unlock-protocol/unlock/issues/13363
Refs #

# Checklist:


clemsos (approved):
LGTM

Issue #13631, opened 2024-04-15
Hardhat plugin broken with ethers 6?


julien51:
I set up a new Hardhat project from scracth. I then added the Unlock hardhat plugin. Then in my scripts/tests I am trying to deploy the whole protocol with `await unlock.deployProtocol();` and I get the following error:

```

  1) TokenUriHook
       "before all" hook in "TokenUriHook":
     TypeError: Cannot read properties of undefined (reading 'wait')
      at deployUpgreadableContract (node_modules/@unlock-protocol/hardhat-plugin/src/utils.ts:56:29)
      at async deployUnlock (node_modules/@unlock-protocol/hardhat-plugin/src/deployProtocol.ts:38:28)
      at async deployProtocol (node_modules/@unlock-protocol/hardhat-plugin/src/deployProtocol.ts:105:18)
      at async Context.<anonymous> (test/TokenUriHook.ts:14:5)


```

I suspect this is an issue with the ethers version?

StephanGerbeth:
@nomicfoundation/hardhat-toolbox@5.0.0 requires ethers@6.4.0. so it is not possible to use hardhat unlock plugin which requires ethers@5

clemsos:
hey @StephanGerbeth the latest version of the plugin (0.2.1) now supports ethers6.

PR #13657, opened 2024-04-22
fix(hardhat-plugin): upgrade to support ethers v6

clemsos:
# Description

Support ethers v6 by default in our hardhat plugin

# Issues

Fixes #
Refs #13631

# Checklist:


packages/hardhat-plugin/test/fixture-projects/hardhat-project/hardhat.config.ts
```diff
@@ -1,5 +1,7 @@
 import { HardhatUserConfig } from 'hardhat/types'
 
+import '@nomicfoundation/hardhat-toolbox'
```
clemsos, line 3: the default test case imports the main hardhat lib to avoid errors

packages/hardhat-plugin/package.json
```diff
@@ -36,6 +36,13 @@
     "directory": "packages/hardhat-plugin"
   },
   "devDependencies": {
+    "@nomicfoundation/hardhat-chai-matchers": "2.0.0",
```
julien51, line 39: Should we list them as peer dependencies?

julien51 (approved):

packages/hardhat-plugin/package.json
clemsos: Yes it will be better to have it optional. This requires a bit of changes to use ethers provider directly as well as contracts function (vs the one provided by the hardhat-ethers plugin currently in use in our lib)

julien51 (approved):

PR #13668, opened 2024-04-24
Update building-token-gated-applications.md

ccarfi:
adding platform support table


julien51:
Hey @ccarfi I am actually not sure this is the right place. I think this belongs in guides more than in the dev docs.
I think you could add a Guide called "Token Gating with Unlock" in which would provide a "general" approach and then this table with pointers to these application.

Wdyt?

julien51:
Closing in favor of Guide : https://unlock-protocol.com/guides/token-gating-with-unlock/

PR #13672, opened 2024-04-24
feat(unlock-app): Favorite lock feature

### LINKED ISSUE #13128: "Favorite" section and indicator for locks in dashboard
As a Lock Manager with many locks across many networks, I should be able to "Favorite" any number of locks and see them first in my dashboard, similar to the way that Google Analytics and other applications handles a similar information overload problem

For example, could mark any Lock with a ⭐ and have a section at the top of the dashboard that showed the ⭐ locks

SVell:
# Description
"Favorite" section and indicator for locks in dashboard

# Issues

Fixes #13128
Refs #13128

# Checklist:

image
image


julien51:
@SVell Can we put the star icon near the other icons by the address or just at the end of the title? Right now it feels a bit off.

unlock-app/src/components/interface/locks/List/elements/LockList.tsx
```diff
@@ -101,8 +123,32 @@ export const LockList = ({ owner }: LockListProps) => {
     )
   }, true)
 
+  const [favoriteLocks, setFavoriteLocks] = useState<FavoriteLocks>(
+    JSON.parse(localStorage.getItem('favoriteLocks') as string)
+  )
+
+  const saveFavoriteLocks = (favoriteLocks: FavoriteLocks) => {
+    console.log('saveFavoriteLocks', favoriteLocks)
```
julien51, line 131: Cleanup :)

unlock-app/src/hooks/useAppStorage.ts
```diff
@@ -40,6 +40,15 @@ export function useAppStorage() {
     return null
   }, [])
 
+  const clearStorageWithoutFavorite = useCallback((): void => {
```
julien51, line 43: No no, this is bad practice, The `useAppStorage` code should not _care_ about what kind of data it stores. 

Instead of that, modify the `clearStorage` function below to accept the list of values it should delete and make sure that when it is called, it only deletes these values.

unlock-app/src/hooks/useProvider.ts
```diff
@@ -203,7 +204,7 @@ export const useProvider = (config: any) => {
     setNetwork(undefined)
     setConnected(undefined)
 
-    clearStorage()
+    clearStorageWithoutFavorite()
```
julien51, line 207: That's where you would pass `clearStorage(["x", "y",...])` where `x`, `y` etc are the keys to delete.

unlock-app/src/hooks/useAppStorage.ts
```diff
@@ -40,27 +40,23 @@ export function useAppStorage() {
     return null
   }, [])
 
-  const clearStorageWithoutFavorite = useCallback((): void => {
+  /*
+   * @param exclude: string[] - Array of keys to exclude from deletion, only for keys with APP_NAME
+   */
+  const clearStorage = useCallback((exclude?: string[]): void => {
```
julien51, line 46: No! Now that means every piece of code that calls `cleaStorage` needs to know of _all_ objects stored in localstorage otherwise it will erase them.
Please do what I have been recommending for the last 2 days: only have *explicit* deletes, either thru `clearStorage` or thru `removeKey`.

julien51 (changes requested):
In a good codebase, there is separation of concerns and decoupling. When using a library the library should NOT need to know how it is being used and the calling code should NOT have to know about other codes are using it.

Here, when a calling function deletes a value it should only delete the values that it has set.

julien51 (approved):

PR #13778, opened 2024-05-03
fix(hardhat-plugin): upgrade to support ethers v6 (again)

julien51:
Reverts unlock-protocol/unlock#13777

This should be tested more as this broke master @clemsos !


clemsos:
That one should prevent this PR tests from failing https://github.com/unlock-protocol/unlock/pull/13784

clemsos (approved):

PR #13793, opened 2024-05-07
fix(unlock-app): Fixed first mint error in checkout

SVell:


unlock-app/src/components/interface/checkout/main/Minting.tsx
```diff
@@ -144,13 +144,12 @@ export function Minting({
       if (tokenIds.filter((tokenId?: string) => !!tokenId).length) {
         return tokenIds
       }
-      await sleeper(1000)
+      await sleeper(5000)
```
julien51, line 147: Hum. Please explain? Asking users to wait for 5 seconds is not right!

PR #13817, opened 2024-05-09
fix(unlock-app): Paywall fix

SVell:
# Description
Urgent paywall fix

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/Stepper.tsx
+++ b/unlock-app/src/components/interface/checkout/Stepper.tsx
@@ -7,6 +7,7 @@ import { CheckoutHookType, CheckoutService } from './main/checkoutMachine'
 import { useStepperItems } from './main/useStepperItems'
 import { useSIWE } from '~/hooks/useSIWE'
 import { useAuth } from '~/contexts/AuthenticationContext'
+import { useSelector } from '@xstate/react'
 
 interface IconProps {
   active?: boolean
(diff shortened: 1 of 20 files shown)


unlock-app/src/components/interface/checkout/Stepper.tsx
```diff
@@ -122,6 +127,7 @@ export const Stepper = ({
               key={idx}
               onClick={() => {
                 if (item.to === 'CONNECT') {
+                  if (useDelegatedProvider) return
```
julien51, line 130: let's add a note here to "hide" the stepper item completely rather?

unlock-app/src/components/interface/checkout/main/Disconnect.tsx
```diff
@@ -36,7 +36,7 @@ const Disconnect = ({ service }: DisconnectProps) => {
     setIsDisconnecting(false)
   }
 
-  if (!account) {
+  if (!account || state.context.paywallConfig.useDelegatedProvider) {
```
julien51, line 39: 👍

unlock-app/src/components/interface/checkout/main/Select.tsx
```diff
@@ -337,7 +341,20 @@ export function Select({ checkoutService }: Props) {
...
+  const { signIn, isSignedIn } = useSIWE()
+  const useDelegatedProvider = paywallConfig?.useDelegatedProvider
+
   useEffect(() => {
+    const signToSignIn = async () => {
+      await signIn()
+    }
+
+    if (!connected && useDelegatedProvider) {
+      console.log('Not connected')
```
julien51, line 354:
```suggestion
```

julien51: I wonder if we should move this logic "higher" in the hierarchy because I worry that when we refactor the `Select` step we will miss it. Can you move it higher please?

julien51 (approved):
ok I think that'll do for now.

PR #13824, opened 2024-05-10
feat(subgraph): Receipts improved

julien51:
# Description

We recently found that accessing the transaction's `value` was not accurate because the call to `pruchaseKeys` could be happening in a "larger" transaction with a different value.
In order to handle this, in the short term, we are looking at the GNPChange event's value.

# Checklist:


subgraph/src/receipt.ts
```diff
@@ -79,6 +80,22 @@ export function createReceipt(event: ethereum.Event): void {
...
     receipt.amountTransferred = event.transaction.value
+    // We cannot trust `event.transaction.value` because the purchase function could in fact
+    // be happening inside of a larger transaction whose value is not the amount transfered,
+    // In that case, we need to look up the GNPChanged event
+    if (logs) {
+      for (let i = 0; i < logs.length; i++) {
+        const txLog = logs[i]
+
+        if (
+          // txLog.address == UNLOCK_ADDRESS &&
```
julien51, line 91: I am not sure how to do that since we don't actually have Unlock's address here?

clemsos (approved):
LG but good to add a test (tests are https://github.com/unlock-protocol/unlock/blob/3fdb26c29bbdcaa47d3e3d5e67ce6016d9a29671/tests/test/subgraph.ts#L291

subgraph/src/receipt.ts
```diff
@@ -79,6 +85,25 @@ export function createReceipt(event: ethereum.Event): void {
     log.debug('Creating receipt for base currency lock {}', [lockAddress])
     receipt.payer = event.transaction.from.toHexString()
     receipt.amountTransferred = event.transaction.value
+    // We cannot trust `event.transaction.value` because the purchase function could in fact
+    // be happening inside of a larger transaction whose value is not the amount transfered,
+    // In that case, we need to look up the GNPChanged event
+    if (logs) {
```
clemsos, line 91: It seems that part is only needed if the tx value is null

julien51: That's what I initially thought, but in fact the `transaction.value` could also be "wrong". Imagine if a safe tx buys 2 keys but starts with a 3rd tx in the buddle which funds the safe? Thent the value would be wrong

clemsos: Yes in the case of a safe or a multicall then the value can be calculated that way. Also for multiple purchases the value is the total of all keys purchased.

julien51: Actually, that is a very good point. For receipts we actually have a single receipt per transaction, even if it ends up minting multiple keys (similar to when you buy multiple things at once you only get a single receipt at the store).
So my approach of looking at the GNP is not great.

I guess I can loop over all the GNP changed events for now.

We _really_ need an event to be triggered for receipt, like `PurchaseReceipt()` that would include the required details: `payer`, `contract`, `totalPaid`, `hash`... etc.

subgraph/tests/mockTxReceipt.ts
```diff
@@ -109,23 +109,66 @@ export function newTransactionReceipt(
...
     defaultAddressBytes
   )
 }
+
+// Create transaction receipt for mock transaction
+export function newCancelKeyTransactionReceipt(
+  tokenAddress: Address,
+  refund: BigInt
+): ethereum.TransactionReceipt {
+  return newTransactionReceipt([
```
julien51, line 132: Ohhh! That makes sense. I am not sure why I assumed there could be a single "log" per mocked transaction.

clemsos:
after some digging, the integration tests are failing because the subgraph is enable to access the logs topics of the `GNPChanged` event. The log are correctly emitted but for some reasons, the event emitted by other contracts that the one tracked by the graph "datasource" is empty -no values in topics, just the `topics[0]` hash.

subgraph/src/receipt.ts
```diff
@@ -76,9 +82,37 @@ export function createReceipt(event: ethereum.Event): void {
...
+      const lockContract = PublicLock.bind(Address.fromString(lockAddress))
+      const unlockAddress = lockContract.try_unlockProtocol()
+      let value = BigInt.zero()
+      for (let i = 0; i < logs.length; i++) {
+        const txLog = logs[i]
+        if (
+          txLog.address == unlockAddress.value &&
+          txLog.topics[0].toHexString() == GNP_CHANGED_TOPIC0
+        ) {
+          // log all topics
```
clemsos, line 103:
```suggestion
```

subgraph/tests/mockTxReceipt.ts
```diff
@@ -0,0 +1,187 @@
...
+// adds a GNPChanged event to the tx receipt
+export function newGNPChangedTransactionReceipt(
+  keyValue: BigInt,
+  totalValue: BigInt
+): ethereum.TransactionReceipt {
+  const eventSignature = Bytes.fromHexString(GNP_CHANGED_TOPIC0)
+  const grossNetworkProduct = BigInt.fromU32(0)
+
+  // as the tx is sent from another contract (NOT the lock)
+  // only the signature is passed as topic
```
clemsos, line 159:
```suggestion
  // only the event signature is passed as topics[0]
```

PR #13828, opened 2024-05-13
feature(unlock-app): Sanitise duplicated email fields in checkout

### LINKED ISSUE #13320: Duplicated email fields
Do not show "twice" the email field on checkout.
If we have both a meatdataInput + `emailRequired` we should only show a single email field!
image

0xTxbi:
# Description
This PR addresses an issue in the checkout form where multiple email input fields were being rendered when `emailRequired` was set to true, despite an email field already being specified in the `metadataInputs`.

## Screen Grabs
- Current Checkout Recipient Step (Before)
unsanitised-email-fields-checkout

**Description**: Shows the checkout form rendering two email fields — one from the default `emailRequired` and another from `metadataInputs`.

- Updated Checkout Recipient Step (After)
sanitised-email-fields-checkout

**Description**: Demonstrates the checkout form after applying the fix, displaying only the default required email field, as duplicated fields are successfully filtered out.

# Issues
Fixes #13320
Refs #13320

# Checklist:


julien51 (commented):
So far so good! Please check what happens when the user is not connected wuth a wallet but rather with an Unlock account?

0xTxbi:
> So far so good! Please check what happens when the user is not connected wuth a wallet but rather with an Unlock account?

sure thing! i already factored this scenario. here's how the recipient(s) checkout flow looks like when a user is connected with their unlock account

sanitised-email-fields-unlock-account

julien51:
Very good!

julien51 (approved):

PR #14015, opened 2024-06-06
feat(unlock-app): Show more option on lock creation during network selection

SVell:
# Description

Added a show more button for Select component

# Issues

Fixes #
Refs #

# Checklist:

## Release Note Draft Snippet

image

### DIFF
--- a/packages/ui/lib/components/Select/Select.tsx
+++ b/packages/ui/lib/components/Select/Select.tsx
@@ -34,7 +34,7 @@ export interface SelectProps<T> {
   customOption?: boolean // show custom option that will show a custom input
   disabled?: boolean
   loading?: boolean
-  moreOptions?: MoreOptions
+  moreOptions?: Option[]
 }
 
 const SIZE_STYLES: SizeStyleProp = {
@@ -107,7 +107,7 @@ export const Select = <T extends unknown>({
   customOption = false,
   disabled: fieldDisabled = false,
   loading = false,
-  moreOptions = { showMoreAfter: 0 },
+  moreOptions = [],
 }: SelectProps<T>) => {
   const [selected, setSelected] = useState<Option | null>(null)
   const [custom, setCustom] = useState<boolean>(false) // value that enables/disable custom field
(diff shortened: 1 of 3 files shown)


packages/ui/lib/components/Select/Select.tsx
```diff
@@ -102,14 +104,24 @@ export const Select = <T extends unknown>({
   customOption = false,
   disabled: fieldDisabled = false,
   loading = false,
+  showMoreButton = false,
+  showMoreAfter = 6,
```
julien51, line 108: Do we really need this? We should only show this option as the last one..

julien51: Actually let's use a new prop on the select, Rather called "moreOptions" and merge with the options when applicable.

unlock-app/src/components/interface/locks/Create/elements/CreateLockForm.tsx
```diff
@@ -77,7 +77,7 @@ export const CreateLockForm = ({
   const { networks } = useConfig()
   const web3Service = useWeb3Service()
   const { account } = useAuth()
-  const networkOptions = useAvailableNetworks()
+  const networkOptions = useAvailableNetworks(true)
```
julien51, line 80: Can you explain this change please?

julien51 (commented):
Looking good! Can you please explain show how the options look like when expanded?

packages/ui/lib/components/Select/Select.tsx
```diff
@@ -264,6 +276,17 @@ export const Select = <T extends unknown>({
...
                   </Listbox.Option>
                 )}
+                {!isOtherSelected && options.length > showMoreAfter && (
+                  <button
+                    className="w-full"
+                    type="button"
+                    onClick={() => setIsOtherSelected(true)}
+                  >
+                    <SelectOption
+                      option={{ label: 'Show more', value: CUSTOM_VALUE }}
```
julien51, line 286: Let's see if we can use a different style for this option, like _italics_ and add `...`. So smething like _Show more..._.

Thanks!

julien51 (changes requested):

julien51:
Please post updated screenshots!

julien51 (changes requested):
Please change `moreOptions` to be the array of additional options!

julien51 (approved):

PR #14031, opened 2024-06-11
feat(unlock-app): Logic for Next Auth 

SVell:
# Description

https://www.loom.com/share/1ef0cd3fdcfb4aebbaa5da5559822e73?sid=e6610289-8899-47d2-9e0c-23fe94e755a3

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/config/app.ts
+++ b/unlock-app/src/config/app.ts
@@ -12,8 +12,8 @@ const staging = {
   unlockApp: 'https://staging-app.unlock-protocol.com',
   googleClientId:
     '801850414021-kk4m4gqm7imtggonhl57tfsj512on6s6.apps.googleusercontent.com',
-  coinbaseProjectId:
-    'organizations/84d8b485-5433-4902-9765-3e5f11bf16d6/apiKeys/740902e7-eaee-4d60-a316-20690272e7e8',
+  // https://portal.cdp.coinbase.com/projects
+  coinbaseProjectId: '4e90fd0f-bb62-4e1c-91a4-f08c76d1b09e',
 }
 
 const dev = {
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -10651,6 +10651,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@panva/hkdf@npm:^1.0.2":
```
julien51, line 10654: This change is probably not right? Can you re-generate the yarn.lock file?

unlock-app/src/components/interface/checkout/main/Disconnect.tsx
```diff
@@ -32,6 +33,7 @@ const Disconnect = ({ service }: DisconnectProps) => {
     setIsDisconnecting(true)
     await signOut()
     await deAuthenticate()
+    await nextSignOut()
```
julien51, line 36: Will that fail if the user did not sign in with next?

unlock-app/src/components/interface/checkout/main/Select.tsx
```diff
@@ -393,6 +399,37 @@ export function Select({ checkoutService }: Props) {
     })
   }
 
+  if (isLoadingWaas) {
```
julien51, line 402: This should not be in the `Select` view IMO but one level "above".
The logic in this file here should be as much as possible only about selecting the lock.

unlock-app/src/utils/getUserAccountType.ts
```diff
@@ -0,0 +1,14 @@
+import { storage } from '~/config/storage'
+import { UserAccountType } from './userAccountType'
+
+export const getUserAccountType = async (
```
julien51, line 4: This does not add a lot of value and adds complexity. Please use `storage.getUserAccountType(emailAddress)` directly where you need this.

unlock-app/src/components/interface/connect/Wallet.tsx
```diff
@@ -91,22 +94,37 @@ export const ConnectViaEmail = ({
...
+  const [userAccountType, setUserAccountType] = useState<UserAccountType[]>([])
+
+  useEffect(() => {
+    const userAccountType = async () => {
+      if (!userEmail) {
+        setIsLoadingUserExists(false)
+        return
+      }
+
+      const userType = await getUserAccountType(userEmail)
```
julien51, line 115: please use `useQuery()` for this. Do not use `useEffect`.

julien51 (changes requested):
ok this is good. A few changes (+ fixes in CI!)

PR #14043, opened 2024-06-13
feat(unlock-app): Decent for swaps

julien51:
# Description

Until now we used Decent for the "crosschain purchase" scenario where the transaction would be sent from different network, but only paid on this network's native currency.
Since Uniswap broke their libraries, we are now also using Decent for "swap" scenario where we now have the following supported flows:

- Same chain, use ERC20 for a native token lock
- Same chain, use ERC20 for lock with a different ERC20
- Same chain, use native token for a lock using an ERC20
- Different chain, use native token for a native lock
- Different chain, use ERC20 for a native token lock
- Different chain, use ERC20 for lock with a different ERC20
- Different chain, use native token for a lock using an ERC20

# Issues

Fixes #
Refs #

# Checklist:


unlock-app/src/components/interface/checkout/main/Payment.tsx
```diff
@@ -165,27 +164,15 @@ export function Payment({ checkoutService }: Props) {
 
   const canAfford = balance?.isGasPayable && balance?.isPayable
 
-  const { data: uniswapRoutes, isInitialLoading: isUniswapRoutesLoading } =
-    useUniswapRoutes({
```
julien51, line 169: I will clean up on a different PR.

unlock-app/src/hooks/useCrossChainRoutes.ts
```diff
@@ -1,12 +1,23 @@
...
+import { networks } from '@unlock-protocol/networks'
+import { BoxEvmChains } from '@decent.xyz/box-common'
+import { Token } from '@unlock-protocol/types'
+
+interface CrossChainRouteWithBalance extends CrossChainRoute {
+  resolvedAt: number
+  userTokenBalance: string
+}
+
+// TheBox returns BigInts as strings with a trailing 'n'
```
julien51, line 18: 🤦

unlock-app/src/components/interface/checkout/main/Payment.tsx
```diff
@@ -425,22 +365,28 @@ export function Payment({ checkoutService }: Props) {
                   >
                     <div className="flex justify-between w-full">
                       <h3 className="font-bold">
-                        Pay with {route.symbol} on {route.networkName}
+                        Pay with {route.tokenPayment.symbol} on{' '}
```
clemsos, line 368: that throws. an error when the lock is priced in native token. Added a fix

unlock-app/src/hooks/useCrossChainRoutes.ts
clemsos: looking into it, yes its a real issue when parsing serialized values...

unlock-app/src/components/interface/checkout/main/Payment.tsx
julien51: ha! Good catch I will look at this gain.

PR #14055, opened 2024-06-16
feat(unlock-app): Allow processing of large csvs 

### LINKED ISSUE #12993: Automate processing of large CSV files
Right now our front-end is able to airdrop many NFTs at once, but due to block limit we still have a cap at 50. 
the user can then "re-upload" the same CSV, but we could also make it easier by automatically sending transactions one by one, without asking the user to re-upload the CSV.

iMac7:
# Description
Currently, only 50 entries are processed at a time and the user has to enter the csv again to process the next 50.
This PR loads the entire csv and queues the transactions into groups of 50 (maximum). Each group of transactions must be signed.

# Issues

Fixes #12993

# Checklist:


julien51:
Thanks @iMac7 ! I believe it would be nice to add some kind of loading indicator to the UI, maybe replacing the the "drag and drop" area, showing that X transactions are being mined?

What do you think?

julien51:
> Good idea, I'll take a look at that. The text that said the user should reupload a csv should be replaced as well, what do I put there?

We should probably remove that text in fact and replace with something like this: 

`Due to block size limit, you can only airdrop at most 50 NFT per transaction, so if your list is longer than that, you will be prompted to approve multiple transactions. You can also easily restart the process at any point, as addresses who already have one NFT will be discard.`

julien51:
@iMac7 Do you mind showing a quick video again with the updated loading state?

julien51:
Amazing! LGTM! 
Please add your wallet address to the issue so we can pay a bounty for this one!

Issue #13917, opened 2024-05-31
Health check reporting


julien51:
We now have a pretty good health-check job that runs regularly and checks a lot of things.
Right now it reports to Sentry (as it should) but does not report in the "correct" format so that means it's hard for us to identify issues. Here are improvements that need to be made:
* [ ] We should report "individual" issues to sentry.
* [ ] Distinguish between "warnings" and "error" (warning are non critical, such as "misconfigired token" or "pending proposal on DAO", while errors are critical, like "bad version"... etc

julien51:
Once we have a better script we should also make sure this runs on PR that add new networks: https://github.com/unlock-protocol/unlock/issues/13463

julien51:
https://github.com/unlock-protocol/unlock/issues/13533

PR #14072, opened 2024-06-19
chore: proper logging for protocol health monitoring with Sentry

clemsos:
# Description

This adds a more refined and useful way to log various protocol checks to Sentry (and then fwd them to discord etc)

# Issues

Fixes #
Refs #

# Checklist:


governance/helpers/logger.js
clemsos, line 1: There are several `logger.js` files with duplicate code but as we discussed , it may be easier to have these for now than to introduce another package just to hold that bit of code.

clemsos:
relates to  #13917

governance/helpers/logger.js
```diff
@@ -0,0 +1,25 @@
...
+  error: '❌ ',
+  warning: '⚠️ ',
+}
+
+const log = (msg, level = 'info') => {
+  //log to stdout
+  console.log(`${symbols[level] || ''}[${level}]: ${msg}`)
+
+  // send to sentry
+  if (process.env.SENTRY_DSN) {
```
clemsos, line 20: logs will be sent to sentry only if `SENTRY_DSN` is exported to the shell. Should document this maybe but not sure where ?

julien51: I think that's ok.

julien51 (approved):

PR #14110, opened 2024-06-26
chore: cleanup uniswap deps

julien51:
# Description

In order to remove a few deps, I am cleaning up some code.

# Checklist:


packages/hardhat-helpers/src/uniswap.js
```diff
@@ -1,104 +1,18 @@
 const { ethers } = require('ethers')
 
 const { AllowanceTransfer } = require('@uniswap/permit2-sdk')
-const {
-  Token,
-  CurrencyAmount,
-  TradeType,
-  Percent,
-} = require('@uniswap/sdk-core')
+const { Token } = require('@uniswap/sdk-core')
```
julien51, line 4: I would love to get rid of this completely. It looks like this (and `nativeOnChain`) are used for `getUniswapTokens` but I think we may not "need" this format? We could easily use the tokens from our own network configs?

What do you think @clemsos ?

clemsos: yes we dont need this anymore. The `getUniswapTokens` is still used in some tests but we could use our package instead

julien51: ok I will try to replace that then!

clemsos (approved):
deps cleanup !

PR #14270, opened 2024-07-16
feat(unlock-app): Set gas refund limit

### LINKED ISSUE #13650: Not allow set gas refund higher than necessary
**Describe the bug**

At the moment when we want others users to renew the membership we need to set gas refund, however if we set a value higher than the lock, we are not able to purchase the lock, giving us an allowance error.

**To Reproduce**

1. Use lock with token to renew memberships.
2. Set gas refund higher than the lock price.
3. It not allow now to buy the lock

**Expected behavior**

Should not allow user to set gas refund higher than the lock price. Additionally, an info field text could be used to explain about it.

iMac7:
# Description

# Issues

Fixes #13650 
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/locks/Settings/elements/SettingPayments.tsx
+++ b/unlock-app/src/components/interface/locks/Settings/elements/SettingPayments.tsx
@@ -51,7 +51,7 @@ export const SettingPayments = ({
 
       <SettingCard
         label="Subscription"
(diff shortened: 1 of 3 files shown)


julien51 (approved):
Have you toroughly tested this with both a native currency lock and an eRC20 lock?

julien51:
Awesome! thanks

PR #14319, opened 2024-07-25
feat(locksmith, unlock-app): Bulk export receipts

### LINKED ISSUE #13001: Bulk export receipts
Currently, receipts have to be downloaded one by one. This is a bit cumbersome for lock managers.
Let's add the ability to download all receipts as an async job.
A background job should be triggered when the user clicks the button to download all receipts. 
the background job will generate every single receipt as a PDF and combine them in a large zip file. 

Once completed, the user should be able to download the file.

Since the export can take a while, we should probably create a specific model in the database that will keep track of all the "exports". The table will keep track of the status of the job (pending, completed). The Settings UI should list any "recent" export (last 30 days) and allow users to download the ones that are completed.

iMac7:
# Description

# Issues

Fixes #13001
Refs #

# Checklist:

## Release Note Draft Snippet

https://www.loom.com/share/e5229ca3686a4eccae2f8b3360c09b67?sid=c8001d67-9056-45e1-b4d2-32ec7a2fb31f

### DIFF
--- a/locksmith/src/controllers/v2/receipts.ts
+++ b/locksmith/src/controllers/v2/receipts.ts
@@ -9,6 +9,7 @@ import { Payload } from '../../models/payload'
 import { addJob } from '../../worker/worker'
 import config from '../../config/config'
 import { downloadFileFromS3 } from '../../utils/s3'
+import logger from '../../logger'
 
 export const allReceipts: RequestHandler = async (request, response) => {
   const network = Number(request.params.network)
@@ -94,23 +95,29 @@ export const downloadReceipts: RequestHandler = async (request, response) => {
     try {
       const fileStream = await downloadFileFromS3(bucketName, key)
 
-      response.setHeader('Content-Type', 'application/zip')
-      response.setHeader(
-        'Content-Disposition',
-        `attachment; filename=receipts.zip`
-      )
+      if (fileStream) {
+        const readableStream = fileStream as NodeJS.ReadableStream
 
-      fileStream.pipe(response)
+        response.setHeader('Content-Type', 'application/zip')
+        response.setHeader(
+          'Content-Disposition',
+          `attachment; filename=receipts.zip`
+        )
 
-      fileStream.on('error', () => {
-        if (!response.headersSent) {
-          response.status(400).send('Failed to download file')
-        }
-      })
+        readableStream.pipe(response)
 
-      response.on('finish', () => {
-        console.log('File downloaded successfully')
-      })
+        readableStream.on('error', () => {
+          if (!response.headersSent) {
+            response.status(400).send('Failed to download file')
+          }
+        })
+
+        response.on('finish', () => {
+          logger.info('File downloaded successfully')
+        })
+      } else {
+        response.status(400).send('Failed to download file')
+      }
     } catch (error) {
       if (!response.headersSent) {
         response.status(400).send('Failed to download file')


julien51:
HJey @iMac7 I am really sorry but this does not meet the requirement stated on the issue:

> Let's add the ability to download all receipts as an async job.
> A background job should be triggered when the user clicks the button to download all receipts.
the background job will generate every single receipt as a PDF and combine them in a large zip file.

> Once completed, the user should be able to download the file.

> Since the export can take a while, we should probably create a specific model in the database that will keep track of all the "exports". The table will keep track of the status of the job (pending, completed). The Settings UI should list any "recent" export (last 30 days) and allow users to download the ones that are completed.

Your approach will break pretty quickly once there is more than a few receipts.

julien51:
I think that is the right approach!

The s3 bucket will not be public, so the zip can only be retrieved based on authentication (this will be an extra API route!)

julien51:
> @julien51 I see quite a few failing tests, some because of the new packages 🤔 is the yarn.lock ok or should I take it out? will be checking those and a few linting issues

Since you updated the `yarn.lock` file, please commit the updated one and that would fix that issue.

julien51:
In terms of UI/UX you have something that's a great start but not in the rigth place. In fact we already have a button to download receipts as CSV in the lock's settings > Payment > Receipt. That is where you should show your new button (replace the existing button please).


Screenshot 2024-09-06 at 10 37 52 AM

locksmith/src/utils/receipts.ts
```diff
@@ -32,3 +37,127 @@ export const getAllReceipts = async ({
...
+      doc.moveDown()
+
+      doc.font('Helvetica-Bold').fillColor('#603DEB').text('Amount')
+      doc.font('Helvetica').fillColor('black')
+      doc.moveDown(0.5)
+      doc.text('TOTAL', { align: 'left', continued: true })
+      doc.font('Helvetica-Bold').text(amount, { align: 'right' })
+      doc.font('Helvetica')
+      doc.moveDown()
+
```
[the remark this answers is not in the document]
julien51: wow! Indeed.
Actually, I wonder if we could not just generate a PDF _from_ the HTML rendering rather than create a custom PDF. That would be simpler and MUCH easier to maintain.

locksmith/package.json
```diff
@@ -106,6 +109,7 @@
     "@types/multer-s3": "3.0.3",
     "@types/node": "20.14.13",
     "@types/node-cron": "3.0.11",
+    "@types/pdfmake": "^0",
```
julien51, line 112: That's not right :) Please pin the latest version here!

packages/unlock-js/openapi.yml
```diff
@@ -2998,6 +2998,147 @@ paths:
                     type: string
         401:
           $ref: '#/components/responses/401.NotAuthenticated'
+  
+  /v2/receipts/all/{network}/{lockAddress}:
+    get:
+      operationId: getReceiptsStatus
+      security:
+        - User: []
+      description: Retrieve the status of receipt generation for a lock.
```
julien51, line 3007:
```suggestion
      description: Retrieve the status of receipts generation for a lock.
```

packages/unlock-js/openapi.yml
```diff
@@ -2998,6 +2998,147 @@ paths:
...
+      parameters:
+        - $ref: '#/components/parameters/Network'
+        - $ref: '#/components/parameters/LockAddress'
+      responses:
+        200:
+          description: Successfully retrieved receipt generation status.
```
julien51, line 3013:
```suggestion
          description: Successfully retrieved receipts generation status.
```

PR #14330, opened 2024-07-30
feat(unlock-app): Price on select screen should show user price

### LINKED ISSUE #14224: Price on select screen should show user price if user is connected
Our `select` step should show the actual price for the connected user (assuming the user is not changing the recipient later)

SVell:
# Description

Price on select screen should show user price

# Issues

Fixes #14224
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/checkout/main/Select.tsx
+++ b/unlock-app/src/components/interface/checkout/main/Select.tsx
@@ -230,17 +230,23 @@ export function Select({ checkoutService }: Props) {
             let price
 
             if (account) {
-              price = await web3Service.purchasePriceFor({
-                lockAddress: lock,
-                userAddress: account,
-                referrer: account,
-                network: networkId,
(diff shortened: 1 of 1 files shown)


unlock-app/src/components/interface/checkout/main/Select.tsx
```diff
@@ -225,16 +226,35 @@ export function Select({ checkoutService }: Props) {
               props.network || paywallConfig.network || 1
 
             const lockData = await web3Service.getLock(lock, networkId)
+
+            let price
+
+            if (account) {
+              price = await web3Service.purchasePriceFor({
```
julien51, line 233: This _may_ fail in some cases (for example when the lock uses a hook like the password hook). In that case, please make show the "default" price which is the `lockData.keyPrice`

julien51 (approved):

PR #14412, opened 2024-08-04
feat(unlock-app): Checkout OG tags and frame

### LINKED ISSUE #13593: Add OG + Frame metatags to Checkout URL
Our checkout builder let's anyone build checkout URL with custom behavior.
We are looking to add Open Graph preview for these URLs, as well as, ideally, frame meta tags as well. 

To support frame, it means that we should "mimick" all the states on the checkout with frame "states".

In a simpler/basic first version, we could only support the "most" basic checkout config: 
* single lock (so no need to chose)
* `skipRecipient` as true (so no need to chose/customize)
* no metadata collection (so no need to enter any data or keep state)
* no hook support (no password, no captcha, no promo code... which would require some kind of user input)
* no support for crosschain payments or swap payment of any kind (because we can't support user inputs)

It should however support ERC20 or native lock, whicg means it should ask 2 transactions from the user: first one to "approve" the purchase (and possible renewals when applicable), and the second one to actually perform the mint.

iMac7:
# Description

# Issues

Fixes #13593 
Refs #

# Checklist:


julien51:
> tested with ograph previewer extension on another page because I couldn't navigate to localhost../checkout , it's always in a loading state, some requests fail.

Can you provide a bit more details here please?

unlock-app/app/api/svg-to-png/route.tsx
```diff
@@ -0,0 +1,41 @@
+import sharp from 'sharp'
+import { NextRequest, NextResponse } from 'next/server'
+
+export async function GET(req: NextRequest) {
```
julien51, line 4: HA! I think introducing a route where we let folks arbitrarily convert any URL of an SVG to a PNG is very risky!

unlock-app/app/frames/checkout/components/Success.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { DefaultImage } from './DefaultImage'
+
+export function Success({ lock }: any) {
```
julien51, line 4: Could you please clarify what this compnent is? Success? of what?

unlock-app/app/frames/checkout/components/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+import networks from '@unlock-protocol/networks'
+import { config as appConfig } from '~/config/app'
+
+export async function getLockDataFromCheckout(id: string) {
+  const { config } = await fetch(
+    `${appConfig.locksmithHost}/v2/checkout/${id}`
+  ).then((res) => res.json())
+
+  const locks = config.locks
+  const lockAddress = Object.keys(locks)[0]
```
julien51, line 12: What happens if there are multkple locks here?

julien51 (approved):
ok this looks good! Please see if you can solve the issue of allowing anyone to render a SVG to png here and then we can merge!

julien51:
> how do you propose I do that, have the route in locksmith instead? I can't think of much else particularly due to the limited nature of og and frames, and both depend on the conversion.

No, what I am worried is that our endpoint can be used by people to turn any SVG into a PNG. Moving it to locksmith would not help

> As for security, maybe I can sanitize the url? 🤔 I'm thinking auth would not work with og/frames because there is no way to pass a token there, and ideally anyone should be able to generate a png from the default image to interact with the public og/frame.

No sanitising the URL does not help either. 

>Another option would be to have the default generated svg image converted to png just before it's stored. Considering that there's already default lock images in svg format in prod, I'm not sure how to proceed.

Is there a way to use a data URI?

unlock-app/app/frames/checkout/components/DefaultImage.tsx
```diff
@@ -0,0 +1,36 @@
...
+    </div>
+  )
+
+  //svg images are not rendered, convert to png first
+  const isSvg = /\/icon\/?$/.test(image)
+  const imgurl = !isSvg
+    ? image
+    : `${config.unlockApp}/api/svg-to-png?url=${encodeURIComponent(image)}`
+  const leftImage = (
+    <img src={imgurl} tw="flex-1 min-h-full border-4 border-white rounded-lg" />
```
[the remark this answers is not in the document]
julien51: > hm I did not know that it had security risks, maybe I can download then convert to base64 for that? 🤔

don't download, just render it inline here as base64 indeed.
Check this maybe? https://stackoverflow.com/questions/8499633/how-to-display-base64-images-in-html

unlock-app/app/frames/checkout/route.tsx
```diff
@@ -0,0 +1,43 @@
...
+  if (success === 'true') {
+    return {
+      image: <TransactionSuccess lock={lock} />,
+      buttons: [
+        <Button action="link" target="https://app.unlock-protocol.com/keychain">
+          View on keychain
+        </Button>,
+        <Button action="tx" target="/txdata" post_url="?success=true">
+          Buy again
+        </Button>,
```
julien51, line 33: Let's remove this since it is usually not possible to buy again!

julien51: If the checkotu as a redirect value, let's just add a button to go there (the checkout config may also have a CTA)

julien51 (commented):
ok, one final change and then we'll test this!

julien51 (approved):
LGTM!

PR #14434, opened 2024-08-18
feat(unlock-app): Transfer key on keychain

### LINKED ISSUE #8653: Transfer key on keychain
On the keychain, we now have the ability to cancel a key which is great!
Let's add the ability to transfer a key too!
The button would offer a modal which should show the following:
* a field where the user can enter the recipient of the membership NFT
* A toggle to set the amount of time to be shared (vs the full NFT)
* if there is a penalty for transfers, show a message indicating how much time the recipient will receive (in days, hours and minutes)
* A checkbox that says "lend" key (with a little explainer indicating that if this is checked the recipient will not be able to transfer the key themselves)
* a submit button

We will use the `transferFrom` or `lendKey` function if the transfer is full and the `shareKey` function if transfer is partial.

Importantly, since not all versions of the Public Lock contract support all these features, we need to make sure this only works for supported versions.

Before changing the front-end, we will need to update the UnlockJS library in order to handle versioning at that level and abstract the complexity there.

iMac7:
# Description

# Issues

Fixes #8653 
Refs #

# Checklist:

I used submit buttons instead of  toggles to  make it more intuitive that each button is a different transaction.
How precise should the time input for selecting share time be (days, hrs, mins, sec?) ? Is there an already existing implementation of a time input component I can reuse? (will default to a custom implementation).
Open to any visual changes, I expect a lot of those :laughing: 

`transferFrom` and `shareKey` work for this PR, but `lendKey` does not.
`lendKey` is apparently not present in the abi version (so it fails the version check `if(version.lendKey)` ) even though I see it on the contract :thinking: 
I'm thinking of calling the contract directly for the `shareKey` function as I explore alternatives (Any suggestions?)

<br><br>
(on issue)
> if there is a penalty for transfers, show a message indicating how much time the recipient will receive (in days, hours and minutes) 

How do I set a transfer penalty on a lock to test with it? I'm only seeing a penalty for `cancelAndRefund` 
<br><br>

https://www.loom.com/share/df6a6ad9fc30420cb987e9cba9f7b759?sid=260c0816-e10e-43da-977e-dd722b0b1c0d

In the screen recording I only transferred time to the other address, so transferring the key back fails.
However, I was not able to tell that it's not the real key in the UI. Is there a reliable method to distinguish between a "normal" key and a "time shared" key?

### DIFF
--- a/unlock-app/src/components/interface/keychain/Key.tsx
+++ b/unlock-app/src/components/interface/keychain/Key.tsx
@@ -111,7 +111,11 @@ function Key({ ownedKey, owner, network }: Props) {
     lockAddress: lock.address,
     network: network,
   })
-  if (!transferFeeLoading && !transferfeeError) {
+  if (
+    !transferFeeLoading &&
+    !transferfeeError &&
+    typeof transferFee === 'number'
+  ) {
     isTransferable = isKeyTransferable(transferFee)
   }
 


julien51:
> I used submit buttons instead of toggles to make it more intuitive that each button is a different transaction.

Please make sure to add screenshots so we can see what changes are done

julien51:
> How precise should the time input for selecting share time be (days, hrs, mins, sec?) ? Is there an already existing implementation of a time input component I can reuse? (will default to a custom implementation).

I think it should be days. That's what we default on everywhere in our UI. This means that if the membership has less than 1 day left, we should disable the transfers.

julien51:
> lendKey is apparently not present in the abi version (so it fails the version check if(version.lendKey) ) even though I see it on the contract 🤔

Please share the contract address so we can look this up.

julien51:
> I'm thinking of calling the contract directly for the shareKey function as I explore alternatives (Any suggestions?)

This should be added in `unlock-js`, probably in another PR for clarity/simplicity.

julien51:
> However, I was not able to tell that it's not the real key in the UI. Is there a reliable method to distinguish between a "normal" key and a "time shared" key?

These are both "real" keys! Let me explain. If I have Key #3 with 7 days left and I send you 1 day. If you already have a key, it will extend yours, and if you don't have one, the contract will mint a new one for you!

packages/unlock-js/src/walletService.ts
```diff
@@ -1171,6 +1173,40 @@ export default class WalletService extends UnlockService {
     if (!params.tokenId) throw new Error('Missing tokenId')
 
     const version = await this.lockContractAbiVersion(params.lockAddress)
+
+    if (params.time) {
+      if (!version.shareKey) {
+        throw new Error('Lock version not supported')
+      }
```
julien51, line 1180: ok so the way to implement this is to *not* do that in `walletService.ts` but to add a `shareKey.ts` and `lendKey.ts` file to implement this based on the versions (check how other methods are working!)

packages/unlock-js/src/walletService.ts
```diff
@@ -1171,6 +1173,40 @@ export default class WalletService extends UnlockService {
...
+      const shareKeyParams = {
+        lockAddress: params.lockAddress,
+        tokenId: params.tokenId,
+        recipient: params.to,
+        duration: params.time,
+      }
+      return version.shareKey.bind(this)(
```
julien51, line 1187: I think you are missing this file?

packages/unlock-js/src/walletService.ts
```diff
@@ -1171,6 +1173,40 @@ export default class WalletService extends UnlockService {
...
+    if (params.lend) {
+      if (!version.lendKey) {
+        throw new Error('Lock version not supported')
+      }
+      const lendKeyParams = {
+        from: params.keyOwner,
+        to: params.to,
+        tokenId: params.tokenId,
+      }
+      return version.lendKey.bind(this)(
```
julien51, line 1203: same?

unlock-app/src/components/interface/keychain/TransferModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+  to: string
+  tokenId: string
+  lockAddress: string
+  time?: number
+  lend?: boolean
+}
+
+type TransferOption = 'transfer' | 'share' | 'lend' | undefined
+
+const MAX_TRANSFER_FEE = 10000 //100% of transfer fees = disabled transfers
```
julien51, line 34: I think the logic to check if a key is trsmaferable exists somewhere else. It would be nice to re-use that and (if needed) isolate in some kind of util so we don't re-invent the wheel!

unlock-app/src/components/interface/keychain/TransferModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+    formState: { errors },
+  } = useForm<TransferFormProps>({
+    mode: 'onChange',
+    defaultValues: {
+      recipientAddress: undefined,
+      time: undefined,
+    },
+  })
+
+  async function fetchTransferFees() {
```
julien51, line 73: Please use `useQuery` like we do everywhere else in the app.

unlock-app/src/components/interface/keychain/TransferModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+        <div className="h-[200px] w-[400px] flex justify-center items-center">
+          <p>This membership is not transferable.</p>
+        </div>
+      ) : (
+        <form
+          className="p-6 flex flex-col gap-8 min-w-[300px] max-w-[700px]"
+          onSubmit={handleSubmit(onHandleSubmit)}
+        >
+          <div className="relative gap-3">
+            <Input
```
julien51, line 144: Please use the `AddressInput` component so that we can use ENS names as well.

unlock-app/src/components/interface/keychain/TransferModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+              type="submit"
+              onClick={() => setOption('transfer')}
+              className="w-full"
+              loading={isLoading && option === 'transfer'}
+            >
+              Transfer
+            </Button>
+          </div>
+
+          <div className="flex flex-col gap-2">
```
julien51, line 176: Let's only show that if the membership has an expiration date (memberships which are set to not expire cannot be "shared")

julien51 (changes requested):
That's a great start! there is a bunch of things to check on top of the existing, especially edge cases!

julien51:
Thanks @iMac7 I am runnig this locally to test a few things and it's looking good. A few comments:

* the loading state is not great : https://www.loom.com/share/998a85c1684e48e0a27d0869603ac999 Any way to make the modal of the right size when starting?
* Please make sure you handle errors with a better error message. Here is what it shows when I am cancelling the transaction in my wallet for example
Screenshot 2024-08-30 at 11 39 28 AM
* The option to share membership time should NOT be displayed at all (not just disabled) when the membership does not expire.
*  The error message color for "Address is required." does not match:  
Screenshot 2024-08-30 at 11 41 45 AM
* Once a key has been transferred (regular transfer) the modal should close.

unlock-app/src/components/interface/keychain/TransferModal.tsx
julien51: > The closest I see to this is isTransferPossible in packages/unlock-js/src/KeyManager.ts, it looks more complex for this specific functionality 🤔

Indeed that is a different thing and indeed I could not find a place where we already have `MAX_TRANSFER_FEE` but please do isliate that logic by creating a `isKeyTransferable` function in a util.

unlock-app/src/components/interface/keychain/TransferModal.tsx
julien51: Please do it :D

unlock-app/src/components/interface/keychain/TransferModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+    let result = await web3service.transferFeeBasisPoints(lock.address, network)
+    result = Number(result)
+    return result
+  }
+
+  const {
+    isLoading: pageIsLoading,
+    error,
+    data: transferFees,
+  } = useQuery({
```
julien51, line 81: Please do that inside of a dedicated hook like `useLockTransferFees` (like the rest of the app does!)

unlock-app/src/components/interface/keychain/TransferModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+    queryKey: ['transferFees', lock.address, network],
+    queryFn: fetchTransferFees,
+    enabled: isOpen,
+  })
+
+  if (error) {
+    console.error(error)
+  }
```
julien51, line 89: We at least need to show a message (toast) to users but please explain that. What kind of error is that?

unlock-app/src/components/interface/keychain/TransferModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+      <h3 className="text-xl font-bold text-center">Transfer Key</h3>
+
+      {pageIsLoading ? (
+        <Placeholder.Root>
+          <Placeholder.Image className="h-[200px] w-[400px]" />
+        </Placeholder.Root>
+      ) : transferFees === MAX_TRANSFER_FEE ? (
+        <div className="h-[200px] w-[400px] flex justify-center items-center">
+          <p>This membership is not transferable.</p>
+        </div>
```
julien51, line 151: If this is not transferable let's just not show the modal at all! (and disable the option in the menu)

unlock-app/src/components/interface/keychain/TransferModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+                    placeholder="Address or ENS"
+                    onResolveName={onResolveName}
+                    {...field}
+                  />
+                )
+              }}
+            />
+            {errors?.recipientAddress && (
+              <span className="absolute text-xs text-red-700">
+                Address is required.
```
julien51, line 175:
```suggestion
                A recipient address is required.
```

unlock-app/src/components/interface/keychain/TransferModal.tsx
julien51: hum, thanks for the heads up. I will check this but that is likely based on the lock's duration, not the individual key expiration

julien51 (changes requested):
More changes required! Thanks!

unlock-app/src/hooks/useTransferFee.ts
```diff
@@ -29,3 +30,27 @@ export const useTransferFee = ({ lockAddress, network }: TransferFeeProps) => {
...
   }
 }
+
+export const useFetchTransferFee = ({
+  lockAddress,
+  network,
+}: TransferFeeProps) => {
+  const web3service = useWeb3Service()
+
+  async function fetchTransferFees() {
```
julien51, line 40: I see there is a `getTransferFeeBasisPoints` above in that file. Can you DRY things up a bit please?

julien51 (approved):
one final cleanup and then I think we're looking good!

julien51 (approved):

PR #14595, opened 2024-09-11
Eslint cleanup

### LINKED ISSUE #14564: Eslint cleanup
Now that we have upgraded to eslint, I noticed we are using different rules in many of our packages, as well as only applying eslint to _some_ files and not all the files in some packages.

For example, we use `yarn run eslint src/` instead of `yarn run eslint .` which would cover _all_ the files. We should make that in all the packages/applications.

Many packages do also include their own rules and we should avoid that.

This 200 USDC bounty is open for anyone who "normalizes" our use of eslint to the max!

Myestery:
# Description

Ensure consistent eslint rules across all packages, also ensure commands are same and reused in all package.jsons

# Issues

Fixes #14564
Refs #14564

# Checklist:


locksmith/scripts/connectedPurchase.ts
```diff
@@ -21,6 +22,7 @@ function generatePurchasePayload(message: any, messageKey: string) {
   }
 }
 
+/* eslint evm-address-to-checksummed/evm-address-to-checksummed: "warn" */
```
julien51, line 25: Let's fix this rather than add an exception please :)

locksmith/scripts/metadataUpload/user_metadata_upload.ts
```diff
@@ -57,6 +58,7 @@ async function main(
   inputFile: string,
   host: string
 ) {
+  /* eslint evm-address-to-checksummed/evm-address-to-checksummed: "warn" */
   const userAddress = '0xe29ec42f0b620b1c9a716f79a02e9dc5a5f5f98a'
```
julien51, line 62: same!

packages/networks/eslint.config.js
```diff
@@ -1,16 +1,3 @@
-/* eslint-disable @typescript-eslint/no-require-imports */
-const sortKeysFix = require('eslint-plugin-sort-keys-fix')
```
julien51, line 2: This is important though so I would want to make sure we keep this rule

packages/core/eslint.config.js
```diff
@@ -1,3 +1,2 @@
-/* eslint-disable @typescript-eslint/no-require-imports */
```
julien51, line 1: I am confused that removing this rule does not seem to create any issue. Did you put it somewhere else?

packages/networks/eslint.config.js
```diff
@@ -1,8 +1,5 @@
-/* eslint-disable @typescript-eslint/no-require-imports */
```
julien51, line 1: same

julien51:
ok, one more thing. I see there is at least one duplicated rule between the `base` and the `next` config. Can you remove the duplicate from the `next` config.
Also, can you please check that all rules in `base` and `next` are needed and remove them if not?
Thanks!

julien51 (approved):

julien51:
Great! Tahnsk a lot!

PR #14609, opened 2024-09-16
fix(networks): some network check fixes

clemsos:
# Description

This adds some corrections to the network validation task, namely 

-  [x] ignore DAO-owned contracts in ownership checks
-  [x] fix etherscan /blockscout source code api parsing
- [x]  basic docs
- [x] check for missing hooks

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/packages/networks/bin/utils/contracts.ts
+++ b/packages/networks/bin/utils/contracts.ts
@@ -1,5 +1,6 @@
 import { ethers } from 'ethers'
 import { queryGraph } from './subgraph'
+import { HookType } from '@unlock-protocol/types'
 
 export const getAllAddresses = async ({ network }) => {
   const {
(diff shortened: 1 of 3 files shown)


packages/networks/bin/utils/contracts.ts
```diff
@@ -44,14 +44,24 @@ export const getAllAddresses = async ({ network }) => {
...
-  // TODO: get all hooks
-  if (hooks) {
-    if (hooks.onKeyPurchaseHook) {
-      hooks.onKeyPurchaseHook.map(({ address, id }) => {
-        addresses[`onKeyPurchaseHook_${id}`] = address
-      })
-    }
-  }
+  // TODO: get hooks from type file
+  const requiredHooks = [
```
julien51, line 48: Do you think you could do that in this PR so this does not slip thru the cracks?

packages/networks/bin/validate.ts
```diff
@@ -13,6 +20,9 @@ import {
   getAllAddresses,
 } from './utils'
 
+const DAO_CHAIN_ID = 1 // mainnet for now
```
julien51, line 23: Since we are moving to using the safe + zodiac module on every chain, do we still need this?

We would, however need to check that the zodiac module receives its instructions from the Timelock contract?

julien51 (changes requested):
I think we need to ad the types for hooks now?

packages/networks/bin/validate.ts
clemsos: Yes but this will currently lead PRs to fail. We can updated once the Base system is in place.

packages/networks/bin/utils/contracts.ts
clemsos: yes sure

clemsos:
> I think we need to ad the types for hooks now?

Ok done !

julien51 (approved):

PR #14617, opened 2024-09-17
fix(unlock-app): fix `lock` index page direct reload

0xTxbi:
# Description
This PR introduces fixes that addresses the critical issue causing the `locks` index page to crash on direct reload. 

Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/components/interface/locks/List/elements/LockList.tsx
+++ b/unlock-app/src/components/interface/locks/List/elements/LockList.tsx
@@ -2,6 +2,7 @@
 import { Disclosure } from '@headlessui/react'
 import { Lock } from '~/unlockTypes'
 import { useConfig } from '~/utils/withConfig'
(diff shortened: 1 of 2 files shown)


unlock-app/src/components/interface/locks/List/elements/LockList.tsx
```diff
@@ -103,30 +101,26 @@ export interface FavoriteLocks {
 }
 
 export const LockList = ({ owner }: LockListProps) => {
+  const config = useConfig()
```
julien51, line 104: Hum. Ant way we can just use the import? I think the him is not the right approach

0xTxbi: I think so too!

julien51 (approved):

PR #14625, opened 2024-09-18
fix(networks): avoid hitting etherscan api rate limit when checking contract on new networks 

clemsos:
# Description

We wait a little longer between etherscan api calls to avoid hitting the limit rate

# Issues

Fixes #
Refs #

# Checklist:


unlock-deployer (approved):
Can we do retries as well maybe?

clemsos:
> Can we do retries as well maybe?

the issue is that etherscan api does not differentiate between an api error and a contract that is not verified. It all returns a code "NOTOK" and even the error message does not have consistent indications (therefore #14622)

PR #14633, opened 2024-09-18
feat:lint ignore .next folder in nextjs packages

### LINKED ISSUE #14564: Eslint cleanup

Myestery:
# Description
This pull request helps to fix the hanging issue that occurs in nextjs apps when lint is run.
The cause was that due to the way we changed the eslint command in all packages, we now have to specify blacklist instead of whitelisted folders. My last PR unfortunately did not cater for the .next folder and it caused eslint to hang when run on nextjs packages.

# Issues

Fixes #14564
Refs #

# Checklist:

### DIFF
--- a/packages/eslint-config/src/next.js
+++ b/packages/eslint-config/src/next.js
@@ -29,4 +29,7 @@ module.exports = [
       '@typescript-eslint/no-unused-expressions': 'warn',
     },
   },
+  {
+    ignores: ['.next'],
+  },
 ]


julien51:
ha! great catch here @Myestery !

julien51 (approved):

PR #14806, opened 2024-10-04
feat(smart-contracts): use struct to parse purchase args

### LINKED ISSUE #13457: Use structs as args for purchase
The current models for the purchase function args is to pass several arrays of identical length - one for each required param of a purchase.

**Current signature**

```js
function purchase(
    uint256[] calldata _values,
    address[] calldata _recipients,
    address[] calldata _referrers,
    address[] calldata _keyManagers,
    bytes[] calldata _data
  ) external payable returns (uint256[] memory tokenIds);

```

This makes unpacking encoding args difficult as array can be of arbitrary length.

**Possible improvement**

A more elegant solution could be to have a single array of Struct containing each the param for a single purchase.

```js

Struct PurchaseArgs {
  uint value
  address recipient 
  address referrer
  address keyManager
  bytes memory data
}

function purchase(PurchaseArgs[]) external payable returns (uint256[] memory tokenIds);

```

Not sure what will be the downsides, but it sure improves readability as well as the ability to loop and decode an encoded array of Structs.

Any thoughts?

clemsos:
# Description

We introduce a simpler definition of arguments passed to the `purchase` function by using structs instead of multiple arrays

# Issues

Fixes #13457
Refs #

# Checklist:


julien51 (approved):
ok LGTM!

PR #14833, opened 2024-10-10
feature(locksmith): Update access control for `checkoutConfigs`

### LINKED ISSUE #14248: Let lock managers edit checkout URL
Right now, we use the `createdBy` field on a checkout URL to list them and/or enable edits.
Instead of relying on this mechanism for "access control" we should use lock managers.

0xTxbi:
# Description
This PR introduces changes to transition the access control mechanism for checkout configurations from relying on the `createdBy` field to using **lock managers**.

Previously, a user could only list, or edit a checkout config if they were its creator. With this update, **lock managers** are now granted the ability to manage configurations for locks they control, even if they are not the original creators.

Key changes include:

- Allowing **Lock Managers** to save and manage configurations tied to locks they manage, regardless of who originally created the configuration.
- Listing configurations for users if they are a lock manager on at least one associated lock, even if they did not create the configuration.
- Refactoring the backend by separating logic into operations and controllers, improving organization.

# Issues
Fixes #14248
Refs #14248

# Checklist:


locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -8,35 +13,101 @@ interface SaveCheckoutConfigArgs {
...
+ * @param config - The checkout configuration object
+ * @returns An array of objects containing lock address and network
+ */
+const extractLockInfo = (
+  config: any
+): { address: string; network: number }[] => {
+  return Object.entries(config.locks).map(
+    ([address, lockConfig]: [string, any]) => ({
+      address,
+      network: lockConfig.network,
```
julien51, line 27: lockconfig.network may be missing. In that case, the paywallconfigwill have a "root" level network, like this:
```json
{
  locks: {
     '0x123' : {}
 },
 network: 123
}
```

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -8,35 +13,101 @@ interface SaveCheckoutConfigArgs {
...
+  checkoutConfig: CheckoutConfig
+): Promise<boolean> => {
+  const lockInfo = extractLockInfo(checkoutConfig.config)
+  for (const { address, network } of lockInfo) {
+    if (await isLockManager(address, userAddress, network)) {
+      return true
+    }
+  }
+  // Fallback: check if user is the creator
+  return checkoutConfig.createdBy === userAddress
```
julien51, line 69: Do we want to do that first since will be faster?

locksmith/src/operations/checkoutConfigOperations.ts
0xTxbi: got it!

locksmith/src/operations/checkoutConfigOperations.ts
0xTxbi: I guess so!

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -55,3 +126,81 @@ export const getCheckoutConfigById = async (id: string) => {
...
+
+/**
+ * Retrieves all checkout configurations a user is authorized to manage
+ * @param userAddress - The address of the user
+ * @returns An array of authorized checkout configurations
+ */
+export const getAuthorizedCheckoutConfigs = async (userAddress: string) => {
+  const allConfigs = await CheckoutConfig.findAll({
+    order: [['updatedAt', 'DESC']],
+  })
```
julien51, line 159: Hum. This will not work well, because there are already a few thousand checkout configs... and iterating thru them will take too much time.
I am not sure there is a way to list configs that a user is allowed to manage right now.

I think the only way to handle this would be on the front-end to let a user enter a checkout config id and then verify that they can indeed manage that specific config id.

julien51 (changes requested):
Let's talk about the way to list... because I don't think we can do that.

0xTxbi:
> Let's talk about the way to list... because I don't think we can do that.

sure thing!

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -20,30 +19,19 @@ export const createOrUpdateCheckoutConfig: RequestHandler = async (
...
-  if (id) {
-    const authorized = await isAuthorizedToManageConfig(userAddress, id)
-    if (!authorized) {
-      return response.status(403).send({
-        message:
-          'Unauthorized: You must be a Lock Manager of at least one associated lock or the creator.',
-      })
-    }
-  }
-
```
julien51: I think that was good though. We should make it so that any manager on any lock in the config can update the config.

0xTxbi: okay okay got it!

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -65,19 +53,10 @@ export const createOrUpdateCheckoutConfig: RequestHandler = async (
...
 export const getCheckoutConfig: RequestHandler = async (request, response) => {
   const id = request.params.id
-  const userAddress = request.user!.walletAddress
-
-  const authorized = await isAuthorizedToManageConfig(userAddress, id)
-  if (!authorized) {
-    return response.status(403).send({
-      message: 'Unauthorized: You do not have access to this configuration.',
-    })
-  }
```
julien51: same, that was good.

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -34,13 +34,20 @@ export const createOrUpdateCheckoutConfig: RequestHandler = async (
...
-      config: createdConfig.config,
-      updatedAt: createdConfig.updatedAt.toISOString(),
-      createdAt: createdConfig.createdAt.toISOString(),
+      id: createdConfig?.id,
+      by: createdConfig?.createdBy,
+      name: createdConfig?.name,
+      config: createdConfig?.config,
+      updatedAt: createdConfig?.updatedAt.toISOString(),
+      createdAt: createdConfig?.createdAt.toISOString(),
     })
```
julien51, line 51: I don't think you need to add the `?` here because if `createdConfig` is not defined, the above will catch it?

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -113,9 +120,8 @@ export const deleteCheckoutConfig: RequestHandler = async (
   const deleted = await deleteCheckoutConfigOperation(userAddress, id)
 
   if (!deleted) {
-    return response.status(403).send({
-      message:
-        'Unauthorized: You do not have permission to delete this configuration.',
+    return response.status(404).send({
```
julien51, line 123: Let's keep 403 here. 404 would be for "not found" . If needed we can add a check to see if the config exists?

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -95,6 +94,17 @@ export const saveCheckoutConfig = async ({
     checkoutConfigData.referrer = createdBy
   }
 
+  // Check if the user is authorized to save/update the config
+  if (id) {
+    const existingConfig = await CheckoutConfig.findOne({ where: { id } })
+    if (
+      existingConfig &&
+      !(await isUserAuthorized(createdBy, existingConfig))
```
julien51, line 102: for clarity, please let's rename `createdBy` to something like `user` because the user could be just a lock manager, not the creator of the config.

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -1,62 +1,93 @@
...
+      config: createdConfig.config,
+      updatedAt: createdConfig.updatedAt.toISOString(),
+      createdAt: createdConfig.createdAt.toISOString(),
+    })
+  } catch (error) {
+    console.error('Error creating/updating checkout config:', error)
+    return response.status(500).send({
+      message: 'An error occurred while processing your request.',
+    })
+  }
```
julien51, line 57: We already have a middleware that handles try/catch so let's not add one here.

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -1,62 +1,93 @@
...
-    id: createdConfig.id,
-    by: createdConfig.createdBy,
-    name: createdConfig.name,
-    config: createdConfig.config,
-    updatedAt: createdConfig.updatedAt.toISOString(),
-    createdAt: createdConfig.createdAt.toISOString(),
-  })
+
+  try {
+    const createdConfig = await saveCheckoutConfig({
```
julien51, line 30: For clarity lety's please rename this variable because it can be an existing one. Maybe just `storedConfig` ?

0xTxbi: got it!

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -1,62 +1,93 @@
...
   response
 ) => {
   const userAddress = request.user!.walletAddress
-  const checkoutConfigs = await CheckoutConfig.findAll({
-    where: {
-      createdBy: userAddress,
-    },
-    order: [['updatedAt', 'DESC']],
-  })
+  const checkoutConfigs = await getCheckoutConfigsByUserOperation(userAddress)
```
julien51, line 89:
```suggestion
  const checkoutConfigs = await getCheckoutConfigsByUser(userAddress)
```

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -1,42 +1,132 @@
...
 export const saveCheckoutConfig = async ({
   id,
   name,
-  createdBy,
+  user,
   config,
-}: SaveCheckoutConfigArgs) => {
-  const generatedId = randomUUID()
+}: SaveCheckoutConfigArgs): Promise<CheckoutConfig | null> => {
+  const generatedId = id || randomUUID()
```
julien51, line 88: I think this is confusing...

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -1,42 +1,132 @@
...
+  if (id) {
+    const existingConfig = await CheckoutConfig.findOne({ where: { id } })
+    if (existingConfig && !(await isUserAuthorized(user, existingConfig))) {
+      return null
+    }
   }
 
   const [createdConfig] = await CheckoutConfig.upsert(
     {
+      id: generatedId,
```
julien51, line 107:
```suggestion
id: id || generatedId,
```

locksmith/src/operations/checkoutConfigOperations.ts
julien51, line 88:
```suggestion
  const generatedId = randomUUID()
```

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -1,42 +1,132 @@
...
       name,
-      id: id || generatedId,
-      config,
-      createdBy,
+      config: checkoutConfigData,
+      createdBy: user,
```
julien51, line 110: If this is an existing config I worry we are overwriting the `createdBy` field...

```suggestion
      createdBy: existingConfig.createdBy | user,
```

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -1,42 +1,132 @@
...
     },
     {
-      conflictFields: ['id', 'createdBy'],
+      conflictFields: ['id'],
```
julien51, line 113:
```suggestion
      conflictFields: ['id', 'createdBy'],
```

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -1,42 +1,132 @@
...
     }
   )
+
+  // Ensure createdConfig is not null before returning
+  if (!createdConfig) {
```
julien51, line 118: Above you return `null`. Let's do the same here for consistency?

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -55,3 +145,44 @@ export const getCheckoutConfigById = async (id: string) => {
   }
   return null
 }
+
+export const getCheckoutConfigsByUserOperation = async (
```
julien51, line 149:
```suggestion
export const getCheckoutConfigsByUser = async (
```

locksmith/src/operations/checkoutConfigOperations.ts
```diff
@@ -55,3 +145,44 @@ export const getCheckoutConfigById = async (id: string) => {
...
+  })
+}
+
+/**
+ * Deletes a checkout configuration if the user is authorized
+ * @param userAddress - The address of the user
+ * @param configId - The ID of the checkout configuration
+ * @returns A boolean indicating whether the deletion was successful
+ */
+export const deleteCheckoutConfigOperation = async (
```
julien51, line 166:
```suggestion
export const deleteCheckoutConfig = async (
```

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -71,24 +102,37 @@ export const getCheckoutConfigsByUser: RequestHandler = async (
...
+
+  if (!existingConfig) {
     return response.status(404).send({
-      message: 'Not found',
+      message: 'Config not found.',
     })
   }
-  await checkoutConfig.destroy()
+
+  const deleted = await deleteCheckoutConfigOperation(userAddress, id)
```
julien51, line 128:
```suggestion
  const deleted = await deleteCheckoutConfig(userAddress, id)
```

julien51 (changes requested):
More cleanup and changes...

locksmith/src/operations/checkoutConfigOperations.ts
julien51: @0xTxbi please check this

julien51 (changes requested):
I think it would be nice to add tests here please

julien51 (approved):

Issue #14102, opened 2024-06-25
Support ENS in checkout builder for Referrer field


julien51:
Currently our checkout builder does not support using ENS for the referrer address. Howeever we already have a `<AddressInput>` component that could be used there!

Additionally, if the config does use an ENS, instead of having an address, the paywall UI should be able to "resolve" that ENS and use the resolved address when sending transactions.

sudheerDev:
@julien51 Can you assign this to me

ManyRios:
Hey @julien51 Can I take this one?

blahkheart:
Hey @julien51 is this issue still open? Does @ManyRios have the nod to work on this? Else I'm open to taking it on.

julien51:
@blahkheart you're on it! this one is 150 USDC

blahkheart:
Alright on it

blahkheart:
Hello @julien51 I've fixed the first part of the issue, can you please share more insight on what you mean when you say `Additionally, if the config does use an ENS, instead of having an address, the paywall UI should be able to "resolve" that ENS and use the resolved address when sending transactions.` so that we are on the same page? At the moment, I'm looking to add ENS resolution to the `setPaywallConfig` method in `Paywall.ts` file found in `packages/paywall/src/Paywall.ts`. Please check my understanding of your request.

xordanx:
Opened PR #16435 for this bounty: https://github.com/unlock-protocol/unlock/pull/16435\n\nIt covers both parts of the issue: the checkout builder resolves ENS entered in the referrer field, and checkout resolves any global or lock-specific ENS referrers in the paywall config before transaction arguments are built.

PR #14834, opened 2024-10-10
feat(unlock-app): Support ENS in checkout builder for Referrer field

sudheerDev:
# Description
* Add changes to use AddressField component to resolve ens in checkout builder form
* Changed getReferrer to getReferrers to support an array of addresses
* Added a new hook useGetReferrer
* Added new util func shouldReferrerResolveForENS to check if a lock config has ENS
* use shouldReferrerResolveForENS in getReferrers to resolve for ENS
* Add changes to instances of getReferrers to be async
 

# Issues

Fixes # https://github.com/unlock-protocol/unlock/issues/14102
Refs #

# Checklist:

348459375-d371d3eb-b37e-4fca-a7a9-dddb78e046bb

### DIFF
--- a/unlock-app/package.json
+++ b/unlock-app/package.json
@@ -113,6 +113,7 @@
     "deploy-fleek": "./scripts/deploy-fleek.sh",
     "start": "yarn build && NODE_ENV=production next start",
     "test": "UNLOCK_ENV=test vitest run --coverage --environment=jsdom",
+    "test:watch": "UNLOCK_ENV=test vitest --coverage --environment=jsdom -w",
     "lint": "eslint",
     "ci": "yarn test && yarn lint && yarn build"
   },
--- a/unlock-app/src/components/interface/locks/CheckoutUrl/elements/BasicConfigForm.tsx
+++ b/unlock-app/src/components/interface/locks/CheckoutUrl/elements/BasicConfigForm.tsx
@@ -1,15 +1,17 @@
 import { BasicPaywallConfigSchema } from '~/unlockTypes'
-import { useForm } from 'react-hook-form'
+import { Controller, useForm, useWatch } from 'react-hook-form'
 import {
   Input,
   Checkbox,
   Button,
   ImageUpload,
   Modal,
+  AddressInput,
 } from '@unlock-protocol/ui'
 import { z } from 'zod'
 import { useState } from 'react'
 import { useImageUpload } from '~/hooks/useImageUpload'
+import { onResolveName } from '~/utils/resolvers'
 
 interface Props {
   onChange: (values: z.infer<typeof BasicPaywallConfigSchema>) => void
@@ -24,6 +26,7 @@ export const BasicConfigForm = ({ onChange, defaultValues }: Props) => {
     watch,
     setValue,
     formState: { errors },
+    control,
   } = useForm<z.infer<typeof BasicPaywallConfigSchema>>({
     reValidateMode: 'onChange',
     defaultValues: defaultValues as any,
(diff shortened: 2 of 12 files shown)


unlock-app/src/utils/checkoutLockUtils.ts
```diff
@@ -145,29 +148,61 @@ export const inClaimDisallowList = (address: string) => {
...
+  recipients: string[],
+  paywallConfig?: PaywallConfigType,
+  lockAddress?: string
+): Promise<string[]> => {
+  const isReferrerAddressEns = shouldReferrerResolveForENS(
+    paywallConfig,
+    lockAddress
+  )
+  if (isReferrerAddressEns) {
+    if (isReferrerAddressEns && paywallConfig && paywallConfig.referrer) {
```
sudheerDev, line 180: The reason to change this to be here is that if the referrer is an ENS it will be the same for all. There is no reason to call this for all recipients since that will unnecessarily make quite a few ENS calls.

sudheerDev: This codepath will be executed for any existing ENS addresses on config before this change goes live. since it will be formatted to address from here on out.

sudheerDev:
@julien51 Let me test this once, and I will add you as a reviewer.

unlock-app/package.json
```diff
@@ -113,6 +113,7 @@
     "deploy-fleek": "./scripts/deploy-fleek.sh",
     "start": "yarn build && NODE_ENV=production next start",
     "test": "UNLOCK_ENV=test vitest run --coverage --environment=jsdom",
+    "test:watch": "UNLOCK_ENV=test vitest --coverage --environment=jsdom -w",
```
sudheerDev, line 116: Added for convenience, Let me know if this has to be reverted

unlock-app/src/hooks/useGetReferrers.ts
```diff
@@ -0,0 +1,17 @@
+import { useQuery } from '@tanstack/react-query'
+import { PaywallConfigType } from '@unlock-protocol/core'
+
+import { getReferrers } from '~/utils/checkoutLockUtils'
+
+export const useGetReferrers = (
```
julien51, line 6: I don't understand why we need a hook for this and we could not call `getReferrers` when we need ?

unlock-app/src/components/interface/locks/CheckoutUrl/elements/BasicConfigForm.tsx
```diff
@@ -85,12 +92,25 @@ export const BasicConfigForm = ({ onChange, defaultValues }: Props) => {
...
+            <AddressInput
+              withIcon={false}
+              placeholder="0x..."
+              label="Referrer Address Or ENS"
+              description={BasicPaywallConfigSchema.shape.referrer.description}
+              onResolveName={onResolveName}
+              error={errors.referrer?.message}
+              value={referrer}
+              onChange={(value: any) => {
+                setValue('referrer', value)
```
julien51, line 109: 👍

unlock-app/src/hooks/useGetReferrers.ts
julien51: (you actually do that in several places and that's correct!)

unlock-app/src/utils/checkoutLockUtils.ts
julien51: I am not sure the referrer should ever be an ENS. You are resolving the ENS in the settings UI and the config only receives the actual address, which, IMO is good!

julien51 (changes requested):
Please, remove the hook and let's assume the paywall config's `referrer` value is always an address. If we ever want to add support for ENS, then we should do it upstream.

unlock-app/src/hooks/useGetReferrers.ts
sudheerDev: since `getReferrers` is not sync and it does make rpc calls, I am using hook. react query  caches it based on the inputs to avoid making extra calls on re-renders causing issues.

PR #14877, opened 2024-10-25
feature(smart-contracts): adding the ability to execute more tx as part of a deployment.

julien51:
# Description

One pattern we have seen multiple times is the need to deploy a lock but customize its behavior immediately (add lock managers, change fees... etc). 
Until now, this requires sending multiple transaction: one to deploy the lock and then all the ones to change the values that need to be changed.

I propose a minor addition to let someone pass a list of transactions to be executed from _Unlock.sol_ on the newly deployed lock. This will let someone deploy and customize a lock in a single transaction.

# Checklist:

### DIFF
--- a/smart-contracts/contracts/Unlock.sol
+++ b/smart-contracts/contracts/Unlock.sol
@@ -1,3 +1,5 @@
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.21;
 
@@ -81,7 +83,7 @@ contract Unlock is UnlockInitializable, UnlockOwnable {
   // The WETH token address, used for value calculations
   address public weth;
 
-  // DEPRECATED: udt was the name of the first governance token. 
+  // DEPRECATED: udt was the name of the first governance token.
   // Kept for backward compatibility; however, the `governanceToken` getter is now preferred.
   address public udt;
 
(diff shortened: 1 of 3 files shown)


smart-contracts/contracts/Unlock.sol
```diff
@@ -315,6 +317,37 @@ contract Unlock is UnlockInitializable, UnlockOwnable {
...
+    bytes[] calldata transactions
+  ) public returns (address newLock) {
+    newLock = createUpgradeableLockAtVersion(data, lockVersion);
+    // Execute all transactions
+    for (uint256 i = 0; i < transactions.length; i++) {
+      (bool success, ) = newLock.call(transactions[i]);
+      if (!success) {
+        revert(
+          string.concat(
+            "LockProxyDeployer: transaction failed",
```
clemsos, line 343: Having a literal string may add to the size of the contract significantly. Better to use a custom error

clemsos (approved):

smart-contracts/contracts/Unlock.sol
clemsos: also both OZ and 3rd web have a way to detect forwarded calls. Maybe we can just use one of these lib  

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Multicall.sol

https://github.com/thirdweb-dev/contracts/blob/main/contracts/extension/Multicall.sol

julien51: Good point! Let me do that then!

julien51: ok so when looking at the OZ  and Thirdweb libs they seem to focus on delegate calls, where the code executed is in the context of the caller (here that would be Unlock), which is not really what I would want to achieve...

PR #14884, opened 2024-10-28
feat(governance): check that SAFE bridge/delay modules setup is correct

### LINKED ISSUE #14852: Heath Report : check "control" of Unlock contracts
Our health check already verifies that the Unlock contracts on each chain are owned by the local multisig, however we don't check that the DAO can indeed send transactions. 
For this we need to query the Zodiac  module and check origin & domainId.

We should also add a check in the health check to verify if there are any transaction pending in that zodiac module.

clemsos:
# Description

This adds checks for SAFE modules used by cross-chain governance process and make sure they are correctly set.

# Issues

Fixes #14852 
Refs #

# Checklist:


governance/scripts/bridge/check.js
```diff
@@ -1,7 +1,14 @@
-const { targetChains, DelayMod: delayModAbi } = require('../../helpers/bridge')
+const {
+  targetChains,
+  DelayMod: delayModAbi,
+  ConnextMod: connextModAbi,
+} = require('../../helpers/bridge')
 const { getProvider } = require('../../helpers/multisig')
 const { Contract } = require('ethers')
 
+const BASE_DAO_TIMELOCK_ADDRESS = '0xB34567C4cA697b39F72e1a8478f285329A98ed1b'
```
julien51, line 9: Can we get that from the networks package please?

governance/scripts/bridge/check.js
```diff
@@ -1,7 +1,14 @@
...
+const BASE_DOMAIN_ID = '1650553709'
```
julien51, line 10: same here?

julien51 (changes requested):
Please let's use the networks package to pull addresses and domain id... this will ensure we're not forgetting to change things when we need to :)

julien51 (approved):

PR #14912, opened 2024-10-31
fix(unlock-app): checkout frame errors

### LINKED ISSUE #14888: Debug errors on frame checkout
https://warpcast.com/~/developers/frames?url=https%3A%2F%2Fapp.unlock-protocol.com%2Fcheckout%3Fid%3D704ace46-ec1e-4c5a-975b-dd0e02f835e5

Screenshot 2024-10-28 at 5 56 03 PM

I suspect one of 2 things:
- the user already has a membership and in that case the frame should indicate that to the user first?
- the transaction needs and ERC20 approval first in order to succeed... if so we should send that transaction first.

iMac7:
# Description

# Issues

Fixes #14888 
Refs #13593 

# Checklist:

Indeed an erc20 approval is needed first. This handles that by having the first button click as the approval and the second as the actual purchase, lmk if that's confusing or a more elegant way, I don't see a way to 'batch' those 2 together in a single transaction so far :thinking: .

As for checking if a user has a membership before the purchase, this can happen by cancelling a transaction midway, because the user's wallet details are not accessible elsewhere. Doing that would throw an error anyway so I'm assuming a failed purchase (because you already have a key) does the same and it's sort of an 'acceptable' error

https://github.com/user-attachments/assets/72ada736-41af-4691-be4a-172774addc6b

## Release Note Draft Snippet

image


unlock-app/app/frames/checkout/components/defaultFrame.tsx
```diff
@@ -1,11 +1,15 @@
...
 
-export function getDefaultFrame(state: State) {
-  const lock = state.lock!
+export function getDefaultFrame(ctx: any) {
+  const lock = ctx.state.lock!
+  const isErc20 = !!ctx.state.lock?.tokenAddress
+  const approved = ctx.searchParams.approved || !isErc20
+  const buttonText = approved
+    ? 'Purchase a key'
+    : `Approve ${lock.currencySymbol} spending`
```
julien51, line 12:
```suggestion
  const buttonText = approved
    ? 'Mint'
    : `Approve ${lock.currencySymbol} spending`
```

unlock-app/app/frames/checkout/components/utils.ts
```diff
@@ -42,6 +73,16 @@ export async function getLockDataFromCheckout(id: string) {
   const res = await web3Service.getLock(lockAddress, network)
   const price = `${res.keyPrice} ${res.currencySymbol}`
 
+  let tokenAddress
+  if (res.currencySymbol !== 'ETH') {
```
julien51, line 77: That's not the right way to check if the lock is priced in the native currency. Thus would fail on Polygon for example. Just check that the `lock.tokenAddress` is the Zero address.

julien51 (changes requested):
Some changes needed.

julien51:
> Indeed an erc20 approval is needed first. This handles that by having the first button click as the approval and the second as the actual purchase, lmk if that's confusing or a more elegant way, I don't see a way to 'batch' those 2 together in a single transaction so far 🤔 .

In the case of an ERC20 lock, you first need to check if there is an existing approval and if that approval is "enough" ( > price). I don't think you can "batch" the 2 tx. The user would _first_ need to approve and 2nd need to send the purchase.

julien51:
> As for checking if a user has a membership before the purchase, this can happen by cancelling a transaction midway, because the user's wallet details are not accessible elsewhere. Doing that would throw an error anyway so I'm assuming a failed purchase (because you already have a key) does the same and it's sort of an 'acceptable' error

No, you need to actually perform a check and can't rely on the fact that there is an error, because the purchase could fail for other reasons (sold out.. etc).
So this means we can't show a `Mint` button right away but start with an intermediary action like `Continue`. This button would then trigger an action where you can verify if the user does indeed own a specific key to the lock. If they do, you can show a message saying "you already have a key"... and if they don't you can check if an approval is needed. If it is, trigger the approval, and if they don't (or once they have done it), just trigger the actual purchase tx.

What do you think?

julien51:
> to get the user address, they are first required to sign a dummy message to prompt the wallet connect.

You can get it from the frame as long as the user has interracted with it, which is why the "Continue" button is useful :)

unlock-app/app/frames/checkout/approve/route.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const POST = frames(async (ctx) => {
+  if (!ctx?.message) {
+    throw new Error('Invalid frame message')
+  }
+
+  const lock = ctx.state.lock!
+  const { address: lockAddress, network, priceForUser } = lock
+
+  const calldata = encodeFunctionData({
```
julien51, line 13: You don't need to use that! unlock-js has the logic you need to use.

julien51:
> This might be specific to the current version of framesjs or only on the local debugger not sure, but only a button with action 'mint' or 'tx' is able to give a prompt to connect a wallet. the continue button would have a 'post' or 'post_redirect' property which doesn't connect a wallet but only navigates to another page (I just confirmed locally that it doesn't).

You have the `fid`, and then you can get the wallet address as soon as the user has interracted wuth the frame https://framesjs.org/reference/js/getAddressesForFid#usage

unlock-app/app/frames/checkout/approve/route.ts
julien51: So I think you should first check if the user has already approved enough (and you can compute the amount that needs to be approved based on the ehckout config (especially if it supports renewals)

Sorry my initial comment was unclear. Check unlock-js where there is already logic around how to compute the amount to be approved.

unlock-app/app/frames/checkout/components/defaultFrame.tsx
```diff
@@ -1,11 +1,69 @@
...
+    lock.priceForUser = keyPrice
+
+    if (tokenAddress) {
+      const allowance = await checkAllowance(
+        lockAddress,
+        Number(network),
+        userAddress as string,
+        tokenAddress!
+      )
+      if (Number(allowance) >= Number(keyPrice)) {
```
julien51, line 43: That's not enough, the user may need to approve MORE than the keyprice to allow for renewals.

unlock-app/app/frames/checkout/txdata/route.ts
```diff
@@ -12,30 +10,14 @@ export const POST = frames(async (ctx) => {
...
-      data: mydata,
-      referrer: userAddress,
-    })
-    price = price.toString()
-    return price
-  }
-
-  const keyPrice = await getKeyPrice()
 
   const calldata = encodeFunctionData({
```
julien51, line 17: Please use unlock'js walletService here as well so this supports any version of the lock contract!

julien51:
Ok thanks! I need a little bot of thinking since this is a bt old and I am sure I forgot some of the thinking! Is the screencapture above up to date?

iMac7:
> Ok thanks! I need a little bot of thinking since this is a bt old and I am sure I forgot some of the thinking! Is the screencapture above up to date?

Functionality is mostly the same, just added a new button "Continue" as the first button displayed.

image

unlock-app/app/frames/checkout/approve/route.ts
julien51: Try using [https://github.com/unlock-protocol/unlock/blob/master/packages/unlock-js/src/PublicLock/v15/preparePurchaseKeyTx.js. Ideally since you need to support nultiple verssions you would just add a `preparePurchaseKeyTx` to `web3Service` ](https://github.com/unlock-protocol/unlock/blob/master/packages/unlock-js/src/web3Service.ts#L1165C9-L1165C20)
It will yield the data you need to use for the tx.

unlock-app/app/frames/checkout/txdata/route.ts
julien51: https://github.com/unlock-protocol/unlock/blob/master/packages/unlock-js/src/web3Service.ts#L1165C9-L1165C20

julien51 (commented):
ok this generally looks like the right approach!
How can I test this?

iMac7:
> ok this generally looks like the right approach! How can I test this?

https://framesjs.org/guides/debugger
with checkout url in the search bar.

sometimes the approval amount is off and needs a few seconds to get back to normal, looks like a bug in metamask.

julien51:
> @julien51 sometimes the erc20 approval amount is off and needs a few seconds to get back to normal, looks like a bug in metamask.

I don't understand what role MetaMask plays in that??

iMac7:
> > @julien51 sometimes the erc20 approval amount is off and needs a few seconds to get back to normal, looks like a bug in metamask.
> 
> I don't understand what role MetaMask plays in that??

Actually just a guess because it takes some seconds to update. Might be the dev environment with framesjs

julien51:
ok @iMac7 I really want to test this. Can you suggest how I can do that please?

iMac7:
@julien51 
First get a lock whose keys are purchased with erc20 tokens, then configure a checkout url with that lock.
The issue was the lack of support for that due to a missing erc20 approval step.

To test locally (like my screen recordings above), you can install [frames.js](https://framesjs.org/) globally on your system then [run it](https://framesjs.org/guides/debugger) . 
It opens a `localhost:3010` tab in your browser where the frames can be tested.
There's a big searchbar in the opened tab where you paste the link and it generates a frame from your checkout url, with actions and buttons and everything.
The local environment is not very stable, sometimes needs a couple refreshes for a result.

You can also use the deploy-preview link or staging link in the local debugger or farcaster like in the original issue #13593 and I'll address any arising issues ASAP.

PR #15017, opened 2024-11-04
feat(smart-contracts): remember referrer for renewals

### LINKED ISSUE #13487: Keep track of referrer address for renewals
Currently, the referrer only gets a fee on the first purchaser. The main reason is that the lock does not keep track of the referrer address on recurring purchases.

However, we already keep track of the price paid, the currency and the duration for renewals... so it should be possible to add the `referrer` address so that the lock "remembers" the initial referrer address and pays the fees on renewals (only on `renewMembershipFor`.

We should keep the signature identical for all functions and, if a referrer is set this one should have the priority for the `renewMemvershipFor` function.

Note: when calling `extend`, the user should have the ability to set/use a different referrer.

clemsos:
# Description

Adds a way to remember the referrer used when first purchasing a key then reuse it when renewing it.

# Issues

Fixes #13487
Refs #

# Checklist:


smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -470,6 +474,27 @@ contract MixinPurchase is
...
+   * Renew a given token
+   * get referrer that was set during the original purchase
+   * @notice only works for non-free, expiring, ERC20 locks
+   * @param _tokenId the ID fo the token to renew
+   * @param _referrer the address of the referrer
+   */
+  function renewMembershipFor(uint _tokenId, address _referrer) public {
+    _renewMembershipFor(_tokenId, _referrer);
+  }
+
```
clemsos, line 497: not sure we really need both signature. We can just reuse the existing referrer when zero address is passed?

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -36,17 +36,23 @@ contract MixinPurchase is
   // default to 0
   uint256 internal _gasRefundValue;
 
-  // Keep track of ERC20 price when purchased
+  // DEPREC: use of these is deprecated
```
julien51, line 39:
```suggestion
  // DEPREC: use of these is deprecated, kept for backward storage compatibility!
```

smart-contracts/contracts/mixins/MixinPurchase.sol
julien51: I think it is better for backwards compact? I would make this one as deprecated .. and do that change:
```suggestion
  /**
   * Renew a given token
   * get referrer that was set during the original purchase
   * @notice only works for non-free, expiring, ERC20 locks
   * @param _tokenId the ID fo the token to renew
   */
  function renewMembershipFor(uint _tokenId) public {
    _renewMembershipFor(_tokenId, _renewalConditions[_tokenId].referrer);
  }

  /**
   * Renew a given token
   * get referrer that was set during the original purchase
   * @notice only works for non-free, expiring, ERC20 locks
   * @param _tokenId the ID fo the token to renew
   * @param _referrer the address of the referrer
   */
  function renewMembershipFor(uint _tokenId, address _referrer) public {
    _renewMembershipFor(_tokenId, _renewalConditions[_tokenId].referrer);
  }

```

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -470,6 +474,27 @@ contract MixinPurchase is
...
+
+  /**
+   * Renew a given token
+   * get referrer that was set during the original purchase
+   * @notice only works for non-free, expiring, ERC20 locks
+   * @param _tokenId the ID fo the token to renew
+   * @param _referrer the address of the referrer
+   */
+  function renewMembershipFor(uint _tokenId, address _referrer) public {
+    _renewMembershipFor(_tokenId, _referrer);
```
julien51, line 495:
```suggestion
    _renewMembershipFor(_tokenId, _renewalConditions[_tokenId].referrer);
```

julien51 (commented):
one minor change and then LGTM!

smart-contracts/contracts/mixins/MixinPurchase.sol
clemsos: a simpler way maybe is to not add a new signature. If a referrer has been specified earlier, it takes precedent. if not, it takes the referrer from function args. wdyt ?

smart-contracts/contracts/mixins/MixinPurchase.sol
```diff
@@ -174,16 +180,17 @@ contract MixinPurchase is
...
-    if (_originalPrices[_tokenId] != _keyPrice) {
-      _originalPrices[_tokenId] = _keyPrice;
-    }
-    if (_originalDurations[_tokenId] != expirationDuration) {
-      _originalDurations[_tokenId] = expirationDuration;
-    }
-    if (_originalTokens[_tokenId] != tokenAddress) {
-      _originalTokens[_tokenId] = tokenAddress;
-    }
+  function _recordTokenTerms(
```
julien51, line 183: @clemsos How will tht work for upgraded contracts though?

julien51 (approved):

PR #15071, opened 2024-11-13
fix(smart-contracts): wrong token id after a key was burnt

clemsos:
# Description

This PR fixes a bug in token id assignation after burning a key. As the token id is based on the existing supply of keys and the `burn` function decrease that supply, the token id calculations should take into account the number of keys that have been burnt.

# Issues

Fixes #
Refs #

# Checklist:

### DIFF
--- a/smart-contracts/test/Lock/burn.js
+++ b/smart-contracts/test/Lock/burn.js
@@ -67,6 +67,18 @@ describe('Lock / burn', () => {
     compareBigNumbers(await lock.totalSupply(), totalSupply - 1n)
   })
 
+  it('tokenId is not used twice', async () => {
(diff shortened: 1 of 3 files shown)


smart-contracts/test/Lock/burn.js
```diff
@@ -67,6 +67,18 @@ describe('Lock / burn', () => {
     compareBigNumbers(await lock.totalSupply(), totalSupply - 1n)
   })
 
+  it('tokenId is not used twice', async () => {
+    const totalSupply = await lock.totalSupply()
+    // assert.equal(tokenId + 1n, totalSupply)
```
clemsos, line 72:
```suggestion
```

julien51 (approved):

PR #15085, opened 2024-11-14
feat(docs): Burn from the docs!

julien51:
# Description

Improving our docs and especially adding the ability to burn from the docs directly!
https://www.loom.com/share/5d728c20c9b444d0928a1a22851a1f50

# Checklist:


docs/src/components/Network.js
```diff
@@ -64,40 +67,178 @@ const getProtocolFee = async (provider, unlockAddress) => {
...
+          if (!tx) {
+            try {
+              tx = await unlock.swapAndBurn.populateTransaction(
+                token.address || ethers.ZeroAddress,
+                balance,
+                fee // pool for the token, not for UDT/UP
+              )
+            } catch (e) {
+              console.error(e)
+            }
```
clemsos, line 185: here you can maybe log only if there is no pool at all, also log the pool fee/token that fails. 

also if a pool exist for two or three fees, will that lead to multiple tx being populated ?

docs/src/components/Network.js
```diff
@@ -64,40 +67,178 @@ const getProtocolFee = async (provider, unlockAddress) => {
...
+      burn()
+    }
+  }, [burning, authenticated])
+
+  return (
+    <li>
+      {Number(formatUnits(balance, token.decimals)).toFixed(4)} {token.symbol}{' '}
+      <Button isLoading={true} onClick={burn}>
+        Burn
+      </Button>{' '}
```
clemsos, line 221: here could add a condition to show the button only if the swap burner is set correctly in Unlock

PR #15234, opened 2024-12-03
feat(unlock-app): Checkout hooks

### LINKED ISSUE #15067: Add a webhook to checkout
Our checkout UI includes several [configuration options](https://docs.unlock-protocol.com/tools/checkout/configuration).
We need to add a new configuration option called `hooks` at the "root" of the config.
This object should include multiple properties, whose key should be the events triggered by the [Paywall library](https://docs.unlock-protocol.com/tools/checkout/paywall#events):

```{
status: "",
authenticated: "",
transactionSent: "",
metadata: ""
}```

Each of the value is a URL.

Then, in the checkout UI we should implement the triggers. They should be triggered at the same time as the Javascript events, as POST requests. The body of the POST request should be a `json` object matching the payload sent by the events. Any 2XX response should be considered a success. Any other should be considered an error and the paywall should do up to 2 retries (maximum of 3 notifications), after 1 and 3 seconds respectively.

### PR DESCRIPTION
**Pull request by iMac7**
# Description

# Issues

Fixes #15067 
Refs #

# Checklist:

## Release Note Draft Snippet

https://www.loom.com/share/3ef701f86f8a4cea9398d05e515be7a6?sid=ef35bc2e-1fc9-4ce6-9479-87b93da5d7bc

### CODE REVIEW

File: unlock-app/src/components/interface/checkout/main/checkoutHookUtils.ts [hunk 1]
```diff
@@ -71,3 +73,58 @@ export const getOnPurchaseHookTypeFromPaywallConfig = (
...
 }
+
+let prevBody: string | null = null
+export const postToWebhook = async (body: any, config: any, event: string) => {
+  const url = config?.hooks && config.hooks[event]
+
+  if (!url) return
+
+  if (JSON.stringify(body) === prevBody) {
+    console.log('DUPLICATE REQUEST!! Skipping...', event, body)
```
julien51, line 84: Please clean this up.

unlock-app/src/components/interface/locks/CheckoutUrl/elements/BasicConfigForm.tsx
```diff
@@ -114,6 +114,37 @@ export const BasicConfigForm = ({ onChange, defaultValues }: Props) => {
...
+          {...register('hooks.transactionSent')}
+        />
+        <Input
+          label="Metadata"
+          size="small"
+          description="URL to be called for metadata updates"
+          {...register('hooks.metadata')}
+        />
+      </fieldset>
+
```
julien51, line 147: Let's do add this. This should be a "developer" only setting and I worry this may complicate things for non-dev users. Hooks should only be "configurable" with the API/

julien51 (commented):
I am a bit intrigued about notifications. Can you explain?

unlock-app/src/components/interface/checkout/main/checkoutHookUtils.ts
```diff
@@ -71,3 +73,73 @@ export const getOnPurchaseHookTypeFromPaywallConfig = (
...
+
+    throw lastError
+  }
+
+  try {
+    await retryRequest(3)
+  } catch (error) {
+    console.error('Webhook request failed:', error)
+  }
+}
```
searchableguy, line 145: This is a security issue as well as reliability. No calls should be made to user given urls from frontend. It should go through a proxy and have some sort of verification. 

Also exponential backoff is preferred in cases such as this and should account for failure by storing the request data in backend. This should be easy to add as a worker to graphile-worker and start from backend.

unlock-app/src/components/interface/layouts/index/NotificationsMenu.tsx
```diff
@@ -46,6 +59,14 @@ export function NotificationsMenu() {
...
   }
 
+  if (!isLoading && !error && hooks?.data?.length && hooks?.data?.length > 0) {
+    notifications.push({
+      id: '2',
+      content: <Hooks hooks={hooks.data} refetch={refetch} />,
+      timestamp: new Date(),
+    })
+  }
+
```
searchableguy, line 69:
```suggestion
const notifications = useMemo<NotificationProps[]>(() => {
    const items: NotificationProps[] = []

    if (!email && !['/checkout', '/demo'].includes(pathname)) {
      items.push({
        id: '1',
        content: <PromptEmailLink setModalOpen={setShowModal} />,
        timestamp: new Date(),
      })
    }

    if (!isLoading && !error && hooks?.data?.length > 0) {
      items.push({
        id: '2',
        content: <Hooks hooks={hooks.data} refetch={refetch} />,
        timestamp: new Date(),
      })
    }

    return items
  }, [email, pathname, isLoading, error, hooks, refetch])
```

useMemo seem appropriate here. There is another if statement above for filtering.

searchableguy, line 69:
```suggestion
const notifications = useMemo<NotificationProps[]>(() => {
    const items: NotificationProps[] = []

    if (!email && !['/checkout', '/demo'].includes(pathname)) {
      items.push({
        id: '1',
        content: <PromptEmailLink setModalOpen={setShowModal} />,
        timestamp: new Date(),
      })
    }

    if (!isLoading && !error && hooks?.data?.length > 0) {
      items.push({
        id: '2',
        content: <Hooks hooks={hooks.data} refetch={refetch} />,
        timestamp: new Date(),
      })
    }

    return items
  }, [email, pathname, isLoading, error, hooks, refetch])
```

useMemo seems appropriate. There is another if statement filtering the items and adding to array.

iMac7:
Add hooks to checkout:
PUT - {locksmith_uri}/v2/checkout/hooks/:checkout_id
body - {event: url to post to}
{"authenticated":"http://localhost:8080"}

iMac7:
@searchableguy

locksmith/src/controllers/v2/checkoutController.ts
```diff
@@ -141,3 +143,160 @@ export const deleteCheckoutConfig: RequestHandler = async (
   })
   return
 }
+
+export const updateCheckoutHooks: RequestHandler = async (
```
julien51, line 147: I would love to see tests added to make sure this all works as expected please!

locksmith/src/worker/tasks/checkoutHooks.ts
```diff
@@ -0,0 +1,38 @@
+import { Task } from 'graphile-worker'
+import { Payload } from '../../models/payload'
+import logger from '../../logger'
+import axios from 'axios'
+import { getCheckoutConfigById } from '../../operations/checkoutConfigOperations'
+
+export const checkoutHookJob: Task = async (payload: any) => {
```
julien51, line 7: I don't think we should send these from the backend to be honest. They should be triggered from the front-end directly

packages/core/src/schema.ts
```diff
@@ -313,6 +313,30 @@ export const PaywallConfig = z
...
+            description: 'URL to be called when a transaction is sent.',
+          })
+          .optional(),
+        metadata: z
+          .string({
+            description: 'URL to be called for metadata updates.',
+          })
+          .optional(),
+      })
+      .optional(),
```
julien51, line 339: 👍

packages/unlock-js/openapi.yml
```diff
@@ -3983,6 +3983,229 @@ paths:
         401:
           $ref: '#/components/responses/401.NotAuthenticated'
 
+  /v2/checkout/hooks/all:
```
julien51, line 3986: I am really not a fan of adding new API endpoints for this

unlock-app/src/components/interface/HooksNotifications.tsx
```diff
@@ -0,0 +1,49 @@
+import dayjs from 'dayjs'
+import { useState } from 'react'
+import { locksmith } from '~/config/locksmith'
+
+export const Hooks = ({
```
julien51, line 5: I really don't undersyand what this component does? Can you explain?

unlock-app/src/components/interface/checkout/CheckoutContainer.tsx
```diff
@@ -52,6 +58,40 @@ export function CheckoutContainer() {
...
+            prevBody = JSON.stringify(body)
+            postToWebhook(body, paywallConfig, 'status')
+          }
+        }
+      )
+
+      window.unlockProtocolConfig = paywallConfig
+
+      script = document.createElement('script')
+      script.src = `${config.paywallUrl}/static/unlock.latest.min.js`
```
julien51, line 84: I don't think it makes sense to embed the Paywall here

unlock-app/src/components/interface/checkout/main/Confirm.tsx
```diff
@@ -31,17 +31,23 @@ export function Confirm({ checkoutService, communication }: Props) {
...
+          network,
+        },
+        paywallConfig
+      )
+      communication?.emitUserInfo(
+        {
+          address: account,
+          signedMessage: messageToSign?.signature,
+        },
+        paywallConfig
```
julien51, line 48: The paywallconfig object is already passed when initializing so I am not sure it makes sense to change the signature

julien51 (changes requested):
I don't think you are taking the correct approach.
First please, re-usethe existing endpoint to manage the paywall config. No need to add new endpoints (or pleaase explain why!)
Then, I think you just need to update the `useCheckoutCommunication` hook to call these , without changing the signatures since the hook is already initialized wtih config.

unlock-app/src/components/interface/HooksNotifications.tsx
iMac7: Receives/deletes notifications related to checkout

unlock-app/src/components/interface/checkout/CheckoutContainer.tsx
iMac7: Trying to use it like the other notifications doesn't trigger the status event.
This is also how it's done somewhere else in the codebase, I could not find an alternative for it

packages/unlock-js/openapi.yml
iMac7: I added backend endpoints mainly to manage notifications. How should notifications work for this? :thinking:

iMac7: @julien51

julien51:
@iMac7 I would really love to close the loop on this one! Can you let me know if this is something you can do?

iMac7:
@julien51 yep I'll work on it.
You had asked earlier about notifications. Before I make some necessary changes, how would those work? (where they'll be stored and who should be notified when an event's data is sent)

iMac7:
sorry unavailable to complete this, up for claims

PR #15249, opened 2024-12-06
feature(locksmith): Decouple event creation from user connection state

### LINKED ISSUE #15243: Better handling of async event creation
We currently rely on the user being able to "witness" the lock deployment before creating an event. This is pretty fragile because the user's connection may break, or they may close the tab... etc, leading to an issue where the event is *not* created even though the lock has been deployed.
We need to handle this better!

0xTxbi:
# Description
This PR introduces a more reliable process for event creation by decoupling it from the user’s connection state during lock deployment. The previous implementation depended on the user witnessing the deployment, which was fragile in cases of disconnection or tab closure. The updated logic ensures events are consistently created regardless of the user’s connection status.

# Issues
Fixes #15243
Refs #15243

# Checklist:


locksmith/src/controllers/v2/eventsController.ts
```diff
@@ -73,8 +79,8 @@ export const saveEventDetails: RequestHandler = async (request, response) => {
     request.user!.walletAddress
   )
 
-  // This was a creation!
-  if (created) {
+  // Only send email if event is deployed
+  if (created && event.status === EventStatus.DEPLOYED) {
```
julien51, line 83: ok let's make sure the email is still sent _after_ it was deployed when this is happening later

julien51 (approved):

locksmith/src/controllers/v2/eventsController.ts
0xTxbi: sure thing! good catch!

PR #15540, opened 2025-02-17
fix(unlock-app): Optimize name resolution

0xTxbi:
# Description
This PR introduces improvements to how we resolve names across the dashboard.

# Issues
Fixes #
Refs #

# Checklist:


julien51 (approved):
Very good! I also wonder if we could/should cache things at the cloudflare worker level. wdyt?

0xTxbi:
> Very good! I also wonder if we could/should cache things at the cloudflare worker level. wdyt?

I think we should!

PR #15583, opened 2025-02-28
feature(provider): Rate limiting

### LINKED ISSUE #15559: Add rate limit to rpc
our RPC service needs to have a better rate-limit:
- limit by IP (exclude locksmith): add a limit in number of requests by seconds and by hour for each IP (excpet locksmith which should be unlimited)
- limit by contracts? Unlock contracts (Public Lock, Unlock, ... and others, but we will have to figure out which ones by trial and error) should NOT be rate limited, but others should be.

`Unlock` contracts adresses are coming from the networks package... but lock addresses cna be checked against thre `Unlock` contract with `locks` function.

0xTxbi:
# Description
This PR introduces rate limiting, and a multi-tier lock address caching system for our `provider` to improve performance, reduce blockchain calls, and prevent abuse.

## Rate Limiting
- IP-based limits (10 req/sec, 1000 req/hour)
- Exemptions for Locksmith IPs and Unlock contracts
- Configurable via environment variables

## Lock Address Caching
- In-memory cache (unlimited) for fast access
- Edge-distributed cache (24h TTL) for frequent lookups
- Persistent KV storage (1-year TTL) for long-term consistency
- Auto-initialization and non-blocking operations for efficiency

These changes improve response times, reduce RPC overhead, and ensure fair resource usage.

# Issues
Fixes #15559
Refs #15559

# Checklist:


provider/src/rateLimit.ts
```diff
@@ -0,0 +1,149 @@
...
+): Promise<boolean> => {
+  // Locksmith IPs are always allowed
+  if (isLocksmithIP(ip, env)) {
+    return true
+  }
+
+  try {
+    // Create a key that combines IP with contract address or method to provide granular rate limiting
+    // This is a more stable identifier than just IP alone, as recommended by Cloudflare
+    const rateKey = contractAddress
```
julien51, line 70: In our case I think it will be almost exclusively `POST` though,,, but not a a big deal!

provider/src/handler.ts
```diff
@@ -201,6 +207,48 @@ const handler = async (request: Request, env: Env): Promise<Response> => {
...
+        {
+          status: 429,
+          headers: {
+            ...headers,
+            'Retry-After': '60', // Suggest retry after 60 seconds
+          },
+        }
+      )
+    }
+  }
```
julien51, line 250: ok for now, in order to not break anything, let;'s not _actaully_ block things but add a log message so we cna see how that would perform. Once we have it running for 10 days or more we can go check at what would have actually been blocked and check that it works as expected. wdyt?

0xTxbi: sounds like a plan!

provider/src/rateLimit.ts
0xTxbi: got it!

provider/src/handler.ts
```diff
@@ -201,6 +207,54 @@ const handler = async (request: Request, env: Env): Promise<Response> => {
...
+      contractAddress,
+      env
+    )
+
+    if (!isRateLimitAllowed) {
+      // TEMPORARY: Log but don't block rate-limited requests for monitoring purposes
+      // After 10+ days, review logs and enable actual blocking
+      console.log(
+        `RATE_LIMIT_WOULD_BLOCK: IP=${getClientIP(request)}, Method=${body.method}, Contract=${contractAddress || 'none'}, ID=${body.id || 'none'}`
+      )
```
julien51, line 233: 💯

julien51: Actually, can we log the full body as well? This could be useful to debug more.

julien51: (insead of just the ID which is not very useful!)

julien51 (approved):
I am feeling a bit anxious about merging this in one go because it's pretty big and we don't have tests.

provider/src/handler.ts
0xTxbi: done!

julien51 (approved):

PR #15695, opened 2025-03-07
feature(unlock-app): `useAuthenticate` loading state

0xTxbi:
# Description
This PR introduces a loading state in the `useAuthenticate` hook. The loading state is true when Privy is not ready or when the user is authenticated but lacks a wallet address. It is false once Privy is ready and either the user is unauthenticated or has a wallet address. This improves state handling for authentication-dependent logic.

# Issues
Fixes #
Refs #

# Checklist:

### DIFF
--- a/unlock-app/src/hooks/useAuthenticate.ts
+++ b/unlock-app/src/hooks/useAuthenticate.ts
@@ -159,5 +159,6 @@ export function useAuthenticate() {
     signInWithPrivy,
     signOut,
     privyReady,
+    isLoading: !privyReady || (privyAuthenticated && !wallets[0]?.address),
   }
 }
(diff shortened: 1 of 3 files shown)


unlock-app/src/hooks/useAuthenticate.ts
```diff
@@ -159,5 +159,6 @@ export function useAuthenticate() {
     signInWithPrivy,
     signOut,
     privyReady,
+    isLoading: !privyReady || (privyAuthenticated && !wallets[0]?.address),
```
julien51, line 162: Should we also do that when there is a wallet BUT we don't have locksmith token ready?

julien51 (approved):

PR #15702, opened 2025-03-10
feature(provider): Optimize batch rpc request handling

### LINKED ISSUE #15680: Better handling of "batched" RPC requests
We need to "unbundle" batch RPC request and only send the ones that are appropruate to the underlying RPC endpoint.

For example, in our RPC provider we already handle "single" requests  for chainId locally, but we don't do it for "batched" requests that includes a single request.

0xTxbi:
# Description
This PR introduces logic to unbundle batch rpc requests and process each request independently. If a request can be handled locally (e.g., chainId, cached ENS/BaseName queries), it is processed immediately. Otherwise, requests are checked against rate limits. If none exceed the limit, they are recombined and forwarded to the RPC provider. Responses are then reassembled before returning to the client.

# Issues
Fixes #15680
Refs #15680

# Checklist:


provider/src/batchProcessor.ts
```diff
@@ -0,0 +1,212 @@
...
+export const processBatchRequests = async (
+  requests: RpcRequest[],
+  networkId: string,
+  originalRequest: Request,
+  env: Env
+): Promise<BatchProcessingResult> => {
+  // Process each request individually
+  const processedRequests: ProcessedRequest[] = []
+
+  for (const request of requests) {
```
julien51, line 138: I wonder if we could do a Promise.all here?

julien51: This would increase concurrency...

provider/src/batchProcessor.ts
```diff
@@ -0,0 +1,212 @@
...
+    processedRequests.push(processedRequest)
+  }
+
+  // Extract requests that need to be forwarded
+  const requestsToForward = processedRequests
+    .filter((processed) => processed.shouldForward)
+    .map((processed) => processed.request)
+
+  // Check if all requests are rate limited
+  const allRateLimited = processedRequests.every(
```
julien51, line 154: I am not sure I understand why we do that here?

provider/src/batchProcessor.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  if (providerResponses) {
+    providerResponses.forEach((response) => {
+      if (response && typeof response === 'object' && 'id' in response) {
+        responseMap.set(response.id, response)
+      }
+    })
+  }
+
+  // Combine the responses in the same order as the original batch
```
julien51, line 187: I am not sure this matters but I think it is[ good to have anyway!

provider/src/cache.ts
```diff
@@ -68,13 +69,65 @@ export const storeRPCResponseInCache = async (
   env: Env
 ): Promise<boolean> => {
   try {
-    // Check if this is a cacheable request
+    // Handle batch requests by storing each response individually
+    if (Array.isArray(body) && Array.isArray(json)) {
```
julien51, line 73: A better aporoach here is just to make sure this function is always called with an array?

julien51: so we mive the logic of handling batch or unique at a higher level?

provider/src/handler.ts
```diff
@@ -187,38 +191,36 @@ const handler = async (request: Request, env: Env): Promise<Response> => {
...
+    // Handle single request for chainId
+    if (
+      !Array.isArray(body) &&
+      body.method?.toLowerCase().trim() === 'eth_chainid'
+    ) {
+      return Response.json(processChainIdRequest(body, networkId), { headers })
+    }
 
-    // Handle both single requests and batch requests
+    // Convert single requests to batch format for uniform processing
```
julien51, line 202: YES!

provider/src/handler.ts
```diff
@@ -187,38 +191,36 @@ const handler = async (request: Request, env: Env): Promise<Response> => {
...
+      request,
+      env
+    )
 
-    // handle rate limiting
-    const rateLimit = await shouldRateLimit(request, env, body, networkId)
-    if (rateLimit) {
-      // TEMPORARY: Log but don't block rate-limited requests for monitoring purposes
-      // After 10+ days, review logs and enable actual blocking
+    // If all requests are rate limited, log but don't block
```
julien51, line 214: I am not sure we need that TBH. Probably better to deal with rate limits at the "individual" request level

provider/src/handler.ts
```diff
@@ -239,107 +241,123 @@ const handler = async (request: Request, env: Env): Promise<Response> => {
...
+      }
 
       // Store the response in the cache if applicable
-      await storeRPCResponseInCache(networkId, body, json, env)
+      await storeRPCResponseInCache(
+        networkId,
+        forwardBody,
+        providerResponse,
+        env
+      )
```
julien51, line 313: I don't think this should happen at this level. Caching should _only_ happen for requests that are being sent to the underlying provider so we should oribably handle this at a lower level IMO.

provider/src/batchProcessor.ts
0xTxbi: good catch! i agree! fixed!

provider/src/batchProcessor.ts
0xTxbi: was a flag to check at the handler-level if some/all requests in a batch were ratelimited. removed as it isn't necessary any longer

provider/src/handler.ts
```diff
@@ -187,38 +191,36 @@ const handler = async (request: Request, env: Env): Promise<Response> => {
       )
     }
 
-    const bodyAsString = JSON.stringify(body)
+    // Handle single request for chainId
```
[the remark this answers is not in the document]
0xTxbi: exactly!

provider/src/handler.ts
0xTxbi: sure!

provider/src/handler.ts
0xTxbi: got it!

julien51 (commented):
Per your comment, let's not worry about caching just yet.

provider/src/cache.ts
```diff
@@ -18,17 +18,17 @@ import { ContractType } from './types'
...
   networkId: string,
-  body: RpcRequest | RpcRequest[],
-  request: Request
+  request: RpcRequest,
+  originalRequest: Request
 ): Promise<Response | null> => {
   try {
     // Check if this is a cacheable request
-    if (!isRequestCacheable(body)) {
+    if (!isSingleRequestCacheable(request)) {
```
julien51, line 26: I am not sure I understand why we need to care about single requests differently here? In my mind, at this stage, every request is a "single request", no?

0xTxbi: remember, we're ignoring caching for the meantime. this will be handled in another pr

provider/src/requestProcessor.ts
```diff
@@ -0,0 +1,167 @@
...
+    console.log(
+      `RATE_LIMIT_WOULD_BLOCK: IP=${getClientIP(originalRequest)}, networkId=${networkId}, Request ID=${request.id}, Method=${request.method}`
+    )
+
+    return {
+      request,
+      response: null,
+      shouldForward: true,
+      rateLimited: true,
+    }
```
julien51, line 73: Can you add a comment that this should go away when we start enforcing rate limiting please?

0xTxbi: sure thing

provider/src/providerClient.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (requestsToForward.length === 0) {
+    return { responses: [] }
+  }
+
+  try {
+    // Determine if this is a single request or a batch
+    const isBatchRequest = requestsToForward.length > 1
+    const forwardBody = isBatchRequest
+      ? requestsToForward
+      : requestsToForward[0]
```
julien51, line 26: I am actually not sure we need to take care of this. Are providers not able to handle batches of 1?

provider/src/requestProcessor.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/**
+ * Combine the locally processed responses with the responses from the provider
+ *
+ * @param processedRequests The array of processed requests
+ * @param providerResponses The responses from the provider (if any)
+ * @returns An array of combined responses in the same order as the original batch
+ */
+export const combineResponses = (
```
julien51, line 127: I would probably move that to `batchProcessor` because that is where the logic to "split" lives but that's nbd

provider/src/providerClient.ts
0xTxbi: yeah actually, we don't need to

provider/src/cache.ts
```diff
@@ -18,17 +18,17 @@ import { ContractType } from './types'
  */
 export const getRPCResponseFromCache = async (
```
julien51, line 19: I think this file is not being called, so maybe let's undo the changes for now and come back to this later?

0xTxbi: sure. exactly what we discussed earlier :)

provider/src/providerClient.ts
julien51: ok then let's remove that (and the same rebundling logic below)

provider/src/requestProcessor.ts
0xTxbi: good catch. done!

julien51 (approved):

PR #15732, opened 2025-03-19
feature(provider): Name resolution caching

0xTxbi:
# Description
This PR introduces caching for ens and basenames resolution requests in the provider.

# Issues
Fixes #
Refs #

# Checklist:

### DIFF
--- a/provider/src/batchProcessor.ts
+++ b/provider/src/batchProcessor.ts
@@ -5,18 +5,22 @@ import {
 } from './errorHandlers'
 import { processBatchRequests } from './requestProcessor'
 import { forwardRequestsToProvider } from './providerClient'
-import { storeENSResponseInKV } from './cache'
+import { storeResponseInCache } from './cache'
 
 /**
  * Combine the locally processed responses with the responses from the provider
  *
  * @param processedRequests The array of processed requests
  * @param providerResponses The responses from the provider (if any)
+ * @param chainId The chain ID
+ * @param env The environment variables
  * @returns An array of combined responses in the same order as the original batch
  */
 export const combineResponses = (
   processedRequests: ProcessedRequest[],
-  providerResponses: any[] | null
+  providerResponses: any[] | null,
+  chainId: string,
+  env: any
 ): any[] => {
   // a map of provider responses by request ID for quick lookup
   const responseMap = new Map<number | string, any>()
@@ -38,6 +42,18 @@ export const combineResponses = (
 
     // Otherwise, look up the response from the provider
     const providerResponse = responseMap.get(processed.request.id)
+
+    // If provider response is found and the request is marked as cacheable, store it
+    if (providerResponse && processed.shouldCache) {
+      storeResponseInCache(
+        processed.request,
+        chainId,
+        providerResponse,
+        env
+      ).catch((error: Error) => {
+        console.error('Error caching response:', error)
+      })
+    }
 
     // If we couldn't find a matching response, return an error
     if (!providerResponse) {
@@ -61,14 +77,14 @@ export const combineResponses = (
  * This function handles the complete processing flow, including forwarding requests and combining responses
  *
  * @param body The original request body (single request or batch)
- * @param networkId The network ID
+ * @param chainId The chain ID
  * @param originalRequest The original HTTP request
  * @param env The environment variables
  * @returns A ProcessingResult containing the responses and any error information
  */
 export const processAndForwardRequests = async (
   body: RpcRequest | RpcRequest[],
-  networkId: string,
+  chainId: string,
   originalRequest: Request,
   env: Env
 ): Promise<ProcessingResult> => {
@@ -80,7 +96,7 @@ export const processAndForwardRequests = async (
     // Process the batch of requests
     const batchResult = await processBatchRequests(
       requests,
-      networkId,
+      chainId,
       originalRequest,
       env
     )
@@ -98,7 +114,7 @@ export const processAndForwardRequests = async (
       // Forward requests to the provider
       const forwardingResult = await forwardRequestsToProvider(
         batchResult.requestsToForward,
-        networkId,
+        chainId,
         env
       )
 
@@ -118,7 +134,9 @@ export const processAndForwardRequests = async (
         if (isBatchRequest) {
           const combinedResponses = combineResponses(
             batchResult.processedRequests,
-            [errorResponse]
+            [errorResponse],
+            chainId,
+            env
           )
 
           return {
@@ -146,32 +164,13 @@ export const processAndForwardRequests = async (
       // Get the provider responses
       const providerResponses = forwardingResult.responses || []
 
-      // Process ENS caching for all relevant responses
-      if (providerResponses.length > 0) {
-        // Find requests that should be cached
-        batchResult.processedRequests
-          .filter((req) => req.shouldCacheENS)
-          .forEach((req) => {
-            // Find the matching response for this request
-            const response = providerResponses.find(
-              (resp) => resp.id === req.request.id
-            )
-            if (response) {
-              // Don't await to avoid blocking the response
-              storeENSResponseInKV(req.request, response, networkId, env).catch(
-                (error: Error) => {
-                  console.error('Error caching ENS response:', error)
-                }
-              )
-            }
-          })
-      }
-
       // For batch requests, combine the local and provider responses
       if (isBatchRequest) {
         const combinedResponses = combineResponses(
           batchResult.processedRequests,
-          providerResponses
+          providerResponses,
+          chainId,
+          env
         )
 
         return {
@@ -199,7 +198,9 @@ export const processAndForwardRequests = async (
       if (isBatchRequest) {
         const combinedResponses = combineResponses(
           batchResult.processedRequests,
-          [errorResponse]
+          [errorResponse],
+          chainId,
+          env
         )
 
         return {
--- a/provider/src/batchProcessor.ts
+++ b/provider/src/batchProcessor.ts
@@ -5,7 +5,7 @@ import {
 } from './errorHandlers'
 import { processBatchRequests } from './requestProcessor'
 import { forwardRequestsToProvider } from './providerClient'
-import { storeResponseInCache } from './cache'
+import { storeResponseInCache, shouldStore } from './cache'
 
 /**
  * Combine the locally processed responses with the responses from the provider
@@ -43,8 +43,8 @@ export const combineResponses = (
     // Otherwise, look up the response from the provider
     const providerResponse = responseMap.get(processed.request.id)
 
-    // If provider response is found and the request is marked as cacheable, store it
-    if (providerResponse && processed.shouldCache) {
+    // If provider response is found and the request qualifies for caching, store it
+    if (providerResponse && shouldStore(processed.request, chainId)) {
       storeResponseInCache(
         processed.request,
         chainId,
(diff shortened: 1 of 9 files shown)


provider/src/requestProcessor.ts
```diff
@@ -50,6 +51,34 @@ export const processSingleRequest = async (
...
+      }
+    }
+
+    // If not in cache, mark for later caching
+    return {
+      request,
+      response: null,
+      shouldForward: true,
+      rateLimited: false,
+      shouldCacheENS: true,
```
julien51, line 78: I don't like that this is specifuc to ENS.

provider/src/batchProcessor.ts
```diff
@@ -146,6 +146,27 @@ export const processAndForwardRequests = async (
...
+            )
+            if (response) {
+              // Don't await to avoid blocking the response
+              storeENSResponseInKV(req.request, response, networkId, env).catch(
+                (error: Error) => {
+                  console.error('Error caching ENS response:', error)
+                }
+              )
+            }
+          })
```
julien51, line 167: This is not the right level to do that IMO. You are adding a lot of coupling

provider/src/requestProcessor.ts
julien51: It should just be `shouldCache` so we can easily add more caching later if needed

provider/src/requestProcessor.ts
```diff
@@ -50,6 +51,34 @@ export const processSingleRequest = async (
     }
   }
 
+  // Check if this is an ENS/Basename request that can be cached
```
julien51, line 54: Here you should just check the cache... no matter what kind of request that is

julien51:
```
const cached = await getCachedResponseForRequest(req, networkId)
if (cached) {
return {
      request,
      response: cached,
      shouldForward: false,
      rateLimited: false,
}
```

provider/src/batchProcessor.ts
```diff
@@ -5,6 +5,7 @@ import {
 } from './errorHandlers'
 import { processBatchRequests } from './requestProcessor'
 import { forwardRequestsToProvider } from './providerClient'
+import { storeENSResponseInKV } from './cache'
```
julien51, line 8: On line 41 below, you call 
`await storeResponseInCache(request, chainId, response)`

provider/src/cache.ts
```diff
@@ -151,3 +50,86 @@ export const storeContractStatusInKV = async (
     console.error('Error storing contract status in KV:', error)
   }
 }
+
```
julien51, line 53: function storeResponseInCache(request, chainId, response) {
if (shouldStore(request)) {
storeResponseInCache(request, chaindId, response)
}
}

julien51: function shouldStore(request) {
// Check if ENS, return true
// Check if Basename, return true

// return false
}

provider/src/cache.ts
```diff
@@ -151,3 +50,86 @@ export const storeContractStatusInKV = async (
...
+ * @param request The RPC request
+ * @param networkId The network ID
+ * @param env The environment variables
+ * @returns The cached response or null if not found
+ *
+ * First attempts to retrieve and parse as JSON directly. If that fails,
+ * falls back to retrieving as text and parsing manually. If the text
+ * parsing also fails, deletes the corrupted cache entry.
+ */
+export const getENSResponseFromKV = async (
```
julien51, line 65:
```suggestion
export const getResponseFromKV = async (
```

provider/src/cache.ts
```diff
@@ -151,3 +50,86 @@ export const storeContractStatusInKV = async (
...
+  request: RpcRequest,
+  networkId: string,
+  env: Env
+): Promise<any> => {
+  if (!env.NAME_RESOLUTION_CACHE) {
+    return null
+  }
+
+  const cacheKey = generateENSCacheKey(request, networkId)
```
julien51, line 74:
```suggestion
  const cacheKey = JSON.stringify(request)
```

provider/src/cache.ts
```diff
@@ -151,3 +50,86 @@ export const storeContractStatusInKV = async (
...
+}
+
+/**
+ * Cache an ENS/Basename response
+ * @param request The RPC request
+ * @param response The response to cache
+ * @param networkId The network ID
+ * @param env The environment variables
+ */
+export const storeENSResponseInKV = async (
```
julien51, line 111:
```suggestion
export const storeResponseInKV = async (
```

provider/src/batchProcessor.ts
julien51: In that funcyiom, make sure you dont store requestId and json2.0

provider/src/cache.ts
```diff
@@ -1,106 +1,5 @@
...
-    // Continue even if caching fails
-    return false
-  }
-}
-
-/**
- * Contract CACHING - KV STORAGE
- */
+import { ContractType, Env, RpcRequest } from './types'
+import { getKVContractTypeKey, generateENSCacheKey, getCacheTTL } from './utils'
```
julien51, line 2: You are going to add `getCachedResponseForRequest` here:
// first use shouldStore() to determine if we should even try to retrieve from the cache (if false, return)
// Then, use getResponseFromKV()
// If there is a reponse, rescontruct the json objec ( replace requestId and json2.0)

provider/src/batchProcessor.ts
```diff
@@ -38,6 +43,18 @@ export const combineResponses = (
     // Otherwise, look up the response from the provider
     const providerResponse = responseMap.get(processed.request.id)
 
+    // If provider response is found and the request is marked as cacheable, store it
+    if (providerResponse && processed.shouldCache) {
+      storeResponseInCache(
```
julien51, line 48: I see you are not "awaiting" here. It may be a good idea to not slow down the request but I want to make sure this is what you are trying to acheve!

provider/src/cache.ts
```diff
@@ -151,3 +55,151 @@ export const storeContractStatusInKV = async (
...
+    // Try to get and parse as JSON
+    const cached = await env.REQUEST_CACHE.get(cacheKey, 'json')
+    if (cached) {
+      return cached
+    }
+  } catch (error) {
+    // If JSON parsing fails, try simple text retrieval and parse it
+    try {
+      const rawText = await env.REQUEST_CACHE.get(cacheKey, 'text')
+      if (rawText) {
```
julien51, line 91: This is bad. We should decide to either store json OR store text and not mix things.

provider/src/cache.ts
```diff
@@ -151,3 +55,151 @@ export const storeContractStatusInKV = async (
...
+    return
+  }
+  const cacheKey = generateRequestCacheKey(request)
+  const ttl = getCacheTTL(env)
+
+  try {
+    const cacheValue = { result: response.result }
+    await env.REQUEST_CACHE.put(cacheKey, JSON.stringify(cacheValue), {
+      expirationTtl: ttl,
+    })
```
julien51, line 134: Can we just store json?

provider/src/cache.ts
```diff
@@ -151,3 +55,151 @@ export const storeContractStatusInKV = async (
...
+ * @param chainId The chain ID
+ * @returns True if the request should be cached, false otherwise
+ */
+export const shouldStore = (request: RpcRequest, chainId: string): boolean => {
+  if (!request.method) return false
+
+  // check if it's a name resolution request
+  if (
+    request.method.toLowerCase() === 'eth_call' &&
+    isNameResolutionRequest(request, chainId)
```
julien51, line 154: `request.method.toLowerCase() === 'eth_call'` shou;d be _inside_ `isNameResolutionRequest` for simplicity

provider/src/cache.ts
```diff
@@ -151,3 +55,151 @@ export const storeContractStatusInKV = async (
...
+ * @param response The response to cache
+ * @param env The environment variables
+ */
+export const storeResponseInCache = async (
+  request: RpcRequest,
+  chainId: string,
+  response: any,
+  env: Env
+): Promise<void> => {
+  if (!shouldStore(request, chainId) || !response?.result) {
```
julien51, line 201: Why would `response?.result` be falsy??

provider/src/requestProcessor.ts
```diff
@@ -36,37 +36,47 @@ export const processChainIdRequest = (
...
+    chainId
   )
 
   if (isRateLimited) {
     // Log the rate limit but still forward the request to maintain current behavior
     // This would later be changed to block rate-limited requests
-    console.log(
-      `RATE_LIMIT_WOULD_BLOCK: IP=${getClientIP(originalRequest)}, networkId=${networkId}, Request ID=${request.id}, Method=${request.method}`
-    )
-
```
julien51, line 67: If you remove this, remove the comment above!

provider/src/utils.ts
```diff
@@ -1,14 +1,44 @@
...
+  let paramsStr = ''
 
-// Cache API TTL in seconds (1 day)
-export const CACHE_API_TTL = 86400
+  try {
+    if (Array.isArray(request.params)) {
+      // Handle each parameter individually to ensure proper serialization
+      paramsStr = request.params
+        .map((param) => {
+          if (param === null) return 'null'
```
julien51, line 23: I don't think `param` can be null or an object?

julien51 (changes requested):
Your changes from `networkId` to `chainId` are making everything murkier :( It's too late for this one, but please refrain from doing that in the future!

provider/src/batchProcessor.ts
```diff
@@ -28,46 +33,62 @@ export const combineResponses = (
...
+          console.error('Error caching response:', error)
+        }
       }
-    }
 
-    return providerResponse
-  })
+      // If we couldn't find a matching response, return an error
+      if (!providerResponse) {
+        return {
```
julien51, line 63: Let's add a log message here because this should not happen.

provider/src/cache.ts
```diff
@@ -151,3 +82,129 @@ export const storeContractStatusInKV = async (
...
+    }
+  } catch (error) {
+    console.error(
+      `Error reading from cache: ${error instanceof Error ? error.message : String(error)}`
+    )
+    try {
+      await env.REQUEST_CACHE.delete(cacheKey)
+    } catch (err) {
+      console.error(`Error deleting corrupted cache entry: ${err}`)
+    }
```
julien51, line 119: good idea

provider/src/cache.ts
```diff
@@ -151,3 +82,129 @@ export const storeContractStatusInKV = async (
...
+): Promise<void> => {
+  if (!env.REQUEST_CACHE || !response?.result) {
+    return
+  }
+  const cacheKey = generateRequestCacheKey(request)
+  const ttl = getCacheTTL(env)
+
+  try {
+    const cacheValue = { result: response.result }
+    await env.REQUEST_CACHE.put(cacheKey, JSON.stringify(cacheValue), {
```
julien51, line 146: does this work? I assume if we can "Read" json, it should store json?
```suggestion
    await env.REQUEST_CACHE.put(cacheKey, cacheValue, {
```

julien51 (commented):
ok that strartrs to look good! Please test the hell out of this with real traffic locally?
I will deploy on cloudflare today!

provider/src/cache.ts
0xTxbi: no it doesn't. the `put` method only accepts strings, or an array buffer

0xTxbi:
> ok that strartrs to look good! Please test the hell out of this with real traffic locally? I will deploy on cloudflare today!

okay got it!

provider/src/cache.ts
0xTxbi: nope!

PR #15863, opened 2025-04-19
Update EventContent.tsx

ccarfi:
Change destination on EVENTS signup page to go to a new form created by @Calla-Ji instead of the current Hubspot lead capture form


ccarfi:
@Calla-Ji i put in a pull request to change the event landing page to go to your new form instead of hubspot

julien51:
@ccarfi Unfortunately this won't work since this does not redirect back to the app once the form is complete (unlike the hubspot form...). I think I will just replace with a regular link to the app for now.

ccarfi:
Screenshot 2025-04-19 at 8 14 01 PM

asked @Calla-Ji to add that capability to the form when i set this up this morning :)

PR #15874, opened 2025-04-22
fix(unlokck-app): minor changes

julien51:
# Description

Minor UI fixes and improvements.

# Checklist:


unlock-app/src/components/content/event/Registration/WalletlessRegistration.tsx
```diff
@@ -373,6 +373,24 @@ export const RegistrationForm = ({
         </>
       )}
 
+      <Controller
```
julien51, line 376: Move to the top so that it does not look too awkward when there is an extra checkbox!

unlock-app/src/components/interface/locks/CheckoutUrl/elements/LocksForm.tsx
```diff
@@ -300,6 +300,17 @@ export const LocksForm = ({
     return orderedLocks
   })
 
+  useEffect(() => {
```
julien51, line 303: This is required because the locks may be missing when we load the component for the first time

PR #16174, opened 2025-09-06
Add Swap ETH to USDC and transfer proposal

blahkheart:
# Description

This proposal is a template for making proposals that swap ETH to USDC using Uniswap V3 and make a transfer of a specific `amount` of USDC to a specified address. This was created to facilitate the funding of Unlock DAO's partnership with Rise Up Morning Show.

# Issues

Fixes #
Refs #

# Checklist:


blahkheart:
@clemsos @julien51 @0xTxbi Please review

julien51 (changes requested):
Hey @blahkheart The code looks good but I am worried that the proposal may fail because the ETH/USDC price will evolve over the days that the proposal is "in progress". 
Does it make sense?

I think the only way to go thru here is to specify either an USDC amount we need (probably better) or a ETH amount we want to spend, but we can have fixed values for both

blahkheart:
> Hey @blahkheart The code looks good but I am worried that the proposal may fail because the ETH/USDC price will evolve over the days that the proposal is "in progress". Does it make sense?
> 
> I think the only way to go thru here is to specify either an USDC amount we need (probably better) or a ETH amount we want to spend, but we can have fixed values for both

@julien51 I've made an update to address this concern. It uses the exact output method to specify the amount of USDC needed and a max ETH amount to be spent, to effectively address both, please review.

PR #16312, opened 2026-03-09
Grey out expired memberships #7294

### LINKED ISSUE #7294: Grey out expired memberships
By default we generate a unique image for each lock which is used by NFT.
We should be able to customize each NFT by showing it more or less greyed out based on how close it is from being expired!

9alekc1:
Description
This PR updates Locksmith’s default membership NFT image generation so generated lock icons can reflect key expiration state.

Specifically:
Generated key metadata now uses an explicit networked image URL: /${network}/lock/${address}/icon?id=${keyId}
Locksmith computes grayscale from the key expiration and the lock expiration duration
The grayscale effect applies only to autogenerated default icons
Custom lock images and custom key icons are left unchanged

Issues
Fixes #7294

Release Note Draft Snippet
Locksmith now generates network-specific default NFT image URLs for keys and can render autogenerated membership icons with expiration-based grayscale. Custom lock and key images are unchanged.

Active and expired (greyed-out) subscriptions:
Screenshot 2026-03-10 001211


julien51 (commented):
Good implementation overall — the grayscale SVG filter approach is clean and the logic for computing saturation from remaining duration is sound. A few things worth addressing:

**Potential issue: performance**
`getKeyIconGrayscale` makes two async calls (keyData + web3Service) on every icon request. For pages displaying many keys this could be slow. Consider caching the result or computing it lazily.

**Missing test coverage**
The existing test is updated but there's no test for `getKeyIconGrayscale` itself — especially edge cases like:
- Key with `expirationDuration === MaxUint256` (non-expiring lock) → should return 0
- Already-expired key → should return 1
- Key expiring imminently

**Minor: breaking API change**
`lockIcon(address, options)` now requires `options` (not optional). All callers are updated in this PR which is good, but worth noting in case other consumers exist outside this repo.

**Network in image URL**
The `defaultMappings` fix to include `network` in the icon URL path is a good correctness fix — was this causing issues in prod?

CI passes. Approve once the missing test coverage is added.

PR #16325, opened 2026-03-15
feat: add governance proposal write flows

julien51:
## Summary
- add connected-wallet voting on proposal detail pages with snapshot voting power
- wire queue and execute actions to the governor using the proposal payload already loaded on the page
- gate write actions behind Privy configuration and Base network wallet switching

## Testing
- yarn workspace @unlock-protocol/governance-app lint .
- yarn workspace @unlock-protocol/governance-app build

### DIFF
--- a/governance-app/app/proposals/[id]/page.tsx
+++ b/governance-app/app/proposals/[id]/page.tsx
@@ -72,7 +72,20 @@ export default async function ProposalDetailPage({
                 {proposal.title}
               </h2>
               <div className="prose prose-sm text-brand-ui-primary/72">
-                <ReactMarkdown>{proposal.description}</ReactMarkdown>
+                <ReactMarkdown
+                  components={{
+                    // Proposal descriptions are user-controlled (on-chain). Open
+                    // links in a new tab with noopener noreferrer to prevent
+                    // tab-napping and mitigate phishing via crafted hrefs.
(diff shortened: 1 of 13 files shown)


julien51:
Addressed all code review feedback:
- Removed `actionMutation.isSuccess` from the lifecycle button's `disabled` condition — it persisted across `router.refresh()` and permanently disabled Execute after Queue
- Added `[0,1,2]` bounds check on `vote.support` from subgraph
- Switched `|| 0n` to `?? 0n` for votingPower fallback (0n is falsy)
- Replaced `(error as any).code === 'ACTION_REJECTED'` with ethers v6 `isError(error, 'ACTION_REJECTED')`
- Strip verbose RPC parenthetical data from error messages shown to users

julien51:
Addressed all issues from review #4027276505:

- **#anchor links**: Added `href?.startsWith('#')` to the safe-link check so anchor links (e.g. `#motivation`) render as links instead of plain text
- **TruncatedId label prop**: Added optional `label` prop (default `'Copy full ID'`); all current callers pass `'Copy full proposal ID'` explicitly — future reuse for tx hashes/addresses will get a correct label automatically
- **setInterval cleanup**: Interval now early-exits when `canExecute` is already true (ETA has passed), preventing indefinite 15s ticks

PR #16326, opened 2026-03-15
feat: add governance proposal composer

julien51:
## Summary
- replace the /propose placeholder with a working composer for governor proposals
- support a guided known-contract builder plus advanced JSON submission mode
- show current voting power versus proposal threshold before submission

## Testing
- yarn workspace @unlock-protocol/governance-app lint .
- yarn workspace @unlock-protocol/governance-app build


julien51:
Addressed latest claude-review feedback:

- **Pre-sign confirmation step**: clicking "Submit proposal" now builds and validates the payload first, showing targets/values/description in a review panel before the wallet prompt fires
- **`advancedJson` reset on success**: `setAdvancedJson(defaultAdvancedJson)` added to `onSuccess`
- **Stale threshold**: added `staleTime: 0` to the threshold query
- **`tx.wait()` receipt**: captured `tx.hash` and included a Basescan explorer link in the success toast
- **`PublicLock` removed from `knownContracts`**: every Lock is a separate instance; users must use "Custom contract" with a specific address
- **Advanced mode `hasRequiredFields`**: now validates that the JSON is parseable before enabling the submit button

julien51:
Addressed latest round:

- **Threshold loading/error states**: message now shows "Checking your voting power…" while pending, "Could not load voting power. Check your RPC connection." on error, and only shows the threshold pass/fail message once data has loaded. `meetsThreshold` gated on `thresholdQuery.isSuccess` to prevent false negatives.
- **Receipt status check**: captured receipt from `tx.wait()` and throws `'Transaction reverted on-chain.'` if `receipt.status === 0`.

julien51:
Addressed latest round:

- **Null receipt guard**: added `if (!receipt) throw new Error('Transaction was dropped or replaced before confirmation.')` before the status check
- **Non-clickable toast URL**: removed URL from toast string; now calls `window.open(explorerLink, '_blank', 'noopener,noreferrer')` after success
- **Duplicated `upTokenAddress`**: extracted to a module-level `const upTokenAddress` shared by `tokenAddress` and the UPToken `knownContracts` entry

julien51:
Addressed latest round:

- **`staleTime`**: changed from `0` to `60_000` (1 min) to avoid refetching on every focus/remount
- **`UPTimelock` address**: tried `base.dao?.timelock` but the `NetworkConfig.dao` type doesn't expose `timelock` yet — kept the hardcoded address and added a comment to update it once the networks package adds the field
- **Stable `CallEditor` key**: added `id: crypto.randomUUID()` to `CallDraft` via `createEmptyCall()`; key is now `call.id`; `id` is preserved across kind switches (known↔custom)

julien51:
Addressed latest round:

- **Explorer link**: replaced `window.open` (blocked by popup blockers in async context) with an in-UI success banner showing a clickable Basescan link after submission
- **Stale threshold**: `thresholdQuery.refetch()` is now awaited before building the payload, so the threshold check reflects current voting power at submission time

julien51:
Addressed latest round:

- **Stale `meetsThreshold`**: now reads the refetch result directly — `const result = await thresholdQuery.refetch()` — and gates on `result.data.votingPower >= result.data.proposalThreshold` before building the payload
- **`receipt.hash`**: returning `receipt.hash` instead of `tx.hash` so the Basescan link points to the actually-mined transaction hash (handles EIP-1559 replacements)
- **Call validation in `hasRequiredFields`**: added `calls.every(c => c.functionName !== '')` so the Submit button stays disabled until all calls have a function selected
- **Double `txExplorerUrl` call**: extracted to a `SubmittedBanner` component that computes the link once

julien51:
Addressed latest round:

- **Calldatas in review panel**: added a Calldatas section showing the 4-byte selector + last 8 chars (e.g. `0xabcd1234…ef012345`) for each call
- **Args completeness gate**: `hasRequiredFields` now also checks `c.args.every(a => a.trim() !== '')` so the Submit button stays disabled until all function arguments are filled
- **Runtime field validation**: added explicit `typeof call.contractAddress !== 'string'` and `typeof call.functionName !== 'string'` checks in `buildAdvancedProposalPayload` for clearer error messages before calling ethers
- **`knownContracts[0]` fallback**: changed to `governanceConfig.knownContracts[0]?.label ?? ''`

julien51:
Addressed latest round:

- **Full calldata in review panel**: replaced the truncated preview with a scrollable (`max-h-24 overflow-y-auto`) break-all monospace block showing the complete calldata hex — users can verify the full encoding before signing
- **Custom address guard in `hasRequiredFields`**: added `(c.kind !== 'custom' || isAddress(c.customAddress))` — Submit button stays disabled until the custom contract address is a valid EVM address

julien51:
Addressed latest round:

- **Empty string args**: removed `c.args.every(a => a.trim() !== '')` from `hasRequiredFields` — `string`-type args may intentionally be empty; `parseArgument` validates values per-type at build time
- **Submit button loading state**: added `loading={thresholdQuery.isPending && !thresholdQuery.isError}` and `disabled` when `thresholdQuery.isPending`
- **Retry button**: added a "Retry" button calling `thresholdQuery.refetch()` when `thresholdQuery.isError`
- **Empty-address contracts filtered from dropdown**: `contractChoices` now filters `knownContracts` to only entries with a non-empty address
- **Non-null assertions**: replaced `thresholdState!` with `thresholdState?.votingPower ?? 0n` and `thresholdState?.proposalThreshold ?? 0n`

julien51:
Addressed latest round:

- **`getFunctionChoices` state-mutability filter (blocker)**: now filters to only `nonpayable` and `payable` fragments — `view`/`pure` functions are no longer offered in the dropdown
- **Empty calls guard**: added `if (calls.length === 0) throw` in `buildSimpleProposalPayload` to guard the latent edge case
- **Explorer link safety comment**: added a comment in `SubmittedBanner` explaining that `txExplorerUrl` validates the 32-byte hex pattern before constructing the URL, so the href is safe from open-redirect/XSS

julien51:
Addressed latest round:

- **Custom ABI parse error gates Submit button**: `hasRequiredFields` now checks `!('error' in parseCustomAbi(c.customAbi))` for custom-kind calls
- **Double-submit prevention**: Submit button now disabled when `thresholdQuery.isFetching`, not just `isPending`
- **`functionArgs` array validation**: added explicit `!Array.isArray(call.functionArgs)` check in `buildAdvancedProposalPayload` for a clear error rather than a cryptic ethers throw
- **Stale-threshold-on-refetch-failure comment**: added inline note explaining React Query returns last cached data on error and the governor enforces on-chain anyway
- **Unit tests**: extracted all pure functions (`buildSimpleProposalPayload`, `buildAdvancedProposalPayload`, `parseArgument`, helpers) to `src/lib/governance/composer.ts` (no React deps), added `vitest 3.0.9` + 30 unit tests covering argument parsing, payload building, and error paths; `yarn test` added to the `ci` script

julien51:
Fixed Vercel deployment: added vitest to yarn.lock (was missing, causing lockfile-would-be-modified error on Vercel).

julien51:
All issues from the Claude review feedback have been addressed across the fix commits since 2026-03-29. Summary of what was fixed:

- **knownContractAddresses drift** — now derived dynamically from `governanceConfig.knownContracts` instead of a parallel hardcoded map (includes PublicLock automatically)
- **Fixed-size arrays in parseArgument** — regex `/^(.+)\[\d*\]$/` handles both `T[]` and `T[N]`
- **Advanced mode description body** — `description` field added to `AdvancedProposal` type; uses same `title\n\nbody` format as simple mode
- **parseCustomAbi swallowing errors** — returns `{ error: string }` discriminated union instead of silently returning null
- **Double success toast** — single toast fires in `onSuccess` after `tx.wait()`
- **Null receipt** — explicit check throws before treating dropped tx as success
- **Stale threshold on submit** — refetches and reads fresh result before building payload
- **view/pure functions in function picker** — filtered by stateMutability
- **Misleading error for custom contracts** — error message now distinguishes custom vs known contract
- **parseAdvancedProposal non-object JSON** — added in this push: guard against `null`/array/primitive JSON with a clear error message + tests (31 tests total, all passing)

julien51:
Addressed the issues from the 2026-04-11 review:

1. **uint negative value validation** — split `uint`/`int` branches in `parseArgument`; `uint*` types now reject negative values with a clear error instead of passing `-1n` through to ethers.
2. **Fixed-size array length validation** — `arrayMatch[2]` (the declared size) is now checked against the parsed array length; e.g. `uint256[3]` with `[1,2]` now throws `requires exactly 3 elements, got 2`.

34 tests, all green. Issue 3 (resolveContractAbi duplication) is intentional — `rpc.ts` imports from `react`, so `composer.ts` keeps its own copy to remain React-free and testable in Node.

julien51:
Addressed the 2026-04-11T16:24 review:

1. **Integer precision loss in advanced mode** — added `assertNoNumberInIntPosition` helper in `composer.ts` that walks `fragment.inputs` and rejects any plain JS `number` in a `uint*`/`int*` position (including arrays thereof) with a clear message: *"use \"1000000\" instead of 1000000"*. Test added.

2. **Uncaught exception in submit handler** — moved `thresholdQuery.refetch()` inside the outer try/catch so any synchronous provider error surfaces as a toast rather than propagating unhandled.

**Issue 3 (resolveContractAbi duplication)** — kept as-is. `rpc.ts` imports from `react` (`cache`), so `composer.ts` can't import from it without breaking its no-React-deps contract (needed for Node/test environments).

35 tests, all green.

julien51:
Addressed the 2026-04-12T14:48 review:

1. **bytes odd-length hex** — tightened regex to `^0x([0-9a-fA-F]{2})*$` requiring pairs of hex chars; `0xabc` now throws *"even number of hex chars"* instead of an opaque ethers error. Test added.

2. **thresholdQuery stale data on RPC failure** — now checks `result.status === 'error'` before reading `result.data`, so a failed refetch shows *"Could not verify voting power"* rather than silently passing through cached data.

**Minor (advanced mode submit button on invalid JSON)** — left as-is; the build-time validation produces a clear toast error and the button's disabled state already covers the most common mis-state (JSON parse error). Adding structural validation to `hasRequiredFields` would require duplicating `buildAdvancedProposalPayload` logic.

**Nit (customAbi '[]' default)** — left as-is per review's "Very low severity" assessment.

36 tests, all green.

julien51:
Addressed the 2026-04-12T15:05 review:

1. **Mode switch clears pendingPayload** — ModeButton onClick handlers now call `setPendingPayload(null)`, preventing the stale simple-mode payload from appearing in the review panel after switching to advanced.

2. **call.value precision guard** — added `typeof call.value !== 'string'` check in `buildAdvancedProposalPayload` before passing to `BigInt()`, matching the `functionArgs` guard. Throws with a clear *"use quoted string"* message. Test added.

**Minor (bit-width validation for uintN/intN)** — deferred; the ethers error on overflow is at least attributable to the encoding step, and adding full bit-width validation per type would require parsing the N from 32 bit-width variants across uint8…uint256. Can be a follow-up.

37 tests, all green.

julien51:
Addressed the 2026-04-12T17:06 review:

1. **Tuple precision guard** — refactored `assertNoNumberInIntPosition` to accept `ParamType` (from ethers) instead of a string. Now recurses into `param.components` for tuples (both positional array and named object encoding) and `param.arrayChildren` for nested arrays. Test added (tuple with number `amount` → throws `precision loss`).

2. **functionArgs length check** — added `args.length < fragment.inputs.length` guard before `encodeFunctionData`; throws `requires N arguments, got M` instead of passing `undefined` to ethers. Test added.

3. **hasRequiredFields structural check** — advanced mode now verifies `proposalName` is a non-empty string and `calls` is a non-empty array before enabling the Submit button, not just JSON parseability.

4. **Stale functionName on ABI edit** — `customAbi` `onChange` now also resets `functionName: ''` and `args: []`, same as the contract-select handler.

**Minor nit (tx.wait() timeout)** — left for now; a dropped tx would require the user to refresh regardless, and the governor's on-chain revert is the authoritative gate.

39 tests, all green.

julien51:
Addressed the 2026-04-12T17:47 review:

1. **Precision guard in simple mode tuples** — added `assertNoNumberInIntPosition` call on `parsedArgs` in `prepareSimpleCall` after the `parseArgument` map. Tuple inputs use `JSON.parse` internally, so this closes the gap where `["0x...", 1000000000000000001]` would silently encode the wrong value.

2. **Extra-args check** — added `args.length > fragment.inputs.length` guard in advanced mode that throws *"extra arguments are silently dropped by ethers"* rather than silently truncating.

3. **Misleading receipt.hash comment** — updated to *"canonical on-chain hash from the mined receipt"* (ethers v6 rejects with TRANSACTION_REPLACED, not a different hash).

4. **useMemo for hasRequiredFields** — wrapped in `useMemo([mode, title, description, calls, advancedJson])` to avoid JSON.parse on every keystroke.

39 tests, all green.

julien51:
Addressed the 2026-04-12T23:33 review:

1. **Negative ETH value** — added `ethValue < 0n` check after `BigInt(call.value)` in both `prepareSimpleCall` and `buildAdvancedProposalPayload`, matching the guard already in `parseArgument` for uint types. Test added.

2. **Anonymous tuple components** — removed the `if (component.name)` guard in the named-object tuple branch of `assertNoNumberInIntPosition`; now always recurses using the index as fallback key for unnamed components.

**Nit (PR description)** — will update below.

40 tests, all green.

julien51:
Addressed review PRR_kwDOBrdp0M70IgUb (2026-04-12T23:53):

1. **Bit-width validation** — `parseArgument` now checks fixed-width integer ranges: `uint8` rejects `> 255`, `int8` rejects outside `[-128, 127]`, etc. Error names the type and the valid range. Tests added.

2. **`value: 0` precision bypass** — moved `typeof call.value !== 'string'` check *before* the truthy guard in `buildAdvancedProposalPayload`; numeric `0` (falsy) now hits the same precision-loss error as non-zero numbers. Test added.

3. **Distinct bytes error messages** — split into two: missing `0x` prefix → *"must start with 0x"*; odd-length hex → *"must have an even number of hex chars after 0x"*.

4. **Custom-contract path tests** — added three tests to `buildSimpleProposalPayload`: valid custom ABI + address produces correct calldata; invalid ABI throws `Custom ABI is invalid`; non-address throws `valid custom contract address`.

46 tests, all green.

julien51:
Addressed PRR_kwDOBrdp0M70Il-V:

1. **Timelock address dedup** — extracted into `const timelockAddress` in `governance.ts`; both `timelockAddress` and the `UPTimelock` entry in `knownContracts` now reference the same const.

2. **tx.wait() timeout** — wrapped in `Promise.race` with a 5-minute timeout that rejects with *"Transaction not mined after 5 minutes — check Basescan for status."*

3. **Advanced hasRequiredFields functionName gate** — added `p.calls.every(c => c.functionName?.trim() !== '')` so the Submit button stays disabled with the default template's empty `functionName: ''`.

4. **Bare uint/int range check** — changed regex from `/^uint(\d+)$/` to `/^uint(\d*)$/` (same for int); bare `uint`/`int` now treated as 256-bit and subject to the same overflow check as sized variants.

46 tests, all green.

julien51:
Addressed all issues from the latest review (PRR_kwDOBrdp0M70Io17):

1. **Bug fixed**: Added `typeof parsed.proposalName !== 'string'` guard before `.trim()` to produce a clean validation error instead of `TypeError` when `proposalName` is a non-string value (e.g. `42`).

2. **Fail-loud on missing addresses**: `upTokenAddress` and `governorAddress` now throw at module load if not provided by `@unlock-protocol/networks`, rather than silently falling back to hardcoded literals. `knownContracts` uses the validated `governorAddress` constant.

3. **Distinct error for missing `calls`**: `calls must be an array` vs `At least one advanced call is required` are now separate error messages.

4. **Comment added**: Tuple passthrough in `parseArgument` now has a comment clarifying that `assertNoNumberInIntPosition` runs post-parse in `prepareSimpleCall`.

Two new tests added (48 total, all green): non-string `proposalName` and missing `calls` key.

julien51:
Addressed issues from review 4099412409:

1. **Governance fallback**: Reverted from throw-at-startup to warn-and-fallback — `console.error` logs the misconfiguration (visible in server/browser logs and CI output) but doesn't crash the app on startup if `@unlock-protocol/networks` accidentally drops `base.dao.governor` or the UP token address.

2. **Zombie `tx.wait()`**: `setSubmittedTxHash(tx.hash)` is now called immediately after `governor.propose()` returns, before `Promise.race`. If the 5-minute timeout fires, the mutation ends in error state but `SubmittedBanner` still renders with the Basescan link so the user can check on-chain status.

3. **`-0` comment**: Added notes to both `buildAdvancedProposalPayload` and `prepareSimpleCall` explaining `BigInt('-0') === 0n` so the missing negative guard is obviously safe.

julien51:
Addressed all issues from review 4101291971:

1. **Submit button UX**: Removed `thresholdQuery.isFetching` from `disabled` and `loading` — only `isPending` (initial load) now blocks the button. Background refetches no longer gray it out.

2. **Misleading value error**: Changed "to avoid precision loss" to "must be a quoted string — use `\"X\"` instead of the number X" since `value: 0` has no precision risk; it's rejected purely for consistent typing.

3. **Missing named-tuple field**: Added `if (fieldValue === undefined) throw new Error(...)` guard in `assertNoNumberInIntPosition` named-object branch, with a clear message like `Missing field "amount" in tuple argument at "params"`. New test covers this case.

4. **`setSubmittedTxHash` in mutationFn**: Replaced with `pendingTxHashRef = useRef<string | null>(null)`. `mutationFn` writes `pendingTxHashRef.current = tx.hash`, `onError` flushes it to state (so Basescan link appears even after timeout), `onSuccess` clears the ref.

julien51:
Addressed all issues from review 4101855787:

1. **Precision loss in `uint[]` simple mode**: Added check in `parseArgument`'s array branch — if a parsed element is a JS `number` and the child type starts with `uint`/`int`, throws `"must be a quoted string to avoid precision loss"`. Updated existing `uint256[3]` test to use `["1","2","3"]` (quoted strings), added 2 new tests (small int and > `MAX_SAFE_INTEGER`). 51 tests total.

2. **Incorrect `-0` comment**: Fixed in both `buildAdvancedProposalPayload` and `prepareSimpleCall` — `'-0'` is truthy (non-empty string) so it enters the block; `BigInt('-0') === 0n` so `0n < 0n` is false and no error fires.

3. **`getContractAbi` deduplication**: Renamed `resolveContractAbi` → `getContractAbi` in `composer.ts` and exported it. `rpc.ts` now imports from `composer.ts` and re-exports (preserving its public API). `proposals.ts` also imports from `composer.ts` and its local copy is removed.

4. **`calls.length > 0` in `hasRequiredFields`**: Added explicit check for simple mode so an empty calls array doesn't vacuously enable the submit button.

julien51:
Addressed all issues from review 4102108479:

1. **Silent extra-arg truncation in simple mode**: Added `if (call.args.length > fragment.inputs.length) throw` check in `prepareSimpleCall`, matching the advanced mode guard. New test covers this.

2. **Positional tuple length guard**: Added `if (element === undefined) throw new Error('Missing element at index N...')` in the positional-array branch of `assertNoNumberInIntPosition`, mirroring the named-field check. New test covers this.

3. **Redundant `rpc.ts` re-export**: Removed `export { getContractAbi }` from `rpc.ts` — nothing imports it from there (both `proposals.ts` and `composer.ts` own the canonical import).

4. **Tuple path warning comment**: Expanded the comment to `WARNING: this function does NOT run assertNoNumberInIntPosition — that guard lives in the caller (prepareSimpleCall)`.

5. **Redundant `thresholdState !== undefined`**: Removed — `thresholdQuery.isSuccess` already guarantees data is defined.

julien51:
Fix: Restored  guard —  guarantees data is defined logically but TypeScript doesn't narrow through the separate const assignment, causing a build failure. Using  keeps both TypeScript and the reviewer happy.

PR #16339, opened 2026-03-20
ci: speed up GitHub Actions with Docker layer caching and change detection

julien51:
Four targeted CI optimizations:\n\n1. Docker deps GHA layer caching\n2. GHA cache warmed in parallel with tests\n3. Yarn cache for networks workflow\n4. Skip unchanged Vercel deploys\n\nUpdated description separately via comment.


julien51:
Addressed the review feedback:

**PR description fixed**: Optimization #2 now correctly describes what happens — the GHA cache is warmed by the first matrix job in parallel with tests, but the Docker Hub push (`cache_deps`) still gates on `needs: run-all-tests` to avoid tagging `unlock-dev:latest` with a broken build.

**`paywall-app` test support confirmed**: Verified before merging — `docker/docker-compose.ci.yml` has a `paywall-app` service, and `paywall-app/package.json` has a `"ci": "yarn test && yarn lint && yarn build"` script. `scripts/tests.sh paywall-app` is a valid invocation.

**Redundant `!= '[]'` check removed** (d17e5e2c0): The `deploy-unlock-protocol-com` and `deploy-paywall-app` conditions no longer include the extra `needs.check-changes.outputs.changed != '[]'` guard, matching the pattern used by the other deploy jobs.

julien51:
Responses to the latest review:

**Race condition / misleading framing**: Correct — sibling matrix jobs in the same run will fall back to Docker Hub on a cold cache, and benefit from GHA on subsequent runs. The "warmed in parallel" framing refers to the GHA cache being populated during the test run rather than in a separate sequential post-test step. Will accept this limitation; it's the best we can do with parallel matrix and GHA's eventual-consistency semantics.

**Boolean type coercion**: GHA `workflow_call` boolean inputs handle this correctly — the expression `github.event.pull_request.head.repo.full_name == github.repository` evaluates to a boolean that maps cleanly to the `type: boolean` input. The `if: ${{ !inputs.export_cache }}` condition is safe.

**Duplicate Docker setup overhead**: Known trade-off. The buildx + build-push-action steps run on all matrix jobs to ensure the deps image is available locally. The Docker Hub fallback (`type=registry`) covers the cold-cache case, so no build failures.

**`.worktrees/` gitignore**: Intentional — this is Claude Code's git worktrees directory pattern (per project conventions), not related to `provider/.wrangler/.worktrees`.

julien51:
Re: the latest review (4259678854) —

**BUILDKIT_INLINE_CACHE=1 in _docker-cache.yml**: Already present. The Build image step in `_docker-cache.yml` includes `build-args: BUILDKIT_INLINE_CACHE=1` (added in the original PR commit), so the Docker Hub push already embeds inline cache metadata. The registry fallback in `_tests.yml` (`type=registry,ref=unlockprotocol/unlock-dev:latest`) should pick it up correctly.

**Docker setup overhead on lightweight targets**: Known trade-off, accepted.

No further changes needed — all three correctness concerns are either already handled or not blocking.

PR #16346, opened 2026-05-01
fix(provider): add public RPC fallbacks for Alchemy CDN SSL failures

julien51:
## Summary

- Alchemy's CDN returns HTTP 525 (SSL Handshake Failed) for 5 chains when called from Cloudflare Workers: Avalanche (43114), BSC (56), zkSync (324), Celo (42220), Scroll (534352)
- Root cause: Alchemy's Cloudflare CDN fails SSL handshake with their own origin servers for these chain backends — not reproducible from non-Cloudflare infrastructure (works fine locally)
- Fix: fall back to the official public RPC endpoint (already listed as \`publicProvider\` in \`@unlock-protocol/networks\`) when the primary provider returns a non-2xx response
- Working chains (ETH, Polygon, Arbitrum, Base, etc.) continue using Alchemy exclusively

Additional changes:
- \`wrangler.toml\`: upgrades \`compatibility_date\` from \`2022-08-17\` → \`2024-09-23\` and adds \`global_fetch_strictly_public\` flag
- \`set-env-vars.sh\`: fails fast with a clear error if any required env var is missing (previously only checked \`LOCKSMITH_SECRET_KEY\`)
- \`package.json\`: \`deploy\` script now runs \`set-env-vars\` before \`wrangler deploy\` to ensure secrets are always current

## Test plan

- [ ] \`npx vitest run\` passes — new \`providerClient.test.ts\` covers fallback logic (primary 525 → fallback success, primary failure with no fallback → error, both fail → error)
- [ ] All 15 chains respond correctly at \`https://rpc.unlock-protocol.com/<chainId>\` (verified in production)
- [ ] https://app.unlock-protocol.com/locks no longer shows RPC errors for Avalanche/BSC/zkSync/Celo/Scroll


julien51:
Addressed review feedback:

- **Fallback now scoped to 5xx only** — 4xx errors (429, 401, 403) surface directly to the caller rather than silently rerouting to public RPCs. Added test to verify this.
- **Null guard added** for unsupported network IDs with a clear error message.
- **ABOUTME comments removed** from source and test files.
- **`global_fetch_strictly_public` context**: this flag prevents Cloudflare from routing outbound `fetch()` calls through its own proxy infrastructure. Without it, Cloudflare Workers route external requests through the same CDN layer that causes the 525 SSL handshake failures on certain Alchemy backends. It is directly related to the fix.

julien51:
Addressed the status propagation issue flagged in the latest review:

- **Propagate upstream HTTP status to callers**: `ForwardingResult.error` now carries a `status` field. `providerClient.ts` sets it from `response.status` when the provider returns a non-2xx. `batchProcessor.ts` forwards it to `ProcessingResult.error.status` instead of hardcoding `500`. A provider 429 now correctly reaches the caller as HTTP 429.

All 39 unit tests pass locally.

julien51:
Addressed the 429 fallback behavior flagged in the latest review:

- **429 now retries fallbacks**: A `429 Too Many Requests` is provider-specific (that provider is throttled), not a request-level error. The fallback loop now treats 429 the same as 5xx — it continues to the next fallback. Only non-retryable 4xx errors (403, 404, etc.) stop the chain immediately.
- **Tests updated**: replaced the 429-stops-chain tests with `403` to cover true auth failures, and added two new tests: one for primary 429 falling back successfully, one for a 429 fallback continuing to level 2 then succeeding.

All 41 unit tests pass locally.

julien51:
Addressed the misleading error message and log wording:

- **Accurate error when all providers throw**: Instead of synthesizing a 503 response when the primary throws, we now track `response = null` for throws. If all providers fail with network exceptions (none returned an HTTP response), the error message is now `'All providers failed with network errors'` instead of the misleading `'Provider returned HTTP 503'`.
- **Improved log wording**: Changed `"returned HTTP X, trying fallback: Y"` → `"returned HTTP X, falling back to: Y"` (and `"threw, falling back to: Y"` when the previous provider threw).

All 41 unit tests pass locally.

julien51:
Addressed the two remaining issues from the latest review:

- **Safe JSON in `set-env-vars.sh`**: Replaced heredoc interpolation with `jq -n --arg` for each variable. Values containing `"`, `\`, or newlines are now properly escaped — no more risk of malformed JSON reaching `wrangler secret bulk`.
- **`oneRpcEndpoints` maintenance comment**: Added a comment pointing to the 1RPC supported-networks docs so that when a new chain is added to `supportedNetworks`, the need to also update `oneRpcEndpoints` is visible.

All 41 unit tests still pass locally.

julien51:
Addressed the three remaining items from the latest review:

- **`global_fetch_strictly_public` comment**: Added a 3-line comment in `wrangler.toml` explaining *why* this flag is required (Cloudflare internal CDN routing causes SSL 525 with Alchemy for certain chains) and warning not to remove it.
- **`deploy:code-only` warning**: The script now echoes a visible warning before deploying so stale-secrets deployments are hard to miss by accident.
- **Fallback attempt log level**: Changed `console.warn` → `console.info` for the "falling back to" message. Successful fallbacks are expected behavior, not warnings. `console.warn` is now reserved for actual failures (throws).

All 41 unit tests still pass.

julien51:
Added a  availability check at the top of `set-env-vars.sh` — if `jq` isn't installed, the script now exits immediately with a clear error message and install instructions instead of a confusing 'command not found'.

julien51:
Addressed the two pre-merge issues from the latest review:

- **Type safety in `getFallbackProviders`**: Removed the `as Record<number, ...>` cast. `networks` is already typed as `NetworkConfigs` (`{ [networkId: string]: NetworkConfig }`), so `networks[networkId]` works directly with full type safety.
- **404 should continue the fallback chain**: A 404 from a public RPC fallback almost certainly means a provider endpoint misconfiguration, not a bad request. Updated the break condition to only stop on definitive client errors (401, 403, 400-428 except 404/429) — 404 now continues to the next fallback like 429 does. Added a test for this behavior.

42 tests pass locally.

julien51:
Addressed the two real issues from the latest review:

- **Response body cancellation**: Added `response?.body?.cancel()` before overwriting `response` in the fallback loop and before returning error results. Cloudflare Workers hold open connections for unconsumed bodies — cancelling them explicitly avoids CPU budget waste and potential `worker exceeded time limit` errors on multi-fallback paths.
- **Per-fetch timeout**: Added `signal: AbortSignal.timeout(10_000)` to every `fetchFromProvider` call. A stalled upstream (hanging TCP/TLS) now fails after 10s instead of consuming the full 30s Worker wall-clock limit, worst-case latency with 3 sequential attempts goes from ~90s to ~30s.

All 42 tests still pass.

julien51:
Addressed the three minor documentation gaps from the latest review:

- **Asymmetric 404 handling**: Added a comment on the primary fallback condition explaining why a 404 from the primary provider surfaces immediately (Alchemy returning 404 on a supported chain indicates a misconfiguration beyond the SSL issue the fallback chain is designed to fix — unlike 404 from a fallback public RPC which is likely an endpoint path issue).
- **`response.text()` + `JSON.parse` rationale**: Added a brief comment explaining the choice: it preserves the raw body so malformed-response errors are easier to debug.
- **1RPC static URLs need periodic review**: Added a note in the `oneRpcEndpoints` comment block.

julien51:
Addressed the two items from the latest review:

- **Timeout reduced to 5s**: `FETCH_TIMEOUT_MS = 5_000` with a comment explaining the math (`5s × 3 max attempts = 15s worst-case`, well within CF Workers' 30s wall-clock limit). The previous 10s risked terminating the Worker on a 3-provider all-timeout scenario.
- **Log message fixed**: Changed to `"Last attempt for network X returned HTTP Y, trying next fallback: Z"` — accurate on both the initial primary failure and on subsequent fallback failures, no longer ambiguously refers to "the provider".

julien51:
Addressed the three items from the latest review:

- **Log rename**: `'Last attempt'` → `'Previous attempt'` — the message fires before the next attempt runs, so it correctly describes the *previous* result.
- **404 asymmetry documented**: Added a cross-reference comment at the primary fallback gate (`if (!response || status >= 500 || status === 429)`) noting that 404 from a *fallback* is handled differently inside the loop, pointing to `isDefinitiveClientError`.
- **AbortError/timeout test added**: New test simulates a `DOMException('TimeoutError')` from `AbortSignal.timeout` being thrown by the primary fetch — verifies it's caught the same as any network throw and the fallback fires successfully.

43 tests pass.

julien51:
Addressed the latest review feedback:

- **status in JSON parse error**: added `status: response.status` to the parse error return path (all error paths now consistently include status)
- **oneRpcEndpoints key drift**: exported `oneRpcEndpoints` and added a cross-reference test verifying all its keys have reachable fallbacks via `getFallbackProviders`
- **primary 404 contract**: added explicit test asserting primary 404 surfaces immediately without fallback attempts
- **misleading comment**: fixed 'retried' → 'skipped, not definitive' in the 404 fallback comment

The compatibility date jump and jq CI availability are operational notes — the jq check at the top of `set-env-vars.sh` will surface a clear error immediately if it's missing, and the Cloudflare changelog between 2022–2024 was reviewed when designing this fix (the flag requires 2024-09-23, which is why the date was bumped).

julien51:
Addressed the two real issues from the latest review:

**1. `await response?.body?.cancel()`** — Added `await` to both call sites. `ReadableStream.cancel()` returns a `Promise<void>`; without `await`, the Workers runtime may see an unhandled rejection or hold the upstream connection open during the next `fetch()`.

**2. Primary 404 now tries fallbacks** — Changed the primary trigger condition to include `|| response.status === 404`, making 404 handling consistent everywhere: both primary and fallback 404s skip to the next provider. The `isDefinitiveClientError` exclusion inside the loop was already correct. Updated the test from 'primary 404 surfaces immediately' → 'primary 404 tries fallbacks' to match the new contract.

The compatibility date comment (issue 3) is noted — the bump was intentional and required for `AbortSignal.timeout` + `global_fetch_strictly_public`; verified against the Cloudflare changelog before merging.

julien51:
Latest round addressed:

- **Outer catch `status`**: Added `status: undefined` explicitly to the outermost catch block — consistent shape across all error returns, and `batchProcessor`'s `error.status || 500` already handles undefined correctly
- **Malformed JSON test**: Added test covering HTTP 200 with non-JSON body (e.g. a Cloudflare HTML error page behind the provider URL) — confirms the parse-error path returns `error.message: 'Failed to parse provider response'` with `status: 200`

The PR description note about `global_fetch_strictly_public` being potentially sufficient on its own and the `deploy:code-only` env-var acknowledgement are noted — skipping code changes for those as they're documentation-level.

julien51:
Addressed:
- Dropped `status: undefined` from outer catch — redundant on an optional field
- Added comment at JSON parse site documenting that HTTP 200 with non-JSON body is intentionally not retried (200 means the provider accepted the request)

julien51:
Addressed all three real issues:

**1. Log raw body on parse error** — hoisted `responseText` before the try block so the catch can log it: `console.error('Error parsing JSON response:', error, 'Raw body:', responseText)`. CDN HTML error pages masquerading as 200s will now be visible in the worker logs.

**2. Stronger 1RPC key drift test** — now calls `supportedNetworks(mockEnv, networkId)` and asserts it returns a truthy URL, in addition to checking `getFallbackProviders` returns the 1RPC URL. This catches entries in `oneRpcEndpoints` that reference chains the worker doesn't actually serve.

**3. `deploy:code-only` gated behind `SKIP_SECRET_SYNC=1`** — now fails fast with a clear error unless the env var is explicitly set, preventing accidental stale-secret deploys.

julien51:
Addressed the two issues flagged as 'worth addressing before merging':

**Issue 1 — compat date documentation**: Added a 4-line comment above `compatibility_date` in `wrangler.toml` explaining why 2024-09-23 is required (minimum for `global_fetch_strictly_public`), that the changelog was reviewed, and that no flags in the 2022→2024 range affect this worker's fetch/JSON usage.

**Issue 3 — timeout comment accuracy**: Updated the `FETCH_TIMEOUT_MS` comment to clarify it covers connection + headers (time-to-first-byte), not body streaming, with a note that CF Workers' 30 s wall-clock limit is the backstop for body reads.

**Bonus — issue 4 (minor)**: Added `triedFallbacks` flag to emit an accurate error message: 'Provider failed with a network error' when only one provider was tried vs. 'All providers failed with network errors' when fallbacks were also exhausted.

julien51:
Fixed the inaccurate timeout comment: `AbortSignal.timeout()` aborts the **entire** fetch lifecycle (connection, headers, and body streaming) — not just TTFB. Previous comment was wrong; corrected.

The other two points are noted:
- **Global flag trade-off**: acknowledged — all outbound fetches now bypass CF CDN routing. Fixing 5 broken chains outweighs any potential latency change for healthy chains, and there's no per-request way to scope it.
- **publicProvider audit**: worth doing post-merge. The `global_fetch_strictly_public` root fix makes this less urgent since the 525s are resolved at the connection level.

julien51:
Updated the timeout comment to document the 5 s trade-off: 5 s × 3 providers = 15 s worst-case within the 30 s CF wall-clock limit, with a note that public fallback RPCs (publicProvider, 1RPC) are typically slower than Alchemy. If P99 on fallbacks proves too high post-deploy, bumping to 8–10 s is the lever.

julien51:
Addressed the two actionable minor issues:

**Issue 4 — `set -u`**: Added `set -eu` to `set-env-vars.sh`. Used `${!VAR:-}` (default empty string) for the indirect variable reference so `set -u` doesn't error on unset variables — the empty-string check still catches them correctly.

**Issue 6 — cancel invariant**: Updated the `response?.body?.cancel()` comment to make the invariant explicit: '`response` here is always the previous attempt's response (or null if it threw) — cancel before overwriting to release the upstream CF connection.'

The other points are operational notes (1Password in CI is pre-existing, compat date jump will be staged, 1RPC rate limits are acceptable for a fallback).

julien51:
**Verified**: `NetworkConfigs` in `@unlock-protocol/types` is typed as `{ [networkId: string]: NetworkConfig }` — string keys — so `networks[networkId]` is correct and the fallback chain works as expected.

**Fixed**: Changed `forwardingResult.error.status || 500` → `forwardingResult.error.status ?? 500` in `batchProcessor.ts`. More precise for a `number | undefined` field.

julien51:
Addressed both real issues:

**1. deploy:code-only readability**: Moved the guard to `scripts/deploy-code-only.sh` with an explicit `if/then/else` — no operator-precedence ambiguity, clear error message with a hint to use `yarn deploy` instead.

**2. Primary body cancel safety net**: Added a comment to the final `await response?.body?.cancel()` explaining it serves as the safety net when `getFallbackProviders` returns empty — makes the invariant explicit without restructuring the code.

Also confirmed: `NetworkConfigs` in `@unlock-protocol/types` uses `[networkId: string]` — string keys — so `networks[networkId]` is correct (no cast needed).

julien51:
Added comment to `deploy-code-only.sh` clarifying that it calls `yarn wrangler deploy` directly (not `yarn deploy`) to intentionally bypass `set-env-vars` — makes the deliberate divergence explicit for future maintainers.

The timeout and jq documentation points are noted; the timeout trade-off is already documented in the `FETCH_TIMEOUT_MS` comment, and jq's install hint is already in the script's error message.

julien51:
**ABOUTME comments**: Keeping them — per the project's CLAUDE.md: 'All code files should start with a brief 2-line comment prefixed with "ABOUTME:".' The reviewer's 'Default to writing no comments' applies to inline comments, not the mandatory file header.

**jq dependency**: Already documented in the error message (`brew install jq / apt install jq`) and the jq check fires immediately at script entry. No further documentation change made.

**global_fetch_strictly_public affects all fetches**: Already captured in the wrangler.toml comment: 'Keep this flag — removing it re-introduces those failures.' The worker is a pure RPC proxy so all outbound fetches are to provider URLs; a future maintainer adding Cloudflare-internal calls would need to revisit this anyway.

**200+non-JSON fallback gap**: Acknowledged. The primary fix (flag) makes this much less likely in practice. Filed as a known gap in the code comment.

PR #16347, opened 2026-05-01
fix: pass new wallet address after Privy embedded wallet creation

julien51:
## Summary

- Brand new email users were getting stuck on the checkout "Connect" step after Privy email verification
- Root cause: in `PrivyMigration.handleMigrationIfNeeded`, after `createWallet()` succeeds the code called `onSignedInWithPrivy(user)` with the **stale** `user` closure argument — which still has `wallet: undefined` because Privy hasn't updated the reactive state yet
- Fix: pass the newly created wallet address as `walletAddressOverride` to `onSignedInWithPrivy` so it doesn't rely on the stale user object

## Changes

- `unlock-app/src/config/PrivyProvider.tsx`: add optional `walletAddressOverride` param to `onSignedInWithPrivy`; use it in `handleMigrationIfNeeded` after wallet creation
- `unlock-app/src/__tests__/config/PrivyProvider.test.ts`: 5 new tests covering the bug scenario and existing behaviour

## Test plan

- [ ] Existing tests pass: `cd unlock-app && yarn test`
- [ ] New user signs up with email via Privy on a checkout page — should advance past step 2 after email verification
- [ ] Returning user with existing wallet still authenticates correctly
- [ ] Legacy account detection flow unaffected


julien51:
Addressed latest review feedback:
- **Override priority flipped**: `user.wallet?.address || walletAddressOverride` — live Privy wallet wins; override only fills in for stale user objects
- **Test updated** to assert user.wallet.address takes precedence when both are present
- **ABOUTME headers removed** from test file

julien51:
Addressed the latest review feedback:

- **Explicit `null` return when Locksmith has no access token**: `onSignedInWithPrivy` now returns `null` (not `undefined`) when `loginWithPrivy` succeeds but the response carries no `accessToken` — no longer falls through silently.
- **Guard against double Locksmith auth**: Added `if (getAccessToken()) return` before the normal login path in `handleMigrationIfNeeded`. After wallet creation succeeds and the token is saved, Privy re-fires the `useEffect([user])` with the updated user object — this guard prevents a redundant second auth call.
- **Test coverage**: Added test case "returns null when Locksmith returns no access token" covering the new `null` return path.

All 6 unit tests pass locally.

PR #16349, opened 2026-05-05
fix: pin @railway/cli to 4.5.4 to restore production deploys

julien51:
## Summary

- Pins `@railway/cli` to `4.5.4` in `.github/actions/railway/action.yml`
- The action was installing the latest CLI on every run (`npm i -g @railway/cli`)
- CLI v4.45.0 changed Dockerfile detection: it now rejects Dockerfiles in subdirectories (`acceptChildOfRepoRoot:false`), so `locksmith/Dockerfile` is silently skipped and the Railway build fails with no Dockerfile to use
- v4.5.4 was the version in use when production deploys last succeeded (2025-07-15)

**Root cause log from failing deploy:**
```
skipping 'Dockerfile' at 'locksmith/Dockerfile' as it is not rooted at a valid path (root_dir=, fileOpts={acceptChildOfRepoRoot:false})
```

## Test plan

- [ ] Merge this and verify the next production deploy has `deploy-locksmith-web-production` and `deploy-locksmith-worker-production` succeeding


julien51:
Closing — branch was accidentally based on production history. Replacing with #https://github.com/unlock-protocol/unlock/pull/16350

PR #16351, opened 2026-05-05
revert: undo railway CLI version pin (4.5.4 token format incompatible)

julien51:
Reverts #16350.

v4.5.4 fails with `Unauthorized` — the Railway token in 1Password was rotated to a newer format that only v4.45+ recognizes. The pin made things worse.

The actual build failure is happening on Railway's side (after successful upload with v4.45.0). Need to investigate Railway build logs directly.

### DIFF
--- a/.github/actions/railway/action.yml
+++ b/.github/actions/railway/action.yml
@@ -22,7 +22,10 @@ runs:
   steps:
     - name: Install Railway CLI
       shell: bash
-      run: npm i -g @railway/cli
+      # Pinned: v4.45.0 is the minimum that accepts the current RAILWAY_TOKEN format.
+      # Dockerfile path is set via railway.toml — verify new versions still support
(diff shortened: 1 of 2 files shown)


julien51:
Addressed both reviews:

- **Review 1**: Pinned CLI explicitly to `v4.45.0` (the version that accepts the current token format) with a comment explaining why, so future maintainers don't blindly upgrade.
- **Review 2**: The pin was already reverted in the previous commit — `action.yml` was already back to unpinned. Now it's pinned to `v4.45.0` explicitly. The second reviewer was likely looking at an older snapshot.

Net diff vs master: `railway.toml` added + CLI pinned to `4.45.0`. Together these fix both problems (Dockerfile detection + token auth).

julien51:
Fixed merge conflict: master had `@4.5.4` (from #16350 merge), kept `@4.45.0` which is what actually works with the current token.

PR #16365, opened 2026-05-05
fix: create EVM wallet for Solana-only Privy users, fix blank checkout Connect step

julien51:
## Problem

When a Privy user's primary wallet is a Solana address, `user.wallet?.address` returns a base58-encoded key (no `0x` prefix). The previous code passed that address to Locksmith's `/v2/auth/privy` endpoint, which returned 401 and left the checkout's Connect step blank.

## Fix

**`PrivyProvider.tsx`**

- Adds a `findEvmWallet(user)` helper that scans `user.linkedAccounts` for `chainType === 'ethereum'` — the correct Privy API for finding an EVM wallet, regardless of which wallet Privy designates as "primary"
- `onSignedInWithPrivy` uses `findEvmWallet` instead of `user.wallet`; falls back to `walletAddressOverride` (used when a wallet was just created and Privy's user object has not yet updated)
- `PrivyMigration.handleMigrationIfNeeded` detects Solana-only users via `findEvmWallet` and creates an embedded EVM wallet via `useCreateWallet`, then calls `onSignedInWithPrivy` with the new address — no error shown to the user
- Cache-clearing on errors scoped to 4xx HTTP responses only (via `isAxiosError`), so transient 5xx/network errors do not wipe a valid session
- Legacy account + Solana wallet edge case fixed: migration modal guard uses `!hasEvmWallet` instead of `!user.wallet?.address`

**`session.ts`**

- Exports `removeCurrentAccount()` helper used by the error-recovery paths

## Edge cases

- **Solana-only user on first login**: EVM wallet created transparently; no error shown
- **Legacy account + Solana wallet**: migration modal appears correctly
- **Returning Solana user with existing EVM wallet**: `getAccessToken()` guard prevents duplicate Locksmith auth calls
- **Transient network/server errors**: 5xx errors do not clear a valid cached session

### DIFF
--- a/unlock-app/src/config/PrivyProvider.tsx
+++ b/unlock-app/src/config/PrivyProvider.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { getAccessToken, saveAccessToken } from '~/utils/session'
+import { getAccessToken, saveAccessToken, removeAccessToken, removeCurrentAccount } from '~/utils/session'
 import {
   getAccessToken as privyGetAccessToken,
   PrivyProvider,
@@ -9,6 +9,7 @@ import {
   User,
   usePrivy,
   LinkedAccountWithMetadata,
+  WalletWithMetadata,
 } from '@privy-io/react-auth'
 import { ReactNode, useContext, useEffect, useState } from 'react'
 import { config } from './app'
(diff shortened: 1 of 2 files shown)


julien51:
Addressed all review comments: (1) hoisted `walletAddress` above try/catch to fix ReferenceError; (2) moved `hasEvmWallet` before the legacy-account check and updated the migration modal guard from `!user.wallet?.address` to `!hasEvmWallet`; (3) downgraded misleading console.error to console.info for the no-EVM-wallet pending path; (4) updated PR description to reflect actual implementation.

julien51:
Addressed the latest round of review feedback: extracted a shared `findEvmWallet()` helper to eliminate the duplicate EVM wallet predicate, and scoped the catch-block cache-clearing to 4xx HTTP responses only so transient 5xx/network errors don't wipe valid sessions.

julien51:
Latest round: used `isAxiosError` from axios to replace the `(error as any)` cast, and updated the PR title + description to accurately describe the implementation (EVM wallet creation via PrivyMigration, not an error banner in ConnectPage). Re: the non-token success path — clearing the cache there is intentional: a 2xx with no token means this wallet can't authenticate with Locksmith, so the cached association is invalid. Happy to discuss if there's a safer approach.

PR #16366, opened 2026-05-05
fix: suppress false wallet-creation error toast; restrict OAuth in iframes

julien51:
## Problems

**1. False \"Failed to create wallet\" toast**

`useCreateWallet`'s `onError` callback and `createWalletForUser`'s `catch` block both called `ToastHelper.error`, so any wallet-creation error showed the toast twice. Worse, Privy can fire `onError` for race-condition cases (e.g. a concurrent call already created the wallet) even when the overall flow succeeds — the user sees an error but the wallet is usable.

**2. Google OAuth appearing in embedded checkout (dappcon.io)**

The `isInPaywall` guard was:
\`\`\`ts
isInIframe() && !window.location.href.includes(config.unlockApp)
\`\`\`
When dappcon.io embeds the checkout, it loads `https://app.unlock-protocol.com/checkout` in an iframe. `isInIframe()` is `true` but `window.location.href.includes('app.unlock-protocol.com')` is also `true`, so `isInPaywall = false` and Google OAuth appeared — which fails on third-party domains.

## Fixes

- `useCreateWallet`'s `onError` now only logs (suppressed from user); `createWalletForUser`'s `catch` is the single source of user-facing errors
- `catch` handles `embedded_wallet_already_exists`: a concurrent call already created the wallet, so return the existing EVM address instead of erroring
- `isInPaywall` simplified to `isInIframe()` — any iframe context is an embedded paywall and should restrict to wallet login only

### DIFF
--- a/unlock-app/src/__tests__/config/PrivyProvider.test.ts
+++ b/unlock-app/src/__tests__/config/PrivyProvider.test.ts
@@ -23,6 +23,12 @@ vi.mock('~/config/locksmith', () => ({
 vi.mock('~/utils/session', () => ({
   saveAccessToken: vi.fn(),
   getAccessToken: vi.fn(),
+  removeAccessToken: vi.fn(),
+  removeCurrentAccount: vi.fn(),
+}))
+
+vi.mock('axios', () => ({
+  isAxiosError: vi.fn().mockReturnValue(false),
 }))
 
 import { getAccessToken as privyGetAccessToken } from '@privy-io/react-auth'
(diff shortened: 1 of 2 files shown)


julien51:
Fixed test failures: updated `makeUser` in `PrivyProvider.test.ts` to include `linkedAccounts` matching the updated `findEvmWallet` logic (which reads from `linkedAccounts` instead of `user.wallet`). Also added mocks for `removeAccessToken`, `removeCurrentAccount`, and `isAxiosError`, and a new Solana-wallet test case.

PR #16415, opened 2026-05-12
Sync event image changes to ticket metadata

### LINKED ISSUE #13545: Updating event image should update NFT image if they were the same before update
**Describe the bug**
When you update an event's image, it doesn't update the NFT image

**To Reproduce**

1. Update event image
2. See that NFT image hasn't changed

**Expected behavior**
If event image and NFT are the same, updating event image should update NFT image. If they're different, it shouldn't (or maybe prompt the user? "Do you want to update the NFT image?"). It's also a little tricky that the NFT image is buried in the lock config section, not obvious for an event manager.

**Screenshots**
Screenshot_033024_015443_PM

N/A

**Smartphone (please complete the following information):**

N/A

**Additional context**
Apologies if this should be a Feature request and not a Bug report

694410194:
## Summary
- Sync event image updates into lock NFT metadata only when the NFT image still matches the previous event image
- Preserve custom NFT images that have already diverged from the event image
- Add regression coverage for matching vs custom lock metadata images

Fixes #13545

## Verification
- `NODE_ENV=test ../node_modules/.bin/vitest run __tests__/operations/eventOperations.test.ts --coverage=false`

Note: I also attempted `yarn workspace @unlock-protocol/locksmith build`; in this sparse checkout it fails because local generated/dist workspace artifacts for `@unlock-protocol/unlock-js`, `@unlock-protocol/ui`, and `@unlock-protocol/contracts` are missing. The focused locksmith event operations test passes after running the repo migrations locally.

### DIFF
--- a/locksmith/__tests__/operations/eventOperations.test.ts
+++ b/locksmith/__tests__/operations/eventOperations.test.ts
@@ -1,6 +1,11 @@
 import { expect, vi } from 'vitest'
 
-import { CheckoutConfig, EventData, KeyMetadata } from '../../src/models'
+import {
+  CheckoutConfig,
+  EventData,
+  KeyMetadata,
+  LockMetadata,
+} from '../../src/models'
 import { getEventFixture } from '../../__tests__/fixtures/events'
 
 import {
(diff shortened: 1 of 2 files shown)


694410194:
Opened the CLA-only PR requested by cla-bot: https://github.com/unlock-protocol/unlock/pull/16417

PR #16530, opened 2026-06-15
fix(locksmith): trim whitespace from custom-email recipients

julien51:
## Problem

Customers reported transactional/custom emails not arriving. Investigation traced it to the **custom email sender** ("email all members" from the dashboard): member email addresses stored with leading/trailing whitespace (e.g. `"igor@gateway.fm "`) were passed unmodified into `sendEmailJob`, whose `recipient: z.string().email()` validation rejected them.

Those recipients were **silently dropped** — the job failed and exhausted retries — while clean addresses went through. In production this is a chronic drip: ~195 member emails currently carry surrounding whitespace, and the graphile-worker failed-jobs table shows a long tail of `sendEmailJob` failures with `"Invalid email"` on whitespace addresses.

> Note: this is **separate** from the worker outage that triggered the original report (the job executor was wedged and was restarted). This PR fixes the underlying data-handling bug so these recipients stop being dropped once the queue is healthy.

## Fix

- `Normalizer.emailAddress` now **trims** in addition to lowercasing (mirrors `ethereumAddress`, which already trims).
- `sendToAllJob` resolves each recipient's email through a new pure helper `resolveRecipientEmail`, which normalizes the value and returns `undefined` when no email is present — so members without an email are **skipped** rather than enqueued as jobs guaranteed to fail.

## Tests (TDD)

- `normalizer.test.ts`: `emailAddress` trims surrounding whitespace.
- `sendToAll.test.ts`: `resolveRecipientEmail` trims/lowercases, reads alternative email keys case-insensitively, and returns `undefined` when absent.

Both written test-first (watched fail, then implemented). `eslint` clean.

### DIFF
--- a/locksmith/__tests__/utils/normalizer.test.ts
+++ b/locksmith/__tests__/utils/normalizer.test.ts
@@ -8,6 +8,13 @@ describe('Normalizer', () => {
         'test@example.com'
       )
     })
+
+    it('trims surrounding whitespace', () => {
+      expect.assertions(1)
+      expect(Normalizer.emailAddress('  igor@gateway.fm  ')).toEqual(
+        'igor@gateway.fm'
+      )
+    })
   })
 
   describe('ethereumAddress', () => {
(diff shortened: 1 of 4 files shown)


julien51:
CI fix: the `Unit Tests locksmith` job failed on a single prettier/prettier error in the new `sendToAll.test.ts` (the 317 other lint items are pre-existing `no-explicit-any` warnings, not from this PR). Reformatted the test file with prettier — `eslint .` now reports 0 errors and the 7 new tests pass locally. No production-code changes in this push.

PR #16593, opened 2026-08-08
fix(locksmith): block private WebSub hub.callback URLs (SSRF)

SashaMIT:
## Summary
- Unauthenticated `POST /api/hooks/:network/*` accepted `hub.callback` with only zod `.url()` validation, then Locksmith `fetch`ed it during WebSub intent verification.
- An attacker could point the callback at loopback / private / link-local / cloud-metadata addresses (SSRF), and successful subscriptions later deliver lock/key events to that URL.
- Add `assertSafeCallbackUrl` (Hop-style IP class checks + DNS resolve), call it before verify/rejection/notify fetches, and set `redirect: 'error'` to block open-redirect pivots.

## Test plan
- [x] `yarn vitest run __tests__/utils/safeCallbackUrl.test.ts --coverage=false` → 3/3
- [ ] CI green on this PR

Tooling assist used for prep; commit authored/signed by me (no Cursor co-author trailer).

Made with [Cursor](https://cursor.com)

PR #16594, opened 2026-08-09
fix(locksmith): block private URLs for Apple Wallet pass thumbnails

SashaMIT:
## Summary

Sibling of #16593 (WebSub `hub.callback` SSRF). `createAppleWalletPass` fetches lock `metadata.image` via bare `fetch` while generating a pass for unauthenticated `GET /v2/pass/.../ios`. A lock manager can point `image` at a private/link-local URL and make Locksmith request it.

This reuses `assertSafeCallbackUrl` (same helper as #16593) before the thumbnail download and sets `redirect: 'error'` so a 3xx cannot bypass the host check.

If #16593 merges first, rebase will drop the shared util add and keep only the pass call-site. If this merges first, #16593 can reuse the util as-is.

## Test plan

- [x] `npx vitest run __tests__/utils/safeCallbackUrl.test.ts` (3 passed)
- [ ] Confirm pass generation still works with a public https thumbnail

Made with [Cursor](https://cursor.com)

### DIFF
--- a/locksmith/__tests__/utils/safeCallbackUrl.test.ts
+++ b/locksmith/__tests__/utils/safeCallbackUrl.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from 'vitest'
+import {
+  assertSafeCallbackUrl,
+  isBlockedIp,
+} from '../../src/utils/safeCallbackUrl'
+
+describe('safeCallbackUrl', () => {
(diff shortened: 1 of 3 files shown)


SashaMIT:
CLA sign PR opened per cla-bot: https://github.com/unlock-protocol/unlock/pull/16596 (single-file `.clabot` add for `SashaMIT`). Once that merges, `verification/cla-signed` should clear on this PR. Vercel authorize remains maintainer-side.

PR #16606, opened 2026-08-28
fix: sign with the authenticated wallet, not whichever Privy lists first

blahkheart:
# Description

Fixes wallet-backed transactions using a different connected wallet from the account shown as authenticated.

Privy can return multiple connected wallets without guaranteeing their order. `useProvider` now resolves the wallet whose address matches the authenticated account, creates the `WalletService` from that wallet provider instead of a stale context provider, and stops the transaction with a clear prompt when the correct wallet is unavailable.

Delegated checkout providers remain owned by the parent page, matching context providers are reused without replacing provider identity, and asset watching uses the authenticated provider. Background network discovery also handles wallet mismatch rejections without leaving an unhandled promise.

Adds focused regression tests covering unordered wallets, stale context, delegated checkout with and without a Privy session, provider reuse, mismatched-wallet handling, and asset watching.

# Checklist

- [x] Documentation changes are not applicable
- [x] I have added tests that prove the fix is effective
- [x] Screenshots are not applicable because this has no visual changes

# Testing

- `yarn workspace @unlock-protocol/unlock-app vitest run --coverage.enabled=false` — 234 passed, 3 skipped
- `yarn workspace @unlock-protocol/unlock-app tsc --noEmit --pretty false`
- ESLint and Prettier checks on the changed files
- Repository pre-commit and pre-push hooks

## Release Note Draft Snippet

Fixed wallet-backed actions, including membership airdrops, so they sign with the authenticated wallet instead of another connected wallet.

### DIFF
--- a/unlock-app/src/__tests__/hooks/useProvider.test.tsx
+++ b/unlock-app/src/__tests__/hooks/useProvider.test.tsx
@@ -1,0 +1,149 @@
+// @vitest-environment jsdom
+
+import { renderHook } from '@testing-library/react'
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import React from 'react'
+
+const AUTHENTICATED = '0x374355b89D26325c4C4Cd96f99753b82fd64b2Bb'
+const OTHER = '0x2268b701DdE76122ff70af751CafA39623799e49'
+
+const makeWallet = (address: string) => ({
+  address,
+  getEthereumProvider: vi.fn().mockResolvedValue({ __address: address }),
+})
+
+// Wallets Privy reports as connected. Reassigned per test.
+let connectedWallets: ReturnType<typeof makeWallet>[] = []
+const connectWallet = vi.fn()
+
+vi.mock('@privy-io/react-auth', () => ({
+  useWallets: () => ({ wallets: connectedWallets }),
+  // Must be callable with no arguments: passing callbacks reaches Privy
+  // internals that require a surrounding PrivyProvider.
+  useConnectWallet: () => ({ connectWallet }),
+}))
+
+const toastError = vi.fn<(message: string) => void>()
+vi.mock('@unlock-protocol/ui', () => ({
+  ToastHelper: {
+    error: (message: string) => toastError(message),
+    success: vi.fn(),
+  },
+}))
+
+const walletServiceConnect = vi.fn().mockResolvedValue(1)
+vi.mock('@unlock-protocol/unlock-js', () => ({
+  WalletService: class {
+    connect = walletServiceConnect
+    getAccount = vi.fn().mockResolvedValue(AUTHENTICATED)
+  },
+}))
+
+// Captures which underlying provider each BrowserProvider was built from, so
+// assertions can tell the two wallets apart.
+vi.mock('ethers', () => ({
+  ethers: {
+    BrowserProvider: class {
+      underlying: unknown
+      constructor(underlying: unknown) {
+        this.underlying = underlying
+      }
+      getNetwork = vi.fn().mockResolvedValue({ chainId: '8453' })
+      send = vi.fn().mockResolvedValue(undefined)
+    },
+  },
+}))
+
+import { useProvider } from '~/hooks/useProvider'
+import ProviderContext from '~/contexts/ProviderContext'
+import AuthenticationContext from '~/contexts/AuthenticationContext'
+
+const setProvider = vi.fn()
+
+// The provider already in context — deliberately the *wrong* wallet, matching
+// the state that produced the original bug.
+const staleProvider = {
+  underlying: { __address: OTHER },
+  getNetwork: vi.fn().mockResolvedValue({ chainId: '8453' }),
+  send: vi.fn().mockResolvedValue(undefined),
+}
+
+const wrapper = ({ children }: { children: React.ReactNode }) =>
+  React.createElement(
+    AuthenticationContext.Provider,
+    { value: { account: AUTHENTICATED, setAccount: vi.fn() } },
+    React.createElement(
+      ProviderContext.Provider,
+      { value: { provider: staleProvider, setProvider } },
+      children
+    )
+  )
+
+describe('useProvider', () => {
+  beforeEach(() => {
+    connectedWallets = []
+    connectWallet.mockClear()
+    toastError.mockClear()
+    setProvider.mockClear()
+    walletServiceConnect.mockClear()
+  })
+
+  it('picks the authenticated wallet by address, not by list position', async () => {
+    // Authenticated wallet is present but NOT first: indexing wallets[0] would
+    // pick the wrong one.
+    connectedWallets = [makeWallet(OTHER), makeWallet(AUTHENTICATED)]
+
+    const { result } = renderHook(() => useProvider(), { wrapper })
+    await result.current.getWalletService(8453)
+
+    expect(connectWallet).not.toHaveBeenCalled()
+    expect(toastError).not.toHaveBeenCalled()
+
+    // The WalletService must be built from the authenticated wallet.
+    const usedProvider = walletServiceConnect.mock.calls[0][0]
+    expect(usedProvider.underlying).toEqual({ __address: AUTHENTICATED })
+  })
+
+  it('signs with the authenticated wallet, not the stale context provider', async () => {
+    // Context still holds the previously-connected wrong wallet.
+    connectedWallets = [makeWallet(AUTHENTICATED)]
+
+    const { result } = renderHook(() => useProvider(), { wrapper })
+    await result.current.getWalletService(8453)
+
+    const usedProvider = walletServiceConnect.mock.calls[0][0]
+    expect(usedProvider.underlying).not.toEqual({ __address: OTHER })
+    expect(usedProvider.underlying).toEqual({ __address: AUTHENTICATED })
+    // And the context is refreshed for subsequent renders.
+    expect(setProvider).toHaveBeenCalled()
+  })
+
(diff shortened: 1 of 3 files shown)


unlock-app/src/hooks/useProvider.ts
```diff
@@ -93,46 +100,55 @@ export const useProvider = () => {
...
+   * callers must not fall back to the ambient provider: it may well be signing
+   * from a different address than the one the UI shows as connected.
    */
-  const ensureCorrectWallet = async () => {
-    if (!account) return true
-    // Get current connected address
-    const currentWalletAddress = wallets[0]?.address?.toLowerCase()
-    const authenticatedAddress = account.toLowerCase()
+  const resolveAuthenticatedProvider = async () => {
+    if (!account) return provider
```
julien51: Delegated-provider checkout (`paywallConfig.useDelegatedProvider`) has `account` set via SIWE but an empty Privy `wallets` list, so the lookup below always misses and throws. Suggest:

```suggestion
    if (!account || provider?.parentOrigin) return provider
```

This also prevents the match path from overwriting the parent-page provider (see comment on `setProvider` below).

unlock-app/src/hooks/useProvider.ts
```diff
@@ -93,46 +100,55 @@ export const useProvider = () => {
...
     }
+
+    const browserProvider = createBrowserProvider(
+      await wallet.getEthereumProvider()
+    )
+
+    // Keep the context in sync for later renders. The returned value is what
+    // this call must use: `provider` from context is captured at render time
+    // and will still hold the previous wallet for the rest of this call.
+    setProvider(browserProvider)
```
julien51: Two issues here:

1. When Privy does have a session in the checkout iframe for the same address in delegated mode, this replaces the parent-page provider with a `BrowserProvider` that has no `parentOrigin`, so `useSIWE.tsx:84` computes the wrong domain and `useCheckoutCommunication.ts:145` routes RPC to the in-iframe wallet instead of the parent's. The `parentOrigin` guard above covers it.
2. This runs on every `getWalletService` call with a fresh provider identity, re-rendering the whole tree under `GlobalWrapper` even when context already holds the right wallet. Please short-circuit when the resolved wallet address matches the one already in context (master did this via the early `===` return).

unlock-app/src/hooks/useProvider.ts
```diff
@@ -93,46 +100,55 @@ export const useProvider = () => {
...
+
+    return browserProvider
+  }
+
+  /**
+   * Ensures the connected wallet matches the authenticated account
+   * Returns true if successful, false otherwise
+   */
+  const ensureCorrectWallet = async () => {
```
julien51: `ensureCorrectWallet` has no callers in the monorepo and is now a wrapper with side effects (overwrites `ProviderContext`) that can reject (`getEthereumProvider()` throws when the wallet is not connected; master caught that and returned `false`). The doc comment above no longer matches. Either delete it or make the doc + error handling match.

unlock-app/src/hooks/useProvider.ts
```diff
@@ -154,11 +170,18 @@ export const useProvider = () => {
...
     try {
-      // Ensure the wallet matches the authenticated account before proceeding
-      await ensureCorrectWallet()
+      // Ensure the wallet matches the authenticated account before proceeding.
+      // Signing with a different address than the UI reports produces failures
+      // that surface as unrelated contract reverts, so this must be fatal.
+      const activeProvider = await resolveAuthenticatedProvider()
+      if (!activeProvider) {
+        throw new Error(
+          `Wrong wallet connected. Please switch to ${account} in your wallet and try again.`
```
julien51, line 195: Users now get two near-identical messages: the toast from `resolveAuthenticatedProvider` plus this thrown message, which callers re-toast (`MessageToSign.tsx:48`, `ConfirmCrypto` → `Confirm.tsx` onError). Suggest dropping the inline toast and letting callers surface the thrown message.

unlock-app/src/__tests__/hooks/useProvider.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+}))
+
+import { useProvider } from '~/hooks/useProvider'
+import ProviderContext from '~/contexts/ProviderContext'
+import AuthenticationContext from '~/contexts/AuthenticationContext'
+
+const setProvider = vi.fn()
+
+// The provider already in context — deliberately the *wrong* wallet, matching
+// the state that produced the original bug.
```
julien51: Nit: these comments (here and at L129-130) describe bugfix history. Please rephrase as the invariant under test, e.g. "context holds a provider for a different address than the authenticated account; the WalletService must not be built from it".

julien51 (changes requested):
Thanks — the core fix is right (looking the wallet up by address instead of `wallets[0]`, and using the resolved provider rather than the render-captured one), and the tests are focused. One regression blocks merging, plus a few smaller things.

### Blocking: breaks checkout with `useDelegatedProvider`

`packages/paywall/src/utils/index.ts:97` sets `useDelegatedProvider = !!provider`, so any integrator passing their own provider hits this path. In that mode the provider is injected by the parent page (`useCheckoutCommunication.ts:259`) and there is no Privy login, so `useWallets()` is **empty**. `Select.tsx:478` signs in via SIWE (works, `account` is still undefined) and then sets `account` — after which every `getWalletService` call (`MessageToSign.tsx:30`, `Returning.tsx:43`, `ConfirmCrypto.tsx:166`, `ConfirmCrossChainPurchase.tsx:109`) misses in `wallets.find`, pops the Privy connect modal this mode deliberately hides, and throws. On master the `false` from `ensureCorrectWallet` was discarded and the delegated provider was used.

Suggested guard at the top of `resolveAuthenticatedProvider` (the parent page owns the account there):

```ts
if (!account || provider?.parentOrigin) return provider
```

Please add a test with `connectedWallets = []`, `account` set and a `parentOrigin` provider in context, asserting the context provider is used and no modal/throw.

### Should fix

- `watchAsset` was not migrated: it still uses the ambient `provider` for `wallet_switchEthereumChain` / `wallet_watchAsset`, so "Add to wallet" (`keychain/Key.tsx:147`) targets `wallets[0]` — the exact scenario this PR fixes.
- `utils/networks.ts:31` (`useAvailableNetworks`) calls `getWalletService()` from a `useEffect` with no `.catch`. Now that a mismatch is fatal, simply rendering Create Lock / New Event / Certification forms in a mismatched state opens the modal, toasts, and leaves an unhandled rejection. Worth a `.catch` there.

### Consider

- The root cause is one layer lower: `useAuthenticate.ts:143` still seeds `ProviderContext` from `wallets[0]`. Resolving by address there (keyed on `[account, wallets]`) would fix every consumer of `provider` once, instead of re-resolving and calling `setProvider` on every `getWalletService` call (fresh `BrowserProvider` identity each time → full-tree re-render mid-transaction, and bulk flows like `ExtendKeysDrawer` call it per item). At minimum, skip `setProvider` when the resolved wallet address already matches what is in context.
- Heads-up on a behavior change: `account` comes from `user.linkedAccounts` (`PrivyProvider.tsx:75`), not `useWallets()`. A user who linked MetaMask but signs in with email now hard-fails every tx until they connect MetaMask, where master silently signed with the embedded wallet. Probably the right outcome, but it is a UX cliff to be aware of.

CI: the `claude-review` failure is an OIDC-token issue on fork PRs and the Vercel checks just need team authorization — neither is on you.

unlock-app/src/hooks/useProvider.ts
blahkheart: Addressed in abc728981. Delegated providers now return before any Privy wallet lookup, and regression tests cover both an empty Privy wallet list and a simultaneous matching Privy session.

unlock-app/src/hooks/useProvider.ts
blahkheart: Addressed in abc728981. The delegated-provider guard preserves parentOrigin behavior, and the context provider is now reused when eth_accounts[0] matches the authenticated address, avoiding a fresh provider identity and setProvider call. Both paths have regression coverage.

unlock-app/src/hooks/useProvider.ts
blahkheart: Addressed in abc728981 by removing the unused ensureCorrectWallet wrapper and export.

unlock-app/src/hooks/useProvider.ts
blahkheart: Addressed in abc728981. The inline toast was removed; callers now receive the single thrown mismatch error. The regression test asserts that no toast is emitted here.

unlock-app/src/__tests__/hooks/useProvider.test.tsx
blahkheart: Addressed in abc728981. The comments and test naming now state the provider/account invariants rather than the bugfix history.

blahkheart:
Addressed the requested changes in abc728981:

- preserved parent-owned delegated checkout providers, with tests for empty and populated Privy wallet lists
- reused a matching context provider instead of replacing its identity
- removed the unused ensureCorrectWallet wrapper and duplicate toast
- routed watchAsset through the authenticated provider
- handled useAvailableNetworks rejection paths
- rewrote test comments as behavioral invariants

Validation: 234 tests passed, 3 skipped; TypeScript, Prettier, commit hooks, and push hooks passed.

unlock-app/src/__tests__/hooks/useProvider.test.tsx
blahkheart: Noted, thanks for the feedback. I'll work on these

